PR #1, opened 2014-09-27
Add a new option immediate= to immediately apply a permanent change

mscherer:
Currently, either you apply the change in the configuration
of firewalld ( without permanent=True ), or you apply it live.
I most of the time want to do the 2 at the same time, ie open the
port ( so I can use the service ) and make sure it stay open on reboot.

### DIFF
--- a/system/firewalld.py
+++ b/system/firewalld.py
@@ -52,6 +52,12 @@ options:
       - "Should this configuration be in the running firewalld configuration or persist across reboots"
     required: true
     default: true
+  immediate:
+    description:
+      - "Should this configuration be applied immediately, if set as permanent"
+    required: false
+    default: false
+    version_added: "1.9"
   state:
     description:
       - "Should this port accept(enabled) or reject(disabled) connections"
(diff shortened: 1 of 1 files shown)


mscherer:
This was first on https://github.com/ansible/ansible/pull/9115 

So I am ok with tweaking the message, but i am a bit short on idea of what to use instead. 
I would suggest: "Operation applied to on-disk configuration", and "Operation applied to live firewall". Instead of "Permanent operation" and "Non permanent operation".  Thoughts ?

mpdehaan:
Thanks!

stick a "version_added: '1.8'" on this argument and it's good to go.

We'll merge core much faster given the split.

ghost:
@mpdehaan : it's there now.

maxamillion:
+1 LGTM.
I added a comment on the old PR before the split and this looks good, I was curious about feedback for the needs_revision tag? (Or maybe that doesn't mean what I think it means.)

josephholsten:
@mpdehaan does this need anything else for merge? Seems like this repo is a bit under water. Anything we can do to help keep on top of outstanding PRs?

bcoca:
can you change 'version added' to 1.9?

mscherer:
Yep, just did.

maxamillion:
Is the `needs_revision` label still applicable here?

ghost:
good question.

PR #2, opened 2014-09-27
Ability to enable SSL when creating MongoDB users

philm:
Adapted from [here](http://docs.mongodb.org/manual/tutorial/configure-ssl-clients/#pymongo).

### DIFF
--- a/database/mongodb_user.py
+++ b/database/mongodb_user.py
@@ -68,6 +68,7 @@ options:
         required: false
         default: null
     ssl:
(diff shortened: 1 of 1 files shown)


mpdehaan:
Can you stick a "version_added: '1.8'" on this argument

We'll be merging these a lot faster, so this is basically good to go.

Thanks!

philm:
All set.

mpdehaan:
Merged, thanks!

PR #4, opened 2014-09-28
zabbix_maintenance: set state to be present by default

resmo:
state=present is the default in many modules.


mpdehaan:
Merged,thanks!

PR #5, opened 2014-09-28
zabbix_maintenance: option for using env vars to provide login data.

resmo:
If no login data is provided by arguments, the environment vars ZABBIX_LOGIN_USER, ZABBIX_LOGIN_PASSWORD, ZABBIX_SERVER_URL are used to login into Zabbix server.


mpdehaan:
I'm wondering if the comments about the env vars could be confusing as ansible normally doesn't load the user's environment?

PR #6, opened 2014-09-28
zabbix_group: new module

resmo:
Add or remove a host group to Zabbix.


cove:
@resmo any interest in combining efforts on a single zabbix_group module? It looks like we still both have PRs in for the same module.

PR #8, opened 2014-09-29
rename list to list_ to avoid shadowing in github_hooks module

bwhmather:
The definition was leaking into ansible.module_utils.basic and causing
type checking to fail when running module as script.  Not entirely clear
why this should be the case.


mpdehaan:
Minor, but code convention wise, I'd like to see this as "_list".

(It's because "list()" is a function in Python)

If you can update this, I'll gladly apply it.

Thanks!

mpdehaan:
closing this as included in linked ticket, #9

PR #9, opened 2014-09-29
make it possible to configure the content type of a github webhook

bwhmather:
(Depends on (and includes) https://github.com/ansible/ansible-modules-extras/pull/8 as I wasn't able to test it without fixing shadowing imports first)

Adds `content_type` argument to `github_hooks` module to allow user to choose between json and form encoding for requests from github.  Default is kept as json so it should not break backwards compatibility.

The patch changes the signature of `create` so that it no longer matches the other commands in the module.  I also haven't spent too much thought about argument ordering.  I'm happy to try to make things match if you are unhappy with the style.

### DIFF
--- a/source_control/github_hooks.py
+++ b/source_control/github_hooks.py
@@ -57,6 +57,12 @@ options:
     required: false
     default: 'yes'
     choices: ['yes', 'no']
+  content_type:
+    description:
+      - Content type to use for requests made to the webhook
+    required: false
+    default: 'json'
+    choices: ['json', 'form']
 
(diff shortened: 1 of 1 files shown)


mpdehaan:
See comments on previous ticket about list_ but does not appear to depend on the other as it includes it, so I'm going to close #8 and just track this one if it's ok.

bwhmather:
Done.
Have added a prefix to all of the action functions for consistency
(Citation needed but always thought convention was prefix for private, suffix for shadowing built-in)

PR #18, opened 2014-09-29
mkfs.xfs uses -f instead of -F flag

jravetch:
mkfs.xfs uses the -f instead of -F flag to force creating a filesystem on a devices that have an existing filesystem


mpdehaan:
Merged, thank you!

PR #19, opened 2014-09-29
Fix composer module checkmode

komuta:
Hi,

While writing and testing playbooks using the composer module, I ran into two issues:
- check mode didn't work, even if it seems supported
- the changed status detection test was not valid

I propose this simple pull request to address those issues.

I've rewritten the way options are parsed and handled to be more consistent with the way other modules work (gem for example), since it was this origin of the check mode problem. I hope it's more appropriate now.

Note: original request was https://github.com/ansible/ansible/pull/8304

### DIFF
--- a/packaging/language/composer.py
+++ b/packaging/language/composer.py
@@ -138,6 +138,7 @@ def main():
 
     if module.check_mode:
         options.add("--dry-run")
+        del module.params['CHECKMODE']
 
     # Get composer command with fallback to default  
     command = module.params['command']


ghost:
change detection fix was merged in https://github.com/ansible/ansible-modules-extras/pull/61 . This PR needs to be rebased now.

komuta:
Thanks, I'll rebase it. By the way, how are pull requests prioritized ? The request #61 has been created the previous week and introduce a patch which is already fixed in my request (which goes back to July).

ghost:
this one is labeled  P3, P1 is the most important.

techtonik:
I believe the PR #61 is just easier to review, because it addresses exactly one problem and explains the cause of that problem and the fix.

I also find that `composer` module is hackish, but reading up on this request, I'd reduce the initial description to:

```
The PR fixes broken `check` mode for composer module.

It is still suboptimal, because the root cause of the bug resides in the for loop which
treats each module parameter. Better way it to build the options for composer by
accessing explicitly each module parameters, as the gem module does. Should I redo
the code as in gem module?
```

komuta:
Well, the [original commit](https://github.com/komuta/ansible/commit/0abf643eb7bc756c5c61dcdee65320b3777d2290) was pretty much simple I think. I followed Michael's advice to refactor the checkmode detection to be more consistent with other modules, and then the diff became a little bit more complicated.

Anyway, I revised the initial description to better describe what i've done in both commit. I use the composer module in a lot of my playbooks, so I'm willing to help this code to be merge as soon as possible. Please contact me anything is missing.

techtonik:
Oh, much better now.

The perfectionist in me still says that even rereading this thread takes time, so I'd just drop the `change detection` change, which is already fixed, filled separate PR with fix for `check mode` and separate PR to `unify option parsing`. That will maximize the chances to make this reviewed ASAP.

komuta:
It's done. I've simplified the fix in this request, and created a dedicated request for the unification of composer options parsing (#94).

komuta:
You're right about the implementation detail: the way check mode is implemented in ansible, it is passed to the module as a **CHECKMODE** parameter. And the way the module was originally implemented, you just can't make this cleaner without taking away the module.params loop. So if you want to make it right at once, you get the complicated patch I submitted before, and if you want to keep it simple, you just have to take the CHECKMODE out of the params array, one way or the other. Following your advice, I tried to keep it simple, and put the cleaning code in another pull request.

As for the default options comment, I was not my code so just left it this way.

techtonik:
Now I see. I am not convinced that hack with CHECKMODE should be introduced. The non-WTF logic is to fetch known stuff from `module.params`. It can be explained that `module.params` contains this CHECKMODE variable and other parameters, so that's why there is a filtering goes on - with a link to documents or code that does it. That would be a perfect code for this PR.

Something like this:

```
mapping = """
prefer_source    --prefer-source
prefer_dist         --prefer-dist
no_dev              --no-dev
""".strip()
for line in mapping.splitlines():
  param, option = line.split()
  if module.params[param]:
      cmdline += ' ' + option
```

And still have a `--dry-run` check managed separately.

komuta:
Thanks.

PR #22, opened 2014-09-30
Add selinux_permissive module, to be able to manage permissive domain

mscherer:
Previously on the main repository ( https://github.com/ansible/ansible/pull/7697 )

### DIFF
--- a/system/selinux_permissive.py
+++ b/system/selinux_permissive.py
@@ -1,0 +1,130 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2015, Michael Scherer <misc@zarb.org>
+# inspired by code of github.com/dandiker/
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: selinux_permissive
+short_description: Change permissive domain in SELinux policy
+description:
+  - Add and remove domain from the list of permissive domain.
+version_added: "1.9"
+options:
+  domain:
+    description:
+        - "the domain that will be added or removed from the list of permissive domains"
+    required: true
+  permissive:
+    description:
+        - "indicate if the domain should or should not be set as permissive"
+    required: true
+    choices: [ 'True', 'False' ]
+  no_reload:
+    description:
+        - "automatically reload the policy after a change"
+        - "default is set to 'false' as that's what most people would want after changing one domain"
+        - "Note that this doesn't work on older version of the library (example EL 6), the module will silently ignore it in this case"
+    required: false
+    default: False
+    choices: [ 'True', 'False' ]
+  store:
+    description:
+      - "name of the SELinux policy store to use"
+    required: false
+    default: null
+notes:
+    - Requires a version of SELinux recent enough ( ie EL 6 or newer )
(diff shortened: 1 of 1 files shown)


groks:
What is the correct prefix for selinux-related modules in ansible: 'se' or 'selinux_'?

There are 'selinux' and 'seboolean' modules in the main index. This module is 'selinux_permissive' and there is a new pull request for 'seport'.

Someone should pick a prefix and then we can all stick to it.

mscherer:
I personnally prefer to be explicit, and I think se prefix is likely going to collide with something else. There is also no seport or sepermissive in selinux.  And prefix_ is the convention that seems to be used most of the time for module in ansible.

dankeder:
I named the module "seport" because there is already a module called "seboolean". I have no problem renaming it to "selinux_port", "semanage_port" or something else. However I would appreciate if the naming was consistent with existing modules.

system/selinux_permissive.py
```diff
@@ -0,0 +1,110 @@
...
+'''
+
+EXAMPLES = '''
+- selinux_permissive: name=httpd_t permissive=true
+'''
+try:
+    import seobject
+except ImportError:
+    print "failed=True msg='policycoreutils-python required for this module'"
+    sys.exit(1)
```
xen0l: Thanks for this module, I needed it myself! However, this module is missing "import sys" statement. The module will emit python traceback, when policycoreutils-python package is not installed.

mscherer:
@xen0l  sorry, for some reason, i totally miss your comment. i fixed it, thanks for noticing.

system/selinux_permissive.py
```diff
@@ -0,0 +1,117 @@
...
+version_added: "1.9"
+options:
+  domain:
+    description:
+        - "the domain that will be added or removed from the list of permissive domains"
+    required: true
+  permissive:
+    description:
+        - "indicate if the domain should or should not be set as permissive"
+    required: true
```
bcoca, line 37: since this is a boolean if so show the choices ['yes','no'] or ['True', 'False']

system/selinux_permissive.py
```diff
@@ -0,0 +1,117 @@
...
+  no_reload:
+    description:
+        - "automatically reload the policy after a change"
+        - "default is 'true' as that's what most people would want after changing one domain"
+        - "Note that this doesn't work on older version of the library (example EL 6), the module will silently ignore it in this case"
+    required: false
+    default: False
```
bcoca, line 45: description says default is tru, but then default says it is false, the same as the code, also add choices field here

system/selinux_permissive.py
```diff
@@ -0,0 +1,117 @@
...
+    store = module.params['store']
+    permissive = module.params['permissive']
+    domain = module.params['domain']
+    no_reload = module.params['no_reload']
+
+    if not HAVE_SEOBJECT:
+        module.fail_json(changed=False, msg="policycoreutils-python required for this module")
+
+    try:
+        permissive_domains = seobject.permissiveRecords(store)
```
bcoca, line 92: should we even try when store = ''?

system/selinux_permissive.py
```diff
@@ -0,0 +1,117 @@
...
+        if domain not in permissive_domains.get_all():
+            if not module.check_mode:
+                permissive_domains.add(domain)
+            changed = True
+    else:
+        if domain in permissive_domains.get_all():
+            if not module.check_mode:
+                permissive_domains.delete(domain)
+            changed = True
+
```
bcoca, line 121: I'm supposing the other calls to the seobject library can also produce exceptions, they should be handled gracefully.

gregdek:

gregdek:
@mscherer friendly ping on the outstanding issues that @bcoca pointed out.

system/selinux_permissive.py
mscherer, line 92: yes, the default value of '' mean that we take the default store:

```
>>> import seobject
>>> seobject.permissiveRecords('')
<seobject.permissiveRecords instance at 0x7f05eb8d9438>
```

system/selinux_permissive.py
mscherer, line 121: So set_reload is just setting a value, so I would consider it doesn't throw a exception. get_all/add/delete could fail, let me rewrite the code

mscherer:
So, I pushed the change, then found out that my repo was in a bad state ( and pushed too much stuff ), and so I finally cleaned and pushed it, taking (if I am not wrong) the remarks in account.

PR #23, opened 2014-09-30
Added module haproxy, to handle enable/disable backend server with shutdown_sessions feature in haproxy.

ravibhure:
Previously on the main repository (ansible/ansible#6889)


marcusramberg:
:+1:  This seems very useful.

jszwedko:
:+1: :+1: agreed.

bcoca:
can you move the file into the 'network' directory and add the version added 1.9?

ravibhure:
@bcoca :+1:

bcoca:
A few of comments:
- Instead of action [enable_server,disable_server] we prefer to use state [enabled,disabled], this is the standard we try to follow across modules.
- The module could always return the weight information and set it when specified, which removes the need for set/get_weight.
- An example not using inventory_hostname (with_items: groups['backend_www']?).
- no need to check if host is not set as the argument spec (#L138), it should already complain about that
- I think you misplaced the checkmode support (#L146)
- check mode implementation is returning always changed, this is not how it should work, it is better that it does not support it and is skipped over giving possibly incorrect information.

ravibhure:
Thanks @bcoca, I have done all of the fixes you suggested, 

but I am considering the one most important you said "The module could always return the weight information and set it when specified, which removes the need for set/get_weight." 
so this needs to be seperate action, as it is conditional check before taking any further action.

bcoca:
I'm not sure I've parsed that correctly, do you mean its 2 diff calls to haproxy? it could still be a single task in ansible. Also I've noticed you have not renamed action to state, but have renamed the options.

ravibhure:
Its been updated.

bcoca:
Nice, looks good functionally with the exception of check_mode, if you are not going to implement it remove the support flag, always returning changed=true is not an acceptable support of this option.

On line 143, there is an improper function call, it should have msg= for the string.

Some comments about the style:
- Normally the main method goes at the end of the module.
- Named arguments argument constructors are preferred over kwargs
- The constant ACTION_CHOICES should just be passed to the argument spec or moved to top level scope (left margin) as with RECV_SIZE
- Method comments should be descriptive and short, no need to copy the haproxy manpages
- many of the indents are not 4 spaces

ravibhure:
@bcoca, replicated changes, please review.

PR #24, opened 2014-09-30
Add Tidy module

rmarchei:
This module removes unwanted files based on specific criteria. Multiple criteria are AND’d together

### DIFF
--- a/files/tidy.py
+++ b/files/tidy.py
@@ -1,0 +1,236 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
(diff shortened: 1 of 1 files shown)


bcoca:
aside from correcting teh shebang to /usr/bin/python, the module looks good, but I'm thinking it is too specific, would it make sense to redo as a 'find' module? It can be used in more scenarios, to do the same thing as tidy, you would just need to feed the list to file: state=absent.

rmarchei:
Hi Brian,
the idea of the find module sounds great, I'll work on it!

bcoca:
closing in favor of find module:
https://github.com/ansible/ansible-modules-core/pull/1025

PR #28, opened 2014-09-30
Add dpkg_selections module

brian-brazil:
This is mainly useful for setting packages to 'hold' to prevent them
from being automatically upgraded.

Copied from https://github.com/ansible/ansible/pull/7233

### DIFF
--- a/packaging/dpkg_selections
+++ b/packaging/dpkg_selections
@@ -1,0 +1,60 @@
+#!/usr/bin/python
+
+DOCUMENTATION = '''
+---
+module: dpkg_selections
+short_description: Dpkg package selection selections
+description:
+    - Change dpkg package selection state via --get-selections and --set-selections.
+version_added: "2.0"
+author: Brian Brazil <brian.brazil@boxever.com>
+options:
+    name:
+        description:
+            - Name of the package
+        required: true
+    selection:
+        description:
+            - The selection state to set the package to.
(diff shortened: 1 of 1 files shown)


gregdek:

packaging/dpkg_selections
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/python
+
+DOCUMENTATION = '''
+---
+module: dpkg_selections
+short_description: Dpkg package selection states
+description:
+    - Change dpkg package state via --get-selections and --set-selections.
+version_added: "1.6"
```
bcoca: version 2.0

packaging/dpkg_selections
```diff
@@ -0,0 +1,60 @@
...
+author: Brian Brazil <brian.brazil@boxever.com>
+options:
+    name:
+        description:
+            - Name of the package
+        required: true
+    state:
```
bcoca: I would rename to selection to avoid confusing with the normal meaning of 'state' in ansible modules

packaging/dpkg_selections
brian-brazil: Updated

packaging/dpkg_selections
brian-brazil: Hmm, this does appear to be the normal use of `state` across almost all the ansible modules I've used and I can't find guidelines on what state is meant to mean - changed anyway.

PR #29, opened 2014-09-30
Add module system/seport.py

dankeder:
Module for managing SELinux network port type definitions.


dankeder:
I updated the PR, check_mode should work correctly now.

I also added the possibility to specify port range instead of a single port and added some docstrings.

gregdek:

PR #30, opened 2014-09-30
add ldap module

grossws:
When I tried to configure ldap with ansible, I have to do it in three phases: copy `ldif` file to target system, perform `ldapsearch` to check if entry exists and call to `ldapadd` or `ldapmodify` after that (conditionally). It's uncomfortable and error-prone. So I implemented ldap module using python ldap library, present in base repos (CentOS base in my case). 

Another advantage over calling `ldapmodify` (and other utils from `openldap-clients`) that module supports fully idempotent ldap operation and check mode. 

Supported operation modes are equal to `ldapadd`, `ldapmodify` and `ldapdelete` openldap utils but will not apply any operations if LDAP entry wouldn't change after these operations (like deleting non-existent attribute, replace attribute with values already present in it etc).

Supported features:
- connect using `ldap`, `ldaps` and `ldapi` protocols;
- _STARTTLS_ over `ldap` protocol;
- anonymous and authenticated operation (using `simple` auth);
- SASL _EXTERNAL_ auth method on `ldapi` protocol (usually used to configure OpenLDAP 2.4+ on-line configs in `cn=config` subtree);
- adding and deleting entries (via `state=present` and `state=absent`);
- adding, deleting and replacing individual attributes, attribute-value pairs or their combinations;
- `simple_args` and `complex_args` format for entry modification command;
- many examples to explain different args format and operation modes;
- simple DSL for one-line ldap entry modification command (with EBNF-like `simple_args` format definition);
- _UTF-8_ for attribute values;
- grouping (by attribute) and collapsing ldap modification operations to get true idempotent behavior;
- parsed and final ldap modifications (as they will be sent to ldap server) are logged to simplify module learing and operations debugging.

Target system requirement:
- python-ldap

It's rebased (with path and extension fixed) of PR ansible/ansible#7409.

### DIFF
--- a/system/ldap.py
+++ b/system/ldap.py
@@ -46,6 +46,13 @@ options:
         required: false
         default: null
         aliases: []
+    auth:
+        description:
+            - Authentication method. C(sasl_ext) is supported only on U(ldapi://) (unix socket connection)
+        required: false
+        aliases: []
+        choices: ["simple", "sasl_ext"]
+        default: "simple"
     dn:
         description:
             - Operation-specific distinguished name (dn)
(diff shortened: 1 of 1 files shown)


dankeder:
Please consider squashing the commits into one, six commits for adding a single file is overkill :-).

grossws:
Ok, I'll squash them with modifications that I want to add to avoid second PR with them. And update this PR in couple days.

gregdek:

grossws:
I'll recreate/reopen pull request when I have enough time to do some alterations to this module args before freezing its api. Now I'm closing this PR to remove it from list of open PRs.

PR #31, opened 2014-09-30
known_hosts: manage ssh known_hosts files with ansible.

mcv21:
The known_hosts module lets you add or remove a host from the
known_hosts file. This is useful if you're going to want to use the
git module over ssh, for example. If you have a very large number of
host keys to manage, you will find the template module more useful.

This was pull request 7840 from the old ansible repo, which was
accepted-in-principle but not yet merged. The mailing list thread
reading it is:

https://groups.google.com/forum/#!topic/ansible-devel/_e7H_VT6UJE/discussion

### DIFF
--- a/system/known_hosts.py
+++ b/system/known_hosts.py
@@ -1,0 +1,234 @@
+#!/usr/bin/env python
+
+"""
+Ansible module to manage the ssh known_hosts file.
+Copyright(c) 2014, Matthew Vernon <mcv21@cam.ac.uk>
+
+This module is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This module is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this module.  If not, see <http://www.gnu.org/licenses/>.
+"""
+
+DOCUMENTATION = '''
+---
+module: known_hosts
+short_description: Add or remove a host from the C(known_hosts) file
+description:
+   - The M(known_hosts) module lets you add or remove a host from the C(known_hosts) file. 
+     This is useful if you're going to want to use the M(git) module over ssh, for example. 
+     If you have a very large number of host keys to manage, you will find the M(template) module more useful.
+version_added: "1.6"
+options:
+  host:
+    description:
+      - The host to add or remove (must match a host specified in key)
+    required: true
+    default: null
+  key:
+    description:
+      - The SSH public host key, as a string (optional if removing the host)
+    required: true
+    default: null
+  path:
+    description:
+      - The known_hosts file to edit
+    required: no
+    default: "(homedir)+/.ssh/known_hosts"
+  state:
+    description:
+      - I(present) to add the host, I(absent) to remove it.
+    choices: [ "present", "absent" ]
+    required: no
+    default: present
+requirements: [ ]
+author: Matthew Vernon
+'''
+
+EXAMPLES = '''
+# Example using with_file to set the system known_hosts file
+- name: tell the host about our servers it might want to ssh to
+  known_hosts: path='/etc/ssh/ssh_known_hosts'
+               host='foo.com.invalid'
+               key="{{ lookup('file', 'pubkeys/foo.com.invalid') }}"
+'''
+
+# Makes sure public host keys are present or absent in the given known_hosts
+# file.
+#
+# Arguments
+# =========
+#    host = hostname whose key should be added
+#    key = line(s) to add to known_hosts file
+#    path = the known_hosts file to edit (default: ~/.ssh/known_hosts)
+#    state = absent|present (default: present)
+
+import os,os.path,tempfile,errno
+
+def enforce_state(module, params):
+    """
+    Add or remove key.
+    """
+
+    host = params["host"]
+    key = params.get("key",None)
+    port = params.get("port",None)
+    #expand the path parameter; otherwise module.add_path_info
+    #(called by exit_json) unhelpfully says the unexpanded path is absent.
+    params["path"]=os.path.expanduser(params.get("path"))
+    path = params.get("path")
+    state = params.get("state","present")
+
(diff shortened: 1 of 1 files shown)


rrtj3:
Looking over this module since I have a need for such a module. One question I have is if the 'key' is really required when removing a host from the known_hosts file? (I would prefer not to have to list the key when removing a host.)

mcv21:
Thanks for your comment, and sorry for taking so long to get back to you!

I went by analogy with the authorized_key module in specifying the key to remove; if you think that it would be substantially better to not bother specifying the key when doing "remove", I could be persuaded, but I thought the current interface was neater...?

mcv21:
On further reflection, I think you're right, so I've adjusted the interface accordingly - the key is optional if you're removing a host.

system/known_hosts.py
```diff
@@ -0,0 +1,234 @@
+#!/usr/bin/env python
```
bcoca: shebang should be /usr/bin/python

system/known_hosts.py
```diff
@@ -0,0 +1,234 @@
...
+options:
+  host:
+    description:
+      - The host to add or remove (must match a host specified in key)
+    required: true
+    default: null
+  key:
+    description:
+      - The SSH public host key, as a string (optional if removing the host)
+    required: true
```
bcoca: this should be false

system/known_hosts.py
```diff
@@ -0,0 +1,234 @@
...
+version_added: "1.6"
+options:
+  host:
+    description:
+      - The host to add or remove (must match a host specified in key)
+    required: true
+    default: null
+  key:
+    description:
+      - The SSH public host key, as a string (optional if removing the host)
```
bcoca: change to: required when state=present, optional when state=absent in which case all keys for a host will be removed

system/known_hosts.py
```diff
@@ -0,0 +1,234 @@
...
+---
+module: known_hosts
+short_description: Add or remove a host from the C(known_hosts) file
+description:
+   - The M(known_hosts) module lets you add or remove a host from the C(known_hosts) file. 
+     This is useful if you're going to want to use the M(git) module over ssh, for example. 
+     If you have a very large number of host keys to manage, you will find the M(template) module more useful.
+version_added: "1.6"
+options:
+  host:
```
bcoca: we might want to rename this to name= to keep consistency, host= can be an alias

system/known_hosts.py
```diff
@@ -0,0 +1,234 @@
...
+
+DOCUMENTATION = '''
+---
+module: known_hosts
+short_description: Add or remove a host from the C(known_hosts) file
+description:
+   - The M(known_hosts) module lets you add or remove a host from the C(known_hosts) file. 
+     This is useful if you're going to want to use the M(git) module over ssh, for example. 
+     If you have a very large number of host keys to manage, you will find the M(template) module more useful.
+version_added: "1.6"
```
bcoca, line 29: - note to update on merge to appropriate

system/known_hosts.py
```diff
@@ -0,0 +1,234 @@
...
+author: Matthew Vernon
+'''
+
+EXAMPLES = '''
+# Example using with_file to set the system known_hosts file
+- name: tell the host about our servers it might want to ssh to
+  known_hosts: path='/etc/ssh/ssh_known_hosts'
+               host='foo.com.invalid'
+               key="{{ lookup('file', 'pubkeys/foo.com.invalid') }}"
+'''
```
bcoca, line 63: it would be nice to add example using pipe lookup to discover the ssh key

system/known_hosts.py
```diff
@@ -0,0 +1,234 @@
...
+# file.
+#
+# Arguments
+# =========
+#    host = hostname whose key should be added
+#    key = line(s) to add to known_hosts file
+#    path = the known_hosts file to edit (default: ~/.ssh/known_hosts)
+#    state = absent|present (default: present)
+
+import os,os.path,tempfile,errno
```
bcoca: we prefer an import per line

system/known_hosts.py
```diff
@@ -0,0 +1,234 @@
...
+            return True, False #current, not-replace
+    #No match found, return current and replace
+    return True, True
+
+def main():
+
+    module = AnsibleModule(
+        argument_spec = dict(
+            host      = dict(required=True,  type='str'),
+            key       = dict(required=True,  type='str'),
```
bcoca: should not be required as per docs

system/known_hosts.py
```diff
@@ -0,0 +1,234 @@
...
+    Add or remove key.
+    """
+
+    host = params["host"]
+    key = params.get("key",None)
+    port = params.get("port",None)
+    #expand the path parameter; otherwise module.add_path_info
+    #(called by exit_json) unhelpfully says the unexpanded path is absent.
+    params["path"]=os.path.expanduser(params.get("path"))
+    path = params.get("path")
```
bcoca: simpler to collapse these 2 lines:
path=os.path.expanduser(params.get("path"))

system/known_hosts.py
```diff
@@ -0,0 +1,234 @@
...
+    state = params.get("state","present")
```
bcoca: no need to default here again, taken care of in module definition

system/known_hosts.py
```diff
@@ -0,0 +1,234 @@
...
+    #The approach is to write the key to a temporary file,
+    #and then attempt to look up the specified host in that file.
+    try:
+        outf=tempfile.NamedTemporaryFile(delete=False)
+        outf.write(key)
+        outf.close()
+    except IOError,e:
+        module.fail_json(msg="Failed to write to temporary file %s: %s" % \
+                             (outf.name,str(e)))
+    rc,stdout,stderr=module.run_command(['ssh-keygen','-F',host,
```
bcoca: use modules find bin utility to locate proper ssh-keygen as PATH might not be set or correct
Why not try ssh-keyscan against the host?

mcv21: Thanks for these comments, which I'll get to shortly. Would you like a commit on top of this commit (for a 2-commit PR) or for me to incorporate changes and still leave a single commit here?

To answer your second question, because ssh-keyscan is in general insecure (it doesn't do any host validation).

mcv21:
@bcoca -  I've made your code changes; are you happy with this module now?
@abadger - you removed this from the next milestone?

@bcoca thanks for the feedback, BTW. If you want me to squash these two commits into 1, just say. I left added_in alone for now; again, if you want me to change that to 1.9, just ask.

PR #37, opened 2014-10-01
win_chocolatey: Windows package management with Chocolatey

petemounce:
Ported from ansible/ansible#8981.


ibratoev:
Can you consider the ability to use your own custom choco repo? Maybe the source parameter can accept an url in this case?

PR #38, opened 2014-10-01
win_updates: Add module to install windows updates

petemounce:
Supports specifying which category to install all updates from.
**NEEDS:** `choco install PSWindowsUpdate` beforehand - does not attempt to install it (or detect it's missing, yet). Can be done via #37.

PSWindowsUpdate: http://gallery.technet.microsoft.com/scriptcenter/2d191bcd-3308-4edd-9de2-88dff796b0bc#content

### DIFF
--- a/windows/win_updates.ps1
+++ b/windows/win_updates.ps1
@@ -1,0 +1,86 @@
+#!powershell
+# This file is part of Ansible
+#
+# Copyright 2014, Trond Hindenes <trond@hindenes.com>
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


petemounce:
@cchurch I noticed while working with this that the `win_reboot` module that you wrote (ansible/ansible#8946) that it disagreed with PSWindowsUpdate's `Get-WURebootStatus` cmdlet for figuring out whether a reboot is needed.

(I found otherwise that `win_reboot` worked a treat, so thanks for that!)

cchurch:
@petemounce Thanks, I'll look at that.  I'd also wanted to address the comments from @trondhindenes and @wozz before re-submitting that one.

PR #42, opened 2014-10-02
adding username/password support along with ssl/starttls to mail

bhcopeland:
Original pull request from here https://github.com/ansible/ansible/pull/7213

The mail module lacks the ability to have SMTP username and password. Also, due to sending credentials support has been added for SSL/TLS and STARTTLS.

Would be really good to get this merged.

### DIFF
--- a/notification/mail.py
+++ b/notification/mail.py
@@ -189,9 +189,13 @@ def main():
         body = subject
 
     try:
-        smtp = smtplib.SMTP_SSL(host, port=int(port))
-    except (smtplib.SMTPException, ssl.SSLError):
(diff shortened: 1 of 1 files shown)


bhcopeland:
will this be merged?

bcoca:
yes, but could you modify it so we still handle the exception if non ssl (line 177) connection fails? Your patch removed this in favor of handling ssl failure exception.

danpilch:
Any update on this? I could really appreciate having this functionality!

bhcopeland:
I can merge changes from #271. If needs be. Tested it and outputted okay.

danpilch:
verified as working.

PR #51, opened 2014-10-03
Add missing __init__.py files for find_package() to work

brenix:
When running `setup.py`, the `packaging` and `source_control` directories are not included with the `find_package()` function due to missing `__init__.py` files. This leaves the installation incomplete, therefore the modules in these directories will not be available.


jimi-c:
Merged, thanks!

PR #54, opened 2014-10-05
Created a module to manage logrotate config files

retr0h:


retr0h:
@jimi-c any chance this will get merged?  It has been quite some time.

bcoca:
First of all, sorry for making you wait. Sadly we have a huge backlog, we are working through it faster now but it has caused delays to many contributions.

That said, we are not going to merge this into ansible, we feel that it is too application specific and that using the template module will easily provide this functionality without adding more code or complexity.

retr0h:
Thanks for getting back to me @bcoca.  Could you explain how it is too application specific?  Many applications want to utilize logrotate.  Sure, it could be a template, but seems like we would want this type of logic in ansible.  I talked with @mpdehaan months ago, and he felt it should be added as well (granted he has moved on now).

I do understand what you are saying, and am on the fence really.  This way just seemed a little cleaner from a usage perspective, but maybe it over complicates things.  If one were to want to setup newsyslog(8), I would agree more about it being in a template, but for a system tool which is common to most linux, seems like this would be useful.

bcoca:
I discussed this in the core team, and we find that a role that installs/configures logrotate would be a better way to manage this and share with others. 

The logrotate format is not that complex (IMHO) that I see the need to add a module specifically to manage this. Also note that module itself requires this knowledge as you pass the specific options in a freeform list. You are also missing ownership and permissions for the file, which already covered by the template module.

retr0h:
@bcoca Sure, makes a lot of sense.  Thanks for getting back to me.

PR #55, opened 2014-10-06
Make sure extras/packaging modules are included in builds

mtr:
A fresh install broke our current Ansible scripts because the homebrew module had been left out.  Adding this **init**.py ensures that the extras/packaging module is included in builds.


sivel:
Duplicate of #51

However #51 also adds it for multiple missing locations.

mtr:
I am sorry.  I didn't see #51.  My bad.

PR #58, opened 2014-10-06
Added Docker Facts module: issue #57

### LINKED ISSUE #57: RFE: Docker Facts module
#### ISSUE TYPE:
- Feature Idea
#### COMPONENT NAME:
- docker_facts
#### ANSIBLE VERSION:

1.8
#### ENVIRONMENT:

Ubuntu 13.04
#### SUMMARY:

Docker facts module. This module returns facts pertaining to a running fleet of containers as well as images.

It would be useful to have a docker_facts module to return information about running containers as well I found an issue with the docker module that I ended up discussing with Paul Durivage.  The essential problem is that there is a list of dictionaries per container that only contains the last running container when using "with_sequence" or "with_items" versus "count", hence making it impossible to utilize information for the entirety of a running fleet of docker containers outside of a task using the docker module.

An idea to deal with this short-coming would be to have a "docker_facts" module for obtaining information from the fleet of Docker containers.
#### STEPS TO REPRODUCE:

N/A
#### EXPECTED RESULTS:

Dictionary containing two primary entries for containers and images
#### ACTUAL RESULTS:

N/A

CaptTofu:
Added docker_facts (fixes #57)

### DIFF
--- a/cloud/docker_facts.py
+++ b/cloud/docker_facts.py
@@ -1,0 +1,246 @@
+#!/usr/bin/python
+
+# vim: ts=4:expandtab:au BufWritePost
+# (c) 2014, Patrick "CaptTofu" Galbraith <patg@patg.net>
+# Code also from rax_facts and the primary docker module
+#
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+# This is a DOCUMENTATION stub specific to this module, it extends
+# a documentation fragment located in ansible.utils.module_docs_fragments
+DOCUMENTATION = '''
+---
+module: docker_facts
+short_description: Gather facts for Docker containers
+description:
+     - Gather facts for Docker containers and images
+version_added: "0.1"
+options:
+  id:
+    description:
+      - container ID to retrieve facts for
+    default: null (all containers if neither name nor id specified)
+  name:
+    description:
+      - container name to retrieve facts for
+    default: null (all containers if neither name nor id specified)
+  images:
+    description:
+      - image name or 'all'. Off by default.
+    default: null
+author: Patrick Galbraith
+'''
+
+EXAMPLES = '''
+- name: Gather info about containers
+  hosts: localhost
+  gather_facts: False
+  tasks:
+    - name: Get facts about containers
+      local_action:
+        module: docker_facts
+        name: container1
+
+- name: Gather info about all containers and images
+  hosts: docker
+  gather_facts: True
+  tasks:
+    - name: Get facts about containers
+      local_action:
+        module: docker_facts
+        images: all
+
+    - name: containers debug info
+      debug: msg="Container Name {{ item.key }} IP Address {{ item.value.docker_networksettings.IPAddress }}"
+      with_dict: docker_containers
+
+    - name: images info
+      debug: msg="Image ID {{ item.key }} Author {{ item.value.docker_author }} Repo Tags {{ item.value.docker_repotags }}"
+      with_dict: docker_images
+
+'''
+
+HAS_DOCKER_PY = True
+
+import re
+from urlparse import urlparse
+try:
+    import docker
+    try:
+        from docker.errors import APIError as DockerAPIError
+    except ImportError:
+        from docker.client import APIError as DockerAPIError
+except ImportError:
+    HAS_DOCKER_PY = False
+
+
+class DockerManager:
+
+    def __init__(self, module):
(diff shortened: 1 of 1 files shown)


daviddyball:
I just wanted to add that I'm currently using @CaptTofu docker_facts module and really like the way it uses a dict for `docker_containers`. It makes it much easier to target specific information without having to resort to `with_items` loops in playbooks.

bcoca:
doesn't this make more sense as part of the docker inventory script?

daviddyball:
I could be wrong, but my understanding was that inventory doesn't update after the play has started. What if `task-1` launches 3 containers and `task-5` needs to know about them? Knowing their state by adding a `docker_facts` task and registering the results seems logical to me.

CaptTofu:
It’s good to have facts to be able to use in your playbooks. Also, we at HP ATG found that we prefer to use facts modules to write static inventory files (templates) so that we don’t have to rely on accessing the Docker daemon (or Nova API) every time we need to run a play against our instances or containers.

gregdek:

gregdek:

# 

gregdek:
ping to @daviddyball for potential review here under the new process

daviddyball:
@CaptTofu @gregdek 

The script has been working for me in all but one case for about 6 months now. The case where it fails is if the container list (equivalent of `docker ps -a`) contains a container which is `Dead`. It seems to fail to pop the `Name` attribute and the script dies. I'm still trying to reproduce it reliably, but I see it frequent enough that I think it could be a problem.

I'll report back if I find anything.

cloud/docker_facts.py
```diff
@@ -0,0 +1,246 @@
...
+        entity_dict = dict()
+        for entity in entities:
+            id = entity.get('Id')
+            short_id = id[:13]
+            name = short_id
+
+            if type == 'containers':
+                try:
+                    name = entity.get('Names', list()).pop(0).lstrip('/')
+                except IndexError:
```
daviddyball, line 180: Just add `AttributeError` so that if the resulting `list()` object is defaulted to, then we catch `.pop(0)` correctly and then set `name = short_name`.

chrismeyersfsu:
@daviddyball Does the argument of "inventory doesn't update after the play has started" still hold with the addition of `refresh_inventory` in Ansible 2.0 ?

cloud/docker_facts.py
CaptTofu, line 180: Hi guys!

Very very sorry for being AWOL - I had my head deeply in getting a code contrib into Kubernetes, plus a back-injury that made it hard to think of anything.

I would LOVE to get docker_facts into Ansible modules extra.

I need to read through these mails— also, would you guys want to do a quick hangout to figure out what I need to do to get this in?

Regards,

Patrick

daviddyball:
@chrismeyersfsu not 100% sure. I haven't been running `develop` or `2.0` branch so couldn't say. My production environment is still on `1.9.x`

chrismeyersfsu:
@CaptTofu With respect to @bcoca suggestion. Does this make more sense to be a docker inventory script with Ansible 2.0 `refresh_inventory` feature in mind?

CaptTofu:
> On Oct 26, 2015, at 11:00 AM, David Dyball notifications@github.com wrote:
> 
> @chrismeyersfsu https://github.com/chrismeyersfsu not 100% sure. I haven't been running develop or 2.0 branch so couldn't say. My production environment is still on 1.9.x
> 
> —
> Reply to this email directly or view it on GitHub https://github.com/ansible/ansible-modules-extras/pull/58#issuecomment-151165763.
> 
> Mail app somehow lost the last email in this chain— the question about this being an inventory script, if I understand the question correctly — the reason I wrote this is for more than just needing inventory. 

Here’s a use case— one that I showed a friend how to use recently: a setup where I was building out an IPtables setup to set up static IP addresses to map to the private IP addresses and forward ports. With facts, he was able loop through all the running containers, obtain the IP, and print out an “iptables restore” type file that allowed him to do this.

There are other cases as well— namely, those where I want to to be able to iterate and write files or do certain steps with that information where this is useful. Hence why I wrote it :)

tima:
@CaptTofu Your use cases aren’t anything you couldn’t do thru an inventory script. Facts and inventory variables are folded together under "magic" hostlers variable[1]. Inventory variables have the benefit of inheritance that could come in handy depending on nature of the data you’re working with. 

[1] http://docs.ansible.com/ansible/playbooks_variables.html#magic-variables-and-how-to-access-information-about-other-hosts

jcassee:
A use-case that I intend to use this module for: stopping all containers that are not explicitly started from Ansible.

tima:
@jcassee: What started a container won't matter to Ansible if the dynamic inventory script has been written properly.

jcassee:
@tima Not completely sure what you mean. When I provision a host and Ansible has a list of all containers that should be running, how do I make sure that running containers that are not on that list are stopped? (It is not my intention to use thing issue as a discussion platform, so maybe we should move this to [ansible-project](https://groups.google.com/forum/#!forum/ansible-project).)

CaptTofu:
That sounds intesting, though using docker_facts is incredibly simple for me and doesn’t require any invocation the an inventory plugin would. There were some other use-cases as well that I’d have to pull out. I wrote this module due to a suggest by Paul Durivage and Michael Dehaan originally and they had some ideas as well.

In any case, if you guys want it, I’ll be glad to clean it up, do whatever, to get it in and help. Otherwise, I’ll maintain it separately and make available to anyone to use it. That’s the beauty of Ansible— you can write whatever you need to in order to get something done.

Kind regards,

Patrick

PR #59, opened 2014-10-07
Allow disabling of autorefresh for zypper repositories

jlec:
In case of release repositories or other special cases you might not
need the autorefreshing of the repos. This patch adds a configure
option instead of hard enabling this.

Signed-off-by: Justin Lecher jlec@gentoo.org


jlec:
This PR is ....

PR #61, opened 2014-10-09
composer: Fix `changed` status that always returns False

techtonik:
re.match in has_changed function never worked properly, because match
requires searched sequence to be present exactly at a start of processed
string, which is not the case here:

```
 "msg": "Loading composer repositories with package information Installing dependencies from lock file Nothing to install or update Generating optimized autoload files"
```


mpdehaan:
Merged, thanks!

PR #62, opened 2014-10-10
Make lvg work when a list of pvs is used

gpocentek:
Use the dev_list variable to generate the vgcreate command, otherwise
the list of pvs is used as a unique argument, and the vgcreate command
fails.


gpocentek:
This is actually the same patch as https://github.com/ansible/ansible-modules-extras/pull/15 , no need for duplicates.

PR #63, opened 2014-10-10
fixed zypper.py and zypper_repository.py for SLE10 platform

bmanojlovic:
SUSE Enterprise Linux 10 use python 2.4 and zypper command does not accept all parameters as newer SUSE distributions, this patch set fixes that.


mpdehaan:
merged, thanks!

PR #66, opened 2014-10-12
Enable "ufw default allow routed"

FauxFaux:
- The policy is shown in `status verbose`, so all the check mode stuff should keep working.
- `--dry-run` works as expected.
- No idea whether it's legal as an argument to `interface`


mpdehaan:
Merged, thanks!

PR #75, opened 2014-10-17
Clarify doc for cpan module

offby1:
... note that you can pass names that include a precise version number.


jimi-c:
Merged, thank you!

PR #77, opened 2014-10-17
Add enabled/disabled support to bigip_node

omgjlk:
This allows one to enable or disable a node, useful for when doing
maintenance on a node to prevent connections from being attempted to it.

This will completely disable the node for any pool it might be in.


bcoca:
this needs a rebase

omgjlk:
Rebased.

PR #87, opened 2014-10-21
[lvol] Add opts parameter

DavidWittman:
Adds the ability to set options to be passed to the lvcreate command
using the `opts` parameter.

### DIFF
--- a/system/lvol.py
+++ b/system/lvol.py
@@ -57,6 +57,10 @@ options:
     - Shrink or remove operations of volumes requires this switch. Ensures that
       that filesystems get never corrupted/destroyed by mistake.
     required: false
+  opts:
+    version_added: "1.9"
+    description:
+    - Free-form options to be passed to the lvcreate command
 notes:
   - Filesystems on top of the volume are not resized.
 '''
@@ -70,6 +74,9 @@ EXAMPLES = '''
 
 # Create a logical volume the size of all remaining space in the volume group
 - lvol: vg=firefly lv=test size=100%FREE
+
+# Create a logical volume with special options
+- lvol: vg=firefly lv=test size=512g opts="-r 16"
 
 # Extend the logical volume to 1024m.
 - lvol: vg=firefly lv=test size=1024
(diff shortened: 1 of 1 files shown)


jimi-c:
Hi @DavidWittman, could you please add a `version_added: "1.8"` field to the new param? If you have any questions about the format, just look at one of the other modules to get an idea of where to position it in the docstring, there are plenty of examples of it in just about every other module.

Thanks!

DavidWittman:
@jimi-c Definitely. I wasn't sure how those were added. Commit inbound.

DavidWittman:
Alright, I just amended/force-pushed my previous commit. Thanks @jimi-c.

jimi-c:
Thanks!

gregdek:
@DavidWittman Sorry, looks like this needs to be rebased. When you've done that, please ping module owner @jhoekx -- if he gives +1 to the rebased PR, we will merge.

DavidWittman:
Thanks @gregdek. I rebased, fixed the version added, and improved the documentation a bit.

@jhoekx Could you please review this PR?

system/lvol.py
```diff
@@ -212,7 +221,11 @@ def main():
                 changed = True
             else:
                 lvcreate_cmd = module.get_bin_path("lvcreate", required=True)
-                rc, _, err = module.run_command("%s %s -n %s -%s %s%s %s" % (lvcreate_cmd, yesopt, lv, size_opt, size, size_unit, vg))
+                if opts is None:
```
jhoekx: This doesn't really scale to multiple parameters. Other parameters have an earlier test to check if they should be used, otherwise they're just the empty string.

Take a look at `yesopt` that was added today for how I would prefer it.

jhoekx:
One question for @gregdek : would this still go in a 1.9 dot release? If not, the version in the docs should be incremented to 2.0 I guess.

system/lvol.py
DavidWittman: Yeah, I wasn't really in love with that as I was fixing the merge conflict. I'll fix that up real quick.

DavidWittman:
Thanks for the review @jhoekx.... Just updated my commit per your suggestions. I can amend the version per @gregdek's instruction as well.

DavidWittman:
@gregdek Do I need to change the version added to 2.0?

gregdek:
@DavidWittman I believe so but let me verify. @bcoca opinion here?

bcoca:
i just corrected after merge, but yes, normally it would be nice to have

PR #89, opened 2014-10-24
Introduce a download_artifact plugin to resolve maven artifacts

ghost:
This plugin will allow ansible users to resolve a Maven artifact from either maven central or a custom repository and download it to a target location on the ansible managed system


chrisisbeef:
I had multiple github accounts and merged them into a single account - apologies for the ghost PR

mpdehaan:
Possibly should be "maven_artifact" or maybe just "maven" ?

chrisisbeef:
I believe there is already something at the "maven" coordinate - maven-artifact makes sense however. I can change this today.

chrisisbeef:
Updated module name - thanks for consideration

bcoca:
to be closer to the ansible convention I would use dest= instead of target=

chrisisbeef:
Changed "target" to "dest"

packaging/maven_artifact.py
```diff
@@ -0,0 +1,376 @@
...
+    downloader = MavenDownloader(repository_url, repository_username, repository_password)
+
+    try:
+        artifact = Artifact(group_id, artifact_id, version, classifier, extension)
+    except ValueError as e:
+        module.fail_json(msg=e.args[0])
+
+    prev_state = "absent"
+    if os.path.isdir(dest):
+        dest = dest + "/" + artifact_id + "-" + version + ".jar"
```
krotscheck: If a specific version is not provided, this path creates a runtime error since version = NoneType. Also, the requested extension is not taken into account.

PR #90, opened 2014-10-24
add cloudtrail module

tedder:
Cloudtrail is the AWS auditing configuration. It's fairly simple, but also very important to configuration management/devops/security to ensure it remains enabled. That's why I created it as a module.

Supports check_mode and all that neat stuff. The only thing I'm unsure on is generating the documentation- I don't know how to do that.

### DIFF
--- a/cloud/cloudtrail.py
+++ b/cloud/cloudtrail.py
@@ -1,0 +1,227 @@
+#!/usr/bin/python
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = """
+---
+module: cloudtrail
+short_description: manage CloudTrail creation and deletion
+description:
+  - Creates or deletes CloudTrail configuration. Ensures logging is also enabled. This module has a dependency on python-boto >= 2.21.
+version_added: "1.7.3"
+author: Ted Timmons
+requirements: ["boto"]
+options:
+  state:
+    description:
+      - add or remove CloudTrail configuration.
+    required: true
+    choices: ['enabled', 'absent']
+  name:
+    description:
+      - name for given CloudTrail configuration.
+      - This is a primary key and is used to identify the configuration.
+  s3_bucket_prefix:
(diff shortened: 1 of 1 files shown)


ada-u:
+1

cloud/cloudtrail.py
```diff
@@ -0,0 +1,227 @@
...
+import os
+from collections import Counter
+
+try:
+    import boto
+    import boto.cloudtrail
+    from boto.regioninfo import RegionInfo
+except ImportError:
+    print "failed=True msg='boto required for this module'"
+    sys.exit(1)
```
bcoca: do not throw exception nor use print nor sys.exit, check other modules and see how they set a variable here and then use module.fail_json to return meaningful error messages.

cloud/cloudtrail.py
```diff
@@ -0,0 +1,227 @@
...
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = """
+---
+module: cloudtrail
+short_description: manage CloudTrail creation and deletion
+description:
+  - Creates or deletes CloudTrail configuration. Ensures logging is also enabled. This module has a dependency on python-boto >= 2.21.
+version_added: "1.7.3"
```
bcoca: update to 2.0

cloud/cloudtrail.py
```diff
@@ -0,0 +1,227 @@
...
+author: Ted Timmons
+requirements: ["boto"]
+options:
+  state:
+    description:
+      - add or remove CloudTrail configuration.
+    required: true
+    choices: ['enabled', 'absent']
```
bcoca: make the either enabled/disabled or present/abset (or all 4) for consistency.

cloud/cloudtrail.py
```diff
@@ -0,0 +1,227 @@
...
+        include_global_events={'default':True, 'required': False, 'type': 'bool' },
+    ))
+
+    module = AnsibleModule(argument_spec=argument_spec, supports_check_mode=True)
+    ec2_url, access_key, secret_key, region = get_ec2_creds(module)
+    aws_connect_params = dict(aws_access_key_id=access_key,
+                              aws_secret_access_key=secret_key)
+
+    if module.params['state'] == 'enabled' and not module.params['s3_bucket_name']:
+        module.fail_json(msg="s3_bucket_name must be specified as a parameter when creating a cloudtrail")
```
bcoca: use module's "required together" function instead

tedder:
rebasing broke @bocca's comments on the pull request, my bad. I've made the updates though. Should this be in `/cloud/` or in `/cloud/amazon/`? The latter doesn't exist in -extras, but I'm making an assumption.

bcoca:
i missed that, yes, cloud/amazon would be the correct place

bocca:
Sorry mates, but you misstyped bcoca as bocca.Un saludoPablo Romero-------- Mensaje original --------Asunto: Re: [ansible-modules-extras] add cloudtrail module (#90)De: Brian Coca Para: ansible/ansible-modules-extras CC: bocca Merged #90.

PR #92, opened 2014-10-26
Add new set of OpenStack modules

emonty:
The new modules do all of the keystone v3 and are using common library
code to be consistent. These are aimed at consumers of the cloud -
subsequent os_ modules for cloud admin may have project names ... like
os_keystone_tenant, for example - with the reasoning being that an
end user of a cloud should not need to know service code names, but a
cloud administrator certainly knows what's running.


cloud/os_image.py
```diff
@@ -0,0 +1,187 @@
...
+try:
+    import shade
+except ImportError:
+    print("failed=True msg='shade is required for this module'")
+
+
+DOCUMENTATION = '''
+---
+module: os_image
+short_description: Add/Delete images from OpenStack Cloud
```
tnoff: This is already a module: glance_image. Why are you creating a new one?

cloud/os_keypair.py
```diff
@@ -0,0 +1,107 @@
...
+try:
+    import shade
+except ImportError:
+    print("failed=True msg='shade is required for this module'")
+
+
+DOCUMENTATION = '''
+---
+module: os_keypair
+short_description: Add/Delete a keypair from OpenStack
```
tnoff: This is already a module, nova_keypair, why are you making a new one?

cloud/os_network.py
```diff
@@ -0,0 +1,220 @@
...
+
+try:
+    import shade
+except ImportError:
+    print("failed=True msg='shade is required for this module'")
+
+DOCUMENTATION = '''
+---
+module: os_network
+short_description: Creates/Removes networks from OpenStack
```
tnoff: Again, already a module, quatum_network

cloud/os_router.py
```diff
@@ -0,0 +1,124 @@
...
+try:
+    import shade
+except ImportError:
+    print("failed=True msg='shade is required for this module'")
+
+
+DOCUMENTATION = '''
+---
+module: os_router
+short_description: Create or Delete routers from PpenStack
```
tnoff: quantum_router already exists

cloud/os_compute.py
```diff
@@ -0,0 +1,396 @@
...
+                if ip not in module.params['floating_ips']:
+                    extra_ips.append(ip)
+            if extra_ips:
+                _delete_floating_ip_list(cloud, server, extra_ips)
+                changed = True
+    return (changed, server)
+
+
+def _get_server_state(module, cloud):
+    server = cloud.get_server_by_name(module.params['name'])
```
omgjlk: If I'm reading this right, idempotence is only on the name of the server, not any other associated data such as metadata or image. Is that intended? In the rax module idempotence was more fine grained, but that had a lot to do with internal needs.

emonty: It would actually be great to get to get a richer story on idempotence. The server-name key is mainly holder from the old code from the refactor. I'll go look at the rax module to see what the logic is.

omgjlk: I think the good story is to provide idempotence on the combination of:
- name
- flavor
- image
- metadata
- nics

Debatable as to whether or not to throw floating_ip in the mix too.

cloud/os_compute_facts.py
```diff
@@ -0,0 +1,90 @@
...
+    import shade
+    from shade import meta
+except ImportError:
+    print("failed=True msg='shade is required for this module'")
+
+DOCUMENTATION = '''
+---
+module: os_compute_facts
+short_description: Retrieve facts about a compute instance
+extends_documentation_fragment: openstack
```
omgjlk: Should mention in here somewhere that the facts will be returned in the variable "openstack".

cloud/os_compute_floating_ip.py
```diff
@@ -0,0 +1,156 @@
...
+   - Associates or disassociates a specific floating IP with a particular instance
+options:
+   state:
+     description:
+        - indicates the desired state of the resource
+     choices: ['present', 'absent']
+     default: present
+   instance_name:
+     description:
+        - name of the instance to which the public IP should be assigned
```
omgjlk: Since many instances can be named the same thing, perhaps it's better to have an option for instance_id, or to add such an option for users who encounter scenarios where instances have the same name.

cloud/os_compute_facts.py
```diff
@@ -0,0 +1,90 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this software.  If not, see <http://www.gnu.org/licenses/>.
+
+try:
+    import shade
+    from shade import meta
+except ImportError:
+    print("failed=True msg='shade is required for this module'")
```
sivel: "Baby JSON" no longer exists in ansible so this print will not work correctly. The correct pattern should be something like:

```
try:
    import something
    HAS_SOMETHING = True
except:
    HAS_SOMETHING = False

# Some other stuff

def main():
    module = ...
    if not HAS_SOMETHING:
        module.fail_json(msg='missing something python module')
```

cloud/os_compute_floating_ip.py
```diff
@@ -0,0 +1,156 @@
...
+
+    try:
+        cloud = shade.openstack_cloud(**module.params)
+        nova = cloud.nova_client
+        neutron = cloud.neutron_client
+
+        state, floating_ip_id = _get_floating_ip_id(module, neutron)
+        if module.params['state'] == 'present':
+            if state == 'attached':
+                module.exit_json(changed = False, result = 'attached', public_ip=module.params['ip_address'])
```
omgjlk: This could be a false positive. You check if the ip is attached, but not that it's attached to the desired instance. If the ip is already attached to a different instance, this should be an error.

cloud/os_compute_floating_ip.py
```diff
@@ -0,0 +1,156 @@
...
+   instance_name:
+     description:
+        - name of the instance to which the public IP should be assigned
+     required: true
+     default: None
+   ip_address:
+     description:
+        - floating ip that should be assigned to the instance
+     required: true
+     default: None
```
omgjlk: There should be an option to define which port to attach the floating IP to, for instances with more than one nic.

cloud/os_compute_facts.py
```diff
@@ -0,0 +1,90 @@
...
+short_description: Retrieve facts about a compute instance
+extends_documentation_fragment: openstack
+description:
+   - Retrieve facts about a compute instance from OpenStack
+options:
+   name:
+     description:
+        - Name of the instance
+     default: None
+   id:
```
omgjlk: Other modules are using options such as "server_name" and "server_id", but this module uses just "name" and "id". If a complete overhaul is being done, these options should align across the modules.

cloud/os_floating_ip.py
```diff
@@ -0,0 +1,210 @@
...
+
+try:
+    import shade
+except ImportError:
+    print("failed=True msg='shade is required'")
+
+DOCUMENTATION = '''
+---
+module: os_floating_ip
+short_description: Add/Remove floating IP from an instance
```
omgjlk: Why is this here, when there is already os_compute_floating_ip ? Which one should be included and reviewed? This one here seems to correct some of the issues I saw in os_compute_floating_ip.

omgjlk: Except this one doesn't allow you to specify a floating IP that already exists, so maybe some convergence of the two modules is in order?

cloud/os_floating_ip.py
```diff
@@ -0,0 +1,210 @@
...
+     description:
+        - Indicate desired state of the resource
+     choices: ['present', 'absent']
+     default: present
+   network_name:
+     description:
+        - Name of the network from which IP has to be assigned to VM. Please make sure the network is an external network
+     required: true
+     default: None
+   instance_name:
```
omgjlk: Should also provide a way to specify instance_id (or server_id to align with other modules) for when multiple instances exist with the same name.

cloud/os_image.py
```diff
@@ -0,0 +1,187 @@
...
+    module = AnsibleModule(argument_spec, **module_kwargs)
+
+    if module.params['state'] == 'present':
+        if not module.params['file'] and not module.params['copy_from']:
+            module.fail_json(msg="Either file or copy_from variable should be set to create the image")
+
+    try:
+        cloud = shade.openstack_cloud(**module.params)
+
+        id = cloud.get_image_id(module.params['name'])
```
omgjlk: How does this handle multiple images existing with the same name?

cloud/os_keypair.py
```diff
@@ -0,0 +1,107 @@
...
+    module = AnsibleModule(argument_spec, **module_kwargs)
+
+    try:
+        nova = shade.openstack_cloud(**module.params)
+
+        if module.params['state'] == 'present':
+            for key in nova.list_keypairs():
+                if key.name == module.params['name']:
+                    if module.params['public_key'] and (module.params['public_key'] != key.public_key ):
+                        module.fail_json(msg = "name {} present but key hash not the same as offered.  Delete key first.".format(key['name']))
```
omgjlk: Perhaps a force=bool option would allow overwriting existing key. Probably not necessary for the first drop but something to think about for a later improvement.

cloud/os_network.py
```diff
@@ -0,0 +1,220 @@
...
+# along with this software.  If not, see <http://www.gnu.org/licenses/>.
+
+try:
+    import shade
+except ImportError:
+    print("failed=True msg='shade is required for this module'")
+
+DOCUMENTATION = '''
+---
+module: os_network
```
omgjlk: Since this is an admin level task, more of a cloud operator/deployer rather than a cloud consumer, shouldn't it use the specific name to the project, os_neutron_network ?

cloud/os_router.py
```diff
@@ -0,0 +1,124 @@
...
+
+try:
+    import shade
+except ImportError:
+    print("failed=True msg='shade is required for this module'")
+
+
+DOCUMENTATION = '''
+---
+module: os_router
```
omgjlk: This feels like a deployer task too.

cloud/os_router.py
```diff
@@ -0,0 +1,124 @@
...
+     default: present
+   name:
+     description:
+        - Name to be give to the router
+     required: true
+   admin_state_up:
+     description:
+        - desired admin state of the created router .
+     required: false
+     default: true
```
omgjlk: Should there be an option to provide a tenant_id like there is with networks?

cloud/os_router_gateway.py
```diff
@@ -0,0 +1,156 @@
...
+
+try:
+    import shade
+except ImportError:
+    print("failed=True msg='shade is required for this module'")
+
+
+DOCUMENTATION = '''
+---
+module: os_router_gateway
```
omgjlk: Feels like a deployers task.

cloud/os_router_interface.py
```diff
@@ -0,0 +1,163 @@
...
+# along with this software.  If not, see <http://www.gnu.org/licenses/>.
+
+try:
+    import shade
+except ImportError:
+    print("failed=True msg='shade is required for this module'")
+
+DOCUMENTATION = '''
+---
+module: os_router_interface
```
omgjlk: Feels like a deployer's task.

cloud/os_subnet.py
```diff
@@ -0,0 +1,208 @@
...
+
+try:
+    import shade
+except ImportError:
+    print("failed=True msg='shade is required for this module'")
+
+
+DOCUMENTATION = '''
+---
+module: os_subnet
```
omgjlk: Another deployer task.

cloud/os_image.py
Shrews: Probably best explained by reading this Ansible dev ML thread:

https://groups.google.com/forum/#!searchin/ansible-devel/openstack/ansible-devel/njBsvZmX43U/CbR5auutumsJ

cloud/os_compute_facts.py
Shrews: Thanks. Fixed in a future revision of this PR.

cloud/os_compute_floating_ip.py
Shrews: Thanks. Fixed in a future revision of this PR.

cloud/os_compute_floating_ip.py
Shrews: Will be removing os_compute_floating_ip.py and just having os_floating_ip.py.

cloud/os_compute_facts.py
Shrews: Thanks. A future revision of this PR will try to use 'server' to contain either a server instance name or ID.

cloud/os_floating_ip.py
Shrews: I'm merging os_compute_floating_ip into this one. The os_compute module lets you set auto_floating_ip to automatically get an IP at server create time. I'm honestly not sure how useful this module will be anyway, other than to maybe remove a floating IP. But it could be enhanced in a future version to accept a known IP.

cloud/os_floating_ip.py
Shrews: See my note about using 'server' parameter to contain either.

cloud/os_compute_facts.py
Shrews: Will do.

emonty:
@bcoca This is now ready to land. The interface is solid, and the library support behind it is good. We've tested on HP, Rackspace and both devstack and tripleo local installs. There will still be bugs for sure (we just learned about a new failure condition in OpenStack yesterday, in fact, that isn't accounted for yet), but we can work those as normal work.

omgjlk:
I'll give one more quick pass today as well, but I'm generally in favor of merging too.

cloud/openstack/os_auth.py
```diff
@@ -0,0 +1,72 @@
...
+
+    if not HAS_SHADE:
+        module.fail_json(msg='shade is required for this module')
+
+    try:
+        cloud = shade.openstack_cloud(**module.params)
+        module.exit_json(
+            changed=False,
+            auth_token=cloud.auth_token,
+            service_catalog=cloud.service_catalog)
```
omgjlk: Instead of just dumping the token and catalog in the return to be registered, why not make this an actual fact module http://docs.ansible.com/developing_modules.html#module-provided-facts and automatically make facts out of the data? Saves a step in playbooks.

cloud/openstack/os_floating_ip.py
```diff
@@ -0,0 +1,244 @@
...
+
+try:
+    import shade
+    HAS_SHADE = True
+except ImportError:
+    HAS_SHADE = False
+
+DOCUMENTATION = '''
+---
+module: os_floating_ip
```
omgjlk, line 29: Hate to bring it up now, but looks like the python-openstackclient folks decided that the verbiage should be "ip floating", such as "ip floating create" and "ip floating add". Should our modules follow suit here, to try and reduce the variety of ways in which these resources are referenced?

cloud/openstack/os_image.py
```diff
@@ -0,0 +1,192 @@
...
+requirements: ["shade"]
+   state:
+     description:
+       - Should the resource be present or absent.
+     choices: [present, absent]
+     default: present
+'''
+
+EXAMPLES = '''
+# Upload an image from an HTTP URL
```
omgjlk: This example says from an http url, but I don't see a URL provided in the example, instead I see a file.

In fact, looks like you can't do one from a URL. Is the intention to have a task to download first, then upload via this task? What about creating an image in the registry that is a reference to a http location? (okay if that's a later feature)

cloud/openstack/os_object.py
```diff
@@ -0,0 +1,124 @@
...
+    import shade
+    HAS_SHADE = True
+except ImportError:
+    HAS_SHADE = False
+
+
+DOCUMENTATION = '''
+---
+module: os_object
+short_description: Create or Delete object objects and containers from OpenStack
```
omgjlk: object said twice here.

cloud/openstack/os_facts.py
```diff
@@ -0,0 +1,81 @@
...
+     description:
+        - Optional list of dicts tying volumes to mount points
+     default: None
+requirements: ["shade"]
+'''
+
+EXAMPLES = '''
+# Create a server, then inject facts about it:
+- os_server_facts:
+    state: present
```
bcoca: since this is a fact gathering plugin, state seems not to serve any function here

cloud/openstack/os_secgroup.py
```diff
@@ -0,0 +1,112 @@
...
+
+try:
+    import shade
+except ImportError:
+    print("failed=True msg='shade is required for this module'")
+
+
+DOCUMENTATION = '''
+---
+module: os_secgroup
```
omgjlk: openstack client calls these security groups, not secgroups. We probably should too, I think "secgroup" nomenclature is going away.

cloud/openstack/os_auth.py
```diff
@@ -0,0 +1,72 @@
...
+
+DOCUMENTATION = '''
+---
+module: os_auth
+short_description: Retrieve an auth token
+extends_documentation_fragment: openstack
+description:
+   - Retrieve an auth token from an OpenStack Cloud
+options:
+  state:
```
bcoca: state seems not used here

cloud/openstack/os_auth.py
```diff
@@ -0,0 +1,72 @@
...
+    description:
+      - Should the resource be present or absent.
+    choices: [present, absent]
+    default: present
+requirements: ["shade"]
+'''
+
+EXAMPLES = '''
+# Fetch facts about an instance called vm1
```
bcoca: comment talks about fact but we are retrieving auth token

cloud/openstack/os_secgroup_rule.py
```diff
@@ -0,0 +1,141 @@
...
+
+def main():
+
+    argument_spec = openstack_full_argument_spec(
+        security_group      = dict(required=True),
+        ip_protocol         = dict(default='tcp', choices=['tcp', 'udp', 'icmp']),
+        from_port           = dict(required=True),
+        to_port             = dict(required=True),
+        cidr                = dict(required=True),
+        state               = dict(default='present', choices=['absent', 'present']),
```
omgjlk: This module should support a cidr /or/ another security group name. Pretty common to link security groups together rather than CIDRs.

cloud/openstack/os_auth.py
```diff
@@ -0,0 +1,72 @@
...
+- os_auth:
+    cloud: rax-dfw
+  register: results
+- debug: var=results
+'''
+
+def main():
+
+    argument_spec = openstack_full_argument_spec(
+        state=dict(default='present', choices=['absent', 'present']),
```
bcoca: we seem to define state, but not really do anything with it

cloud/openstack/os_server_volume.py
```diff
@@ -0,0 +1,129 @@
...
+      volume: mysql-data
+      device: /dev/vdb
+'''
+
+
+def main():
+    argument_spec = openstack_full_argument_spec(
+        server=dict(required=True),
+        volume=dict(required=True),
+        device=dict(default=None),
```
omgjlk: Device should default to auto perhaps, as a convenience, since the data is returned about where it's mounted anyway.

cloud/openstack/os_facts.py
```diff
@@ -0,0 +1,81 @@
...
+     required: true
+   mounts:
+     description:
+        - Optional list of dicts tying volumes to mount points
+     default: None
+requirements: ["shade"]
+'''
+
+EXAMPLES = '''
+# Create a server, then inject facts about it:
```
bcoca: does it really create a server?

cloud/openstack/os_facts.py
```diff
@@ -0,0 +1,81 @@
...
+notes:
+   - Facts are placed in the C(openstack) variable.
+options:
+   server:
+     description:
+        - Name or ID of the instance
+     required: true
+   mounts:
+     description:
+        - Optional list of dicts tying volumes to mount points
```
bcoca: unclear on what mounts does? it is passed to the metadata function, does it mean we get info about those mounts? does it default to returning all mount info?

cloud/openstack/os_auth.py
emonty: YES

cloud/openstack/os_floating_ip.py
emonty, line 29: It seems on this one it might be a bit more awkward for the ansible user. People talk about floating-ips out loud. I think the osc usage is more about having commands and subcommands. I agere in generally with alignment, but I think it would be substantially strange if this was called os_ip_floating - I don't feel adamant though - just kindof a feel?

cloud/openstack/os_image.py
emonty: Yup. Later feature. Should fix the commnent.

cloud/openstack/os_facts.py
emonty: Thanks for the catch - this should actually be removed for now - it's left over from a previous attempt to get and end to end flow of "create server, create volume, mount volume, install software" - but since cinder volumes don't have a way to express what mount point they were created for, it gets awkward. However, this can be added later if it turns out it's actually needed.

cloud/openstack/os_server_volume.py
emonty: Yup - None == auto. I'll add a comment.

emonty:
@j2sol @bcoca Thanks a ton! Great reviews - I cleaned up the reviews, added some comment and also added some todo list items. As a follow up note from the security_group linking comment-  I realized I was still tracking the nova parameter names, which was not the intent. I cleaned up the parameter names, but will come back by in a follow on to do the appropriate "do you neutron, yes, awesome, no, cool we'll use nova" logic.

omgjlk:
LGTM. :+1: :mountain_bicyclist:

cloud/openstack/os_server.py
```diff
@@ -0,0 +1,440 @@
...
+        module.params['image'], module.params['image_exclude'])
+
+    if flavor:
+        flavor_id = cloud.get_flavor(flavor)
+    else:
+        flavor_id = cloud.get_flavor_by_ram(flavor_ram, flavor_include)
+
+    nics = _network_args(module, cloud)
+
+    bootkwargs = dict(
```
marszall87, line 287: Is it possible to add "block_device_mapping_v2" option to bootkwargs (probably should be mutually exclusive with root_volume)? It is a new syntax for attaching block devices introduced in the Nova v2 API, old syntax will be removed as of v3. Details here: https://wiki.openstack.org/wiki/BlockDeviceConfig

emonty, line 287: I think that's a fantastic idea-- honestly, I think there are 2 things we want to add (and should do in a follow-on PR):
- something like block_device_mapping_v2 (although wow what a horrible parameter name that would be) to support "create root volume from image and boot server from that volume"
- something like ansible/ansible-modules-core#394 which allows specification of a list of volumes to create and attach at boot.

marszall87, line 287: Both features are supported by this horrible block_device_mapping_v2 parameter. I use it like this to create new bootable volume from .iso (/dev/vda) and attach existing volume (/dev/vdb):

```
  - device_name: /dev/vda
    uuid: "{{ image_id }}"
    boot_index: 0
    volume_size: 40
    source_type: image
    destination_type: volume
    delete_on_termination: true
  - device_name: /dev/vdb
    uuid: "{{ volume_id }}"
    source_type: volume
    destination_type: volume
```

bcoca:
@emonty so in an attempt to make this a bit saner and so we can make progress, can you make a separate request per module and in that add the deprecation of the old module?

this way we avoid the situation of having a bunch of deprecated modules w/o the substitute being in place, also allows us to review/merge a module at a time and make some progress.

emonty:
@bcoca Ok. I split them into PR per module, combined with deprecations and submitted to -core. For reference, it's ansible/ansible-modules-core#1040 ansible/ansible-modules-core#1041 ansible/ansible-modules-core#1042 ansible/ansible-modules-core#1043 ansible/ansible-modules-core#1044 ansible/ansible-modules-core#1045 ansible/ansible-modules-core#1046 ansible/ansible-modules-core#1047 ansible/ansible-modules-core#1048 ansible/ansible-modules-core#1051 ansible/ansible-modules-core#1052 ansible/ansible-modules-core#1053 ansible/ansible-modules-core#1054 ansible/ansible-modules-core#1055 ansible/ansible-modules-core#1056 ansible/ansible-modules-core#1057 ansible/ansible-modules-core#1058 ansible/ansible-modules-core#1059

Also - we should land ansible/ansible#10237 before we land any of these.

bcoca:
just noting here cause i keep forgetting, ansible/ansible#10237 was merged.

gregdek:
Is there any reason keeping this one open since everything has been split out and submitted as other PRs?

gregdek:
(But going through the process of adding "owner_pr" since emonty is an Extras module owner)

omgjlk:
We will go through this at Summit if it doesn't get closed before then. 
- jlk

bcoca:
using this ticket to keep track of the list

gregdek:
OK, folks. What should we do with this very large and confusing PR?  It looks like everything that matters has been split out into separate PRs. Should we close this?

omgjlk:
I vote to close it yes. We've got everything in other PRs and we can come up with a better way (maybe tags?) to track those PRs.

bcoca:
1/2 have been merged already

gregdek:
That's good enough for me.

PR #93, opened 2014-10-27
Added OpenStack Quantum port module.

rosmo:


jimi-c:
cc @emonty, as this may conflict with some of the work he's doing around the OpenStack modules.

johnpbatty:
I also just submitted an OpenStack quantum_port module (#107), before I came across this one!
Looking at this implementation it does support more features/parameters than mine, so I recommend you take this in preference to mine.

gregdek:
Adding "owner_pr" since @rosmo is an Extras module owner

gregdek:
OK, so, according to new policy, since @rosmo is already a module owner, we need one more reviewer for this one to bring it in. Boilerplate follows:

---

Because you’re already an approved reviewer, we assume that you’ve reviewed your own module; you just need to find one other person to review, and you’re all set.

emonty:
I'd really prefer to not merge this, as it's in the format of the modules we're deprecating. We have a set of neutron modules going in to ansible-modules-core right now, and I believe @dguerri has a port module coming. Also - please see https://github.com/ansible/ansible-modules-core/blob/devel/cloud/openstack/README.md for more background about what's going on in more detail

gregdek:
Good enough for me. I'll close it. Thanks for the PR, @rosmo, and apologies.

PR #96, opened 2014-10-30
Add pam_limits module to ansible : modify PAM limits for Linux

usawa:
I coded a pam_limits ansible module to modify the /etc/security/limits.conf file, for Linux. It's common to all linux distros. 
If rule is absent, it's added to file
If rule is present, it's modified
use_max and use_min permits to take min or max value between the old or new value.

### DIFF
--- a/system/pam_limits.py
+++ b/system/pam_limits.py
@@ -1,0 +1,222 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2014, Sebastien Rohaut <sebastien.rohaut@gmail.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+import os.path
+import shutil
+import re
+
+DOCUMENTATION = '''
+---
+module: pam_limits
+version_added: "historical"
+short_description: Modify Linux PAM limits
+description:
+     - The M(pam_limits) module modify PAM limits, default in /etc/security/limits.conf.
+       For the full documentation, see man limits.conf(5).
+options:
+  domain:
+    description:
+      - A username, @groupname, wildcard, uid/gid range.
+    required: true
+    default: null
+  limit_type:
+    description:
+      - Limit type : hard or soft.
+    required: true
+    choices: [ "hard", "soft" ]
+    default: null
+  limit_item:
+    description:
+      - The limit to be set : core, data, nofile, cpu, etc.
+    required: true
+    choices: [ "core", "data", "fsize", "memlock", "nofile", "rss", "stack", "cpu", "nproc", "as", "maxlogins", "maxsyslogins", "priority", "locks", "sigpending", "msgqueue", "nice", "rtprio", "chroot" ]
+        default: null
+  value:
+    description:
+      - The value of the limit.
+    required: true
+    default: null
+  backup:
+    description:
+      - Create a backup file including the timestamp information so you can get
+        the original file back if you somehow clobbered it incorrectly.
+    required: false
+    choices: [ "yes", "no" ]
+    default: "no"
+  use_min:
+    description:
+      - If set to C(yes), the minimal value will be used or conserved.
+        If the specified value is inferior to the value in the file, file content is replaced with the new value,
+        else content is not modified.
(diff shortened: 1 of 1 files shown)


SiegfriedEhret:
:+1: that's a good idea. Please merge this !

prune998:
working great ! I needed exactly this !
Thanks

mateusduboli:
Could you add the [ '-' ] option to the type option, it enables setting both hard and soft limits in one statement. The docs are here.
http://linux.die.net/man/5/limits.conf

usawa:
Note : the module doesn't deal with :

"All items support the values -1, unlimited or infinity indicating no limit, except for priority and nice."

I'll work on it when I'll have some time.

system/pam_limits.py
```diff
@@ -0,0 +1,242 @@
...
+
+import os
+import os.path
+import shutil
+import re
+
+DOCUMENTATION = '''
+---
+module: pam_limits
+version_added: "historical"
```
bcoca: 2.0 should be version

system/pam_limits.py
```diff
@@ -0,0 +1,242 @@
...
+    description:
+      - A username, @groupname, wildcard, uid/gid range.
+    required: true
+    default: null
+  limit_type:
+    description:
+      - Limit type : hard or soft.
+    required: true
+    choices: [ "hard", "soft" ]
+    default: null
```
bcoca: required fields do not have defaults

system/pam_limits.py
```diff
@@ -0,0 +1,242 @@
...
+    if not found:
+        changed = True
+        new_limit = domain + "\t" + limit_type + "\t" + limit_item + "\t" + str(new_value) + new_comment + "\n"
+        message = new_limit
+        nf.write(new_limit)
+
+    f.close()
+    nf.close()
+
+    # Copy tempfile to newfile
```
bcoca, line 221: use atomic_move function to put file into place

bcoca:
I'm tempted to just say 'use template module instead' and you get to also use limits.d, but this has some nice features in it.

usawa:
I'll correct problems pointed in your comments. But for the last one, no, we can't use a template module for pam limits, because he template will totally remove the old content. Multiple playbooks/roles must change only specified values, not all content. It should be possible to use something like lineinfile module, but without control on the values.

gregdek:
@usawa happy to merge this if you fix the issues pointed out by @bcoca.

system/pam_limits.py
usawa: Can you explain ? Just remove "default" line from documentation ?

system/pam_limits.py
usawa, line 221: done !

usawa:
@gregdek I floowed your comments. It should be ok now, just test it.

A little test playbook :

## ▽
- hosts: localhost
  sudo: yes
  tasks:
  - name: "Pam limits"
    pam_limits: domain="{{ item.domain }}" limit_type="{{ item.limit_type }}" limit_item="{{ item.limit_item }}" value="{{ item.value }}" use_min=yes
    with_items:
    - { domain: "joe", limit_type: "soft", limit_item: "nofile", value: "10000" }
    - { domain: "joe", limit_type: "hard", limit_item: "nofile", value: "20000" }

gregdek:
@bcoca re-review pls?

eeroniemi:
I use this module on production, and it works great.

system/pam_limits.py
```diff
@@ -0,0 +1,236 @@
...
+     - The M(pam_limits) module modify PAM limits, default in /etc/security/limits.conf.
+       For the full documentation, see man limits.conf(5).
+options:
+  domain:
+    description:
+      - A username, @groupname, wildcard, uid/gid range.
+    required: true
+  limit_type:
+    description:
+      - Limit type : hard or soft.
```
alxgu, line 41: Can't open the documentation with ansible-doc:

ansible-doc pam_limits
 ...
  File "/usr/bin/ansible-doc", line 112, in get_man_text
    desc = " ".join(opt['description'])
TypeError: sequence item 0: expected string, dict found

Please replace the second colon trough dash or whatever :-)

system/pam_limits.py
```diff
@@ -0,0 +1,236 @@
...
+    required: true
+    choices: [ "hard", "soft" ]
+  limit_item:
+    description:
+      - The limit to be set : core, data, nofile, cpu, etc.
```
alxgu, line 46: Here too
Please replace the second colon trough dash or whatever :-)

vamfoom:
I'm attempting to use the pam_limits module as included in Ansible 2.1 but getting this error:

  TASK [system_configs : pam_limits] *********************************************
  fatal: [10.0.107.1]: FAILED! => {"failed": true, "msg": "module (pam_limits) is missing interpreter line"}
  fatal: [10.0.97.1]: FAILED! => {"failed": true, "msg": "module (pam_limits) is missing interpreter line"}
  fatal: [10.0.70.1]: FAILED! => {"failed": true, "msg": "module (pam_limits) is missing interpreter line"}
        to retry, use: --limit @/etc/ansible/main.retry

What am I missing?  This is the task:
- pam_limits: domain=\* limit_item=nofile limit_type=- value=65536

alxgu:
limit_type could be "hard" or "soft" and not -

PR #98, opened 2014-11-01
Added support for --name and --type arguments when following a new log

rubenvdb:
The logagent supports passing name and type arguments when following a log. This was not possible when following logs using this logentries module. 
Two additional parameters (name and type) were added to the module, which are passed to the 'le follow' command when present. In order to do so, the undocumented alias 'name' for 'path' had to be removed.


JeanFred:
Tested it, looks good to me.

PR #99, opened 2014-11-03
Snmp module for Ansible

ogenstad:
An Ansible module to gather Ansible facts through SNMP polls. Currently gathers basic information from SNMPv2-MIB such as system description, snmp location and snmp contact. Along with interface information and ip addresses. Could be expanded with more polls later on.

The module supports snmp version 2 and 3.

Has to be used with "connection: local" and "gather_facts: false"


bcoca:
can you move this to the network directory and set version added to 1.9?

ogenstad:
Sure. However I'm not really sure how I actually move it. Do I close this request and submit another one?

bcoca:
just to a 'git  mv' and add a commit to this request

ogenstad:
I think I did the correct thing now (new with git). Let me know if you want me to do anything else.

PR #64, opened 2014-10-11
Added module for managing Apple Mac OSX user defaults

frenck:
##### Issue Type:

New Module Pull Request
##### Ansible Version:

1.7.2
##### Environment:

Apple Mac OS X Yosemite 10.10 (build 14A379b)
##### Summary:

Created a new module for managing Mac's user defaults. With this module you'll be able to tailor most of the settings of your Mac using Ansible. It supports most of defaults datatypes. Array, Boolean, Date, Float, Integer and String types.
##### Steps To Reproduce:

N/A
##### Expected Results:

N/A
##### Actual Results:

N/A

PR #102, opened 2014-11-08
add osx_defaults module

josephholsten:


josephholsten:
and #64 is a much better implementation, closing.

PR #104, opened 2014-11-10
module to enable/disable apache2 sites

paulczar:
This is mostly a copy&paste from the already existing apache_module module with a few bits changed to support the `a2ensite` and `a2dissite` commands provided by apache2 on ubuntu.

### DIFF
--- a/web_infrastructure/apache2_site
+++ b/web_infrastructure/apache2_site
@@ -1,0 +1,90 @@
+#!/usr/bin/python
+#coding: utf-8 -*-
+
+# (c) 2014, Paul Czarkowski <pczarkowski@bluebox.net>
+# (c) 2013-2014, Christian Berendt <berendt@b1-systems.de>
+#
+# This site is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This software is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


bcoca:
I think it might be better to patch the current module do be able to do both

ghost:
should ansible modules in this repository be distro specific?

bcoca:
we try to avoid that, only things like packaging should be distro specific

ghost:
@bcoca : so should this PR be closed then? The a2*site commands are only available by default on debian derivatives

bcoca:
no, I'm just suggesting it be expanded to work in more scenarios

ghost:
I mistakenly submitted a [similar PR](https://github.com/ansible/ansible-modules-core/pull/1242) to the core modules repo. My PR is a cut and paste job too, but does add the option to specify an apache instance in the same way as the PR for [apache2_module](https://github.com/ansible/ansible-modules-core/pull/1194)

gregdek:
@bcoca when you say this should be expanded to work in more scenarios, did you have any details?

bcoca:
@gregdek what the description states, being able to manage sites and modules from a single plugin

gregdek:
So is the right approach, then, to add this functionality to the existing apache module, or no?

michaeljs1990:
It looks the maintainer is no longer interested in carrying this one though or has not checked github in a long time. I would be willing to take this one over and submit a PR with functionality for RHEL and Debian.

gregdek:
@michaeljs1990 I agree. I'll close this one. Looking forward to your new PR.

PR #107, opened 2014-11-14
New quantum_port module

johnpbatty:
New module for managing OpenStack network ports.

Allows network ports to be added/removed, in the same way as other
Ansible OpenStack modules.

I would rather have named this module neutron_port, as OpenStack
networking was renamed from quantum to neutron over a year ago.
However, I’m naming it quantum to fit in with the existing modules.


johnpbatty:
I just found an earlier pull request (Added OpenStack Quantum port module. #93) that implements the same function (plus some more).  So recommend that you take that one in preference to mine.

PR #108, opened 2014-11-14
Allow filtering of routed/forwarded packets

n11c:
MAN page states the following :

```
Rules for traffic not destined for the  host  itself  but  instead  for
traffic  that  should  be  routed/forwarded through the firewall should
specify the  route  keyword  before  the  rule  (routing  rules  differ
significantly  from  PF  syntax and instead take into account netfilter
FORWARD chain conventions). For example:

    ufw route allow in on eth1 out on eth2
```

This commit introduces a new parameter "route=yes/no" to allow just that.

### DIFF
--- a/system/ufw.py
+++ b/system/ufw.py
@@ -113,6 +113,11 @@ options:
       - Specify interface for rule.
     required: false
     aliases: ['if']
+  route:
+    description:
+      - Apply the rule to routed/forwarded packets.
+    required: false
+    choices: ['yes', 'no']
 '''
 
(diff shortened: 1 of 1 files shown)


gregdek:
@ovcharenko please review this PR; if you give +1, we will merge

ovcharenko:
@gregdek it looks good. Go ahead with merge.

gregdek:
Merging! Thanks @zacbri for the PR and @ovcharenko for the review.

PR #119, opened 2014-11-25
Fix some logical issues with enabling/disabling a server on the A10.

swimlappy:
Hello,

I tried testing the A10 module and ran into a few issues.  I'm hoping perhaps these changes will get accepted.  Specifically the problems I ran into were
- When trying to enable/disable a server on the A10, both the server name and IP address had to be specified.  If the IP address was not specified in the playbook, the code would fail thinking you were trying to "create a server" and this was required. If you only want to enable/disable a server, which is a pretty simple thing that happens a LOT, this code will make it easier to do that.  See lines 198-200.
- Add a new function to determine if the status of a server needs to be updated.  See additional lines 237-246.
- Only add optional parameters to the json_post array if they were specified in the module as parameters.  My AXAPI calls were including a ports parameter when they didn't need one.  See additional lines 190-199.
- The a10_server.py code is looking for a module parameter called "write_config", but the documentation doesn't mention that.  Make that an optional parameter, or perhaps just update the documentation.  Either way, this code allows you to be more flexible.  See lines 254-255.

Hope this helps others!

Jason

### DIFF
--- a/network/a10/a10_server.py
+++ b/network/a10/a10_server.py
@@ -251,9 +251,7 @@ def main():
             # - in case ports are missing from the ones specified by the user
             # - in case ports are missing from those on the device
             # - in case we are change the status of a server
-            if port_needs_update(defined_ports, slb_server_ports)
-              or port_needs_update(slb_server_ports, defined_ports)
-              or status_needs_update(current_status, axapi_enabled_disabled(slb_server_status)):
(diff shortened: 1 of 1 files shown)


gregdek:
@mischapeters please review this PR; if you give +1, we will merge

mischapeters:
:+1: on 797d889, changes for 1011565 are not PEP, but if that doesn't matter, +1 as well.

gregdek:
From http://docs.ansible.com/developing_modules.html#module-checklist:

"Being pep8 compliant is nice, but not a requirement. Specifically, the 80 column limit now hinders readability more that it improves it"

Thus, merging. Thx @mischapeters for review, and thanks @swimlappy for the patch!

PR #123, opened 2014-11-27
Added module to support managing LXC containers

cloudnull:
The new module will allow users to control LXC containers from ansible. The module was built for use in LXC >= 1.0 and implements most of what can be done using the various lxc clients with regards to running containers. This first module is geared only at managing LXC containers.

The module provides:
- build containers
- destroy containers
- archive containers with the ability to compress the archive
- info from a single container
- start / stop / freeze /restart containers
- run commands within containers
- enable logging when creating a container
- add / modify lxc config for a container
- supports backends including LVM, directory, loop, btrfs

This PR is a continuation of the work that was done in PR: "https://github.com/ansible/ansible/pull/7825". In this PR the module has been refactored to remove the command pattern in favor of being declarative, I've removed fact gather parts of the module, and changed most of the shell commands to use the LXC python library where appropriate.

### DIFF
--- a/cloud/lxc/lxc-container.py
+++ b/cloud/lxc/lxc-container.py
@@ -1015,9 +1015,12 @@ class LxcContainerManagement(object):
         :returns: size and measurement of an LV
         :type: ``tuple``
         """
+
         build_command = [
             'vgdisplay',
-            name
+            name,
+            '--units',
+            'g'
         ]
         rc, stdout, err = self._run_command(build_command)
         if rc != 0:
@@ -1041,11 +1044,14 @@ class LxcContainerManagement(object):
         :returns: size and measurement of an LV
         :type: ``tuple``
         """
+
         vg = self._get_lxc_vg()
         lv = os.path.join(vg, name)
         build_command = [
             'lvdisplay',
-            lv
+            lv,
+            '--units',
+            'g'
         ]
         rc, stdout, err = self._run_command(build_command)
         if rc != 0:
@@ -1072,6 +1078,7 @@ class LxcContainerManagement(object):
         :param snapshot_size_gb: Size of snapshot to create
         :type snapshot_size_gb: ``int``
         """
+
         vg = self._get_lxc_vg()
         free_space, messurement = self._get_vg_free_pe(name=vg)
 
@@ -1112,6 +1119,7 @@ class LxcContainerManagement(object):
         :param mount_point: path on the file system that is mounted.
         :type mount_point: ``str``
         """
+
         vg = self._get_lxc_vg()
 
         build_command = [
@@ -1181,6 +1189,7 @@ class LxcContainerManagement(object):
         :param name: The name of the logical volume
         :type name: ``str``
         """
+
         vg = self._get_lxc_vg()
         build_command = [
             self.module.get_bin_path('lvremove', True),
@@ -1204,6 +1213,7 @@ class LxcContainerManagement(object):
         :param temp_dir: path to the temporary local working directory
         :type temp_dir: ``str``
         """
+
         build_command = [
             self.module.get_bin_path('rsync', True),
             '-aHAX',
@@ -1225,6 +1235,7 @@ class LxcContainerManagement(object):
         :param mount_point: path on the file system that is mounted.
         :type mount_point: ``str``
         """
+
         build_command = [
             self.module.get_bin_path('umount', True),
             mount_point,
--- a/cloud/lxc/lxc-container.py
+++ b/cloud/lxc/lxc-container.py
@@ -1,0 +1,1469 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2014, Kevin Carter <kevin.carter@rackspace.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+
+DOCUMENTATION = """
+---
+module: lxc-container
+short_description: Manage LXC Containers
+version_added: 1.8.0
+description:
+  - Management of LXC containers
+author: Kevin Carter
+options:
+    name:
+        description:
+          - Name of a container.
+        required: true
+    backing_store:
+        choices:
+          - dir
+          - lvm
+          - loop
+          - btrfs
+        description:
+          - Backend storage type for the container.
+        required: false
+        default: dir
+    template:
+        description:
+          - Name of the template to use within an LXC create.
+        required: false
+        default: ubuntu
+    template_options:
+        description:
+          - Template options when building the container.
+        required: false
+    config:
+        description:
+          - Path to the LXC configuration file.
+        required: false
+        default: /etc/lxc/default.conf
+    lv_name:
+        description:
+          - Name of the logical volume, defaults to the container name.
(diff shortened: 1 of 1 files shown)


cloud/lxc/lxc-container.py
```diff
@@ -0,0 +1,1469 @@
...
+    def _get_vg_free_pe(self, name):
+        """Return the available size of a given VG.
+
+        :param name: Name of volume.
+        :type name: ``str``
+        :returns: size and measurement of an LV
+        :type: ``tuple``
+        """
+        build_command = [
+            'vgdisplay',
```
hughsaunders, line 1020: I think `--units g` is needed here, otherwise invalid results will be returned when vgs decides to use other units. For example: 

```
(.venv)root@node06:~/ansible# vgs
  VG             #PV #LV #SN Attr   VSize  VFree
  cinder-volumes   1   0   0 wz--n- 93.13g 93.13g
  lxc              1  21   0 wz--n-  1.63t  1.53t
  node06-vg        1   2   0 wz--n- 93.13g     0
(.venv)root@node06:~/ansible# vgs --units g
  VG             #PV #LV #SN Attr   VSize    VFree
  cinder-volumes   1   0   0 wz--n-   93.13g   93.13g
  lxc              1  21   0 wz--n- 1670.07g 1565.07g
  node06-vg        1   2   0 wz--n-   93.13g       0g
```

This leads to the following error:

```
(.venv)root@node06:~/ansible# ansible-playbook  -i inventory test_playbook.yml

PLAY [hosts] ******************************************************************

GATHERING FACTS ***************************************************************
ok: [localhost]

TASK: [Create a filesystem container] *****************************************
ok: [localhost]

TASK: [Create a filesystem container] *****************************************
ok: [localhost]

TASK: [Create a container, run a command, and leave in a frozen state] ********
changed: [localhost]

TASK: [Create filesystem container] *******************************************
changed: [localhost]

TASK: [Create an lvm container.] **********************************************
failed: [localhost] => {"failed": true, "parsed": false}
{"msg": "Snapshot size [ 5 ] is > greater than [ 1.53 ] on volume group [ lxc ]", "failed": true, "error": "Not enough space to create snapshot", "rc": 2}
{"msg": "failed to unmount [ /tmp/tmpjwe9Vu/test-container-lvm/rootfs ]", "failed": true, "command": "/bin/umount /tmp/tmpjwe9Vu/test-container-lvm/rootfs", "err": "umount: /tmp/tmpjwe9Vu/test-container-lvm/rootfs: not mounted\n", "rc": 1}


           to retry, use: --limit @/root/test_playbook.retry

localhost                  : ok=5    changed=2    unreachable=0    failed=1
```

cloudnull:
@hughsaunders  I've gone and updated the module to always use _--units g_ for both `lvdisplay` and `vgdisplay` commands. This should fix the issue where sizes are no compared correctly.

hughsaunders:
@cloudnull thanks for the update, LGTM.

ant31:
Very nice module. Can't wait to see it merged

yongjunj:
+1

bcoca:
soo, this is already in the repo, not sure if there were 2 identical requests or what happend.

closing this ticket as it is a noop.

bcoca:
sorry, my bad, i had been using it for a few weeks and had a local merge, NOW it is in repo.

cloudnull:
Thanks for the merge @bcoca its really appreciated!

bcoca:
@cloudnull thanks for your contribution!

fghaas:
@cloudnull, I'm just looking at this as this module has been recently merged. Do I understand correctly that there is presently no support for managing overlayfs-based containers?

It looks like there wouldn't be anything that would create or support this syntax:

``` ini
lxc.rootfs = overlayfs:<lower>:<upper>
```

Is that observation accurate? Just checking to see if I'm not overlooking something.

nrb:
@fghaas I double checked, and I do not see it either. More generally, I don't see support for allowing the selection of any filesystem type at the moment.

cloudnull:
@fghaas when i created this module i did not build in support for overlayfs. To extend it I think we'd need to add two things.
- First would be to add an _overlayfs_ option in "https://github.com/ansible/ansible-modules-extras/blob/devel/cloud/lxc/lxc_container.py#L35-L44" 
- Second would be to create an incompatible variable choice entries in "https://github.com/ansible/ansible-modules-extras/blob/devel/cloud/lxc/lxc_container.py#L358-L373". Which would be something similar to this.

``` python
'overlayfs': ['lv_name', 'vg_name', 'fs_type', 'fs_size', 'thinpool', 'zfs_root'],
```

With those changes we'd should be able to create containers using _overlayfs_ too.

fghaas:
@cloudnull I _think_ that sounds about right, yes. :)

jpic:
I had to change:

```
    parsed_options = [i.split('=') for i in options_dict]
```

With:

```
    badly_parsed_options = [i.split('=') for i in options_dict]
    parsed_options = [(o[0], ' '.join(o[1:])) for o in badly_parsed_options]
```

To support mounting in:

```
- name: setup container to run tests in
  sudo: yes
  lxc-container:
    name: '{{ lxc_container_name }}'
    template: debian
    state: started
    template_options: --release jessie
    container_config:
    - "lxc.aa_profile=unconfined"
    - "lxc.mount.entry={{ host_src_path }} /var/lib/lxc/novasalt-ci/rootfs/usr/src none, defaults,bind,uid=0 0 0"
```

Else I'd get:

```
*************
failed: [localhost] => {"failed": true, "parsed": false}
BECOME-SUCCESS-ogwroucjldahxzjmvaipuvmwtbahkayp
  File "/home/jpic/.ansible/tmp/ansible-tmp-1426595904.53-45363252440675/lxc-container", line 3075, in <module>
    main()
  File "/home/jpic/.ansible/tmp/ansible-tmp-1426595904.53-45363252440675/lxc-container", line 1476, in main
    lxc_manage.run()
  File "/home/jpic/.ansible/tmp/ansible-tmp-1426595904.53-45363252440675/lxc-container", line 1373, in run
    action()
  File "/home/jpic/.ansible/tmp/ansible-tmp-1426595904.53-45363252440675/lxc-container", line 966, in _started
    self._config()
  File "/home/jpic/.ansible/tmp/ansible-tmp-1426595904.53-45363252440675/lxc-container", line 624, in _config
    for key, value in parsed_options:
ValueError: too many values to unpack
```

Because my container_config was parsed as:

```
(Epdb) print parsed_options
[['lxc.aa_profile', 'unconfined'], ['lxc.mount.entry', '~/work/novapost /var/lib/lxc/novasalt-ci/rootfs/usr/src none, defaults,bind,uid', '0 0 0']]
```

cloudnull:
@jpic I think that you could simplify your `badly_parsed_options` functions by simply keeping the original `parsed_options` list comprehension and adding a split on the first occurrence of the '=' instead of all of them, which seems to have been a bit of an oversight when the option has more than one '=' sign.

The simpler solution would be like so.

``` python
parsed_options = [i.split('=', 1) for i in options_dict]
```

This would return what we need to for better option parsing.

``` python
[['lxc.aa_profile', 'unconfined'], ['lxc.mount.entry', '{{ host_src_path }} /var/lib/lxc/novasalt-ci/rootfs/usr/src none, defaults,bind,uid=0 0 0']]
```

All said, this is a good find within the module and we should patch it :+1:

cloudnull:
@fghaas if you get a chance please have a look at "https://github.com/ansible/ansible-modules-extras/pull/328" - the PR adds _overlayfs_ and the ability to clone a container. I hope that fits your usecase.

fghaas:
@cloudnull I would love to, but for my use case I eventually had to go with managing lxc via libvirt. We might be able to migrate to lxc userland at some point though, in which case this would absolutely help. Thanks for putting this in!

leonty:
The module uses a python2-lxc import which is not in Pypi library. Is a user supposed to install it manually from git on every target machine?

cloudnull:
@leonty yes this is what we've been doing for our deployments. When using this module we've been using the core ansible `pip` module to resolve the deps by simply installing the python libs from the upstream git sources. That said, I noticed that you've uploaded this package to pypi as noted in issue "https://github.com/ansible/ansible-modules-extras/issues/550" as such I have updated the documentation to indicate where a user can get the requirements and how to installed the package. The associated PR can be found here: "https://github.com/ansible/ansible-modules-extras/pull/567".

PR #124, opened 2014-12-01
gtid_replication parameter added 

banyek:
Environments where MySQL uses GTID replication cannot managed with the original module, because there is a 'MASTER_AUTO_POSITION = 1' switch in all the CHANGE MASTER TO ... commands.

### DIFF
--- a/database/mysql/mysql_replication.py
+++ b/database/mysql/mysql_replication.py
@@ -106,7 +106,6 @@ options:
     master_auto_position:
         descrtiption:
             - does the host uses GTID based replication or not
-        possible values: 0,1
 '''
 
 EXAMPLES = '''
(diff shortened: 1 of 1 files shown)


mruoss:
This would be great! However, I'd go with the module's natural style and use the mysql variable name **master_auto_position** and the same declaration as for master_ssl (boolean):

```
master_auto_position=dict(default=False, type='bool')
```

instead of

```
gtid_replication=dict(default=None, choices=['0', '1'])
```

Anyway, I'd really like to see this merged.

banyek:
True. I'll changet this

banyek:
So, it's 2015 now, and GTID replication is out for like ... a year. I think this feature should added to mysql_replication module, because there is a lot of instances (the GTID enabled ones) which cannot use the old mysql_replication module.

database/mysql/mysql_replication.py
```diff
@@ -103,7 +103,10 @@
     master_ssl_cipher:
         description:
             - same as mysql variable
-
+    master_auto_position:
+        descrtiption:
+            - does the host uses GTID based replication or not
+        possible values: 0,1
```
bcoca: you are missing requried: False and default: , also use choices: instead of possible values:

mruoss: How about type='bool' instead of choices=['0', '1']?

PR #125, opened 2014-12-01
Filtering out warning levels added

banyek:
If mysql gets replication user data in CHANGE MASTER TO command, it warns user. It is a valid scenario, when you want filter out this messages.


banyek:
It's still needed to merge I guess

bcoca:
instead of warnings filter this looks more like an errors filter in which you can have a level of warnings

banyek:
Originally it was only a warnings filter, but lately I just add a feature of filter different levels of errors/warnings, true.
Should I rename it?

bcoca:
yes, its not good when a new parameter is misleading

PR #126, opened 2014-12-02
Removed references to slack domain (no longer relevant)

kgodey:
Slack webhooks now use `hooks.slack.com`, so the domain parameter is irrelevant, but it's still listed as required in the module. This pull request fixes that.

### DIFF
--- a/notification/slack.py
+++ b/notification/slack.py
@@ -33,7 +33,7 @@ options:
     required: false    
   token:
     description:
-      - Slack integration token
+      - The token part of the Webhook URL, i.e. https://hooks.slack.com/services/C(thetoken/generated/byslack)
     required: true
   msg:
     description:
@@ -81,19 +81,24 @@ options:
     choices:
       - 'yes'
       - 'no'
+notes:
+  - Due to a change in the Slack API, the C(domain) parameter for this module was deprecated.
 """
 
 EXAMPLES = """
+The token can be found in the SlackHQ interface when configuring an Incoming WebHook.
+https://hooks.slack.com/services/`thetoken/generated/byslack`
+
 - name: Send notification message via Slack
   local_action:
     module: slack
-    token: thetokengeneratedbyslack
+    token: thetoken/generated/byslack
     msg: "{{ inventory_hostname }} completed"
 
 - name: Send notification message via Slack all options
   local_action:
     module: slack
-    token: thetokengeneratedbyslack
+    token: thetoken/generated/byslack
     msg: "{{ inventory_hostname }} completed"
     channel: "#ansible"
     username: "Ansible on {{ inventory_hostname }}"
--- a/notification/slack.py
+++ b/notification/slack.py
@@ -26,11 +26,6 @@ description:
 version_added: 1.6
 author: Ramon de la Fuente <ramon@delafuente.nl>
 options:
-  domain:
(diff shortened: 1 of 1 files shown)


abadger:
Clarification question: Is the presence of the domain causing an issue?  Or is it just no longer relevant to slack usage?  If I  specify a domain, does it cause errors?

kgodey:
Specifying a domain does not cause errors (actually, it's a required variable), but it's not actually used when the call to Slack is made, so no longer relevant. 

Pull request #101 updated the call to Slack to no longer use the domain, but didn't actually remove the variable.

bcoca:
instead of removing it we should deprecate it (and make it not required), so we don't break existing playbooks.

kgodey:
@bcoca makes sense, is there a best practice to deprecating it, other than adding "DEPRECATED" to the documentation for the variable and making it not required?

sivel:
I haven't seen it done very often. It was done in the rax module a few releases back.

The option was removed from the docs and we actually added a fail_json if it was set:

https://github.com/ansible/ansible-modules-core/blob/devel/cloud/rackspace/rax.py#L709

Not sure if it is still the best solution but just wanted to give an example.

abadger:
If we simply want to deprecate because the domain is simply no longer needed then we probably don't want to fail_json if it's present (the idea of deprecation in this case is to keep user's playbooks from failing if they do not update their parameters).    If the parameter was harmful or we needed the user to migrate from using one parameter in one format to another parameter in a different format then it would probably make sense to use a fail_json to alert them to the requirement to update their playbooks.

So things we can do:
- marking as deprecated in the module docs
- making the parameter optional instead of required
- adding a message to the json return data that the domain parameter should no longer be used (Not very useful as output is usually hidden but I recall a module I worked on recently doing this.)

kgodey:
Updated the pull request according to suggestions by @abadger.

ramondelafuente:
Besides removing the parameter I updated the documentation a little bit in my pull-request. I found that Slack no longer refers to a 'token' in their setup, so it seems confusing that you have to add part of the URL.
Perhaps you could take a look? https://github.com/ansible/ansible-modules-extras/pull/138

kgodey:
Updated docs to match @ramondelafuente's pull request.

kgodey:
@abadger @sivel @bcoca is there anything else I need to do to have this be ready for merge?

bcoca:
we might need to review and add domain back as a useful option, some people have complained about the update breaking their slack notifications (still looking into it).

ramondelafuente:
As far as I can tell, SlackHQ used to do the webhooks at `[team].slack.com`, but are now _always_ directing to `hooks.slack.com`

Do you have the complaints? Maybe I could offload some of the research, you guys seem busy enough.

kgodey:
:+1: to @ramondelafuente, I can help with research too.

gregdek:
This is an extras module now, and under the owner_pr regime, @ramondelafuente can give a +1 and we will merge this.  Right now it needs to be rebased, though, so @kgodey if you can rebase, we can give the ball to @ramondelafuente to make the merge decision.

kgodey:
@gregdek @ramondelafuente looks like the changes I made in this PR have already been implemented and merged in separately. Closing this.

PR #128, opened 2014-12-04
Patch multiple versions rpm zypper

robinro:
Fix bug in ansible get_package_state and get_current_version that breaks when there are multiple versions of a package installed and there is a list of packages to install.
The previous implementation used 'zip' to match requested names to installed names which fails, because rpm outputs multiple lines per package when there are multiple versions.

Testcase: Install opensuse, install multiple kernel versions (happens by update)
Before patch a playbook like the following fails:
zypper: state=present for name={{item}} 
with_items: 
- kernel-desktop
- git

After the patch ansible performs as expected and makes sure both packages are present.
Also the last version number is used for further update information in this version (before if only one package name was given (to not trigger the bug) the oldest version number was used instead of the current).

### DIFF
--- a/packaging/os/zypper.py
+++ b/packaging/os/zypper.py
@@ -95,11 +95,11 @@ def zypper_version(module):
         return rc, stderr
 
 # Function used for getting versions of currently installed packages.
-def get_current_version( packages):
+def get_current_version(m, packages):
     cmd = ['/bin/rpm', '-q', '--qf', '%{NAME} %{VERSION}-%{RELEASE}\n']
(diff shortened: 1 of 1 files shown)


alxgu:
Thanks for the PR. It works great, but in line 144 you used tabs and not spaces, right?

```
    if package not in installed_state:
                                     ^
TabError: inconsistent use of tabs and spaces in indentation
```

robinro:
I fixed the tab/space mixup, sorry for that.
Thanks @alxgu for looking at the patch.

PR #129, opened 2014-12-04
Make sure portage doesn't ask for confirmation

nmeum:
If EMERGE_DEFAULT_OPTS in make.conf(5) contains '--ask' then the portage
module doesn't work correctly, this commit fixes that.


nmeum:
Any comments on this?

PR #131, opened 2014-12-05
Jail support for pkgng module

davidfritzsche:
Add a _jail_ parameter to the pkgng module that is passed as a `--jail` parameter to `pkg`. With jail support in `pkgng` it is easier to setup new jails (e.g., by making it easy to install Python into the new jail) so that the new jail can be used with Ansible in following plays.

Besides jail support some small bugs were fixed and the error reporting was improved.


mpdehaan:
Hmm, I don't know if we should do this, because there are other things that happen in jails, and soon all modules may need a "jail" parameter.

Let's discuss on ansible-project.

The "jail" connection module may help here, though it does need to be run locally.  This may be something we try to do after the v2/ code tree is in place as it somewhat relates to a _perhaps challenging and minimally rewarding_ case of chaining action plugins.

Anyway, stop by -devel and we can figure out what may be the way to go here.

Thanks!

PR #133, opened 2014-12-06
Fix rabbitmq_plugin.py: broken prefix path

pcarinhas:
Old code only looks for /sbin while many current distros use /bin in the generic location.
Adding $prefix/bin so that it can find these binaries

### DIFF
--- a/messaging/rabbitmq_plugin.py
+++ b/messaging/rabbitmq_plugin.py
@@ -59,12 +59,23 @@ EXAMPLES = '''
 - rabbitmq_plugin: names=rabbitmq_management state=enabled
 '''
 
+import os
+
 class RabbitMqPlugins(object):
     def __init__(self, module):
         self.module = module
 
         if module.params['prefix']:
-            self._rabbitmq_plugins = module.params['prefix'] + "/sbin/rabbitmq-plugins"
+            if os.path.isdir(os.path.join(module.params['prefix'], 'bin')):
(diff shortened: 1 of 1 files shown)


mpdehaan:
Can you please share how to replicate the problem that this fixes - and it's symptoms?

Thanks!

pcarinhas:
## Replication of the problem: 

When using the rabbitmq_plugin module, specify with the prefix parameter (which identifies an alternative installation path for rabbitmq). The existing code only allows for the binary folder $prefix/sbin, so ansible is unable to locate the rabbitmq-plugins binary. 

In particular if you set prefix=/usr/lib/rabbitmq, the bin folder in $prefix/bin contains links to all the correct binaries. This location apprears to be quite generic. Those binaries are in fact linked to the actual location in a version-specific folder's sbin. 

## Symptoms

In Centos7 for example (and older versions of Centos),
there is no sbin subfolder so the command fails.

In Centos6.5 for example:
/usr/lib/rabbitmq/bin/rabbitmq-plugins -> ../lib/rabbitmq_server-2.8.7/sbin/rabbitmq-plugins

In Centos 7.X:
/usr/lib/rabbitmq/bin/rabbitmq-plugins -> ../lib/rabbitmq_server-3.1.5/sbin/rabbitmq-plugins

The module fails when it can not find the appropriate folder.

gregdek:
@chrishoffman please review this PR; if you give +1 we will merge

chrishoffman:
+1

PR #139, opened 2014-12-08
Zpool module

freym:
Create and modify a Zpool.

This Pull Request is based on ansible/ansible#8205. I think it was lost when the Module Repos were reorganized.
I fixed some Errors and add a new Feature: Add new devices to an existing pool. This allows Pools with nested Raid Levels like Raid10 or Raid50.


system/zpool.py
```diff
@@ -0,0 +1,374 @@
...
+- zpool: name=rpool raid_type=raidz state=present zpool_devs='sdc sdd sde' zil_dev='mirror sdb7 sda7'
+
+# Destroy a zpool named rpool
+- zpool: name=rpool state=absent
+
+# Add sdb to rpool
+- zpool: name=rpool zpool_devs='sdb' state=add
+
+# Create a new RAID10 zpool called rpool using sda, sdb, sdc and sdd
+- zpool: name=rpool state=present raid_type=mirror zpool_devs='sda sdb'
```
awreece, line 147: Why not do this in one line:

```
zpool: name=rpool state=present zpool_devs='mirror sda sdb mirror sdc sdd'
```

In general, the `zpool: state=present` then `zpool: state=add` idiom does not seem like the cleanest way to represent the storage hierarchy in zfs. The zfs command line uses `mirror sda sdb` to represent hierarchy in a flat line -- you could use that, or maybe you can do something better w/ structured data.

system/zpool.py
```diff
@@ -0,0 +1,374 @@
...
+        l2arc_dev=dict(default=None),
+        listsnapshots=dict(default=None, choices=['on', 'off']),
+        raid_type=dict(default=None, choices=['mirror', 'raidz', 'raidz2', 'raidz3']),
+        spare_devs=dict(default=None),
+        version=dict(default=None),
+        zil_dev=dict(default=None),
+        zpool_devs=dict(default=None)
+    )
+
+    argument_spec['feature@async_destroy'] = dict(default=None, choices=['enabled'])
```
awreece, line 329: Is the plan to document and add flags for every feature potentially used by zfs, or just the features supported by all platforms, or just the ones people use? If it were possible to provide generic feature support that would be much better. I don't know the ansible idiom, but this could look something like:

`feature=com.delphix:async_destroy feature=org.illumos:lz4_compress`

system/zpool.py
```diff
@@ -0,0 +1,374 @@
...
+  zpool_devs:
+    description:
+      - The devices to use for the zpool
+    required: False
+author: Stan Hiller, based on ZFS module from Johan Wiren
+'''
+
+EXAMPLES = '''
+# Create a new RAIDZ zpool called rpool using sdc, sdd, and sdc
+- zpool: name=rpool state=present raid_type=raidz zpool_devs='sdc sdd sde'
```
awreece, line 132: adding disk by `sdc`, etc is only recommended for test or development pools: http://zfsonlinux.org/faq.html#WhatDevNamesShouldIUseWhenCreatingMyPool

system/zpool.py
```diff
@@ -0,0 +1,374 @@
...
+        if prop in ['CHECKMODE']:
+            continue
+        if value:
+            properties[prop] = value
+
+    result = dict()
+    result['name'] = name
+    result['state'] = state
+
+    zpool = Zpool(module, name, properties)
```
awreece, line 350: Is the idiom in ansible to create one mega object that wraps all the parameters, or to create a backing object and call methods on it? My untrained eye sees the code here as being the former structure and would prefer something more akin to:

```
if state == 'present':
    if zpool.exists(pool_name):
        zpool.ensure_properties(pool_name, features)
    else:
        zpool.create(pool_name, pool_config)
```

system/zpool.py
```diff
@@ -0,0 +1,374 @@
...
+    def __init__(self, module, name, properties):
+        self.module = module
+        self.name = name
+        self.properties = properties
+        self.changed = False
+
+        self.immutable_properties = ['ashift']
+
+    def exists(self):
+        cmd = [self.module.get_bin_path('zpool', True)]
```
awreece, line 162: What happens if a pool of the same name but a different configuration exists? e.g a pool with one device exists, but you want to make a pool with two devices exist?

I don't know what the correct thing to do, but ignoring it and moving on seems bad (although better than recreating the pool and throwing away any data on the pool)

nmeum:
:+1: What's the status of this?

galindro:
I'd like to test this module. How Can I install it? I'm using master version of ansible official repository (ansible 2.0.0)

jhawkesworth:
Have a look at this page on the ansible documentation site - hopefully
enough to get you started

http://docs.ansible.com/developing_test_pr.html

gregdek:

bcoca:
@galindro, just download it to a 'library' dir in a role or adjacent to play you use it in

gregdek:
@freym Just a friendly ping to remind you that this PR needs revisions. :)

freym:
Sorry I don`t have the time to work on this module.

gregdek:
Closing due to contributor inactivity.

Thanks @freym -- if you choose to reopen, feel free.

system/zpool.py
olbohlen: I would recommend removing all the *_devs properties and combine them into a general vdevs_specs property where you simply put in zpool arguments. that way you could also use it to setup a pool consisting of two mirrors.

PR #140, opened 2014-12-09
Fall back to simplejson if ImportError on json

steinim:
Simplejson works on more Python versions (2.4+).


bcoca:
actually you should not need to import json at all as the common module code already does this, it might be better to just remove the reference in the module itself

steinim:
Ok. Let us do that then :)

PR #151, opened 2014-12-12
added module for OpenStack Cinder

hhamalai:


sivel:
This is partially a duplicate of https://github.com/ansible/ansible-modules-extras/pull/92

#92 also moves forward with the newly decided naming convention of openstack modules.

cc @emonty

gregdek:
@sivel @emonty if this is actually deprecated by #92 (which itself has been broken out into many core modules), then we should probably just close this. Any reason not to?

emonty:
Yes - we should just close this

gregdek:
Thanks @hhamalai for the PR, and sorry we couldn't ultimately merge it.

PR #152, opened 2014-12-15
Add ipa_group module

ToBeReplaced:
There was some talk a while back on ansible-devel about coming up with FreeIPA modules to work with ansible.

This is a step towards that. It is a module to provide idempotent management of ipa groups.


ToBeReplaced:
This commit needs minor changes for FreeIPA 4, which has now been adopted in RHEL 7.1 and has been in Fedora for two releases.

If this PR is of interest to the project, I will contribute these changes.

gregdek:

mscherer:
So, I fail to understand how the auth is supposed to work, I need to use kerberos for that, no login/password to connect ?

ToBeReplaced:
@mscherer - That is correct. If the user already has credentials locally and you want to forward them to the host running the role, you could add `GSSAPIDelegateCredentials=yes` to your `ssh_args` in `ansible.cfg`. If you need to fetch credentials, `kinit` or `kinit username` is required.

That said, I'm not sure if this role is a good idea or not...  For us, user administration changed hands, so this role is no longer in use. IPA already has a great administration interface, so maybe just use that directly?

mscherer:
I think there is some use for the module, but if you do not use it yourself and do not maintain, then I would indeed not push it.

PR #161, opened 2014-12-17
fix at.py module on non-glibc systems

lifanov:
Put the "-f" positional parameter in the "at" invocation first.
Glibc allows positional parameters after data, but FreeBSD libc (and others) do not.
It is a security feature, and it is also more portable to list positional paramenters first.
This change makes the "at" module work on FreeBSD.


abadger:
Verified problem occurs on freebsd10 and that the patch works on freebsd10 and also fedora20.  Thanks for the fix!

# Merged

Hi!

issue is resolved.

Thank you!

PR #162, opened 2014-12-17
corrected version added

bcoca:
they were actually merged in 1.9


cchurch:
@bcoca I also made the following change for the same reason: https://github.com/ansible/ansible-modules-core/commit/0a297e54bac75d86cd33b0c299f95d82b643e0ab

PR #171, opened 2014-12-29
add zabbix_host module

rongzeng54:
zabbix_host module uses the Zabbix API to add and remove host.
And it don't need any other python module.
And it support all Zabbix API, so expand is easy.


dj-wasabi:
:+1: 
My ansible role could really use this...

jensdepuydt:
This looks like something I was thinking to do myself. Thanks

gregdek:
An alternate zabbix_host module was merged in March. Thanks for the code, @rongzeng54, and our apologies that it wasn't selected for inclusion; we hope the alternate module suits your needs.

PR #174, opened 2014-12-30
win_unzip Module

schwartzmx:
`win_unzip` module:
- Can be useful for extracting libraries or needed sources.
- Also for unzipping self-extracting `.exe`'s for hot fixes such as http://support.microsoft.com/kb/2842230.
- Would most likely be used in conjunction with `win_get_url` to first grab the file.
- All paths specified must be on windows node machine.  Absolute path required.

Initial need for this module was for extracting gzip files that were being pulled down from s3 using AWS and also remotely installing hot fix for initialization of Windows 2008 R2 Servers.

**Updates 1/11**:
- Renamed `zip` param to `src` to be consistent across modules.
- Added `recurse` param to allow for recursive decompression.
  - For example, provided `C:\Example.zip` as a `src`, which contains many `.gz` files.  Specifying `recurse=true` will carry out a recursive decompression of all files within the decompressed `Example.zip` (step 1) to be decompressed as well (step 2).  Further, specifying `rm = true` will remove all compressed files after decompression is complete.
- Added ability to unzip `tar`, `gzip`, and `bzip` files to coincide with added functionality to `win_zip` module: [#177 commit:fc4d2c0d34e321d929b9c751add8c8cd38cfad9c](https://github.com/schwartzmx/ansible-modules-extras/commit/fc4d2c0d34e321d929b9c751add8c8cd38cfad9c)
  - `.msu` files will also be extracted when dealing with self-extracting `.exe` files such as with the hotfix previously mentioned.
- If a `.zip` file is specified and `recurse=false`, then the module will use PowerShell's built-in methods to extract the `.zip` file.  **Otherwise** `PSCX` will be installed, first by attempting a `chocolatey` install, if failure, then install from `.msi`.  This is also a requirement of `win_zip`: [#177](https://github.com/ansible/ansible-modules-extras/pull/177) 

Questions, comments, or suggestions for additions please let me know.

Phil

### DIFF
--- a/windows/win_unzip.ps1
+++ b/windows/win_unzip.ps1
@@ -33,7 +33,7 @@ If ($params.src) {
         Fail-Json $result "src file: $src does not exist."
     }
 
-    $ext = [System.IO.Path]::GetExtension($dest)
+    $ext = [System.IO.Path]::GetExtension($src)
 }
 Else {
     Fail-Json $result "missing required argument: src"
@@ -93,7 +93,7 @@ Else {
     If (-Not ($list -match "PSCX")) {
         # Try install with chocolatey
         Try {
-            cinst -force PSCX
+            cinst -force PSCX -y
             $choco = $true
         }
         Catch {
@@ -109,9 +109,7 @@ Else {
                 Fail-Json $result "Error downloading PSCX from $url and saving as $dest"
             }
             Try {
-                msiexec.exe /i $msi /qb
-                # Give it a chance to install, so that it can be imported
-                sleep 10
+                Start-Process -FilePath msiexec.exe -ArgumentList "/i $msi /qb" -Verb Runas -PassThru -Wait | out-null
             }
             Catch {
                 Fail-Json $result "Error installing $msi"
@@ -127,7 +125,12 @@ Else {
     # Import
     Try {
         If ($installed) {
-            Import-Module 'C:\Program Files (x86)\Powershell Community Extensions\pscx3\pscx\pscx.psd1'
+            Try {
+                Import-Module 'C:\Program Files (x86)\Powershell Community Extensions\pscx3\pscx\pscx.psd1'
+            }
+            Catch {
+                Import-Module PSCX
+            }
         }
         Else {
             Import-Module PSCX
--- a/windows/win_unzip.ps1
+++ b/windows/win_unzip.ps1
@@ -1,7 +1,7 @@
 #!powershell
 # This file is part of Ansible
 #
-# Copyright 2014, Phil Schwartz <schwartzmx@gmail.com>
+# Copyright 2015, Phil Schwartz <schwartzmx@gmail.com>
 #
 # Ansible is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -80,43 +80,13 @@ If ($ext -eq ".zip" -And $recurse -eq $false) {
         Fail-Json $result "Error unzipping $src to $dest"
     }
 }
-# Need PSCX
+# Requires PSCX
 Else {
-    # Requires PSCX, will be installed if it isn't found
-    # Pscx-3.2.0.msi
-    $url = "http://download-codeplex.sec.s-msft.com/Download/Release?ProjectName=pscx&DownloadId=923562&FileTime=130585918034470000&Build=20959"
-    $msi = "C:\Pscx-3.2.0.msi"
-
     # Check if PSCX is installed
     $list = Get-Module -ListAvailable
-    # If not download it and install
+
     If (-Not ($list -match "PSCX")) {
-        # Try install with chocolatey
-        Try {
-            cinst -force PSCX -y
-            $choco = $true
-        }
-        Catch {
-            $choco = $false
-        }
-        # install from downloaded msi if choco failed or is not present
-        If ($choco -eq $false) {
-            Try {
-                $client = New-Object System.Net.WebClient
-                $client.DownloadFile($url, $msi)
-            }
-            Catch {
-                Fail-Json $result "Error downloading PSCX from $url and saving as $dest"
-            }
-            Try {
-                Start-Process -FilePath msiexec.exe -ArgumentList "/i $msi /qb" -Verb Runas -PassThru -Wait | out-null
-            }
-            Catch {
-                Fail-Json $result "Error installing $msi"
-            }
-        }
-        Set-Attr $result.win_zip "pscx_status" "pscx was installed"
-        $installed = $true
+        Fail-Json "PowerShellCommunityExtensions PowerShell Module (PSCX) is required for non-'.zip' compressed archive types."
     }
     Else {
         Set-Attr $result.win_zip "pscx_status" "present"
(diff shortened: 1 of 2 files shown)


elventear:
I tried to use this module as is with only `src` and `dest` arguments and I was not able to make it work with a regular `zip` file.

```
failed: [PLAYGROUND-WIN-2008R2-X64] => {"failed": true, "parsed": false}
This installation package could not be opened.  Contact the application vendor to verify that this is a valid Windows Installer package.
{
    "msg":  "Error recursively expanding C:\\Tools\\Downloads\\SysinternalsSuite.zip to C:\\Tools\\SysinternalsSuite",
    "failed":  true
}
#< CLIXML
<Objs Version="1.1.0.1" xmlns="http://schemas.microsoft.com/powershell/2004/04"><S S="Error">Import-Module : The specified module 'C:\Program Files (x86)\Powershell _x000D__x000A_</S><S S="Error">Community Extensions\pscx3\pscx\pscx.psd1'
 was not loaded because no valid _x000D__x000A_</S><S S="Error">module file was found in any module directory._x000D__x000A_</S><S S="Error">At C:\Users\Administrator\AppData\Local\Temp\ansible-tmp-1428440725.66-11251461_x000D__x000A_</S>
<S S="Error">2752776\win_unzip.ps1:295 char:13_x000D__x000A_</S><S S="Error">+             Import-Module 'C:\Program Files (x86)\Powershell Community _x000D__x000A_</S><S S="Error">Extension ..._x000D__x000A_</S><S S="Error">+ ~~~~~~~~~~~
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~_x000D__x000A_</S><S S="Error">~~~_x000D__x000A_</S><S S="Error">    + CategoryInfo          : ResourceUnavailable: (C:\Program File...\pscx\ps _x000D__x000A_</S><S S="Erro
r">   cx.psd1:String) [Import-Module], FileNotFoundException_x000D__x000A_</S><S S="Error">    + FullyQualifiedErrorId : Modules_ModuleNotFound,Microsoft.PowerShell.Comm _x000D__x000A_</S><S S="Error">   ands.ImportModuleCommand_x000D__x0
00A_</S><S S="Error"> _x000D__x000A_</S></Objs>
```

schwartzmx:
Hi @elventear, I actually just made an update to my module in my repo that may solve this problem, and hadn't pushed it over here.  Could you try it with https://github.com/schwartzmx/ansible-module-devel/commit/ad2b9aecc3c95025af855feea323f4cd9fe76940 update.  It addresses an issue where the pscx install was not given enough time to finish and is immediately attempted to be imported and fails.  This may fix your issue.  Let me know please,  thanks.

elventear:
@schwartzmx I didn't work for me, but there wasn't much output other than it had failed.

schwartzmx:
@elventear that error is occurring as if `recurse = 'true'`, because the error message occurs here in the code:

```
+        If ($recurse) {
+            Fail-Json "Error recursively expanding $src to $dest"
+        }
+        Else {
+            Fail-Json "Error expanding $src to $dest"
+        }
```

which means that it will attempt to recursively decompress all files within the specified `src`,  the real reason for this is if you have something like a `zip` file that contains a bunch of other `zip` files that you would also like to decompress.  Can you give me more of an example, such as the playbook run and also what the contents of the zip are.

Or pscx is not being installed correctly on your machine, I'm not sure.

schwartzmx:
@elventear were you able to trouble shoot this some more?  Or any more feedback you can give me?  I can't reproduce your error.

gregdek:
See discussion in #553 about which module we choose.

gregdek:
I presume this PR is ready for re-review -- @elventear and friends?

perzizzle:
I tested out the unzip portion without any issues

gregdek:
@trondhindenes @petemounce @elventear @smadam813 @jhawkesworth @angstwad @sivel @chrishoffman @cchurch one more review and we can get this pulled in...

trondhindenes:
the module works as it should. I have a few comments however:
1. I don't get the "restart" option. Why would I restart my box after expanding a zip? The module doesn't execute the unpacked contents or anything, so that option will just end up confusing users.
2. Idempodency: I'd like to see some idempodency here, even a simple "creates" option that lets me control whether the task will always rerun or not.

Apart from that, the module works as advertised on my machine.

schwartzmx:
@trondhindenes good point about the restart, that should be handled in another task in the playbook (if needed).  As for the changes what exactly do you mean?  

Check if the dest exists, and if so return `changed = false` and exit?  I think if that were implemented as well, we would need another param to specify whether to overwrite, or delete the dest (if exists) and then unzip the artifact/package?  Thoughts?  I will implement it, no problem just looking for a little more clarification.  Thanks for your review.

trondhindenes:
"Check if the dest exists, and if so return  changed = false  and exit?". Yep, pretty much. You can of course go fancy and do all kinds of checks to find out if the target folder already contains the contents of the zip, but as a minimal way of implementing idempodency, I like the "creates" option which is used in a bunch of other Ansible modules - see http://docs.ansible.com/command_module.html for instance. 
If a certain file exists --> return changed=false and skip execution. I'm a sucker for avoiding "false" changes :-)

schwartzmx:
makes complete sense.  seems like pretty easy implementation will have it done pretty quickly.  Thanks again

schwartzmx:
@trondhindenes hope I understood correctly, added a new commit removing the `restart` ability and added the `creates` param

trondhindenes:
Yep, looks good! I'll do another test tomorrow, not expecting any
surprises. Thanks!

trondhindenes:
Darnit, I'm out of azure credits, so my test env is down for a few days. Will have to look at these during the weekend :-(

schwartzmx:
@trondhindenes no worries.  thanks

gregdek:
Hey @trondhindenes any luck here?

trondhindenes:
Sorry for the delay. I'l take a look tomorrow!

trondhindenes:
@schwartzmx looks good. I just have one more comment before +1-ing this: The "rm" parameter doesnt do validation, so it's possible to do "rm=roger" or whatever. Please use the convertto-bool function to validate boolean expressions (this will take care of cases where the playbook specifies "true","yes" or "1" and the corresponding "false","no" and "0". The function will return $false if it's unable to convert the argument to boolean, so you can use it like this (sorry for the crappy line breaks):
`If ($params.rm -and (!(convertto-bool ($params.rm)))
{
fail-json etc
}
Elseif ($params.rm)
$rm = convertto-bool ($params.rm)
}
Else
{
$rm = $false
}`

schwartzmx:
@trondhindenes the `rm` does do validation [here](https://github.com/ansible/ansible-modules-extras/pull/174/files#diff-27f1011a4fdde39510433b0090452c9aR72) but I think your method is probably the better way.  Will have the changes done today.

Edit: ah, nevermind.  I see your point that it won't validate that it is an actual boolean.  :+1:

schwartzmx:
@trondhindenes added changes, thanks for your feedback.

trondhindenes:
@schwartzmx Phil, my bad. I just realized now testing your module that the convertto-bool actually behaves differently than what I had thought. It will always return false unless the param is 1, true or "yes" so your original logic was fine. That's a possible weakness in the powershell.ps1 file we might have to address later. So, your code should probably just be 

`If ($params.rm)  { 
    $rm = ConvertTo-Bool ($params.rm)
}  
Else { 
    $rm = $false 
}`

@gregdek totally my bad on that last one. When Phil makes that change I'm +1-ing this one without further ado.

schwartzmx:
@trondhindenes no worries :).

I `--amend`ed the changes to the last commit.

trondhindenes:
two thumbs up, this looks good :+1: :+1:  :-)

gregdek:
Thanks all!

PR #177, opened 2014-12-31
win_zip Module

schwartzmx:
`win_zip` module:
- Uses PowerShell Community Extensions' (PSCX) `Write-Zip` cmdlet to zip the specified file or directory.
  - The reason for this, is implementing zipping of files/directories in PowerShell natively is ugly. Files are sometimes skipped and other weird anomalies have been noted of happening.
  - PSCX `Write-Zip` uses 7zip as the core library, and is much more reliable.
- If PSCX is not detected on the machine, it will be downloaded and installed, then imported for the script. 
- Also takes a `rm` argument, for specifying whether to remove the `src` file after compression is completed.
- The `dest` argument's path must exist, but the basename does not need to exist.
  - The script detects whether the `.zip` extension for `dest` was supplied, and adds it if need be.
- I've tried to make exception handling and the resultant error messages as specific and helpful as possible.

**Updates 1/11**:
- Added `tar`, `gzip`, `bzip` functionality.
  - Added param `type` to take care of this.
    - Ex: `type: gzip`
- Fixes to `PSCX` import problems after installation
- Updated docs
- Also added attempt to install `PSCX` with `chocolatey` if present on the machine, if not continue with the `.msi` installation.
- Updates to `win_unzip` module PR: [#174](https://github.com/ansible/ansible-modules-extras/pull/174) have been made to coincide with the changes here.

Questions, comments, or suggestions for additions please let me know.

Phil

### DIFF
--- a/windows/win_zip.ps1
+++ b/windows/win_zip.ps1
@@ -95,7 +95,7 @@ If ($params.dest) {
     }
 
 
-    If (-Not ([System.IO.Path]::GetExtension($dest) -match ".zip")) {
+    If (-Not ([System.IO.Path]::GetExtension($dest) -eq ".zip")) {
         $dest = $dest + ".zip"
     }
 
@@ -105,7 +105,7 @@ Else {
 }
 
 #RM
-If ($params.rm){
+If ($params.rm -eq "true" -Or $params.rm -eq "yes"){
     $rm = $true
 }
 Else {
--- a/windows/win_zip.ps1
+++ b/windows/win_zip.ps1
@@ -141,6 +141,7 @@ Catch {
 
 Set-Attr $result.win_zip "src" $src.toString()
 Set-Attr $result.win_zip "dest" $dest.toString()
+Set-Attr $result.win_zip "rm" $rm.toString()
 
 Exit-Json $result;
 
--- a/windows/win_zip.ps1
+++ b/windows/win_zip.ps1
@@ -39,20 +39,32 @@ $isContainer = $false
 $list = Get-Module -ListAvailable
 # If not download it and install
 If (-Not ($list -match "PSCX")) {
+    # Try install with chocolatey
     Try {
-        $client = New-Object System.Net.WebClient
-        $client.DownloadFile($url, $dest)
+        cinst -force PSCX -y
+        $choco = $true
     }
     Catch {
-        Fail-Json $result "Error downloading PSCX from $url and saving as $dest"
+        $choco = $false
     }
-    Try {
-        msiexec.exe /i $dest /qb
-    }
-    Catch {
-        Fail-Json $result "Error installing $dest"
+    # install from downloaded msi if choco failed or is not present
+    If ($choco -eq $false) {
+        Try {
+            $client = New-Object System.Net.WebClient
+            $client.DownloadFile($url, $dest)
+        }
+        Catch {
+            Fail-Json $result "Error downloading PSCX from $url and saving as $dest"
+        }
+        Try {
+            Start-Process -FilePath msiexec.exe -ArgumentList "/i $dest /qb" -Verb Runas -PassThru -Wait | out-null
+        }
+        Catch {
+            Fail-Json $result "Error installing $dest"
+        }
     }
     Set-Attr $result.win_zip "pscx_status" "pscx was installed"
+    $installed = $true
 }
 Else {
     Set-Attr $result.win_zip "pscx_status" "present"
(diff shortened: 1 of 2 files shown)


t-v:
Is there a specific reason you auto-download/install and install pscx?
I believe the strength of ansible relies on the fact that you don't need software on your clients to run it, only configuration.

I would rather see it happen like this:
http://blogs.msdn.com/b/daiken/archive/2007/02/12/compress-files-with-windows-powershell-then-package-a-windows-vista-sidebar-gadget.aspx

Especially since I don't like things to be automatically installed on my systems.

(I have the same argument for: https://github.com/ansible/ansible-modules-extras/pull/174)

schwartzmx:
What you linked will only work for `.zip` files, and it is very fragile depending on the size of the files hence the code:

```
foreach($file in $input) 
    { 
            $zipPackage.CopyHere($file.FullName)
            **Start-sleep -milliseconds 500**
    }
```

That doesn't look reliable to me or ensure consistency.

PSCX is very powerful and allows much more added functionality than compressing and decompressing files/folders.  I understand that you don't want things to be automatically installed.  

I can change it to fail if PSCX isn't installed, but the reason I did it this way is because I was constantly bringing up new machines, and I didn't want it to fail when doing application configuration so it was autoinstalled.  I can change it though and just add more steps in my playbook, using `win_get_url` and use `win_msi` to install PSCX.  

Pertaining to these modules, I can change it to do a check for PSCX, and use a similar method to that article for just `.zip` files and if the extension is otherwise, then I could check if it's installed to be used for `.msu` `.bz2` `.tar` and `.gz` files.

gregdek:
See discussion in #553 about which module we choose.

schwartzmx:
@gregdek, @trondhindenes replicated the discussed (from #174) changes and proposals to this module as well.

gregdek:
Thanks @schwartzmx. Windows reviewers, please take a look: @trondhindenes @petemounce @elventear @smadam813 @jhawkesworth @angstwad @sivel @chrishoffman @cchurch

gregdek:

# 

gregdek:
Another ping to our Windows friends: @trondhindenes @petemounce @elventear @smadam813 @jhawkesworth @angstwad @sivel @chrishoffman @cchurch

trondhindenes:
@schwartzmx I wouldn't assume that the pscx module is in any specified location. From how I read the code you first test that it exists, and if so you should be able to just import it (or at least use the psmodulepath attribute from the module object to get to the actual module dir instead of assuming its in C:\program files (x86). 

I would Update the documentation to specify that pscx needs to be install and in a path where Powershell can auto-load it ($env:PSModulePath), and then import the module using auto-loading.

windows/win_zip.ps1
```diff
@@ -0,0 +1,241 @@
...
+Else {
+    $pscxPresent = $true
+    Set-Attr $result.win_zip "pscx_status" "present"
+}
+
+# Import
+Try {
+    If ($pscxPresent) {
+        Try {
+            Import-Module 'C:\Program Files (x86)\Powershell Community Extensions\pscx3\pscx\pscx.psd1'
```
trondhindenes: This is the part I don't get

schwartzmx: Yep good point @trondhindenes,  the reason that was like this is because when I originally wrote this I was auto-installing PSCX and the module couldn't import it as it was running in the same PS session, so I had to import it from the install location.  Just a bad lingering snippet.  Will fix.

trondhindenes:
There are now two modules in PR that do pretty much the same: #999 and #177. Maybe figure out which one to go forward with so we don't waste anyone's coding cycles? ping @schwartzmx @davidobrien1985

davidobrien1985:
Hey @trondhindenes mine only uses native PowerShell and doesn't need any external prereqs. Then again, probably not my call?!

schwartzmx:
@trondhindenes @davidobrien1985 We can use whomever's.  They both have the same idea in mind but have different functionalities. 

This one does bzip (pscx), gzip (pscx), tar (pscx) and zip (natively if pscx isn't available).  The other only does zip it appears.

Both have dependencies,  here PSCX (if doing non-zip format, i.e. gzip).
The other, .NET 4.5.

Whichever way we go is fine with me.  I don't mind if we close this, but since https://github.com/ansible/ansible-modules-extras/pull/174 got merged and is consistent with this one,  if we close this one, we should revert https://github.com/ansible/ansible-modules-extras/pull/174 as well.

jhawkesworth:
Not really contributed to either, but how about we have both?  @schwartzmx 's modules support other formats so they are have utility.  I'd suggest adding _pscx onto the end of @schwartzmx's module names the so they are differentiated, or perhaps call them win_archive/ win_unarchive.  I guess it would fairest if this was named something like win_zip_net45 but seems a little overkill when you get .net 4.5 or later installed with most recent windows.  Ultimately as long as its clear which version to chose from the module docs I see no harm in having both modules.

bcoca:
actually having a single module that can do it 'both ways' might be more interesting, not sure what advantages/disadvantages each way has. It seems it would be easy enough to detect and use one and fail to the other or add an option to let the user decide which one to use.

PR #179, opened 2014-12-31
Fix syntax error in gluster_volume docs

gundalow:
Previously make webdocs failed

The documentation would read better if you could put newlines in, but I can't see any way of doing that.

```
ScannerError: while scanning a simple key
  in "<string>", line 16, column 9:
            started/stopped make sure it is  ... 
            ^
could not found expected ':'
  in "<string>", line 17, column 3:
      cluster:
      ^
*** ERROR: MODULE MISSING DOCUMENTATION: ../lib/ansible/modules/extras/system/gluster_volume.py, gluster_volume ***
make[1]: *** [modules] Error 1
```


bcoca:
aside from a merge conflict, this just seems to change the outer quoting. Also a previous commit seems to already have fixed the error you mention

Issue #13, opened 2014-09-29
locale_gen does not generate locales on Arch Linux


tummychow:
##### Issue Type:

Bug report
##### Ansible Version:

1.6.10
##### Environment:

Arch Linux (both master and slave)
##### Summary:

locale_gen doesn't generate locales on Arch Linux.

The [locale_gen](https://github.com/ansible/ansible/blob/devel/library/system/locale_gen#L66) module looks for lines starting with a hash and a space, and removes the hash to enable locale generation for those lines. But Arch's locale.gen doesn't have spaces (only the hash) so no lines get uncommented. Since no locales are enabled, locale-gen does nothing.
##### Steps To Reproduce:

Create a fresh Arch Linux system and install openssh/python2 on it. Add it to your Ansible hosts.

```
$ ansible -m locale_gen -a "name=en_US.UTF-8 state=present" --extra-vars "ansible_python_interpreter=/usr/bin/python2" <your Arch Linux target here>
$ ssh <your arch linux target here> locale -a
```
##### Expected Results:

The Ansible ad-hoc command should succeed and indicate "changed"
The ssh command `locale -a` should list all the locales on the system. `en_US.UTF-8` should be on the list.
##### Actual Results:

The Ansible command does succeed, but C and POSIX are the only locales listed aftewards. `en_US.UTF-8` does not appear.

Note: clone of ansible/ansible#8387

bcoca:
closing this as fixed by #181

PR #181, opened 2015-01-03
Allow every type of locales + archlinux fix

sterfield:
Hi,

Here's a pull request in order to fix several issues on locale_gen module:
- The previous version of this code was supporting only locales using the format "[language]_[territory].[charset]". But all the locales that doesn't have this format were not installable (for example, "fr_FR" or "fr_FR@euro").
- If an invalid locales was provided, the module kept sending a "changed" status.
- This module was not working for archlinux (see Issue #13 )

Now:
- if the user provides an invalid locales, the module failed. Locales are verified using /etc/locale.gen or /usr/share/i18n/SUPPORTED if Ubuntu
- Every types of valid locales are now supported, even if there's no "." in it (which was the previous verification)
- The locale module was not working on Archlinux, as there's no space between the "#" and the locale. This is now supported. Credits goes to danderson189, this is his code.

This module was tested on debian jessie, ubuntu 14 LTS and last Archlinux.

Thanks !

### DIFF
--- a/system/locale_gen.py
+++ b/system/locale_gen.py
@@ -4,6 +4,7 @@
 import os
 import os.path
 from subprocess import Popen, PIPE, call
+import re
 
 DOCUMENTATION = '''
 ---
--- a/system/locale_gen.py
+++ b/system/locale_gen.py
@@ -36,6 +36,26 @@ EXAMPLES = '''
 # location module specific support methods.
 #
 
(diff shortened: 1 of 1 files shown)


jpic:
Tested -> working for me !

Note that on my box I also need #221 for locale_gen to work properly on Arch.

Let's hope it's in the next Ansible release.

PR #187, opened 2015-01-07
Update lvol doc to clarify no impact on file system

### LINKED ISSUE #113: Document whether resize2fs is used with lvol
Or any other filesystem resizing.  It's not clear from the documentation whether I need to run this as a separate step or whether ansible does it automagically.

daysofwineandroses:
Updates doc description of the lvol module that it does not affect the file system when resizing.
Fixes #113.

### DIFF
--- a/system/lvol.py
+++ b/system/lvol.py
@@ -25,6 +25,7 @@ module: lvol
 short_description: Configure LVM logical volumes
 description:
   - This module creates, removes or resizes logical volumes.
+  - Note that when resizing, only the logical volume will be resized, not the file system on top.
 version_added: "1.1"
 options:
   vg:


alxgu:
Or should we add the function to extend the overlaying filesystem?

daysofwineandroses:
Even though it would be convenient to have lvol be able to do that, I'm not really sure if it's the best idea to implement it in the lvol module itself.

File system resizing is not completely without risk and it feels safer to me to have to make the conscious decision to resize in a different task.

`man 8 lvresize` also suggest to first shrink the file system on the volume before shrinking the volume itself and this is something I definitely would not want to happen by accident/automagically.

Just my two cents.

alxgu:
There is already a note in the documentation

> Notes:  Filesystems on top of the volume are not resized.

So if you put/extend it in the description, it should be removed from the bottom.

bcoca:
closing as notes already document this

daysofwineandroses:
Agreed, missed that.

PR #191, opened 2015-01-09
Allow disabling of autorefresh for zypper repositories

jlec:
In case of release repositories or other special cases you might not
need the autorefreshing of the repos. This patch adds a configure
option instead of hard enabling this.

Signed-off-by: Justin Lecher jlec@gentoo.org

### DIFF
--- a/packaging/os/zypper_repository.py
+++ b/packaging/os/zypper_repository.py
@@ -2,6 +2,7 @@
 # encoding: utf-8
 
 # (c) 2013, Matthias Vogelgesang <matthias.vogelgesang@gmail.com>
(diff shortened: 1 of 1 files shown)


bcoca:
to keep current behavior I would make this default to yes

jlec:
Fair enough, I will update the PR soon.

jlec:
@bcoca isn't it that way already? 

refresh = dict(required=False, default='yes', type='bool'),

bcoca:
not in the docs:
        default: "no"

jlec:
True, fixed now.

PR #41, opened 2014-10-02
adding pacman system upgrade feature

n0vember:
Adds the feature of system upgrade, equivalent to "pacman -Su".

PR #193, opened 2015-01-10
fix pacman upgrade bug

newportd:
**Issue Type:**
Bugfix Pull Request

**Ansible Version:**
1.8.2

**Environment:**
Controlling and managing nodes are all Arch Linux (latest)

**Summary:**
When attempting to upgrade a specific package via pacman the package was not upgraded.

**Steps to Reproduce:**
Install an out-of-date package (e.g. openssl 1.0.1.j).

Run the following play:

``` yaml
# This should cause openssl to be upgraded to 1.0.1.k
- name: upgrade openssl
  pacman: name=openssl state=present update_cache=yes
```

**Expected Results:**
The package should be updated to the latest version in the repository (in the above example to 1.0.1.k).

**Actual Results:**
The package was not upgraded to the latest version.

### DIFF
--- a/packaging/os/pacman.py
+++ b/packaging/os/pacman.py
@@ -42,7 +42,7 @@ options:
             - Desired state of the package.
         required: false
         default: "present"
-        choices: ["present", "absent"]
+        choices: ["present", "absent", "latest"]
 
     recurse:
         description:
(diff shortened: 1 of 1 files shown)


novln:
I would prefer an upgrade option like apt module or this pull request #41 

However, thank you, from another pacman lover, for this pull request.

newportd:
Good point. After reviewing the apt and yum modules I like the state=latest approach rather than what I have implemented here. I can change course with this PR if that seems agreeable.

I'm planning to keep this one focused on upgrading a single package (or list of packages) since it seems like #41 handles full system upgrades quite well.

PR #195, opened 2015-01-11
Add Elasticsearch plugin module

ThePixelDeveloper:
First time really playing with Python and Ansible modules.

This manages elasticsearch plugins and will allow you to add and remove them easily. This has been a bit of issue with the elasticsearch playbooks out there as none of them support existential checks as well as removal. Main reason for that being the [name -> directory transformation](https://github.com/elasticsearch/elasticsearch/blob/master/src/main/java/org/elasticsearch/plugins/PluginManager.java#L657-L689)

Conforms to pep8.

Suggestions appreciated.

### DIFF
--- a/packaging/elasticsearch_plugin.py
+++ b/packaging/elasticsearch_plugin.py
@@ -50,8 +50,9 @@ options:
         default: None
     timeout:
         description:
-            - Timeout setting: 30s, 1m, 1h... (1m by default)
+            - Timeout setting: 30s, 1m, 1h...
         required: False
+        default: 1m
     plugin_bin:
         description:
             - Location of the plugin binary
--- a/packaging/elasticsearch_plugin.py
+++ b/packaging/elasticsearch_plugin.py
@@ -3,8 +3,6 @@
 
 import os
 
-from ansible.module_utils.basic import *
-
 """
 Ansible module to manage elasticsearch plugins
 (c) 2015, Mathew Davies <thepixeldeveloper@googlemail.com>
@@ -158,5 +156,6 @@ def main():
 
     module.exit_json(changed=True, cmd=cmd, name=name, state=state, url=url, timeout=timeout, stdout=out, stderr=err)
 
-if __name__ == "__main__":
-    main()
+from ansible.module_utils.basic import *
+
+main()
--- a/packaging/elasticsearch_plugin.py
+++ b/packaging/elasticsearch_plugin.py
@@ -130,8 +130,6 @@ def main():
     version = module.params["version"]
 
     present = is_plugin_present(parse_plugin_repo(name), plugin_dir)
-    
-    print state
 
     # skip if the state is correct
     if (present and state == "present") or (state == "absent" and not present):
--- a/packaging/elasticsearch_plugin.py
+++ b/packaging/elasticsearch_plugin.py
@@ -1,0 +1,160 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+import os
+
+from ansible.module_utils.basic import *
+
+"""
+Ansible module to manage elasticsearch plugins
+(c) 2015, Mathew Davies <thepixeldeveloper@googlemail.com>
+
+This file is part of Ansible
+
+Ansible is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+Ansible is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+You should have received a copy of the GNU General Public License
+along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+"""
+
+DOCUMENTATION = '''
+---
+module: elasticsearch_plugin
+short_description: Manage Elasticsearch plugins
+description:
+    - Manages Elasticsearch plugins.
+version_added: ""
+author: Mathew Davies (@ThePixelDeveloper)
+options:
+    name:
+        description:
+            - Name of the plugin to install
+        required: True
+    state:
+        description:
+            - Desired state of a plugin.
+        required: False
+        choices: [present, absent]
+        default: present
+    url:
+        description:
+            - Set exact URL to download the plugin from
+        required: False
+    timeout:
+        description:
+            - Timeout setting: 30s, 1m, 1h... (1m by default)
+        required: False
+    plugin_bin:
+        description:
+            - Location of the plugin binary
+        required: False
+        default: /usr/share/elasticsearch/bin/plugin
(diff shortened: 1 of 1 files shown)


mscherer:
Could you squash all the commit in one ?

ThePixelDeveloper:
@mscherer Done

andsens:
At @secoya we needed something like that as well, [our module](https://github.com/andsens/ansible-modules-extras/blob/b9343fdf006cd93f17d3e2fef29e345c746d0c3d/database/misc/es_plugin.py) additionally manages the plugin _version_. I thought I'd post this here if you'd like some inspiration on how to manage that.

It doesn't really make sense for us to submit a PR that duplicates what you have done.

ThePixelDeveloper:
You can manage the version in the official ES way of doing it, ie

```
lukas-vlcek/bigdesk/2.4.0
```

although I think it's a good idea to split that out into a version variable and have it deal with behind the scenes. Will work on something this afternoon.

gregdek:
@ThePixelDeveloper thanks for submitting this module to Ansible Extras. Apologies that it’s taken a while to get your module reviewed.

To help facilitate reviews, we’ve broadened the number of people who can approve modules for inclusion into the Extras repository. The list of official reviews can be found here:

skornehl:
@ThePixelDeveloper: I like the idea but I believe this module should have a state=latest and/or a version. Nevertheless we can add this later.

But for the Checklist:

timeout and url need a default
Author should be set, name and github id at least

ThePixelDeveloper:
Cool thanks, will sort this out today sometime.

ThePixelDeveloper:
@skornehl Made the changes and squashed into the first commit. 

Url doesn't need a specified default (other than None) because it's not a required argument.

bcoca:
@ThePixelDeveloper reverse that. 
if requried == true => no default needed, makes no sense since user must always input
if require == false => default needs to be documented, even if just null or None

You need a default because the user is NOT forced to input a value

ThePixelDeveloper:
@bcoca You're right, documented

packaging/elasticsearch_plugin.py
```diff
@@ -0,0 +1,162 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+import os
+
+from ansible.module_utils.basic import *
```
bcoca: so this is not a real import, we recommend you add it to the bottom of the module to allow for easier debugging.

packaging/elasticsearch_plugin.py
```diff
@@ -0,0 +1,162 @@
...
+
+    rc, out, err = module.run_command(cmd)
+
+    if rc != 0:
+        reason = parse_error(out)
+        module.fail_json(msg=reason)
+
+    module.exit_json(changed=True, cmd=cmd, name=name, state=state, url=url, timeout=timeout, stdout=out, stderr=err)
+
+if __name__ == "__main__":
```
bcoca: this won't make the module importable, as the 'basic' is not a real import

packaging/elasticsearch_plugin.py
```diff
@@ -0,0 +1,162 @@
...
+        default: present
+    url:
+        description:
+            - Set exact URL to download the plugin from
+        required: False
+        default: None
+    timeout:
+        description:
+            - Timeout setting: 30s, 1m, 1h... (1m by default)
+        required: False
```
bcoca, line 52: default in description, but we want a field for use in formatting in ansible-doc and the docs.ansible.com site

packaging/elasticsearch_plugin.py
```diff
@@ -0,0 +1,162 @@
...
+            - Location of the plugin binary
+        required: False
+        default: /usr/share/elasticsearch/bin/plugin
+    plugin_dir:
+        description:
+            - Your configured plugin directory specified in Elasticsearch
+        required: False
+        default: /usr/share/elasticsearch/plugins/
+    version:
+        description:
```
bcoca, line 65: missing required

packaging/elasticsearch_plugin.py
```diff
@@ -0,0 +1,162 @@
...
+    plugin_dir = module.params["plugin_dir"]
+    name = module.params["name"]
+    state = module.params["state"]
+    url = module.params["url"]
+    timeout = module.params["timeout"]
+    version = module.params["version"]
+
+    present = is_plugin_present(parse_plugin_repo(name), plugin_dir)
+    
+    print state
```
bcoca: modules should not print like this, I'm guessing its left over debug?

packaging/elasticsearch_plugin.py
```diff
@@ -0,0 +1,162 @@
...
+
+    # skip if the state is correct
+    if (present and state == "present") or (state == "absent" and not present):
+        module.exit_json(changed=False, name=name)
```
bcoca, line 143: this does not account for version

packaging/elasticsearch_plugin.py
```diff
@@ -0,0 +1,162 @@
...
+
+    plugin_bin = module.params["plugin_bin"]
+    plugin_dir = module.params["plugin_dir"]
+    name = module.params["name"]
+    state = module.params["state"]
+    url = module.params["url"]
+    timeout = module.params["timeout"]
+    version = module.params["version"]
+
+    present = is_plugin_present(parse_plugin_repo(name), plugin_dir)
```
bcoca, line 139: this should probably also check for version

packaging/elasticsearch_plugin.py
ThePixelDeveloper, line 65: Added

packaging/elasticsearch_plugin.py
ThePixelDeveloper: Moved

packaging/elasticsearch_plugin.py
ThePixelDeveloper, line 139: Unfortunately ES doesn't store versions installed. It just extracts the plugin into a directory. The plugin tool doesn't specify versions either. Wouldn't make sense to pass the version here.

```
➜  ~elasticsearch  ./bin/plugin -l
Picked up JAVA_TOOL_OPTIONS: -javaagent:/usr/share/java/jayatanaag.jar 
Picked up _JAVA_OPTIONS: -Dawt.useSystemAAFontSettings=lcd                       -Dsun.java2d.xrender=true
Installed plugins:
    - head
    - bigdesk
```

bcoca, line 139: then we might wan't to make a note: if plugin exists with previous version, it will NOT be updated

ThePixelDeveloper, line 139: Done

bcoca:
adding to my list of ' todo when i have free time ' trying to figure out how to get version info to be able to implement an update/downgrade for 'present' when versions are different.

ThePixelDeveloper:
Thanks @bcoca 

One thing I missed was defining a version_added as I didn't know when it was going to be merged: https://github.com/ansible/ansible-modules-extras/blob/devel/packaging/elasticsearch_plugin.py#L31

PR #200, opened 2015-01-15
win_acl Module

schwartzmx:
module `win_acl`:  **Access Control**
- Module allows for adding or removing `Allow`/`Deny` rules (permissions) on files and/or directories on Windows hosts.

See `win_acl.py` for documentation on `params`.

Suggestions, concerns, comments, anything? Let me know please.

Phil

### DIFF
--- a/windows/win_acl.ps1
+++ b/windows/win_acl.ps1
@@ -30,8 +30,8 @@ $result = New-Object psobject @{
 If ($params.src) {
     $src = $params.src.toString()
 
-    If (-Not (Test-Path -Path $src -PathType Leaf -Or Test-Path -Path $src -PathType Container)) {
-        Fail-Json $result "$src is not a valid file or directory on the host"
+    If (-Not (Test-Path -Path $src)) {
+        Fail-Json $result "$src file or directory does not exist on the host"
     }
 }
 Else {
--- a/windows/win_acl.ps1
+++ b/windows/win_acl.ps1
@@ -1,0 +1,146 @@
+#!powershell
+# This file is part of Ansible
+#
+# Copyright 2014, Phil Schwartz <schwartzmx@gmail.com>
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+# WANT_JSON
+# POWERSHELL_COMMON
+
+# win_acl module (File/Resources Permission Additions/Removal)
+$params = Parse-Args $args;
+
+$result = New-Object psobject @{
+    win_acl = New-Object psobject
+    changed = $false
+}
+
+If ($params.src) {
+    $src = $params.src.toString()
+
+    If (-Not (Test-Path -Path $src -PathType Leaf -Or Test-Path -Path $src -PathType Container)) {
+        Fail-Json $result "$src is not a valid file or directory on the host"
+    }
+}
+Else {
+    Fail-Json $result "missing required argument: src"
+}
+
+If ($params.user) {
+    $user = $params.user.toString()
+
+    # Test that the user/group exists on the local machine
+    $localComputer = [ADSI]("WinNT://"+[System.Net.Dns]::GetHostName())
+    $list = ($localComputer.psbase.children | Where-Object { (($_.psBase.schemaClassName -eq "User") -Or ($_.psBase.schemaClassName -eq "Group"))} | Select-Object -expand Name)
+    If (-Not ($list -contains "$user")) {
+        Fail-Json $result "$user is not a valid user or group on the host machine"
+    }
+}
+Else {
+    Fail-Json $result "missing required argument: user.  specify the user or group to apply permission changes."
+}
+
+If ($params.type -eq "allow") {
+    $type = $true
+}
+ElseIf ($params.type -eq "deny") {
+    $type = $false
+}
+Else {
+    Fail-Json $result "missing required argument: type. specify whether to allow or deny the specified rights."
+}
+
+If ($params.inherit) {
+    # If it's a file then no flags can be set or an exception will be thrown
+    If (Test-Path -Path $src -PathType Leaf) {
+        $inherit = "None"
+    }
+    Else {
+        $inherit = $params.inherit.toString()
+    }
+}
+Else {
+    # If it's a file then no flags can be set or an exception will be thrown
+    If (Test-Path -Path $src -PathType Leaf) {
+        $inherit = "None"
+    }
+    Else {
+        $inherit = "ContainerInherit, ObjectInherit"
+    }
+}
+
+If ($params.propagation) {
+    $propagation = $params.propagation.toString()
+}
+Else {
+    $propagation = "None"
+}
+
+If ($params.rights) {
+    $rights = $params.rights.toString()
+}
+Else {
+    Fail-Json $result "missing required argument: rights"
+}
+
+If ($params.state -eq "absent") {
+    $state = "remove"
+}
+Else {
+    $state = "add"
+}
+
+Try {
+    $colRights = [System.Security.AccessControl.FileSystemRights]$rights
+    $InheritanceFlag = [System.Security.AccessControl.InheritanceFlags]$inherit
+    $PropagationFlag = [System.Security.AccessControl.PropagationFlags]$propagation
+
+    If ($type) {
+        $objType =[System.Security.AccessControl.AccessControlType]::Allow
+    }
+    Else {
+        $objType =[System.Security.AccessControl.AccessControlType]::Deny
+    }
+
+    $objUser = New-Object System.Security.Principal.NTAccount($user)
+    $objACE = New-Object System.Security.AccessControl.FileSystemAccessRule ($objUser, $colRights, $InheritanceFlag, $PropagationFlag, $objType)
+    $objACL = Get-ACL $src
+
+    If ($state -eq "add") {
+        Try {
+            $objACL.AddAccessRule($objACE)
+        }
+        Catch {
+            Fail-Json $result "an exception occured when adding the specified rule.  it may already exist."
+        }
+    }
+    Else {
+        Try {
+            $objACL.RemoveAccessRule($objACE)
+        }
+        Catch {
+            Fail-Json $result "an exception occured when removing the specified rule.  it may not exist."
+        }
+    }
+
+    Set-ACL $src $objACL
+
+    $result.changed = $true
+}
+Catch {
+    Fail-Json $result "an error occured when attempting to $state $rights permission(s) on $src for $user"
+}
+
+Exit-Json $result
--- a/windows/win_acl.ps1
+++ b/windows/win_acl.ps1
@@ -1,7 +1,7 @@
 #!powershell
 # This file is part of Ansible
 #
-# Copyright 2014, Phil Schwartz <schwartzmx@gmail.com>
+# Copyright 2015, Phil Schwartz <schwartzmx@gmail.com>
 #
 # Ansible is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


gregdek:
@schwartzmx, thanks for submitting this module to Ansible Extras. Apologies that it’s taken a while to get your module reviewed.

To help facilitate reviews, we’ve broadened the number of people who can approve modules for inclusion into the Extras repository. The list of official reviews can be found here:

gregdek:
Also pinging potential Windows-expert reviewers: @trondhindenes @petemounce @elventear @smadam813 @jhawkesworth @cchurch

windows/win_acl.ps1
```diff
@@ -0,0 +1,146 @@
...
+            $objACL.RemoveAccessRule($objACE)
+        }
+        Catch {
+            Fail-Json $result "an exception occured when removing the specified rule.  it may not exist."
+        }
+    }
+
+    Set-ACL $src $objACL
+
+    $result.changed = $true
```
elventear: Always sets the results to changed? Even if the same ACL permissions where applied?

schwartzmx: Would it be better to have it moved into the `Try` blocks where the rules are either added or removed?

Also instead of fail at https://github.com/ansible/ansible-modules-extras/pull/200/files#diff-557730cf2740ef29928a75ee8482055aR134 and https://github.com/ansible/ansible-modules-extras/pull/200/files#diff-557730cf2740ef29928a75ee8482055aR126,  it should probably just `Exit-Json`

elventear: I haven't tested it, but what happens if you run the module two times in a row? Ideally the second time should not be marked as changed.

I glanced quickly at the API, and you could use `$objACL.GetAccessRules` to get a collection of `AccessRule`s. Then it is a matter of testing if the `AccessRule` you wanted to add already exists or not.

schwartzmx: Got it.  I found a way to compare them pretty easily.

``` PowerShell
$objACL.Access | % { If (($_.FileSystemRights -eq $objACE.FileSystemRights) -And ($_.AccessControlType -eq $objACE.AccessControlType) -And ($_.IdentityReference -eq $objACE.IdentityReference) -And ($_.IsInherited -eq $objACE.IsInherited) -And ($_.InheritanceFlags -eq $objACE.InheritanceFlags) -And ($_.PropagationFlags -eq $objACE.PropagationFlags)) { $match = $true }}
```

should do the trick, and based on that I can add/remove or notadd/notremove and set changed appropriately.  Thanks for the feedback.

Made it a little more readable in my commit (instead of a one liner) d72bb17.

schwartzmx:
@elventear would you might re-reviewing my latest changes from last week?  Also pinging @trondhindenes for a possible review.  Thanks

elventear:
:+1:

gregdek:
OK, we have a +1. Need another from any of @trondhindenes @petemounce @smadam813 @jhawkesworth @angstwad @sivel @chrishoffman @cchurch. 

Once we have that next +1, we can bring @schwartzmx into the family of reviewers, and he has to pay it forward. Incentive for you all. :)

trondhindenes:
@schwartzmx I just noticed this module reports changed=true on second run of the same playbook. Is this by design?

schwartzmx:
@trondhindenes hmm... no it should not.  I added this check [here](https://github.com/ansible/ansible-modules-extras/pull/200/files#diff-557730cf2740ef29928a75ee8482055aR124) which should detect if the rule exists or does not exist and add/remove appropriately.  I will have to investigate...  Could you leave me an example of the task/module call with args?  I will do some testing and see if I find a bug.

Edit: I just did another test in PS using the same method in the module for testing:
test

which seems fine.  Objects aren't equal but the comparison works.

Wonder if it's something in the logic.  Will see if I can find out.

schwartzmx:
@trondhindenes, looking for a little help here...  The only thing I can think of is the `IsInherited` flag changing somehow between your runs as this is not explicitly stated when adding an `ACL`.  If the top level permissions changed to force inheritance it would have reported `changed` again.  To fix that I could just remove that from the comparison because it isn't explicitly altered from this module and is independent, tbh. What do you think?

thanks

trondhindenes:
Sorry this took so long. Looking into it now.

windows/win_acl.ps1
```diff
@@ -0,0 +1,165 @@
...
+
+$result = New-Object psobject @{
+    win_acl = New-Object psobject
+    changed = $false
+}
+
+If ($params.src) {
+    $src = $params.src.toString()
+
+    If (-Not (Test-Path -Path $src -PathType Leaf -Or Test-Path -Path $src -PathType Container)) {
```
trondhindenes: This line actually fails when I run the compiled script on my node. Are you simply testing whether the path exists - either in form of a file or container? In that case you can just do `if (test-path -path $src) {`

schwartzmx: yep just testing that the path exists.  Will change to the shorter check

gregdek:
Looks like revisions mean it's time to go back into community review... @trondhindenes ?

trondhindenes:
@schwartzmx, I just retested this. I have a local user on my node, "Peter". I also have a folder, C:\acltest.

When I run
`ansible -m win_acl -a "user=peter src=C:\acltest type=allow rights='FullControl'" thhighperf4`

It works, but it reports changed every time. The reason for this seems to be that when you're doing the comparison against rules in  `$objACL.Access`, those objects report the username as "REALM\User", such as "TH-HIGHPERF4\Peter" which is the case here.

So I trued running ansible with the user in that format:
`ansible -m win_acl -a "user=TH-HIGHPERF4\Peter src=C:\acltest type=allow rights='FullControl'" thhighperf4`

That failed, since Ansible isn't able to find the user using that format.
I think you might need to look into a little bit more "smartness" around finding the user. Using psbase basically lists a truckload of things we don't need and slows down your script considerably, so I'd recommend you googling around a bit and look at the different methods for listing/verifying local(and domain) users. Let me know if you need any help with that.

schwartzmx:
@trondhindenes thanks for the testing and figuring out the issue.  If you are familiar with this could you help me out?  I've trying some other approaches from searching and came up with these, but they still don't seem complete.

``` PowerShell
#Quick way to get local domain accounts, but nothing else
Get-Wmiobject -class Win32_UserAccount | ForEach-Object {Write-Host $_.Caption}
```

Some other tries..

``` PowerShell
#Using net localgroup
$localGroups = net localgroup | where {$_ -and $_ -notmatch "command completed successfully" } | select -skip 2 
ForEach ($group in $localGroups) {
    $group = $group.SubString(1)
    $fmtCmd = "net localgroup '${group}' | where {`$_ -and `$_ -notmatch 'command completed successfully' } | select -skip 4"
    $(Invoke-Expression "$fmtCmd")
}
```

``` PowerShell
$localComputer = [System.Net.Dns]::GetHostName()
$groups = Get-Wmiobject -class Win32_Group 
ForEach ($groupname in $groups) {
      $groupname = $groupname.Name
      $group = [ADSI]("WinNT://$localComputer/$groupname,group")
      $group.PSBase.Invoke('Members') | % {
        $_.GetType().InvokeMember('Name', 'GetProperty', $null, $_, $null)
      }  
}
```

Any suggestions?

trondhindenes:
I'll take a look at it. The way I see it (and to emulate the "permissions" dialog in Windows), any "resolvable" entity (account or group) can be given explicit permissions. This can be a local user/group, or a domain user/group in a realm the computer trusts (if it's domain-joined). So all we need is basically to be able to verify that the supplied user/group is resolvable in some way or another - probably easiest by using a adsi lookup and just ensure that the lookup returns something valid. I'll fork your PR see if I can come up with a simple way to do this.

trondhindenes:
@schwartzmx I wasnt able to fork your repo since I already have a fork with the same name, so:
I created a generic function for verifying a local/domain user/group (can be found here: https://gist.github.com/trondhindenes/d99180d886a4e570b890). I've implemented this in your win_acl and put that whole thing here: https://gist.github.com/trondhindenes/05c12636dfac4da08a38

Would you have a look at it and commit it if it looks good? The UserSearch function is implemented so that you can throw at it:
-local user without realm: "Administrator"
-local group without realm: "Administrators"
-local user with realm: "MYPC\Administrator"
-local group with realm: "MYPC\Administrators"
-Domain user with realm: "MYDOMAIN\JohnB"
-Domain group with realm: "MYDOMAIN\Domain Users"

UserSearch will normalize this and output a user with realm, which is the format get-acl uses so comparison should work.

trondhindenes:
Yup, just tested all kinds of combos now, and it seems to be working with my humble additions.

schwartzmx:
@trondhindenes awesome work! And thanks for taking the time to help me out I really appreciate it.  Will be committing it shortly.  Also would you mind if I added you as an additional author in the documentation?

trondhindenes:
@schwartzmx no worres! I don't mind being lister as an author.

Btw, +1 from me on this now.

windows/win_acl.py
```diff
@@ -0,0 +1,147 @@
...
+
+DOCUMENTATION = '''
+---
+module: win_acl
+version_added: "2.0"
+short_description: Set file/directory permissions for a system user or group.
+description:
+     - Add or remove rights/permissions for a given user or group for the specified src file or folder.
+options:
+  src:
```
bcoca: src is normally used for files you are copying from, in this case path/dest seem more appropriate

windows/win_acl.py
```diff
@@ -0,0 +1,147 @@
...
+    description:
+      - File or Directory
+    required: yes
+    default: none
+    aliases: []
```
bcoca: no need for empty aliases

windows/win_acl.py
```diff
@@ -0,0 +1,147 @@
...
+version_added: "2.0"
+short_description: Set file/directory permissions for a system user or group.
+description:
+     - Add or remove rights/permissions for a given user or group for the specified src file or folder.
+options:
+  src:
+    description:
+      - File or Directory
+    required: yes
+    default: none
```
bcoca: if requried is true, default makes no sense

bcoca:
minor doc notes, once resolved we can merge, looks great

schwartzmx:
@bcoca thanks for the feedback, I made the appropriate changes.  Let me know if there is anything else.  
Thanks

schwartzmx:
Any more updates needed here guys?  @bcoca @gregdek @trondhindenes

trondhindenes:
I'm all for getting this merged.

cchurch:
Trying this one against Windows Server 2008 R2 with the following task: 

```
- win_acl:
    path: "C:\\mywebdir"
    user: "IIS_IUSRS"
    rights: FullControl
    type: allow
    state: present
    inherit: "ContainerInherit, ObjectInherit"
    propagation: "None"
```

Gives me the following error (I also had to modify the try/catch blocks to see):

`Exception calling "AddAccessRule" with "1" argument(s): "Some or all identity references could not be translated.`

I was able to get it to work by returning the group SID from the `UserSearch` function, then using `System.Security.Principal.SecurityIdentifier($user)` in place of `System.Security.Principal.NTAccount($user)`.

With this modification, it works for my particular use case, but always returns "changed" even when no additional ACE was added.

I would suggest trying to return the exception message from your try/catch blocks instead of a generic "an exception occured" message.

And... your examples in the docs use `src` instead of `path`.

schwartzmx:
@cchurch Thanks for the test and feedback.  The exception message and docs is an easy fix (nice catch).  Any ideas on how this should be solved?  Should we just use the SID instead of the UserName?  I'm not sure why it always returns true in that instance, it should return false from my check where it iterates through all permissions doing a compare.  Maybe the SID is transformed to the Username once the rule is added, so it is never caught as a match when accessing the current permissions (which have Usernames).

thanks again.  I will try to think of a good way of doing this, I just don't have a Windows 2008 instance to test on.

gregdek:
Adding new process boilerplate. We will be evaluating all new module PRs according to this process, effective immediately.

# 

gregdek:
@cchurch the issue you describe for this module -- is it a blocking issue, in your opinion?

If it is, please respond with "needs_revision" in the text and we will tag appropriately.

If not, seems like this is pretty close to mergeable -- I'd say in that case respond with "shipit" and we will ship it.

Thanks @schwartzmx for the patience here. Seems like we're pretty close on this one.

schwartzmx:
@gregdek thanks I updated the docs with @cchurch's findings,  and Chris I haven't been able to think of a good way to solve for Windows 2008 case.  Any ideas would be appreciated from any of the windows reviewers.

Thanks!

h0nIg:
i would appreciate if this module will be included regardless if there is a problem with a specific windows version. this bug should be addressed later, since its a pain to work with acls and powershell :(

bcoca:
talked to @cchurch, merging this for now, need to open ticket to address the 2008 issue

h0nIg:
added pull request to fix win 2008 issue: https://github.com/ansible/ansible-modules-extras/pull/1111

PR #208, opened 2015-01-20
created openstack heat module for ansible

RCFans:
Hi guys!

I've created an openstack heat module for ansible, please kindly review and accept, thanks!

Cheers,

Jiaxing

### DIFF
--- a/cloud/openstack/head.py
+++ b/cloud/openstack/head.py
@@ -1,0 +1,239 @@
+#!/usr/bin/python
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


emonty:
Hi! Thanks for the module! We've been working on refactoing most of the OpenStack modules - so I'd like to get this refactored to match before be land it. I'll take a stab at doing that and put up a response.

emonty:
Ok. I've cherry-picked this on to #92 and updated it to use the common openstack stuff and the shade library. I've also taken the code that does the logic around how to interact with heat and put it in to the shade library in upstream OpenStack here: https://review.openstack.org/#/c/151900/ I changed a couple of the parameters for consistency - so you probably want to try it out and see if it still does what you need and/or give feedback if I've destroyed the world.

PR #214, opened 2015-01-21
Add EUC-JP locale name normalization

zinrai:
The function normalizes checks for UTF-8, but the same issue exists for
other locales as well.  This fix adds normalization for EUC-JP, a Japanese
locale.


bcoca:
can we look at a more general fix vs adding each particular case? otherwise i can see a very long replace chain.

zinrai:
Here is a version that puts the normalization rules in a dict, making is more general.  What do you think?

bcoca:
probably the best we can do as I cannot find a rule that would work for all or even most cases

Issue #146, opened 2014-12-10
monitoring/monit.py won't recognize the name of a process which contains uppercase letters


TobiasPottgueter:
##### Issue Type:

Bug Report
##### Ansible Version:

1.7.2 (tested)
branch: devel (i looked into the code, see Possible Solution)
##### Environment:

Centos 6.6
##### Summary:

monitoring/monit.py won't recognize a process name if the name contains uppercase letters.
sorry, but that's all.
##### Steps To Reproduce:

```
- name: myMonitProcess
  monit: name=myMonitProcess state=unmonitored
```

will result in

```
msg: myMonitProcess process not presently configured with monit
```

if "monit summary" is displaying:

```
Process 'myMonitProcess'  Running
System 'localhost' Running
```
##### Expected Results:

I expect it to work with uppercase process names as it is working with lowercase only process names, and mixedcase process names self speaking.
##### Actual Results:

Unless i use completely lowercase process names it won't work.
##### Possible Solution:

Sorry, i don't have a test environment, but i managed to identify a possible solution:

line 78 and line 79

```
parts = line.lower().split()
if len(parts) > 2 and parts[0] == 'process' and parts[1] == "'%s'" % name:
```

'line' should NOT be converted to lower and 'process' should then be 'Process' (uppercase P)

wimnat:
TobiasPottgueter can you please test my PR resolves your issue?

bcoca:
closing via merged PR

PR #216, opened 2015-01-22
Changed status() to be case-sensitive of process

wimnat:
Code change as per solution proposed in https://github.com/ansible/ansible-modules-extras/issues/146.

### DIFF
--- a/monitoring/monit.py
+++ b/monitoring/monit.py
@@ -75,8 +75,8 @@ def main():
             # Sample output lines:
             # Process 'name'    Running
             # Process 'name'    Running - restart pending
-            parts = line.lower().split()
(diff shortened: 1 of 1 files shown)


bcoca:
a better fix would be to just make parts[0].lower(), this will be resilient to upstream case changes

wimnat:
Do you mean to add parts[0].lower() in addition to the current change?  lower() has to be removed from the parts = statement so that the actual process name is unchanged.

bcoca:
this patch:

```
-            parts = line.lower().split()
-            if len(parts) > 2 and parts[0] == 'process' and parts[1] == "'%s'" % name:
+            parts = line.split()
+            if len(parts) > 2 and parts[0].lower() == 'process' and parts[1] == "'%s'" % name:
```

wimnat:
Yep, agreed.  PR updated.

PR #218, opened 2015-01-22
Allow sending direct messages with slack module

wkf:


azat-aimfund:
Why this is not in production yet??

PR #223, opened 2015-01-23
New Module: Yum config manager

maxamillion:


system/yum_config_manager.py
```diff
@@ -0,0 +1,134 @@
...
+    rc, out, err = module.run_command(cmd)
+
+    if rc != 0:
+        module.fail_json(
+            msg="rc of commmand is non-zero",
+            cmd=args,
+            stdout=out,
+            stderr=err,
+            rc=rc,
+            changed=True,
```
alxgu: changed should be "False" in a failed case

alxgu:
I get an error with following arguments: -a "state=enabled name=unix"
NameError: global name 'args' is not defined

Check line 121,122 and 112
NameError: global name 'run_result' is not defined.

maxamillion:
Yeah, that was a horrible oversight on my part. I admittedly wrote this for someone, they said it worked great and asked if I would submit it upstream. However I'm starting to reconsider that because the `yum-config-manager` util doesn't seem to provide useful exit codes. It's basically always 0 even if you provide a reponame that doesn't exist on the machine.

maxamillion:
For sake of personal sanity and keeping things idempotent and reporting honestly, I'm going to close this. I might re-open later if we can get some improvements into yum-config-manager upstream though.

PR #224, opened 2015-01-24
Support empty locale.gen

jpic:
Apparently, locale.gen is really distro-specific and is generated by the glibc distro package, so there's not really a sane default we can rely on. Anyway, if `existing_line` is not in `/etc/locale.gen`, then the module will always report as having "changed" and will never make a change. This patch fixes that issue by appending `new_line` if it's not in `lines` before writing to `/etc/locale.gen`.


jpic:
Closed in favor of #181

PR #225, opened 2015-01-24
Ansible support Consul

sgargan:
##### Issue Type:

Feature Pull Request
##### Ansible Version:

1.9
##### Environment:

OSX 10.10 (Yosemite)
Linux (Ubuntu 14.04 Trusty Tahr) 3.13.0-37-generic
##### Summary:

Initial commit of Ansible support for the Consul clustering framework (http://consul.io).

Submission includes support for
- creating and registering services and checks
- reading, writing and lookup for values in consul's kv store
- creating and manipulating sessions for distributed locking on values in the kv
- creating and manipulating ACLs for restricting access to the kv store
- inventory support that reads the Consul catalog and group nodes according to
  - datacenters
  - exposed services
  - service availability
  - arbitrary groupings from the kv store

This submission makes extensive use of the python-consul library and this is required
as a dependency and can be installed from pip.
##### Steps To Reproduce:

The tests were written to target a vagrant cluster which can be setup by following the
instructions here http://github.com/sgargan/consul-vagrant


sgargan:
Tests for these modules are included in this pull request ansible/ansible/pull/10078

hoffoo:
+1 for this

clustering/consul
```diff
@@ -0,0 +1,463 @@
...
+    ''' registers a service with the the current agent '''
+    result = service
+    changed = False
+
+    consul_api = get_consul_api(module)
+    existing = get_service_by_id(consul_api, service.id)
+
+    # there is no way to retreive the details of checks so if a check is present
+    # in the service it must be reregistered
+    if service.has_checks() or not(existing or existing == service):
```
hoffoo: I am wondering about this - in my use case I always want the service to be registered even if there aren't any checks. Could has_checks() be optional or make it so the service is always reregistered?

Removing this line works for me (ie if i change tags on a service they are properly updated in consul)

matlockx:
+1

fourcube:
+1

clustering/consul
```diff
@@ -0,0 +1,463 @@
...
+
+    def __init__(self, check_id, name, node=None, host='localhost',
+                    script=None, interval=None, ttl=None, notes=None):
+        self.check_id = self.name = name
+        if check_id:
+            self.check_id = check_id
+        self.script = script
+        self.interval = str(interval)
+
+        if not self.interval.endswith('m') or self.interval.endswith('s'):
```
mattupstate: I'm not really a fan of this conditional. It appears as if you're trying to set a reasonable default but this could have bad side effects. The different between `1s` and `1m` is quite significant and as the value increases the difference gets larger. I would suggest an error condition when the unit is not specified to avoid confusion, and thus this line and the following should be removed (see additional comment). Additionally, if I specify `interval: '10s'`, the service never gets registered.

clustering/consul
```diff
@@ -0,0 +1,463 @@
...
+       return ConsulCheck(
+            module.params.get('check_id'),
+            module.params.get('check_name'),
+            module.params.get('check_node'),
+            module.params.get('check_host'),
+            module.params.get('script'),
+            module.params.get('interval'),
+            module.params.get('ttl'),
+            module.params.get('notes')
+        )
```
mattupstate, line 322: I would recommend changing the above return to:

``` python
check = ConsulCheck(
    module.params.get('check_id'),
    module.params.get('check_name'),
    module.params.get('check_node'),
    module.params.get('check_host'),
    module.params.get('script'),
    module.params.get('interval'),
    module.params.get('ttl'),
    module.params.get('notes')
)
duration_units = ['ns', 'us', 'ms', 's', 'm', 'h']
if not all([(check.interval.endswith(suffix) for suffix in duration_units)]):
    module.fail_json(
        msg='you must specify a unit (%s) for check intervals' %
        duration_units.join(', '))
return check
```

clustering/consul
sgargan: Marin,

The intention here is that the service should only be registered if it is
different from what is already registered. This is so we can report
'changed' correctly You can't retrieve the details for the checks from the
api so there is now way to tell if the checks have changed so you have to
assume so and re-register the service.

 if service.has_checks() or not(existing or existing == service):

This is saying if the user supplies checks or the service provided is not
the same as the existing registered check. The tags are included in the
equals comparison so it should re-register if you change a tag. If you're
seeing different this is definitely an issue. I'll put together a test to
make sure this is working.

hoffoo:
It would be really nice if we could show a human readable error if connecting to consul failed. May be its easiest to try /v1/status/leader or something of the like when first connecting, rather than checking every method's return.

The current error is

```
invalid output was: Traceback (most recent call last):
  File "<stdin>", line 1806, in <module>
  File "<stdin>", line 453, in main
AttributeError: 'ConnectionError' object has no attribute 'read'
```

edit: i guess what im actually talking about is here: https://github.com/ansible/ansible-modules-extras/pull/225/files#diff-4e08945c99d467491755b42981552a74R457

clustering/consul
sgargan: I took a look at this and the logic is incorrect, if you change the tags it
does not reregister the service. I've make a fix and added some tests which
I'll submit soon.

sgargan: Yes this is not very useful, I'll get it to emit something more informative
if the connection fails.

clustering/consul
sgargan, line 322: I've tweaked what you've supplied and made this change in the check class. Thanks.

mattupstate, line 322: Right on. Just be sure to update your doc sting above.

clustering/consul
```diff
@@ -0,0 +1,464 @@
...
+            notes=dict(required=False),
+            port=dict(default=8500, type='int'),
+            script=dict(required=False),
+            service_id=dict(required=False),
+            service_name=dict(required=False),
+            service_port=dict(required=False, type='int'),
+            state=dict(default='present', choices=['present', 'absent']),
+            tags=dict(required=False, type='list'),
+            token=dict(required=False),
+            url=dict(default='http://localhost:8500')
```
hoffoo: Doesn't look like this is being used, not sure if its even needed. I guess maybe if someone has a proxy in front of consul and the url may be different?

edit: i mean the url var

dublx:
+1

ratherScript:
+1

clustering/consul
```diff
@@ -0,0 +1,465 @@
...
+
+try:
+    import json
+except ImportError:
+    import simplejson as json
+
+try:
+    import consul
+except ImportError, e:
+    print "failed=True msg='python-consul required for this module. "\
```
bcoca: this needs to be a fail_json using the module, no prints and no exits in modules

clustering/consul
```diff
@@ -0,0 +1,465 @@
...
+    state:
+        description:
+          - register or deregister the consul service, defaults to present
+        required: true
+        choices: ['present', 'absent']
+    service_id:
+        description:
+          - the ID for the service, must be unique per node, defaults to the
+            service name
+        required: false
```
bcoca, line 61: when not required default: should be present in docs

clustering/consul
```diff
@@ -0,0 +1,465 @@
...
+    host:
+        description:
+          - host of the consul agent with which to register the service,
+            defaults to localhost
+        required: false
+    notes:
+        description:
+          - Notes to attach to check when registering it.
```
bcoca, line 75: requried?

clustering/consul
sgargan: This took the lead from other modules importing third party libraries and It does fail if the library is not present. I'll look at importing the library in the scope of the module and use the fail_json instead.

sgargan:
I've been back through the documentation annotating the required and default flags. It might be an idea to introduce 'optional' and 'depends_on' states for parameters
- Optional is roughly ~ required false, default None, 'I'll use it if provided but you don't have to'
- depends_on, is a soft required, required if another parameter is present.

I've also modified the imports to fail via module.fail_json to avoid the print and exit(1).

bcoca:
i like the depends_on, not sure about the optional as it seems redundant (though shorthand) for existing options.

PR #226, opened 2015-01-24
Add pushbullet module

willybarro:
Add module to send push notifications via https://www.pushbullet.com/

Pushbullet is a free service for sending push notifications. They have native apps for android, ios and a chrome extension.

I've followed all the guidelines and tested the "make webdocs", so, it may be an easy pull request to review :)

I'm not sure what should be put on "version_added", so I added 1.8.

### DIFF
--- a/notification/pushbullet.py
+++ b/notification/pushbullet.py
@@ -95,6 +95,7 @@ EXAMPLES = '''
 
 try:
     from pushbullet import PushBullet
+    from pushbullet.errors import InvalidKeyError, PushError
 except ImportError:
     pushbullet_found = False
 else:
@@ -131,8 +132,11 @@ def main():
         module.fail_json(msg="Python 'pushbullet.py' module is required. Install via: $ pip install pushbullet.py")
 
     # Init pushbullet 
-    pb = PushBullet(api_key)
-    target = None
+    try:
+        pb = PushBullet(api_key)
+        target = None
+    except InvalidKeyError:
+        module.fail_json(msg="Invalid api_key")
 
     # Checks for channel/device
     if device is None and channel is None:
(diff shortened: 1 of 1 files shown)


theshapguy:
+1

notification/pushbullet.py
```diff
@@ -0,0 +1,174 @@
...
+
+DOCUMENTATION = '''
+---
+author: Willy Barro
+requirements: [ pushbullet.py ]
+module: pushbullet
+short_description: Sends notifications to Pushbullet
+description:
+   - This module sends push notifications via Pushbullet to channels or devices.
+version_added: "1.8"
```
resmo: this should be "2.0"

notification/pushbullet.py
```diff
@@ -0,0 +1,174 @@
...
+
+    # Init pushbullet 
+    pb = PushBullet(api_key)
+    target = None
+
+    # Checks for channel/device
+    if device is not None and channel is not None:
+        module.fail_json(msg="You can't use both device and channel at the same time.")
+
+    if device is None and channel is None:
```
resmo, line 142: this is ok. just wanted to point out there is a option for AnsibleModule:

```
mutually_exclusive = (
    ['device', 'channel'],
),
```

willybarro:
Hey guys... for some reason I didn't receive an email/nofitication with those considerations. I'll fix that today and update the branch :) Thanks for you considerations.

willybarro:
And here we go... code updated.

willybarro:
@bcoca Travis CI was failing this one. @sivel just added (5 days ago) a python 2.4 compatibility requirement. What do you think: Should I "downgrade" my code to remove the dict comprehension or not?

bcoca:
is the pushbullet lib 2.4 compatible?

willybarro:
@bcoca Haven't tested that myself because I just found out this new requirement. But their tests are working on 2.6 and up.

bcoca:
we are 'upping' testing in preparation for v2, hoping to help minimize the impact, of course this points out some existing discrepancies. If the underlying library your module uses does not support 2.4 we can flag this module as requiring python >=2.6 and ignore it in the 2.4 tests.

willybarro:
Great. Those flags will work just fine! It just needs to be >= 2.7, because I'm using dict comprehensions (that's why tests for this module was failing on 2.4).

abadger:
If the underlying libs support 2.6 then our code should as well.  I can't
remember if dict comprehensions were added in 2.7; if so we shouldn't use
them here.

sivel:
dict comprehensions were added in 2.7, and are not available in 2.6.  The pushbullet lib appears to support python 2.6.  So the dict comprehension should probably be replaced.

willybarro:
@sivel @abadger Here we go https://github.com/ansible/ansible-modules-extras/pull/492

PR #229, opened 2015-01-28
Add gce_img module for utilizing GCE image resources

tanpeter:


cloud/google/gce_img.py
```diff
@@ -0,0 +1,174 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+"""An Ansible module to utilize GCE image resources."""
+
+import sys
+
+# import module snippets
```
sivel: These module snippets should actually appear as the line directly before the call to `main()`

cloud/google/gce_img.py
```diff
@@ -0,0 +1,174 @@
...
+from ansible.module_utils.gce import *
+
+try:
+  from libcloud.compute.types import Provider
+  from libcloud.compute.providers import get_driver
+  from libcloud.common.google import GoogleBaseError
+  from libcloud.common.google import ResourceNotFoundError
+  _ = Provider.GCE
+except ImportError:
+  print('failed=True '
```
sivel: This type of "error" here will not be interpreted by ansible correctly any more since "Baby JSON" support was removed in 1.8.

The new pattern should be something like:

``` python
try:
    import foo
    HAS_FOO = True
except ImportError:
    HAS_FOO = False

# Additional code here

def main():
    module = AnsibleModule(...)
    if not HAS_FOO:
        module.fail_json(msg="Python foo module is required")

    # remaining code
```

cloud/google/gce_img.py
```diff
@@ -0,0 +1,174 @@
...
+
+# Delete an image named test-image in zone us-central1-a.
+- gce_img:
+    name: test-image
+    zone: us-central1-a
+    state: deleted
+'''
+
+
+def main():
```
sivel, line 170: Perhaps just my opinion, but I like to see less code in `main` and more code moved into other functions.  It's not a whole lot of code, but I'd still recommend splitting the code up into smaller functions if possible.

cloud/google/gce_img.py
```diff
@@ -0,0 +1,174 @@
...
+  module = AnsibleModule(
+      argument_spec=dict(
+          name=dict(required=True),
+          source=dict(),
+          state=dict(default='present'),
```
sivel: You might want to also add in `choices` here.

cloud/google/gce_img.py
```diff
@@ -0,0 +1,174 @@
...
+  zone = module.params.get('zone')
+  changed = False
+
+  try:
+    image = gce.ex_get_image(name)
+  except GoogleBaseError, e:
+    module.fail_json(msg=str(e), changed=False)
+
+  # user wants to create an image.
+  if state in ['active', 'present'] and not image:
```
sivel: Along with adding `choices` to the `state` argument, you can simplify this logic, as well as similar logic to this farther down.

sivel:
cc @erjohnso

cloud/google/gce_img.py
```diff
@@ -0,0 +1,174 @@
...
+  from libcloud.common.google import ResourceNotFoundError
+  _ = Provider.GCE
+except ImportError:
+  print('failed=True '
+        "msg='libcloud with GCE support is required for this module.'")
+  sys.exit(1)
+
+DOCUMENTATION = '''
+---
+module: gce_img
```
erjohnso, line 23: Please add `version_added: "1.9"` here and remove it from the options below.

cloud/google/gce_img.py
```diff
@@ -0,0 +1,174 @@
...
+short_description: utilize GCE image resources
+description:
+    - This module can create and delete GCE private images from gzipped
+      compressed tarball containing raw disk data or from existing detached
+      disks in any zone. U(https://cloud.google.com/compute/docs/images)
+options:
+  name:
+    description:
+      - the name of the image to create
```
erjohnso: s/image to create/image to create or delete/

cloud/google/gce_img.py
```diff
@@ -0,0 +1,174 @@
...
+    image = gce.ex_get_image(name)
+  except GoogleBaseError, e:
+    module.fail_json(msg=str(e), changed=False)
+
+  # user wants to create an image.
+  if state in ['active', 'present'] and not image:
+    if not source:
+      module.fail_json(msg='Must supply a source', changed=False)
+
+    if source.startswith('https://storage.googleapis.com'):
```
erjohnso: Not mandatory, but might be more Google friendly if you also supported `gs://bucket/path/to/image.tgz` too.

cloud/google/gce_img.py
```diff
@@ -0,0 +1,174 @@
...
+  name = module.params.get('name')
+  source = module.params.get('source')
+  state = module.params.get('state')
+  zone = module.params.get('zone')
+  changed = False
+
+  try:
+    image = gce.ex_get_image(name)
+  except GoogleBaseError, e:
+    module.fail_json(msg=str(e), changed=False)
```
erjohnso: I'm guessing a 404 doesn't raise an exception or we wouldn't get very far creating an image.  But, I think this block is unnecessary and I'd remove the check in favor of just trying to create/destroy the image.

cloud/google/gce_img.py
```diff
@@ -0,0 +1,174 @@
...
+
+requirements: [ "libcloud" ]
+author: Peter Tan <ptan@google.com>
+'''
+
+EXAMPLES = '''
+# Create an image named test-image from the disk 'test-disk' in zone us-central1-a.
+- gce_img:
+    name: test-image
+    source: test-disk
```
erjohnso, line 89: Perhaps show an example with URI too (especially if you support `https://` and `gs://`).

erjohnso:
Thanks for contributing this @tanpeter! And thanks for the quick review @sivel.

tanpeter:
Ok, I got your idea now. My only concern is when <code>gce.ex_create_image()</code> returns successfully, I can't effectively know is it a new image or an existing image, then how do I flip the <code>changed</code> flag?

erjohnso:
I usually set `changed = False` early on and then only flip it to true if something has changed. In this case, when you call `gce.ex_create_image()` and the call succeeds, you can set `changed = True`. If the call errors out because the resource already exists, you can leave `changed = False`, but not throw a module error.  That gives us the idempotency that ansible strives to follow.

tanpeter:
This looks like a libcloud bug: when I pass in string parameter to `gce.ex_delete_image()` and the image doesn't exist, I got:

```
File ".ansible_module_generated", line 151, in delete_image return gce.ex_delete_image("fedora1")
File "/usr/local/lib/python2.7/dist-packages/libcloud/compute/drivers/gce.py", line 2172, in ex_delete_image
    request = '/global/images/%s' % (image.name)
AttributeError: 'NoneType' object has no attribute 'name'
```

And libcloud does the following when the parameter is not a GCENodeImage:

```
if not hasattr(image, 'name'):
            image = self.ex_get_image(image)
```

So there will be an `ex_get_image()` call in either cases.

erjohnso:
Hmm... what version of libcloud are you using?  It looks like trunk handles string and object OK unless I'm missing something: https://github.com/apache/libcloud/blob/trunk/libcloud/compute/drivers/gce.py#L3294

FWIW, the plan is to release libcloud 0.17.0 this weekend, so if there is a bug I can get it merged in today.

tanpeter:
Please look at line 3307, if the image does not exist, `self.ex_get_image()` would return None; then referencing None.name will crash.

erjohnso:
Ah, gotcha!  Thanks, I'll wrap that with a try/except NotFound error.  I'll /cc you on the PR.

Then you'll be able to do something like,

```
def delete_image(gce, image, module):
  """Delete a specific image resource."""
  try:
     image = gce.ex_get_image(image)
     image.delete()
  except ResourceNotFoundError:
     pass
  except GoogleBaseError, e:
     module.fail_json(msg=str(e), changed=False)
```

erjohnso:
@tanpeter - ok, merged https://github.com/apache/libcloud/pull/441

erjohnso:
@bcoca - Hi Brian, new GCE module ready for merge (unless you want to review). With the recent org change, I wanted an ACK before just hitting the merge button.  Are you (in general) OK with me helping maintain GCE modules?

bcoca:
looks good to me. It is fine if you continue in charge of the GCE modules as you have up to now.

erjohnso:
Thanks Brian.

PR #231, opened 2015-01-31
Correctly report "changed: false" when trying to install a package not i...

### LINKED ISSUE #230: The pkgutil provider reports state: present when a package is not in the catalog
When I try to install a non-existing package with something like

```
# ansible unstable9s -m pkgutil -a "name=CSWlibpsl-dev state=present"
```

The result is success and the state "present" is reported:

```
unstable9s | success >> {
    "changed": true, 
    "name": "CSWlibpsl-dev", 
    "state": "present", 
    "stderr": "Package CSWlibpsl-dev not in catalog. Exiting.\n", 
    "stdout": "Solving needed dependencies ...\n"
}
```

The package was not installed on the machine and is not installed afterwards:

```
unstable9s% pkginfo -x CSWlibpsl-dev
ERROR: information for "CSWlibpsl-dev" was not found
```

dago:
...n the catalog. This fixes #230

### DIFF
--- a/packaging/os/pkgutil.py
+++ b/packaging/os/pkgutil.py
@@ -162,10 +162,15 @@ def main():
             (rc, out, err) = package_uninstall(module, name)
             out = out[:75]
 
-    if rc is None:
-        result['changed'] = False
-    else:
+    if rc == 0:
         result['changed'] = True
+    else:
+        result['changed'] = False
+
+    if rc is not None and rc != 0:
+        result['failed'] = True
+    else:
+        result['failed'] = False
 
     if out:
         result['stdout'] = out
--- a/packaging/os/pkgutil.py
+++ b/packaging/os/pkgutil.py
@@ -162,15 +162,14 @@ def main():
             (rc, out, err) = package_uninstall(module, name)
             out = out[:75]
 
-    if rc == 0:
-        result['changed'] = True
-    else:
+    if rc is None:
+        # pkgutil was not executed because the package was already present/absent
         result['changed'] = False
-
-    if rc is not None and rc != 0:
+    elif rc == 0
+        result['changed'] = True
+    else
+        result['changed'] = False
         result['failed'] = True
-    else:
-        result['failed'] = False
 
     if out:
         result['stdout'] = out


abadger:
I don't have a solaris box with pkgutil to check this but some of this looks a little strange.
- Is rc ever going to be None?
- Is rc a dependable indicator for failure?  ie: is trying to install an already installed package going to return rc=0?  Will attempting to upgrade a package that's already at the latest version return rc=0?  Will uninstalling a package that's not present on the system return rc=0?
- How do we differentiate between failure and no action?  It seems like return code might not be enough to do that?

dago:
Regarding your questions:
- If the package is already installed `package_install` is not called and rc stays with the preinitialized value `None`
- A package is not tried to be installed if it is already present as the existence is checked previously with `pkginfo -q`. From pkginfo the returned is dependable. If the operation was successful 0 is returned, otherwise a different returncode. Otherwise it is considered a bug in pkgutil. I am also in contact with Peter Bonivart, the author of pkgutil, to fix issues as they occur.
- If pkgutil was invoked for action and rc == 0 then something was changed, otherwise an error occurred or rc was still None and pkgutil was not invoked. The operation is failed if pkgutil was invoked which means rc != None where rc == 0 means success.

I do acknowledge that the twist with rc == None if pkgutil was not invoked is not clearly visible in the code, if you think this is useful I can refactor it.

abadger:
Ah, okay.  So maybe something like this:

``` python
if rc is None:
    # Did not run commands as no change needed
    result['changed'] = False
elif rc == 0
    result['changed'] = True
else:
    result['changed'] = False
    result['failed'] = True
```

It's a little quirky but by and large we don't set failed=False when everything succeeds.  Just set failed=True when there is a failure.

dago:
Yes, this is exactly the definitions I had in mind.

abadger:
@dago if you'd like to update the PR I'll merge this.

dago:
Pull request updated.

abadger:
Thanks!

# Merged

Hi!

issue is resolved.

Thank you!

PR #232, opened 2015-02-02
[slack] Remove unused "domain" option, clarify use of token

Jell:
The code for slack notification has changed, and the "domain" option was removed. At the same time, the meaning of "token" has been changed as well.

This PR cleans things up and give a short explanation of what the token should look like.

### DIFF
--- a/notification/slack.py
+++ b/notification/slack.py
@@ -26,14 +26,9 @@ description:
 version_added: 1.6
 author: Ramon de la Fuente <ramon@delafuente.nl>
 options:
-  domain:
-    description:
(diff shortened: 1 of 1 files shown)


phumpal:
:+1:

dltj:
:+1:

gregdek:
Looks like this needs to be rebased. @Jell, please rebase and then ping module owner @ramondelafuente.  If he gives +1, we will merge.

Jell:
this appears to actually no longer be relevant and has already been fixed in parallel of this PR (hence resulting in the current merge conflict).

Closing!

PR #233, opened 2015-02-02
Add Swap

umurkontaci:
The module allows creating and removing (the ones created with this module), temporary and permanent swaps.
## Overview

2 major steps.
1. Create swap file through fallocate (with fallback to dd) and call mkswap.
2. If it's permanent, create fstab and to persist on reboot.

Example use cases:
1. You need a permanent swap since some cloud hosts don't provide by default.
2. You need to make a build that requires a lot of RAM you add swap and remove it after the build.

PS: First PR on this repo, let me know if something is doesn't fit.

### DIFF
--- a/system/swap.py
+++ b/system/swap.py
@@ -1,0 +1,266 @@
+#!/usr/bin/python
+from subprocess import Popen
+import os
+
+DOCUMENTATION = '''
+---
+# If a key doesn't apply to your module (ex: choices, default, or
+# aliases) you can use the word 'null', or an empty list, [], where
+# appropriate.
+module: swap
+short_description: Add and remove swap
+description:
+    - Add and remove swaps
+    - Resize swaps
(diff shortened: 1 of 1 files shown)


bcoca:
Hi!

Thanks very much for your submission to Ansible.  It sincerely means a lot to us.

We are interested in this idea and would like to see a wider discussion on it
on one of our lists. 
- There is already a mount module that handles fstab, we would like to avoid having 2 modules updating the same config file
- There is already a filesystem module, it does not currently handle swap but it could be modified to do so easily.

So I'm going to close this ticket for now, we would like to incorporate the ability to manage swap, but in an effort to avoid module sprawl we would like to see it done as part of the existing modules above.

Can you please post on ansible-development list so we can talk about this idea with the wider group?
- https://groups.google.com/forum/#!forum/ansible-devel

Thank you once again for this and your interest in Ansible!

PR #234, opened 2015-02-03
pkgng: add 'batch' parameter

mwpher:
When installing certain packages such as BSDstats, you can get prompts asking for installation details. Thus, this adds a parameter to the pkgng module to skip mid-install interactive package prompts, by setting the environment variable $BATCH to "yes". This will automatically accept the defaults for any query posed during the installation. Obviously, this won't work for questions that have no defaults, or that do not use the ports system that recognizes this variable.

A good matching feature would be the ability to specify answers to these questions through the pkgng module.

Hopefully I implemented this correctly! Please let me know if I did anything wrong; I'm a fairly inexperienced programmer.

Thanks,
Matt


mwpher:
I'm sorry I left the merge commit in; I'll close this request and resubmit it once I've fixed it.

bcoca:
you can fix on same PR by just using --force on push

PR #235, opened 2015-02-03
pkgng: add 'batch' parameter

mwpher:
Some packages attempt to prompt the user for certain settings during
installation. Thus, this parameter sets the environment variable
$BATCH to 'yes', which forces package installation scripts to accept
default values for these interactive prompts. This should work for all
prompts that have a default value and aren't implemented through a
custom script (as this variable is built into the ports/package system).
This feature was created once I realized that package installs hang
when installing an interactive package without this variable set.
Hopefully someone smarter then me can fix that issue.

A good matching feature would be allowing the user to specify custom
answers to these package prompts.

Although this variable is integrated into the pkg/port system and thus
should work with most prompts, I have NOT tested this with any
packages besides bsdstats. Please let me know what I can improve
upon next time.

### DIFF
--- a/packaging/os/pkgng.py
+++ b/packaging/os/pkgng.py
@@ -63,6 +63,13 @@ options:
               for newer pkgng versions, specify a the name of a repository
               configured in /usr/local/etc/pkg/repos
         required: false
+    batch:
+        description:
+            - for packages with interactive prompts during installation,
+              this makes pkgng automatically accept all default options for
+              the installation of the package.
+        default: yes
+        required: false
 author: bleader
 notes:
     - When using pkgsite, be careful that already in cache packages won't be downloaded again.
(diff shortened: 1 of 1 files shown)


bcoca:
default should be true, not false (docs are correct, but code is not)

I would even be fine with this NOT being a parameter as ansible is always in "batch" when running modules.

mwpher:
While it's true I contradicted myself, it's actually the other way around; the code is correct for what I intended. I wanted to make it false by default, since I figured it would be better if the user explicitly enabled it (to make sure they understood that prompts were being skipped). At least, I figured that if it was on by default, the user should be notified that ansible was taking that action, but I wasn't sure how to make the module recognize when the user would've been prompted for input.

As you suggest, I'll change it so that it's on automatically. I'll change the pull request if possible.

bcoca:
All of the other package manager plugins do this also (apt has some that are not "skippable", but there is debconf for that).

updating pull request from shell: you just update the branch and push again against the branch that was used to create the PR

mwpher:
Committed! Anything else to change?

PR #238, opened 2015-02-06
New module apache2_site

webstandardcss:
This module will allow you to enable or disable your Apache virtualhosts
by wrapping a2ensite and a2dissite which creates or removes symbolic
links from /etc/apache2/sites-enabled/.

It is based on the core Ansible module called apache2_module which
similarly wraps a2enmod and a2dismod.

### DIFF
--- a/web_infrastructure/apache2_site.py
+++ b/web_infrastructure/apache2_site.py
@@ -1,0 +1,96 @@
+#!/usr/bin/python
+#coding: utf-8 -*-
+
+# (c) 2015-2016, Trey Brister <webstandardcss@gmail.com>
+#
+# This module is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


webstandardcss:
I have tested both the present and absent states and both appear to work just fine.   Please let me know if there are any questions.   I hope this module gets in core.

sivel:
I believe this is a duplicate of #104

webstandardcss:
@sivel and Paul Loundes, Thanks for bringing this to my attention. I closed the pull request. 

Note: Actually the other one was more heavily modified than mine to customize the text within the code.  I just made the minimum required edits to do the job. Mainly because I was confused whether the code was referencing an "ansible module" or an "apache2 module". That is why i only changed the required text.

PR #242, opened 2015-02-09
Add new module mongodb_replset

jirutka:
This module allows to initiate a new replica set or add new members to an existing one. It can’t remove members from a replica set, this should be always done with caution by hand.

_Copied from https://github.com/ansible/ansible/pull/8153._

### DIFF
--- a/database/misc/mongodb_replset.py
+++ b/database/misc/mongodb_replset.py
@@ -77,19 +77,17 @@ except ImportError:
     pymongo_found = False
 
 
-def load_mongocnf_creds():
-    """ Read credentials from ~/.mongodb.cnf file, when exists. """
+def read_mongocnf_creds():
+    """ Read credentials from ~/.mongodb.cnf file, when exists.
+
+    :return: tuple of username and password
+    """
     config = ConfigParser.RawConfigParser()
     try:
         config.read(os.path.expanduser('~/.mongodb.cnf'))
-        creds = {
-            'user': config.get('client', 'user'),
-            'password': config.get('client', 'pass')
-        }
+        return (config.get('client', 'user'), config.get('client', 'pass'))
     except (ConfigParser.Error, IOError):
-        return False
-
-    return creds
+        return (None, None)
 
 
 def replset_conf(client):
@@ -175,6 +173,9 @@ def main():
                                    'login_port', 'hosts', 'replica_set'])
     nodes = split_hosts(hosts)
 
+    if not user and not password:
+        user, password = read_mongocnf_creds()
+
     initiated = False
     try:
         try:
@@ -186,12 +187,6 @@ def main():
                 client = MongoClient(host, int(port), read_preference=ReadPreference.SECONDARY)
             else:
                 module.fail_json(msg="Unable to connect: %s" % e)
-
-        if not user and not password:
-            credentials = load_mongocnf_creds()
-            if credentials is not False:
-                user = credentials['user']
-                password = credentials['password']
 
         if user and password:
             try:
--- a/database/misc/mongodb_replset.py
+++ b/database/misc/mongodb_replset.py
@@ -1,0 +1,217 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2015, Jakub Jirutka <jakub@jirutka.cz>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: mongodb_replset
+author: Jakub Jirutka
+version_added: "1.9"
+short_description: Initiate or extend replica set.
+description:
+  - This module allows to initiate a new replica set or add new members to an
+    existing one. It can't remove members from a replica set, this should be
+    always done with caution by hand.
+options:
+  login_user:
+    description:
+      - The username used to authenticate with.
+    aliases: [ "user" ]
+    required: false
+  login_password:
+    description:
+      - The password used to authenticate with.
+    aliases: [ "password" ]
+    required: false
+  login_host:
+    description:
+      - The first host to connect when initiating the replica set.
+    aliases: [ "host" ]
+    required: false
+    default: localhost
+  login_port:
+    description:
+      - Port of the first host to connect when initiating the replica set.
+    aliases: [ "port" ]
+    required: false
+    default: "27017"
+  hosts:
+    description:
+      - A comma delimited list of replica set members.
+    aliases: [ "members" ]
+    required: true
+    example: mongo0,mango1:27011,mango2:27012
+  replica_set:
+    description:
+      - Name of the replica set to create or connect to.
+    aliases: [ "replset" ]
+    required: true
+    default: "rs0"
+'''
+
+import ConfigParser
+try:
+    from pymongo.errors import ConnectionFailure, OperationFailure, ConfigurationError
+    from pymongo import MongoClient
+    from pymongo.mongo_replica_set_client import MongoReplicaSetClient
+    from pymongo.read_preferences import ReadPreference
+    from pymongo.uri_parser import split_hosts
+    pymongo_found = True
+except ImportError:
+    pymongo_found = False
+
+
+def load_mongocnf_creds():
+    """ Read credentials from ~/.mongodb.cnf file, when exists. """
+    config = ConfigParser.RawConfigParser()
+    try:
+        config.read(os.path.expanduser('~/.mongodb.cnf'))
+        creds = {
+            'user': config.get('client', 'user'),
+            'password': config.get('client', 'pass')
+        }
+    except (ConfigParser.Error, IOError):
+        return False
+
+    return creds
+
+
+def replset_conf(client):
+    """ Return replica set configuration; the same as rs.conf() in shell.
+
+    :param client: initialized Mongo client
+    :return: replica set config
+    """
+    return client['local'].system.replset.find_one()
+
+
+def add_members(client, members):
+    """ Add new members to the replica set.
+
+    :param client: initialized Mongo client
+    :param members: list of tuples that defines hostnames and ports of the
+                    replica set members;
+                    example: `[(mango0, 27017), (mango1, 27018)]`
+    """
+    conf = replset_conf(client)
+    curr_hosts = [m['host'] for m in conf['members']]
+    new_hosts = [join_colon(x) for x in members]
+
+    new_id = max([int(x['_id']) for x in conf['members']]) + 1
+    for host in (set(new_hosts) - set(curr_hosts)):
+        conf['members'] += [{'_id': new_id, 'host': host}]
+        new_id += 1
+
+    conf['version'] += 1
+    client.admin.command('replSetReconfig', conf)
+
+
+def replset_initiate(client, name, members):
+    """ Initiate replica set with the specified members.
+
+    :param client: initialized Mongo client
+    :param name: name of the replica set to initiate
+    :param members: list of tuples that defines hostnames and ports of the
+                    replica set members;
+                    example: `[(mango0, 27017), (mango1, 27018)]`
+    """
+    hosts = [{'_id': idx, 'host': join_colon(val)} for idx, val in enumerate(members)]
+    conf = {'_id': name, 'members': hosts}
+    client.admin.command('replSetInitiate', conf)
+
+
+def join_colon(iterable):
+    return ':'.join(str(s) for s in iterable)
+
+
+def main():
+    module = AnsibleModule(
+        argument_spec={
+            'login_user':     {'aliases': ['user']},
+            'login_password': {'aliases': ['password']},
+            'login_host':     {'aliases': ['host'], 'default': 'localhost'},
+            'login_port':     {'aliases': ['port'], 'default': 27017},
+            'hosts':          {'aliases': ['members'], 'required': True},
+            'replica_set':    {'aliases': ['replset'], 'required': True}
+        }
+    )
+
+    if not pymongo_found:
+        module.fail_json(msg='Python module "pymongo" must be installed.')
+
+    user, password, host, port, hosts, replset = (
+        module.params[k] for k in ['login_user', 'login_password', 'login_host',
+                                   'login_port', 'hosts', 'replica_set'])
+    nodes = split_hosts(hosts)
+
+    initiated = False
+    try:
+        try:
+            client = MongoReplicaSetClient(hosts, replicaSet=replset,
+                                           read_preference=ReadPreference.PRIMARY)
+            initiated = True
+        except ConfigurationError, e:
+            if 'is not a member of replica set' in e.message:
+                client = MongoClient(host, int(port), read_preference=ReadPreference.SECONDARY)
+            else:
+                module.fail_json(msg="Unable to connect: %s" % e)
+
+        if not user and not password:
+            credentials = load_mongocnf_creds()
+            if credentials is not False:
+                user = credentials['user']
+                password = credentials['password']
+        elif not password and user:
+            module.fail_json(msg='When supplying login arguments, both login_user and login_password must be provided')
+
+        if user and password:
+            try:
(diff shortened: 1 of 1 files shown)


database/misc/mongodb_replset.py
```diff
@@ -0,0 +1,218 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: mongodb_replset
+author: Jakub Jirutka
+version_added: "1.8"
```
sivel: This should be `1.9`

jirutka: Okay, changed.

database/misc/mongodb_replset.py
```diff
@@ -0,0 +1,218 @@
...
+    pymongo = None
+
+
+def load_mongocnf_creds():
+    """ Read credentials from ~/.mongodb.cnf file, when exists. """
+    config = ConfigParser.RawConfigParser()
+    mongocnf = os.path.expanduser('~/.mongodb.cnf')
+
+    try:
+        config.readfp(open(mongocnf))
```
sivel: I would advocate for safer opening and closing of the file.  Right now there is no explicit close.

database/misc/mongodb_replset.py
```diff
@@ -0,0 +1,218 @@
...
+    conf = {'_id': name, 'members': hosts}
+    client.admin.command('replSetInitiate', conf)
+
+
+def join_colon(iterable):
+    return ':'.join(str(s) for s in iterable)
+
+
+def main():
+    module = AnsibleModule(
```
sivel, line 152: As of now, the current recommended way for instantiating dicts is by using the `dict` function instead of `{}`

database/misc/mongodb_replset.py
jirutka: Me too, I don’t know why I’ve written it so stupidly… I’ll fix it at evening.

database/misc/mongodb_replset.py
```diff
@@ -0,0 +1,218 @@
...
+            'login_port':     {'aliases': ['port'], 'default': 27017},
+            'hosts':          {'aliases': ['members'], 'required': True},
+            'replica_set':    {'aliases': ['replset'], 'required': True}
+        }
+    )
+
+    if not pymongo:
+        module.fail_json(msg='Python module "pymongo" must be installed.')
+
+    user, password, host, port, hosts, replset = (
```
sivel, line 167: Is this necessary?  Seems overly complicated and too "clever".  I'd advocate for something more explicit if this is really needed.

database/misc/mongodb_replset.py
jirutka, line 152: I prefer to follow the PEP8 style guide.

database/misc/mongodb_replset.py
jirutka, line 167: What’s wrong with it? I’m sure that pythonists are able to read list comprehensions…

database/misc/mongodb_replset.py
```diff
@@ -0,0 +1,218 @@
...
+    try:
+        try:
+            client = MongoReplicaSetClient(hosts, replicaSet=replset,
+                                           read_preference=ReadPreference.PRIMARY)
+            initiated = True
+        except ConfigurationError, e:
+            if 'is not a member of replica set' in e.message:
+                client = MongoClient(host, int(port), read_preference=ReadPreference.SECONDARY)
+            else:
+                module.fail_json(msg="Unable to connect: %s" % str(e))
```
sivel: Is the double cast to string necessary?  I believe `%s` should already handle it

database/misc/mongodb_replset.py
```diff
@@ -0,0 +1,218 @@
...
+                                 members="'%s'" % added_hosts)
+        else:
+            changed = False
+    else:
+        try:
+            replset_initiate(client, replset, nodes)
+        except OperationFailure, e:
+            module.fail_json(msg='Unable to initiate replica set: %s' % str(e))
+
+    module.exit_json(changed=changed)
```
sivel: This `exit_json` contains very little info.  Some of the `fail_json` statements contain more info.

database/misc/mongodb_replset.py
jirutka: To be honest, I don’t know, it’s a long time since I’ve written this. I’ll look at it later.

database/misc/mongodb_replset.py
jirutka: What additional info should it contain?

database/misc/mongodb_replset.py
sivel, line 152: Unfortunately ansible has historically not agreed.  Additionally, I'm pretty sure pep8 doesn't mention the use of `dict` vs `{}`

jirutka, line 152: There are already several modules with this dict style, so it should be okay.

database/misc/mongodb_replset.py
sivel: The replset configuration may be beneficial.  Most modules generally return more than just a success/changed.  The more info the better.

jirutka: I agree that more info is better, but what exactly should it return?

jirutka: Maybe a list of added members (hostname)?

database/misc/mongodb_replset.py
jirutka: BTW Is it allowed to use Python 2.5+ in the extra modules, or should I stick to ancient 2.4 like in core modules? I would like to use a context manager here (aka the “with statement”).

database/misc/mongodb_replset.py
bcoca: you could return the status of the replica set, hosts, state, timestamps, etc.

database/misc/mongodb_replset.py
bcoca: modules should be python2.4, the only exception is cloud modules as they execute normally on 'master' vs on targets.

jirutka: Fixed. I’ve refactored code to use [RawConfigParser.read](https://docs.python.org/2.6/library/configparser.html#ConfigParser.RawConfigParser.read) (instead of readfp) that takes care about closing the file itself.

database/misc/mongodb_replset.py
jirutka: You’re right, it’s unnecessary, so I’ve removed the double cast.

database/misc/mongodb_replset.py
jirutka: @sivel I’ve added members state and list of added hosts to exit info. Example:

``` json
{
  "added_hosts": ["database-1.vpn:27017"],
  "changed": true,
  "members": {
    "database-1.vpn:27017": {"state": "SECONDARY"},
    "database-2.vpn:27017": {"state": "PRIMARY"}
  }
}
```

jirutka:
@sivel I’ve implemented your suggestions and also improved the code a little bit — I’ve kept these additional changes in separate commits for simpler review. Feel free to squash them before merge, if you like.

jirutka:
Gentle ping @sivel, @bcoca. Ready for merge.

sivel:
Patience is a virtue.  I don't handle merges in this tree.  This is tagged with P3, all P2, P1 issues and P3 issues submitted earlier and such get higher priority.  This has only been here 5 days.  I generally expect most PRs of this nature to take a fair amount longer.

jirutka:
Ah, so #241 has been just a bright exception, it still takes incredibly long time to merge a PR… :crying_cat_face: Well, I give up, it’s not worth the effort.

bcoca:
yes, sometimes when we already have tests made (i've had patch module since 2013) and the PR is easy, it can get done faster than normal. In this case I don't have a test setup ready for mongo replica sets.

jirutka:
@bcoca You have written tests for the patch module? That’s great, but where it is? I don’t see _any_ tests in this repo.

bcoca:
i had plays to deploy updates to one of my internal apps,I  also had the
action plugin side which i added to allow for local patch sources

## 

Brian Coca

jirutka:
Just out of curious, if you had patch module since 2013, why you haven’t added it yourself back then? o.O

bcoca:
looong todo list, stuff just drops

UnderGreen:
Hello guys.

I did't see this module, before writing and adding my PR (https://github.com/ansible/ansible-modules-extras/pull/265)
But I see, 2 problems with your module:
1. Absent don't work
2. If you try to configure replica host and it host is down(wrong mongo port or really host down), your module will fail. It is hard to use in mongodb role. Only possible use - when all mongodb hosts configured and it should be task outside role. My module can work in role, I will publish my role soon.

gregdek:
Adding "owner_pr" since @jirutka owns a module in Extras

database/misc/mongodb_replset.py
```diff
@@ -0,0 +1,227 @@
...
+
+def join_colon(iterable):
+    return ':'.join(str(s) for s in iterable)
+
+
+def main():
+    module = AnsibleModule(
+        argument_spec={
+            'login_user':     {'aliases': ['user']},
+            'login_password': {'aliases': ['password']},
```
resmo: It is recommended to set `no_log: True` for passwords.

gregdek:
@jirutka Sorry for the long delay on this. 

The good news: @resmo is an approved reviewer, and you yourself are an approved reviewer because you're already a module owner. So we assume you'll pass your own review; if you can address @resmo's concerns and get his +1, we will merge.

UnderGreen:
I will again recommend to see https://github.com/ansible/ansible-modules-extras/pull/265 instead this, because my replica set module can init RS, can delete members, have more configurable options for RS members and I have popular mongodb role, which based on my module.

gregdek:
@UnderGreen I have no objection to seeing you reopen #265. Really, it's a question of which one gets reviewed and approved first -- and @jirutka has not yet made revisions to this PR as requested by the least reviewer.

jirutka:
Uh, okay, I’ll review it this week.

jirutka:
@UnderGreen I still think that you _shouldn’t_ remove members from replica set using Ansible, this should be always done with caution by hand.

UnderGreen:
I think users should have this feature and they should decide themselves use it or not...

gregdek:
Putting both this and #265 into core_review, since we've basically got differing patches with differing viewpoints, and I think we need to escalate and make a call here.

gregdek:

# 

gregdek:
OK, @jirutka, we're choosing to pursue this PR over #265.  I know it's taken a while, but please bear with us as we work through this process.

Please make the suggested revisions by @resmo above, and when you're done, please add a comment with the text "ready_for_review" and we will continue with the review process.  Thanks!

jirutka:
ready_for_review

gregdek:
Thanks @jirutka, and thanks especially for your patience. 

@resmo can you take a look?

UnderGreen:
Where is examples section?

database/misc/mongodb_replset.py
```diff
@@ -0,0 +1,227 @@
...
+      - A comma delimited list of replica set members.
+    required: true
+    aliases: [ members ]
+    example: mongo0,mango1:27011,mango2:27012
+  replica_set:
+    description:
+      - Name of the replica set to create or connect to.
+    required: true
+    aliases: [ replset ]
+    default: "rs0"
```
resmo, line 61: default is not needed for required: true

database/misc/mongodb_replset.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+def main():
+    module = AnsibleModule(
+        argument_spec={
+            'login_user':     {'aliases': ['user']},
+            'login_password': {'aliases': ['password'], no_log: True},
+            'login_host':     {'aliases': ['host'], 'default': 'localhost'},
+            'login_port':     {'aliases': ['port'], 'default': 27017},
+            'hosts':          {'aliases': ['members'], 'required': True},
```
resmo, line 158: ansible has some magic included to handle lists, no need to pass a comma separated list and split it: set `type='list'` and I would also add an alias `host`, this way you use it like

```
tasks:
- mongodb_replset
  host: your_host
```

as well as

```
tasks:
- mongodb_replset
  hosts: 
  - your_host
  - your_2nd_host
```

and also as comma separated list as well...

database/misc/mongodb_replset.py
```diff
@@ -0,0 +1,227 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: mongodb_replset
+author: Jakub Jirutka
```
resmo, line 24: please add your github handle as well, like
`author: "Jakub Jirutka (@jirtuka)"`

database/misc/mongodb_replset.py
```diff
@@ -0,0 +1,227 @@
...
+version_added: "1.9"
```
resmo, line 25: This would be "2.0" now, hopefully you apply the changes soon, so it will make it into 2.0

database/misc/mongodb_replset.py
```diff
@@ -0,0 +1,227 @@
...
+short_description: Initiate or extend replica set.
+description:
+  - This module allows to initiate a new replica set or add new members to an
+    existing one. It can't remove members from a replica set, this should be
+    always done with caution by hand.
+options:
+  login_user:
+    description:
+      - The username used to authenticate with.
+    aliases: [ user ]
```
resmo, line 35: missing:

```
required: false
default: null
```

database/misc/mongodb_replset.py
```diff
@@ -0,0 +1,227 @@
...
+  login_password:
+    description:
+      - The password used to authenticate with.
+    aliases: [ password ]
```
resmo, line 39: missing:

```
required: false
default: null
```

database/misc/mongodb_replset.py
```diff
@@ -0,0 +1,227 @@
...
+  login_host:
+    description:
+      - The first host to connect when initiating the replica set.
+    aliases: [ host ]
+    default: localhost
```
resmo, line 44: missing:

```
required: false
```

database/misc/mongodb_replset.py
```diff
@@ -0,0 +1,227 @@
...
+  login_port:
+    description:
+      - Port of the first host to connect when initiating the replica set.
+    aliases: [ port ]
+    default: 27017
```
resmo, line 49: missing:

```
required: false
```

database/misc/mongodb_replset.py
```diff
@@ -0,0 +1,227 @@
...
+  hosts:
+    description:
+      - A comma delimited list of replica set members.
+    required: true
+    aliases: [ members ]
+    example: mongo0,mango1:27011,mango2:27012
```
resmo, line 55: remove example here... add and example section.

resmo:
needs_revision

gregdek:
Thanks @resmo for the review.

@jirutka After review, it looks as though this PR needs some revisions. 

database/misc/mongodb_replset.py
```diff
@@ -0,0 +1,227 @@
...
+
+def join_colon(iterable):
+    return ':'.join(str(s) for s in iterable)
+
+
+def main():
+    module = AnsibleModule(
+        argument_spec={
+            'login_user':     {'aliases': ['user']},
+            'login_password': {'aliases': ['password'], no_log: True},
```
UnderGreen, line 155: NameError: global name 'no_log' is not defined

UnderGreen:
This module really need a lot of improvements. How to setup member as a arbiter? How to setup params like priority, hidden, etc?
I can rewrite this module to support this improvements, but I don't see a normal use case in mongodb role. In playbook it's possible, call this module after configuration of all nodes with mongo. If you want to use this module inside mongodb role, then you should create new group of hosts with flag like a 'mongodb_replication' and play with it after configuration of all another mongodb hosts. Or we need to create a new role like mongodb-replication, but why we should do it, if we can use one role, than two. And for this new role we need to create a new group of hosts, etc...

This is my opinion about this PR and about closing my PR, my module not a prototype like this, it's used by a lot of peoples with my EXISTING and WORKING role.

robynbergeron:
@jirutka A friendly reminder: this pull request has been marked as needing your action. If you still believe that this PR applies, and you intend to address the issues with this PR, just let us know in the PR itself and we will keep it open pending your changes.

gregdek:
@jirutka Another friendly reminder: this pull request has been marked as needing your action. If you still believe that this PR applies, and you intend to address the issues with this PR, just let us know in the PR itself and we will keep it open. If we don't hear from you within another 14 days, we will close this pull request.

gregdek:
OK, this PR has been in needs_revision for almost 3 months with no comment from the submitter, and we originally closed a PR advocating the same functionality in favor of this one. Closing this PR in favor of #265.

PR #249, opened 2015-02-11
homebrew: Package name is not required.

simono:
e.g. `- homebrew: update_homebrew=yes upgrade_all=yes' is a valid task.

### DIFF
--- a/packaging/os/homebrew.py
+++ b/packaging/os/homebrew.py
@@ -31,7 +31,8 @@ options:
     name:
         description:
             - name of package to install/remove
-        required: true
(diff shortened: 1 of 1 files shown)


packaging/os/homebrew.py
```diff
@@ -31,7 +31,7 @@
     name:
         description:
             - name of package to install/remove
-        required: true
+        required: false
```
resmo, line 34: add `default: None`

simono, line 34: Done, thanks.

PR #253, opened 2015-02-12
svc: Ensure python 2.5 compatibility

SkaveRat:
- eg. debian lenny

Ansible has python 2.5 as minimum version

### DIFF
--- a/system/svc.py
+++ b/system/svc.py
@@ -198,10 +198,12 @@ class Svc(object):
             self.module.fail_json(msg="failed to execute: %s" % str(e))
         return (rc, out, err)
 
-
     def report(self):
         self.get_status()
-        return {k: self.__dict__[k] for k in self.report_vars}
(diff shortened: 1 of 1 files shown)


bcoca:
actually its 2.4, but this works for that also.

jirutka:
> eg. debian lenny

@SkaveRat Security updates for Lenny have been discontinued as of February 6th, 2012. We have the year 2015 now, so why the hell are you arguing with Debian Lenny?

SkaveRat:
not arguing that it's a good thing, just that there is some old python out there

jirutka:
I’m sure that there’s also some Python 1.1 out there… So what?

SkaveRat:
but ansible doesn't run on 1.1 - it runs on 2.4+

so this script should also be 2.4+

not even a big change, so I don't see your problem

PR #255, opened 2015-02-12
svc: fixing missing restart method

SkaveRat:


bcoca:
you seem to have an extra commit here

SkaveRat:
whoops. fixed in #256

PR #263, opened 2015-02-16
make examples more consistent

yauh:
instead of the alias `name` the proper attribute `user should be used in the examples


bcoca:
actually, we prefer name as it is more consistent with other modules

yauh:
Might I ask why you advocate `user` in the detailed parameters listing if you prefer `name`? Consistency should be achieved, so either use `name` in the examples or `user` in the options. I am fine with both, but having this mixture is confusing.

bcoca:
that is what the PR #264 is for

yauh:
Wonderful, thanks!

PR #268, opened 2015-02-17
Rackspace Cloud Load Balancer SSL termination module

smashwilson:
This is a reworking of ansible/ansible#7944.

I've replaced the weird "nested setting" setup that I'd done before with an independent module, `rax_clb_ssl`. I think it's much more natural this way.

Even though `https_redirect` is an attribute on the load balancer itself, I'm including it here rather than in `rax_clb` proper. You can't actually enable HTTPS redirection until SSL termination is added, so if I didn't include it, you'd need to invoke `rax_clb`, `rax_clb_ssl`, then `rax_clb` again to enable it properly.

Here's an example task:

``` yaml
  - name: ssl termination
    rax_clb_ssl:
      state: present
      credentials: .credentials
      balancer_name: "{{ the_clb.balancer.id }}"
      private_key: "{{ ssl_private_key }}"
      certificate: "{{ ssl_certificate }}"
      secure_traffic_only: yes
      https_redirect: yes
```

/cc @sivel, @claco, @ckrough

### DIFF
--- a/cloud/rackspace/rax_clb_ssl.py
+++ b/cloud/rackspace/rax_clb_ssl.py
@@ -116,9 +116,6 @@ def cloud_load_balancer_ssl(module, loadbalancer, state, enabled, private_key,
         else:
             certificate = certificate.strip()
 
-    if state not in ('present', 'absent'):
-        module.fail_json(msg="State must be either 'present' or 'absent'.")
-
     attempts = wait_timeout / 5
 
     # Locate the load balancer.
--- a/cloud/rackspace/rax_clb_ssl.py
+++ b/cloud/rackspace/rax_clb_ssl.py
@@ -21,7 +21,7 @@ module: rax_clb_ssl
 short_description: Manage SSL termination for a Rackspace Cloud Load Balancer.
 description:
 - Set up, reconfigure, or remove SSL termination for an existing load balancer.
-version_added: "1.8.2"
+version_added: "1.9"
 options:
   balancer_name:
     description:
--- a/cloud/rackspace/rax_clb_ssl.py
+++ b/cloud/rackspace/rax_clb_ssl.py
@@ -1,0 +1,284 @@
+#!/usr/bin/python
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+# This is a DOCUMENTATION stub specific to this module, it extends
+# a documentation fragment located in ansible.utils.module_docs_fragments
+DOCUMENTATION='''
+module: rax_clb_ssl
+short_description: Manage SSL termination for a Rackspace Cloud Load Balancer.
+description:
+- Set up, reconfigure, or remove SSL termination for an existing load balancer.
+version_added: "1.8.2"
+options:
+  balancer_name:
+    description:
+    - Name or ID of the load balancer on which to manage SSL termination.
+    required: true
+  state:
+    description:
+    - If set to "present", SSL termination will be added to this load balancer.
+    - If "absent", SSL termination will be removed instead.
+    choices:
+      - present
+      - absent
+    default: present
+  enabled:
+    description:
+    - If set to "false", temporarily disable SSL termination without discarding
+    - existing credentials.
+    default: true
+  private_key:
+    description:
+    - The private SSL key as a string in PEM format.
+  certificate:
+    description:
+    - The public SSL certificates as a string in PEM format.
+  intermediate_certificate:
+    description:
+    - One or more intermediate certificate authorities as a string in PEM
+    - format, concatenated into a single string.
+  secure_port:
+    description:
+    - The port to listen for secure traffic.
+    default: 443
+  secure_traffic_only:
+    description:
+    - If "true", the load balancer will *only* accept secure traffic.
+    default: false
+  https_redirect:
+    description:
+    - If "true", the load balancer will redirect HTTP traffic to HTTPS.
+    - Requires "secure_traffic_only" to be true. Incurs an implicit wait if SSL
+    - termination is also applied or removed.
+  wait:
+    description:
+    - Wait for the balancer to be in state "running" before turning.
+    default: false
+  wait_timeout:
+    description:
+    - How long before "wait" gives up, in seconds.
+    default: 300
+author: Ash Wilson
+extends_documentation_fragment: rackspace
+'''
+
+EXAMPLES = '''
+- name: Enable SSL termination on a load balancer
+  rax_clb_ssl:
+    balancer_name: the_loadbalancer
+    state: present
+    private_key: "{{ lookup('file', 'credentials/server.key' ) }}"
+    certificate: "{{ lookup('file', 'credentials/server.crt' ) }}"
+    intermediate_certificate: "{{ lookup('file', 'credentials/trust-chain.crt') }}"
+    secure_traffic_only: true
+    wait: true
+
+- name: Disable SSL termination
+  rax_clb_ssl:
+    balancer_name: "{{ registered_lb.balancer.id }}"
+    state: absent
+    wait: true
+'''
+
+from ansible.module_utils.basic import *
+from ansible.module_utils.rax import *
+
+try:
+    import pyrax
+    HAS_PYRAX = True
+except ImportError:
+    HAS_PYRAX = False
+
+def cloud_load_balancer_ssl(module, balancer_name, state, enabled, private_key,
+                            certificate, intermediate_certificate, secure_port,
+                            secure_traffic_only, https_redirect,
+                            wait, wait_timeout):
+    # Validate arguments.
+
+    if not balancer_name:
+        module.fail_json(msg='balancer_name is required.')
+
+    if state == 'present':
+        if not private_key:
+            module.fail_json(msg="private_key must be provided.")
+        else:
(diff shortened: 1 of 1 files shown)


cloud/rackspace/rax_clb_ssl.py
```diff
@@ -0,0 +1,284 @@
...
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+# This is a DOCUMENTATION stub specific to this module, it extends
+# a documentation fragment located in ansible.utils.module_docs_fragments
+DOCUMENTATION='''
+module: rax_clb_ssl
+short_description: Manage SSL termination for a Rackspace Cloud Load Balancer.
+description:
+- Set up, reconfigure, or remove SSL termination for an existing load balancer.
+version_added: "1.8.2"
```
sivel: `version_added` should be `1.9`

cloud/rackspace/rax_clb_ssl.py
```diff
@@ -0,0 +1,284 @@
...
+    wait: true
+
+- name: Disable SSL termination
+  rax_clb_ssl:
+    balancer_name: "{{ registered_lb.balancer.id }}"
+    state: absent
+    wait: true
+'''
+
+from ansible.module_utils.basic import *
```
sivel: These import lines need to be the last lines before the call to `main()`

Look at the other modules for examples.

cloud/rackspace/rax_clb_ssl.py
```diff
@@ -0,0 +1,284 @@
...
+except ImportError:
+    HAS_PYRAX = False
+
+def cloud_load_balancer_ssl(module, balancer_name, state, enabled, private_key,
+                            certificate, intermediate_certificate, secure_port,
+                            secure_traffic_only, https_redirect,
+                            wait, wait_timeout):
+    # Validate arguments.
+
+    if not balancer_name:
```
sivel: If this is required, it should be marked as such in the argument spec, and this can go away.

cloud/rackspace/rax_clb_ssl.py
```diff
@@ -0,0 +1,284 @@
...
+
+    if success:
+        module.exit_json(**result)
+    else:
+        module.fail_json(**result)
+
+def main():
+    argument_spec = rax_argument_spec()
+    argument_spec.update(dict(
+        balancer_name=dict(type='str'),
```
sivel: It looks like this should be required.

Additionally, I'd rather this parameter be called something slightly more generic.  In `rax_dns_record` it is just called `loadbalancer`

cloud/rackspace/rax_clb_ssl.py
```diff
@@ -0,0 +1,284 @@
...
+
+    # Locate the load balancer.
+
+    clb = pyrax.cloud_loadbalancers
+    if not clb:
+        module.fail_json(msg='Failed to instantiate client. This '
+                             'typically indicates an invalid region or an '
+                             'incorrectly capitalized region name.')
+
+    balancers = []
```
sivel: Instead of this, we should probably use `rax_find_loadbalancer` from the shared module utils.

cloud/rackspace/rax_clb_ssl.py
smashwilson: Done.

Just out of curiosity, what's the rationale for this? Is it just for consistency?

cloud/rackspace/rax_clb_ssl.py
smashwilson: Ah, right. I wasn't aware of `required=True` when I first wrote this. Fixed and much cleaner.

cloud/rackspace/rax_clb_ssl.py
smashwilson: `loadbalancer` it is!

cloud/rackspace/rax_clb_ssl.py
smashwilson: :sparkles: Yeah, that's much nicer. The `if not clb:` check isn't needed any more, either, I see.

cloud/rackspace/rax_clb_ssl.py
sivel: It is a "fake" import.  The import is never actually called, instead the ModuleReplacer code replaces those lines with the full content of the actual file, building out a single python file with all ansible stuff it needs.

Putting it at the bottom ensures that tracebacks identify the correct line number in a module, as the line numbers would be offset.

smashwilson:
@sivel Thanks for the review! Changes have been made and re-tested.

sivel:
I'll do another once over in the near future, then I'll need to write a full battery of integration tests for this module before merging.

smashwilson:
> I'll do another once over in the near future, then I'll need to write a full battery of integration tests for this module before merging.

Sure, whenever you get the chance.

Where do the integration tests live, by the way?

sivel:
The integration tests exist at: https://github.com/ansible/ansible/tree/devel/test/integration

cloud/rackspace/rax_clb_ssl.py
```diff
@@ -0,0 +1,259 @@
...
+    if new_ssl_termination:
+        new_ssl = dict(
+            enabled=new_ssl_termination['enabled'],
+            secure_port=new_ssl_termination['securePort'],
+            secure_traffic_only=new_ssl_termination['secureTrafficOnly']
+        )
+    else:
+        new_ssl = None
+
+    result = dict(
```
sivel, line 202: I think the `result` should also include some data about what load balancer it just acted on.

cloud/rackspace/rax_clb_ssl.py
```diff
@@ -0,0 +1,259 @@
...
+                    continue
+
+                if value is not None and existing_ssl.get(ssl_attr) != value:
+                    # module.fail_json(msg='Unnecessary change', attr=ssl_attr, value=value, existing=existing_ssl.get(ssl_attr))
+                    needs_change = True
+        else:
+            needs_change = True
+
+        if needs_change:
+            balancer.add_ssl_termination(**ssl_attrs)
```
sivel: Need to catch exceptions here, such as `pyrax.exceptions.ClientException`, otherwise an LB in `PENDING_UPDATE` could cause a traceback.

cloud/rackspace/rax_clb_ssl.py
```diff
@@ -0,0 +1,259 @@
...
+            balancer.delete_ssl_termination()
+            changed = True
+
+    if https_redirect is not None and balancer.httpsRedirect != https_redirect:
+        if changed:
+            # This wait is unavoidable because load balancers are immutable
+            # while the SSL termination changes above are being applied.
+            pyrax.utils.wait_for_build(balancer, interval=5, attempts=attempts)
+
+        balancer.update(httpsRedirect=https_redirect)
```
sivel: Need to catch exceptions here, such as `pyrax.exceptions.InvalidLoadBalancerParameters`, otherwise a traceback can be raised here.

gregdek:
Marking this needs_revision -- @smashwilson if you can satisfy @sivel's requests, we will merge.

smashwilson:
Thanks - I'll ping here once I get time to push fixes. :bow:

cloud/rackspace/rax_clb_ssl.py
```diff
@@ -0,0 +1,259 @@
...
+    - How long before "wait" gives up, in seconds.
+    default: 300
+author: Ash Wilson
+extends_documentation_fragment: rackspace
+'''
+
+EXAMPLES = '''
+- name: Enable SSL termination on a load balancer
+  rax_clb_ssl:
+    balancer_name: the_loadbalancer
```
grahamrhay: has this changed?

grahamrhay: I needed to specify a region too

grahamrhay:
works for me, thanks!

cloud/rackspace/rax_clb_ssl.py
smashwilson, line 202: I've added a `rax_to_dict(balancer)` element. There's some redundant information in there (`https_redirect` occurs at the top-level _and_ within the balancer element) but that didn't seem too terrible to me.

cloud/rackspace/rax_clb_ssl.py
```diff
@@ -0,0 +1,269 @@
...
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+# This is a DOCUMENTATION stub specific to this module, it extends
+# a documentation fragment located in ansible.utils.module_docs_fragments
+DOCUMENTATION='''
+module: rax_clb_ssl
+short_description: Manage SSL termination for a Rackspace Cloud Load Balancer.
+description:
+- Set up, reconfigure, or remove SSL termination for an existing load balancer.
+version_added: "2.0"
```
smashwilson, line 24: Is this the correct `version_added` to specify, now?

cloud/rackspace/rax_clb_ssl.py
```diff
@@ -0,0 +1,269 @@
...
+                balancer.add_ssl_termination(**ssl_attrs)
+            except pyrax.exceptions.PyraxException, e:
+                module.fail_json(msg='%s' % e.message)
+            changed = True
+    elif state == 'absent':
+        # Remove SSL termination if it's already configured.
+        if existing_ssl:
+            try:
+                balancer.delete_ssl_termination()
+            except pyrax.exceptions.PyraxException, e:
```
smashwilson, line 167: I noticed that `rax_clb` blanket-catches `Exception`. Is catching the somewhat-more-specific `PyraxException` instead okay?

:thought_balloon: Actually, it might be nice to have better behavior for handling uncaught exceptions from Python modules in general, maybe in a parent context rather than relying on each module to catch its own... using `module.fail_json()` strips out a traceback, for example.

smashwilson:
@sivel I've finally gotten around to implementing fixes for your comments and pushed my changes out. I also gave it a quick spin with a test playbook to catch any obvious regressions and it still worked for me.

sivel:
LGTM.

sivel:
Merged, thanks!

grahamrhay:
:rainbow:

grahamrhay:
any idea what version of ansible this will make it into?

sivel:
It should be included in the next major release which is 2.0

grahamrhay:
great, thanks!

PR #269, opened 2015-02-18
dnsmadeeasy module fixes

mlandauer:
Included in this PR are some fixes that allow the dnsmadeeasy module to work correctly with MX and TXT records, root domain A records, CNAME records that point to the root of the domain.

Each fix is separated into its own commit:
- Now we can set an A record on the root of a domain by passing for instance `record_name="" record_type="A" record_value="1.2.3.4"` - If we leave out record_name completely it reverts to its previous behaviour
- Now we can set a CNAME record that points to the root domain by passing `record_name="www" record_type="CNAME" record_value=""` - If we leave out record_value completely it reverts to its previous behaviour
- It also now can update MX and TXT records correctly. Previously it assumed that there could only be a single record (of any type) for a particular `record_name`. This assumption is clearly not true in practise and is especially not true for MX and TXT records for which there are typically many records for a single name. Now, when updating a TXT or MX record it will only update a record that matches the name, type and value, otherwise it treats it as a new record.

Let me know if I need to make any further changes or update the coding style - I'm not a Python native - I did my best to match the coding style in the preexisting code.

### DIFF
--- a/network/dnsmadeeasy.py
+++ b/network/dnsmadeeasy.py
@@ -134,6 +134,7 @@ class DME2:
         self.domain_map = None      # ["domain_name"] => ID
         self.record_map = None      # ["record_name"] => ID
         self.records = None         # ["record_ID"] => <record>
+        self.all_records = None
 
         # Lookup the domain ID if passed as a domain name vs. ID
         if not self.domain.isdigit():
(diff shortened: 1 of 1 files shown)


gregdek:
@briceburg please review this PR; if you +1, we will merge

briceburg:
Looks great.

+1 ready to merge

Issue #109, opened 2014-11-14
bigip_facts:  get_fw_rule NotImplemented Error


lndevnull:
Getting the following error after a recent F5 upgrade to 11.6 from 11.5

```
failed: [localhost -> 127.0.0.1] => {"failed": true}
msg: received exception: Server raised fault: 'Exception caught in LocalLB::urn:iControl:LocalLB/VirtualServer::get_fw_rule()
Common::NotImplemented
```

Role is defined as the following: 

```
  tasks:
  - name: Get F5 Virtual Servers
    local_action: >
      bigip_facts
      server=esilo-f5-corp
      user={{ f5_user }}
      password={{ f5_passwd }}
      include=virtual_server
    register: bigip_vars
```

Ansible 1.7.2

lndevnull:
Temporary fix was to comment out both get_fw_rule definitions.

srgvg:
ping @mhite

mhite:
Haven't moved past 11.5 yet but will grab the 11.6 VM and see what I can figure out.

mhite:
From iControl documentation:

```
get_fw_rule *IMPORTANT* This method has been deprecated immediately. As of v11.6.0, attempting to perform this method will return a not-implemented exception. Gets the firewall rules for the specified virtual servers.
```

Removal seems the best course of action.

mhite:
Feel free to give this a whirl:

https://github.com/ansible/ansible-modules-extras/pull/272

mhite:
Or #273, which is a bit more of a general fix which should help if F5 does this to me again. ;)

bcoca:
closing vea merged #273

PR #272, opened 2015-02-21
Fixes get_fw_rule NotImplemented error on 11.6 BIG-IP

mhite:
This pull request fixes issue #109. According to the BIG-IP iControl documentation, the get_fw_rule() API call is no longer supported on virtual servers as of 11.6. I have removed the reference to it.


bcoca:
I'm not that familiar with obsolescence in the F5 family, does it make sense to detect the version and keep the call for those that have older models?

mhite:
@bcoca - Looking at the API documentation, there is a ton of get_fw_rule_\* API calls that were introduced in 11.3 and 11.5. The bigip_facts module was only calling a single one of them (get_fw_rule, introduced in 11.3). All of their get_fw_rule_\* related API calls were then ripped out in 11.6 breaking all backward compatability. In the documentation for the API calls they removed, they refer you to use another API call which is also removed in the same version. It's a mess on their end right now as far as I can tell.

Until there is some stability in how F5 implements the get_fw_rule\* stuff in the VirtualServer API, I am happy just removing the single call we do today. Alternatively, I can catch the suds WebFault exception like I do with MethodNotFound and just ignore it. I'm happy to go that route if you are more comfortable with that approach.

BTW, here is the link to the VirtualServer API information:

https://devcentral.f5.com/wiki/iControl.LocalLB__VirtualServer.ashx

You can see just how much they ripped out if you just search for get_fw_rule on that page.

mhite:
I've also opened pull-request #273 as an alternative solution.

bcoca:
merged #273 instead

PR #277, opened 2015-02-23
Windows module to unzip zip files

perzizzle:
Windows module to unzip zip files. Discussed in google group at https://groups.google.com/forum/#!starred/ansible-devel/3wSl03GLGGY

### DIFF
--- a/network/citrix/netscaler.py
+++ b/network/citrix/netscaler.py
@@ -97,7 +97,6 @@ ansible host -m netscaler -a "nsc_host=nsc.example.com user=apiuser password=api
 '''
 
 
-import json
 import base64
 import socket
 


perzizzle:
I created a test playbook but wasn't sure where to put it. It's dependent on a zip file being staged on the remote machine (I don't have win_copy working on my test machine). The zip file is called file.zip and contains parentFolder/childFolder/file.txt. The gist below has the test playbook.

https://gist.github.com/perzizzle/c25daebfb66dcb16be42

elventear:
Tried the module; works for me. It would be nice to have a feature where we could specify the destination is a directory and it should be created if it doesn't exist already.

perzizzle:
Makes sense, I can add the ability to create the destination path as an option.

perzizzle:
Commit 164c541 checks to see if the destination path exists and creates it if it doesn't. Didn't seem worth creating another parameter for this. I believe it follows the unarchive paradigm. @elventear

elventear:
:+1:

PR #282, opened 2015-02-25
Added session and monitor state enabling/disabling

mhite:
This P.R. implements node based session and monitor state manipulation. The BIG-IP GUI does not map directly to session and monitor state API calls but I did manage to find a forum post explaining how they are mapped behind the scenes.

https://devcentral.f5.com/questions/icontrol-equivalent-call-for-b-node-down

Enabled (all traffic allowed):
monitor_state=enabled, session_state=enabled

Disabled (only persistent or active connections allowed):
monitor_state=enabled, session_state=disabled

Forced offline (only active connections allowed):
monitor_state=disabled, session_state=disabled

This needs testers.

### DIFF
--- a/network/f5/bigip_node.py
+++ b/network/f5/bigip_node.py
@@ -61,6 +61,22 @@ options:
         default: present
         choices: ['present', 'absent']
         aliases: []
+    session_state:
+        description:
+            - Set new session availability status for node
+        version_added: "1.9"
+        required: false
+        default: null
+        choices: ['enabled', 'disabled']
+        aliases: []
+    monitor_state:
+        description:
+            - Set monitor availability status for node
+        version_added: "1.9"
+        required: false
+        default: null
+        choices: ['enabled', 'disabled']
+        aliases: []
     partition:
         description:
             - Partition
@@ -137,6 +153,31 @@ EXAMPLES = '''
       partition=matthite
       name="{{ ansible_default_ipv4["address"] }}"
 
+# The BIG-IP GUI doesn't map directly to the API calls for "Node ->
+# General Properties -> State". The following states map to API monitor
+# and session states.
+#
+# Enabled (all traffic allowed):
+# monitor_state=enabled, session_state=enabled
+# Disabled (only persistent or active connections allowed):
+# monitor_state=enabled, session_state=disabled
+# Forced offline (only active connections allowed):
+# monitor_state=disabled, session_state=disabled
+#
+# See https://devcentral.f5.com/questions/icontrol-equivalent-call-for-b-node-down
+
+  - name: Force node offline
+    local_action: >
+      bigip_node
+      server=lb.mydomain.com
+      user=admin
+      password=mysecret
+      state=present
+      session_state=disabled
+      monitor_state=disabled
+      partition=matthite
+      name="{{ ansible_default_ipv4["address"] }}"
+
 '''
 
 try:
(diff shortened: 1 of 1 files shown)


perzizzle:
I have successfully tested this module using ansible 1.8 and BIG-IP 11.6.0.

msabramo:
Node disabling/enabling is working great, once I disable the SSL cert validation (see #334), so I'd say :+1:  to merging this. I am using ansible 1.9 from git `devel` branch and a `BIG-IP 11.6.0 Build 3.0.412 Hotfix HF3`

msabramo:
@perzizzle and I have both tested this and it works for us. Should it be merged?

mhite:
I think it has seen sufficient testing. @bcoca -- merge at your leisure!

bcoca:
still needs a rebase

mhite:
I see. It actually looks like similar code was merged recently from an Oct. 2014 PR (merged in March 2015). The unfortunate part is that the merged code does not fully supported all 3 session/monitor disable states. Not ideal, but not certainly not the end of the world.

bcoca:
it seems you just would need to add monitoring: enabled/disabled at this point

network/f5/bigip_node.py
```diff
@@ -61,6 +61,22 @@
         default: present
         choices: ['present', 'absent']
         aliases: []
+    session_state:
+        description:
+            - Set new session availability status for node
+        version_added: "1.9"
```
bcoca, line 67: 2.0 at this point

bcoca:
closing this as dupe of already merged #288

msabramo:
I could be wrong, but I believe that this is not a dupe of #288 as they target two different modules. Unfortunately, there is a lot of duplication between the various F5 modules.
- This PR modifies `network/f5/bigip_node.py`.
- #288 modifies `network/f5/bigip_pool_member.py`.

Both are useful. This one allows you to disable a node entirely (across all pools). #288 was for disabling a node in a single pool.

bcoca:
my bad, reopening and now asking ... how much overlap do we have and where should these things live? should we open ansible-devel thread?

bcoca:
in any case, this PR still needs a rebase

mhite:
I wouldn't say there is overlap in bigip_node vs. bigip_pool_member regarding session and monitor_state. @msabramo's summary of what they do is correct.

The problem at this point is that what got merged in https://github.com/ansible/ansible-modules-extras/commit/f86ce495c7e8b29c02b25ebfabc6bdb2adc5e3b6 overlaps with this, and only partially implements the functionality. Where the PR that got merged partially incorporates this functionality as an extension of the 'state' parameter, this PR doesn't touch that and creates two new parameters to manipulate. 

You just merged https://github.com/ansible/ansible-modules-extras/pull/288 which does not touch the state parameter.

Not to upset the author (@j2sol) of the original PR you merged into this one, I kind of think you should back that one out and just merge this one for consistency sake with bigip_pool_member.py. 

Similarly, trying to control monitor state and session state out of the single 'state' parameter just won't take us far. As mentioned in this PR, there are various combinations of session and monitor state that when combined do certain things:

Enabled (all traffic allowed):
monitor_state=enabled, session_state=enabled

Disabled (only persistent or active connections allowed):
monitor_state=enabled, session_state=disabled

Forced offline (only active connections allowed):
monitor_state=disabled, session_state=disabled

bcoca:
ok, reverting the previous commit

PR #283, opened 2015-02-25
New module for AWS Elastic Transcoder pipelines

wimnat:
New module to create or delete elastic transcoder pipelines in AWS.

It works using the transcoder name although AWS do not enforce this as unique. This could of course cause issues for those not using unique names but referencing transcoder by ID is not what I've seen in the real world.

Comments welcome.

### DIFF
--- a/cloud/amazon/elastictranscoder.py
+++ b/cloud/amazon/elastictranscoder.py
@@ -1,0 +1,204 @@
+#!/usr/bin/python
+#
+# (c) 2014, Rob White <wimnat@gmail.com>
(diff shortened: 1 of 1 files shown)


gregdek:
Fails travis with error:
Sorry: IndentationError: unexpected indent (elastictranscoder.py, line 186)

wimnat:
I'm not happy with this module any more in its present state.  I'll reopen the PR after some work.

gregdek:
Thanks @wimnat.

PR #288, opened 2015-02-25
Support for monitor and session state manipulation added to bigip_pool_m...

mhite:
This P.R. implements pool member based session and monitor state manipulation. The BIG-IP GUI does not map directly to session and monitor state API calls but I did manage to find a forum post explaining how they are mapped behind the scenes.

https://devcentral.f5.com/questions/icontrol-equivalent-call-for-b-node-down

Enabled (all traffic allowed):
monitor_state=enabled, session_state=enabled

Disabled (only persistent or active connections allowed):
monitor_state=enabled, session_state=disabled

Forced offline (only active connections allowed):
monitor_state=disabled, session_state=disabled

This needs testers.

### DIFF
--- a/network/f5/bigip_pool_member.py
+++ b/network/f5/bigip_pool_member.py
@@ -63,6 +63,22 @@ options:
         default: present
         choices: ['present', 'absent']
         aliases: []
+    session_state:
+        description:
+            - Set new session availability status for pool member
+        version_added: "1.9"
+        required: false
+        default: null
+        choices: ['enabled', 'disabled']
+        aliases: []
+    monitor_state:
+        description:
+            - Set monitor availability status for pool member
+        version_added: "1.9"
+        required: false
+        default: null
+        choices: ['enabled', 'disabled']
+        aliases: []
     pool:
         description:
             - Pool name. This pool must exist.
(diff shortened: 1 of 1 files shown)


perzizzle:
I have successfully tested this module using ansible 1.8 and BIG-IP 11.6.0.

msabramo:
Okay, I'm testing this now. I ran into this issue of SSL certificate verification (because the test F5 that I've been given access to has a bogus self-signed certificate with the wrong common name), which I also ran into while working on [f5-cli](https://github.com/jdubs/f5-cli).

```
$ ansible-playbook disable_node.yaml

PLAY [mt1-pyweb01] ************************************************************

GATHERING FACTS ***************************************************************
ok: [mt1-pyweb01]

TASK: [Force pool member offline] *********************************************
failed: [mt1-pyweb01 -> 127.0.0.1] => {"failed": true}
msg: received exception: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:581)>

           to retry, use: --limit @/Users/marca/disable_node.yaml.retry

mt1-pyweb01                : ok=1    changed=0    unreachable=0    failed=1
```

For f5-cli, I worked around this with https://github.com/jdubs/f5-cli/pull/9, which does:

``` python
    def disable_ssl_cert_validation(self):
        # From https://www.python.org/dev/peps/pep-0476/#id29
        import ssl
        ssl._create_default_https_context = ssl._create_unverified_context
```

mhite:
Yeah, it looks like monkey patching is the only viable option to disable verification. I am happy to add this to the BIG-IP modules, although we should track this through a separate issue.

msabramo:
I'm working on it now. I'll send a PR shortly.

mhite:
Even better! Thanks for the help! ;)

msabramo:
@mhite: I've started a PR in https://github.com/ansible/ansible-modules-extras/pull/334; I need to copy my code to the other `bigip` modules I think, but I thought I'd post what I have now, in case you have suggestions.

msabramo:
OK, well the pool member disabling is working great, once I disable the SSL cert validation (see #334), so I'd say :+1:  to merging this. I am using ansible 1.9 from git `devel` branch and a `BIG-IP 11.6.0 Build 3.0.412 Hotfix HF3`

msabramo:
@perzizzle and I have both tested this and it works for us. Should it be merged?

davidmc24:
This looks good to me as well.  Tested using Ansible 1.8.4 and `BIG-IP 11.3.0 Build 3144.0 Hotfix HF8`.

magnusheino:
Any reason why this great new feature hasn't been merged?

bcoca:
@mhite as you authored both can you weigh in?

PR #291, opened 2015-02-27
lldp - Ignore lines that do not have the key in them

beardymcbeards:
Some switches return multi-line output, which breaks the split function,
the change seeks to only preform the split on a line that has the key.


ahill00:
+1

antonym:
I can confirm this fixes a previously opened bug that was closed:
https://github.com/ansible/ansible/issues/6935

PR #294, opened 2015-03-02
New module: nfsexport

randynobx:
Here is a new module, nfsexport, for working with nfs export entries in /etc/exports.


randynobx:
Rebasing commits and will open new pull request with clean patching.

PR #295, opened 2015-03-03
composer: Fixes changed status

jails:
Fixes changed status that always returns `False` with composer.

This [previous PR](https://github.com/ansible/ansible-modules-extras/pull/61) had fixed the issue but because of a [Composer recent change](https://github.com/composer/composer/commit/cb336a5416595efa321c024735e6452c9c7df106) stderr is now used for reporting information meant for humans while stdout is more for the output of the command.

This PR would solve this "changed status" detection in a more definitive way.

### DIFF
--- a/packaging/language/composer.py
+++ b/packaging/language/composer.py
@@ -159,7 +159,8 @@ def main():
         output = parse_out(err)
         module.fail_json(msg=output)
     else:
-        output = parse_out(out)
+        # Composer version > 1.0.0-alpha9 now use stderr for standard notification messages
+        output = parse_out(out + err)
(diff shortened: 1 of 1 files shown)


abadger:
Code looks good.  Could you add a comment about which version of Composer the output we're searching for switches from stdout to stderr?  Someday we might want to remove old compatibility code or someone new to composer might not know why we search both strings.

jails:
I added the following comment: "Composer version > 1.0.0-alpha9 now use stderr for standard notification messages".

Feel free to keep me updated if you need something more.

abadger:
Thanks for the fix!  Merged.

# Merged

Hi!

issue is resolved.

Thank you!

PR #296, opened 2015-03-04
New Module: nfsexport

randynobx:
Created a new module for working with entries in /etc/exports (or nfs exports file in an otherwise specified location).

Module is nfsexport.py under the systems group.


gregdek:

robynbergeron:

# 

robinro:
`/etc/exports` allows to have the same path in multiple lines with different clients, that are then combined. This module does not take this into account so far.

As far as I'm aware this is a seldomly used feature, so it might be ok to fix this later.

system/nfsexport.py
```diff
@@ -0,0 +1,126 @@
...
+            - Address o list of addresses of the nfs clients
+        required: false
+        default: "'*', which means `all`"
+    options:
+        description:
+            - nfs export options.
+        required: false
+        default: ro,root_squash
+        aliases: ['opt', 'opts']
+examples:
```
Spredzy, line 54: Based on the guidelines[1] this should be in its own section

[1] http://docs.ansible.com/ansible/developing_modules.html#example

system/nfsexport.py
```diff
@@ -0,0 +1,126 @@
...
+            dest      = dict(required=False, default='/etc/exports'),
+            clients   = dict(required=False, default='*'),
+            options   = dict(required=False, default='ro,root_squash', aliases=['opt', 'opts']),
+            ),
+        supports_check_mode=True
+        )
+
+    params = module.params
+    params['path'] = params['path'].strip()
+    client_list = params['clients'] \
```
Spredzy, line 78: Can we add an example where the clients is a list in the documentation. So people know how to use this module in such a case

system/nfsexport.py
```diff
@@ -0,0 +1,126 @@
...
+    try:
+        fd = open(params['dest'], 'a')
+    except IOError, err:
+        module.fail_json(msg="failed to open dest for writing: %s" % str(err))
+    fd.write(exportline)
+    fd.close()
+    module.exit_json(changed=True)
+
+from ansible.module_utils.basic import *
+main()
```
Spredzy, line 126: It seems to be a community practice to have 

```
if __name__ == '__main__':
    main()
```

Can this be added here ?

system/nfsexport.py
```diff
@@ -0,0 +1,126 @@
...
+def main():
+    module = AnsibleModule(
+        argument_spec = dict(
+            path      = dict(required=True),
+            dest      = dict(required=False, default='/etc/exports'),
+            clients   = dict(required=False, default='*'),
+            options   = dict(required=False, default='ro,root_squash', aliases=['opt', 'opts']),
+            ),
+        supports_check_mode=True
+        )
```
Spredzy, line 74: nit: Wrong indentation here

system/nfsexport.py
```diff
@@ -0,0 +1,126 @@
+#!/usr/bin/env python
```
Spredzy, line 1: The shebang should always be #!/usr/bin/python, this allows ansible_python_interpreter to work[1]

[1] http://docs.ansible.com/ansible/developing_modules.html#module-checklist

Spredzy:
Thanks @randynobx for this PR. I've tested it and it works as advertised. Few remarks:
- The module doesn't match the guidelines listed here http://docs.ansible.com/ansible/developing_modules.html#module-checklist
- What if I want to have a line like the following with nfsexport module

```
/nfs/share 127.0.0.1(rw) 192.168.0.0/24(ro)
```

How would I specify that ?

Else, once the comment addressed it will look good to me.

randynobx:
Thanks for the comments Spredzy and robinro

The module has been updated to address all concerns presented by Spredzy.

I am currently working on finding a way to handle such a scenario as robinro brought up. That may end up being a future update.

A new PR will be opened on the new proper repo with the aforementioned comments addressed.

PR #297, opened 2015-03-06
homebrew_tap: Fix valid tap pattern

aereal:
ref. https://github.com/Homebrew/homebrew/blob/493edcf1/Library/Homebrew/tap_constants.rb#L2

### DIFF
--- a/packaging/os/homebrew_tap.py
+++ b/packaging/os/homebrew_tap.py
@@ -52,7 +52,7 @@ homebrew_tap: tap=homebrew/dupes,homebrew/science state=present
 
 def a_valid_tap(tap):
     '''Returns True if the tap is valid.'''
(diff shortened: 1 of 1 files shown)


abadger:
I don't know anything about homebrew -- In the link you post i see this regex, though: 

```
HOMEBREW_TAP_ARGS_REGEX = %r{^([\w-]+)/(homebrew-)?([\w-]+)$}
```

Is making the two match what's necessary?  If so, the initial portions of the regex are different... For the homebrew one, it starts with ([\w-]+) (one or more word character or dashes)  whereas ours starts with (\S+) (one or more of any non-string characters). It seems like ours is more accepting than the homebrew regex.  Is that a problem?  Should we change it now as well?

Probably would be nice to also include a comment with that Homebrew link as documentation of where we're getting this regex from.

PR #298, opened 2015-03-06
pkgin: added check mode support, laid some groundwork for future `state=latest' support

L2G:
I think this may be the first significant pull request I've made to any Python-based project, let alone Ansible. So please don't be shy about pointing out anything that violates Ansible or Python style.

Functionally, all this does is call `pkgin -n` (answer "no" to everything) when check mode is in effect, instead of `pkgin -y`. But I also tried to gussy up the messages in the output and to get the code prepared for adding in support for `state=latest` another time.

Please note the "Known bug" that I described in the documentation block. This is not a bug I introduced, it's something that apparently has always been there due to the way pkgin works. I raised the issue in the pkgin project (see NetBSDfr/pkgin#42).

### DIFF
--- a/packaging/os/pkgin.py
+++ b/packaging/os/pkgin.py
@@ -1,8 +1,10 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
 
-# (c) 2013, Shaun Zinck
-# Written by Shaun Zinck <shaun.zinck at gmail.com>
+# Copyright (c) 2013 Shaun Zinck <shaun.zinck at gmail.com>
+# Copyright (c) 2015 Lawrence Leonard Gilbert <larry@L2G.to>
+#
+# Written by Shaun Zinck
 # Based on pacman module written by Afterburn <http://github.com/afterburn>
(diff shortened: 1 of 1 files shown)


packaging/os/pkgin.py
```diff
@@ -60,64 +67,97 @@
...
-        if query_package(module, pkgin_path, package):
+        if not module.check_mode and query_package(module, pkgin_path, package):
             module.fail_json(msg="failed to remove %s: %s" % (package, out))
 
         remove_c += 1
 
     if remove_c > 0:
-
-        module.exit_json(changed=True, msg="removed %s package(s)" % remove_c)
+        module.exit_json(changed=True, msg=format_action_message(module, "removed", remove_c))
```
bcoca, line 160: ~~why remove the information about # of packages?~~
never mind, it gets formatted in external function now

bcoca:
changes look good, could you squash the commits?

L2G:
@bcoca Sure thing.

PR #302, opened 2015-03-09
[cloud][amazon][new_module] Add module for managing IAM server certificates

Jell:
Add module to Upload/Delete/List IAM server certificate om AWS for use with ELB.

### DIFF
--- a/cloud/amazon/ec2_cert
+++ b/cloud/amazon/ec2_cert
@@ -1,0 +1,219 @@
+#!/usr/bin/python
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: ec2_cert
+short_description: Administrate IAM server certificates
+description:
+    - List, upload and delete IAM server certificates for use with ELB.
+version_added: "1.9"
+options:
+  aws_secret_key:
+    description:
+      - AWS secret key. If not set then the value of the AWS_SECRET_KEY environment variable is used.
+    required: false
+    default: None
+    aliases: ['ec2_secret_key', 'secret_key' ]
+  aws_access_key:
+    description:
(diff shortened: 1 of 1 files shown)


gregdek:
Hi @jell -- thanks for submitting this module to Ansible Extras. Apologies that it’s taken a while to get your module reviewed.

gregdek:
Calling potential AWS reviewers, if any of you have time to review, it would be appreciated: @defionscode @jsmartin @tombamford @garethr @lorin @silviud @zeekin @willthames @bpennypacker

cloud/amazon/ec2_cert
```diff
@@ -0,0 +1,219 @@
...
+            aws_secret_key = os.environ['AWS_SECRET_KEY']
+        elif 'EC2_SECRET_KEY' in os.environ:
+            aws_secret_key = os.environ['EC2_SECRET_KEY']
+
+    if not aws_access_key:
+        if 'AWS_ACCESS_KEY' in os.environ:
+            aws_access_key = os.environ['AWS_ACCESS_KEY']
+        elif 'EC2_ACCESS_KEY' in os.environ:
+            aws_access_key = os.environ['EC2_ACCESS_KEY']
+
```
bpennypacker, line 170: Functions for reading AWS credentials have been standardized in the module ansible.module_utils.ec2.  To be consistent with the other AWS modules you should import that module and use the function get_aws_connection_info to fetch AWS_ACCESS_KEY, AWS_SECRET_KEY, etc. rather than doing it yourself.  See, for example, the [ec2 module](https://github.com/ansible/ansible-modules-core/blob/devel/cloud/amazon/ec2.py#L1226).

willthames:
Use `get_aws_connection_info` and `connect_to_aws`.

There really is no need to reimplement the gathering of aws keys. `connect_to_aws` also copes with things like security_tokens and boto profiles so that you don't have to

cloud/amazon/ec2_cert
```diff
@@ -0,0 +1,219 @@
...
+'''
+
+import sys
+import time
+
+try:
+    import boto.iam
+except ImportError:
+    print "failed=True msg='boto required for this module'"
+    sys.exit(1)
```
bpennypacker, line 91: Rather than call sys.exit() you should call module.fail_json() if you can't import these modules.  [Here](https://github.com/ansible/ansible-modules-core/blob/devel/cloud/amazon/ec2.py#L507) is an example.

juliedavila, line 91: Duplicate PR of https://github.com/ansible/ansible-modules-core/pull/918

I'm up for ppl comparing and choosing one ove the other.

gregdek:
Both checked in Mar 9, too. Funny.

Well, we need to choose one, and the one in Core was written essentially as part of a suite of three IAM modules -- two of which have now been committed. Which tells me that the one currently awaiting review in Core is probably the way to go, for consistency's sake -- and also because IAM really is core functionality and belongs in Core regardless.

Sorry, @Jell -- sincerely appreciate the PR, but I think it makes more sense to pursue core/918 in this case. Thanks for submitting it. Sorry we didn't realize earlier that it was a duplicate.

PR #307, opened 2015-03-11
gluster_volume: fixes, new "force" option

nailgun:
Several fixes in system/gluster_volume module. Also added an option to force gluster to create bricks on rootfs.

### DIFF
--- a/system/gluster_volume.py
+++ b/system/gluster_volume.py
@@ -262,8 +262,11 @@ def main():
     def set_volume_option(name, option, parameter):
         run_gluster([ 'volume', 'set', name, option, parameter ])
 
-    def add_brick(name, brick):
-        run_gluster([ 'volume', 'add-brick', name, brick ])
+    def add_brick(name, brick, force):
+        args = [ 'volume', 'add-brick', name, brick ]
+        if force:
+            args.append('force')
+        run_gluster(args)
 
     def rebalance(name):
         run_gluster(['volume', 'rebalance', name, 'start'])
@@ -363,7 +366,7 @@ def main():
                     removed_bricks.append(brick)
 
             for brick in new_bricks:
-                add_brick(volume_name, brick)
+                add_brick(volume_name, brick, force)
                 changed = True
 
             # handle quotas
--- a/system/gluster_volume.py
+++ b/system/gluster_volume.py
@@ -292,7 +292,7 @@ def main():
             brick=dict(required=False, default=None),
             start_on_create=dict(required=False, default=True, type='bool'),
             rebalance=dict(required=False, default=False, type='bool'),
-            options=dict(required=False, default=None, type='dict'),
+            options=dict(required=False, default={}, type='dict'),
             quota=dict(required=False),
             directory=dict(required=False, default=None),
             force=dict(required=False, default=False, type='bool'),
@@ -318,7 +318,7 @@ def main():
     if not myhostname:
         myhostname = socket.gethostname()
 
-    options = module.params['options'] or {}
+    options = module.params['options']
     quota = module.params['quota']
     directory = module.params['directory']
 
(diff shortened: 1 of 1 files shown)


system/gluster_volume.py
```diff
@@ -91,7 +96,7 @@
 
 EXAMPLES = """
 - name: create gluster volume
-  gluster_volume: state=present name=test1 brick=/bricks/brick1/g1 rebalance=yes hosts:"{{ play_hosts }}"
+  gluster_volume: state=present name=test1 brick=/bricks/brick1/g1 rebalance=yes cluster:"{{ play_hosts }}"
```
nailgun, line 99: Looks like 'hosts' was renamed to 'cluster' formerly. Fixing docs!

system/gluster_volume.py
```diff
@@ -281,10 +288,11 @@ def set_quota(name, directory, value):
             transport=dict(required=False, default='tcp', choices=[ 'tcp', 'rdma', 'tcp,rdma' ]),
             brick=dict(required=False, default=None),
             start_on_create=dict(required=False, default=True, type='bool'),
-            rebalance=dict(required=False, default=False, taype='bool'),
+            rebalance=dict(required=False, default=False, type='bool'),
```
nailgun, line 294: fix typo

system/gluster_volume.py
```diff
@@ -300,13 +308,14 @@ def set_quota(name, directory, value):
     replicas = module.params['replicas']
     transport = module.params['transport']
     myhostname = module.params['host']
-    start_volume = module.boolean(module.params['start_on_create'])
+    start_on_create = module.boolean(module.params['start_on_create'])
```
nailgun, line 314: fix name conflict
`start_volume` is defined as function above

system/gluster_volume.py
```diff
@@ -300,13 +308,14 @@ def set_quota(name, directory, value):
...
     rebalance = module.boolean(module.params['rebalance'])
+    force = module.boolean(module.params['force'])
 
     if not myhostname:
         myhostname = socket.gethostname()
 
-    options = module.params['options']
+    options = module.params['options'] or {}
```
nailgun: options may be None, but code below assumes it's always a dict

system/gluster_volume.py
```diff
@@ -329,11 +338,12 @@ def set_quota(name, directory, value):
...
         # create if it doesn't exist
         if volume_name not in volumes:
-            create_volume(volume_name, stripes, replicas, transport, cluster, brick_path)
+            create_volume(volume_name, stripes, replicas, transport, cluster, brick_path, force)
+            volumes = get_volumes()
             changed = True
 
         if volume_name in volumes:
-            if volumes[volume_name]['status'].lower() != 'started' and start_volume:
+            if volumes[volume_name]['status'].lower() != 'started' and start_on_create:
```
nailgun, line 349: fix name collision

system/gluster_volume.py
```diff
@@ -329,11 +338,12 @@ def set_quota(name, directory, value):
 
         # create if it doesn't exist
         if volume_name not in volumes:
-            create_volume(volume_name, stripes, replicas, transport, cluster, brick_path)
+            create_volume(volume_name, stripes, replicas, transport, cluster, brick_path, force)
+            volumes = get_volumes()
```
nailgun, line 345: update volumes list after volume creation. Code below assumes the list is up-to-date

nailgun:
gluster_volume module is broken in devel

system/gluster_volume.py
bcoca: we should just make the default {} instead of None above, this would be redundant

system/gluster_volume.py
bcoca, line 99: is this mixing notations?

nailgun, line 99: No, it was renamed. It was named `hosts` in previous version of this module, named `glusterfs` (it's in stable-1.9 branch now)

nailgun:
@bcoca Fixed! Thanks!

PR #310, opened 2015-03-11
Add new module 'zookeeper'

jbowes:
This module allows creating, updating, and removing keys from ZooKeeper
(http://zookeeper.apache.org).

You may optionally specify a list of ACLs to apply to the key, and
either ensure that they exist, or that they are the only ACLs on the
key.

### DIFF
--- a/clustering/zookeeper.py
+++ b/clustering/zookeeper.py
@@ -1,0 +1,279 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2015, James Bowes <jbowes@repl.ca>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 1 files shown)


gregdek:

clustering/zookeeper.py
```diff
@@ -0,0 +1,279 @@
...
+logging.getLogger("kazoo").propagate = False
+
+try:
+    from kazoo.client import KazooClient
+    from kazoo.exceptions import NoAuthError, NoNodeError, ZookeeperError
+    from kazoo.security import make_acl, make_digest_acl_credential
+except ImportError:
+    print "failed=True msg='kazoo is required for this module. "\
+          "see https://kazoo.readthedocs.org/en/latest/install.html'"
+    sys.exit(1)
```
bcoca, line 116: modules should never exit or throw exceptions, capture the import error and use fail_json to return the message, you can look at ec2/openstack modules for an example

clustering/zookeeper.py
```diff
@@ -0,0 +1,279 @@
...
+
+# Initialiize Ansible Module
+# import module snippets
+from ansible.module_utils.basic import *
```
bcoca, line 120: you might want to do this at the bottom of the file, this is not a real import and will mess up line numbers for error reporting

jbowes:
Except for ACLs, this is covered by #811. Closing!

jcftang:
-It's not just ACL's that this PR covers, but setting of znodes is covered here that isn't in the znode plugin. It would be nice if the current znode plugin also supported setting of values for znodes-

update: the znode plugin does create/update znodes, it wasn't obvious to me when reading through the docs.

PR #312, opened 2015-03-13
homebrew_tap: Make valid tap check more accepting

jayanthm:
Homebrew tap's [naming convention](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/brew-tap.md#naming-conventions-and-limitations) basically tries to validate that the taps match `username/repo_name`. It's not worth it to try to validate github's naming convention for username/repos or homebrew's slight tweaks to it. One thing that does seem to be required here is the `/` since it is looking for a repo within a namespace of some sort; in this case it's username. All other checks we should leave it up to the hombrew tap library instead of doing it in ansible. The basic check for `/` at least verifies that the users aren't trying to tap a homebrew package. Although, I would be even be ok with getting rid of the is valid check completely from this module.

Some version of this is important to merge in soon as it's preventing users from actually tapping valid libraries such as `homebrew/homebrew-head-only`

### DIFF
--- a/packaging/os/homebrew_tap.py
+++ b/packaging/os/homebrew_tap.py
@@ -52,7 +52,7 @@ homebrew_tap: tap=homebrew/dupes,homebrew/science state=present
 
 def a_valid_tap(tap):
     '''Returns True if the tap is valid.'''
-    regex = re.compile(r'^(\S+)/(homebrew-)?(\w+)$')
+    regex = re.compile(r'^(\S+)/(homebrew-)?(\S+)$')
     return regex.match(tap)
 
 


abadger:
This looks like a dupe of:  https://github.com/ansible/ansible-modules-extras/pull/297

In that PR the reporter linked to: https://github.com/Homebrew/homebrew/blob/493edcf1/Library/Homebrew/tap_constants.rb#L2

The regex at that link location is:

```
HOMEBREW_TAP_ARGS_REGEX = %r{^([\w-]+)/(homebrew-)?([\w-]+)$}
```

whereas ours is currently:

```
regex = re.compile(r'^(\S+)/(homebrew-)?(\w+)$')
```

Should we just be making our regex match that?

(And yes, maybe we should get rid of our check and just look at error codes from talking to homebrew.... but that's going to be a more invasive change and I probably don't have time to review that now... I'd rather get something working into 1.9 and then review something like that for the next version.)

jayanthm:
@abadger We can set it to be the same as what Homebrew defines it as in their constants. However, that feels like a loosing battle as this repo will have to be kept up to date with this one. What I've done is make that check fairly loose (even less strict than theirs). I'll let you make the call on which you prefer.

abadger:
<nod>  We're going to be tagging 1.9rc2 any day now so I'll make the regex match what's in the homebrew repo for stable-1.9.  If you'd like to work up a PR that removes the is_valid checks and instead processes the output of the homebrew command to do the same thing I can look at merging that for 1.10/2.0 (whichever we end up calling the next release).

jayanthm:
@abadger Sounds good. I'll try to do that when I have some downtime

abadger:
Cool.  If I don't appear to see it, feel free to ping me on IRC to take a look.  I'm abadger1999 on freenode

PR #315, opened 2015-03-15
Add openstack keystone modules

emonty:
Add modules for managing keystone users, roles and projects. These are
end-user modules because in keystone v3, cloud consumers can be domain
admins and as such can create users, roles and projects in their domain.
In earlier keystone v2 clouds, only admin users would use these modules.

### DIFF
--- a/cloud/openstack/os_project.py
+++ b/cloud/openstack/os_project.py
@@ -1,0 +1,166 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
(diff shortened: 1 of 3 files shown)


emonty:
os_role is not done yet - but I pushed these up so that @dguerri could see them.

gregdek:
Adding "owner_pr" because @emonty owns other modules

gregdek:
@emonty is this module ready for review?

emonty:
Nope - but have a whole new set coming any day now that will be.

PR #316, opened 2015-03-15
Require semicolon to support pkgin version 0.7.0 because of parsable output

drscream:
The newest version of pkgin use parsable output by non tty sessions. Because
some description have also hyphens it result in an error for checking installed
packages. For example: "openvpn-2.3.6;=;Easy-to-use SSL VPN daemon"

Details about the pkgin version could be found here:
  https://github.com/NetBSDfr/pkgin/blob/master/CHANGES#L48
  https://github.com/NetBSDfr/pkgin/blob/master/main.c#L62

### DIFF
--- a/packaging/os/pkgin.py
+++ b/packaging/os/pkgin.py
@@ -93,7 +93,7 @@ def query_package(module, pkgin_path, name):
         #     '<' - installed but out of date
         #     '=' - installed and up to date
         #     '>' - installed but newer than the repository version
-        pkgname_with_version, raw_state = out.split(' ')[0:2]
+        pkgname_with_version, raw_state = out.split(';')[0:2]
 
         # Strip version
         # (results in sth like 'gcc47-libs')


drscream:
##### Issue Type:

Bugfix Pull Request

##### Ansible Version:

ansible 1.8.4 (with the devel `pkgin.py` file)

##### Environment:

```
SunOS router 5.11 joyent_20140717T041004Z i86pc i386 i86pc Solaris
pkgin 0.7.0 for SunOS-5.11 x86_64 (using SQLite 3.8.7.4)
```

I'm using the last version 2014Q4 release of the MIBE Image from Joyent SmartOS called `base64-lts`.

##### Summary:

Because the new `pkgin` version (since 0.7.0) support parseable output we should use it in the `pkgin.py`. The biggest problem is `pkgin` automatically use the parseable output if no TTY is used.

##### Steps To Reproduce:
1. Install `pkgin` version 0.7.0 or newer on the destination system
2. Use an example `pkgin` task for `openvpn`

```
---
  - pkgin: name=openvpn state=present
```

##### Expected Results:

The package should be installed, if it's already installed it should be shown ans present.

##### Actual Results:

```
failed: [192.168.201.1] => {"failed": true}
calculating dependencies... done.
nothing to do.
```

By debugging the issue a little bit more I found out that the `pkgin.py` module try to parse the output `openvpn-2.3.6;=;Easy-to-use SSL VPN daemon`. The module try to split by ` `and`-`. Because the` `split could not be applied it will split by the last`-`result in`openvpn-2.3.6;=;Easy-to`.

bcoca:
should this not be conditional both on version and if currently using a tty or not?

drscream:
Yes we should check the version of `pkgin` and decide if we've parseable output or not. But I'm not sure if it's a good idea to run `pkgin -v` on each command before installing a package. An idea would be to check for semicolons but maybe it's also allowed in the description of a package.

jeffrizzo:
A better solution might be to use the '-p' flag unless pkgin errors out (I'm about to excoriate the pkgin author for changing the output if it's not a tty, which is what's causing this issue in the first place)

Or, split on either ';' or ' '.  In any case, this needs fixing - I ran into it on ansible 1.9.1 on NetBSD :(

Do the modules for other package managers do any sort of version checking? Or just assume things like the pkgin module does?

bcoca:
closing this PR in favor of competing #454 as it seems to address this in a backwards compatible manner

PR #318, opened 2015-03-16
windows firewall control

t-v:
A module to control the windows firewall through ansible


windows/win_fw.ps1
```diff
@@ -0,0 +1,346 @@
+#!powershell
+#
+# WANT_JSON
+# POWERSHELL_COMMON
+
+function getFirewallRule ($fwsettings) {
+    try {
+            
+        #$output = Get-NetFirewallRule -name $($fwsettings.name);
+        $rawoutput=@(netsh advfirewall firewall show rule name=$($fwsettings.Name))
```
petemounce: Why not the powershell `*-NetFirewallRule` cmdlets...?

t-v:
I originally wrote it to use the *-NetFirewallRule cmdlets, but it gave issues on making them idempotent and that's why I continued using netsh since that uses the same variables for the set as the get

gregdek:
ping for review: @trondhindenes @petemounce @elventear @smadam813 @jhawkesworth @angstwad @sivel @chrishoffman @cchurch

jhawkesworth:
I have two pretty minor suggestions...

Windows firewall has the concept of 'profiles' - public, private and domain.
I guess not everybody needs to set rules that apply to a specific profile, but it would be nice if the documentation mentioned that the module manipulates rules for all profiles only.

Also I wonder if win_firewall_rule would be a better name for the module, since it doesn't change whether the firewall is on or off as far as I can tell.

t-v:
@jhawkesworth instead of just updating the docs with the fact that it did not support profiles, I actually added the option and updated the docs with the new option.
I also renamed the module from win_fw to win_firewall_rule

trondhindenes:
@TimothyVandenbrande what were your issues using the powershell cmdlets? BTW: this dsc resource does pretty much the same thing, maybe use for inspiration: https://github.com/PowerShell/xNetworking/blob/dev/DSCResources/MSFT_xFirewall/MSFT_xFirewall.psm1

t-v:
as far as I can remember (it's been a while), there were 2 things:
- compatibility with my win 2008 servers
- and the fact that either the input or the output values were different for the get module than for the set module.
  These 2 things were the main reason for using netsh, especially the backwards compatibility.

gregdek:
FYI, looks like we've got broken docstring in your branch, due to breakage in the bundler module which is our fault and unrelated... can you rebase? Sorry for the confusion.

t-v:
OK, I was already looking what I did wrong, because I kniw my module worked
but for some reason it said there was an error.
But I will rebase 1st thing tomorrow.

trondhindenes:
@TimothyVandenbrande great work! A few things we need to take care of before I proposed to get this merged in. Some are important, some are just me being picky.
Must-changes:
-The $Profile variable is special in PowerShell. Please replace it with another name. Using $profile makes scripts unstable, and makes it next to impossible to debug.

Recommended changes:
-use the "failifempty switch for required arguments to clean up code. This gives us a standardized output for missing required arguments:
`Get-Attr -obj $params -name attributename -failifempty $true -resultobj $result`

I can't get this module to run on PSv4. I'm suspecting the problem is the use of $profile. Would you update the code and push a new version? I'd be happy to test again once that's done.

t-v:
I will do that 1st thing tomorrow together with the debase.

gregdek:
Also needs "version_added: 2.0"

t-v:
so far I have changed the $profile variable to $winprofile
(worked on PS3 and PS4, since last time I only tested it with the $profile variable on my win2008 server)
I have rebased it to the latest upstream devel
and I added the version.

t-v:
@trondhindenes I tried using the fail if empty, but I find the downside that you don't see a list of missing arguments, but you recieve 1 by 1. That was the main reason I first check all of them and than report the missing ones.

If you prefer that I change it to a more normalized structure it's ok, but this is why it has my preference.

trondhindenes:
@TimothyVandenbrande I see your point. We should probably add support for allowed argument values in the pars-args function, but that's a different discussion.

gregdek:
So @trondhindenes -- is this a +1 yet? Any clear revisions that need to be made?

trondhindenes:
Yep, +1

robynbergeron:
Apologies for the delay here. :( The good news is.... yay, marking for inclusion! pinging @ansible/core

windows/win_firewall_rule.ps1
```diff
@@ -0,0 +1,354 @@
+#!powershell
+#
```
bcoca, line 2: missing license

t-v:
@robynbergeron @bcoca 
Added the license in the powershell part.

robynbergeron:
@TimothyVandenbrande thanks for the quick update.  Moving back to shipit, the label previously known as priority_reviewed ... since I'm assuming this does not need retesting or anything since it's just adding a license. @bcoca and @jhawkesworth can holler at me if I am assuming wrong :)

cc @ansible/core for inclusion.

PR #228, opened 2015-01-27
enable setting downtimes for Service-Group services and hosts

michaelgroening:
Hi I enabled setting of downtimes for servicegroup services and hosts in the nagios module. I would appreciate if this PR could be included in the nagios module of ansible-modules-extras.

PR #324, opened 2015-03-18
monitoring/nagios: Allow comment to be specified

tmzullinger:
The default remains 'Scheduling downtime' but can be overridden.

Adding comments about why hosts/services are in downtime helps those who view nagios/icinga understand issues.

I have used this in production this with 1.8.4 on a CentOS 6 host with Icinga.  I ran the unit tests from a Fedora 20 system.

### DIFF
--- a/monitoring/nagios.py
+++ b/monitoring/nagios.py
@@ -51,6 +51,11 @@ options:
        Only usable with the C(downtime) action.
     required: false
     default: Ansible
+  comment:
+    description:
+     - Comment for C(downtime) action.
+    required: false
+    default: Scheduling downtime
   minutes:
     description:
       - Minutes to schedule downtime for.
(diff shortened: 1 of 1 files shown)


gregdek:
@tbielawa please review this PR; if you +1, we will merge.

monitoring/nagios.py
```diff
@@ -51,6 +51,11 @@
        Only usable with the C(downtime) action.
     required: false
     default: Ansible
+  comment:
```
bcoca, line 54: ~~I would default it to null (None/empty) as that follows current behavior, adding it with a default changes this.~~
nvmd, you took this out of the function and into the spec, I'm fine with that then.

timlnx:
Sorry I'm late to the party. Looks like everything is good on these two PR's then eh? (also: #228 )

PR #328, opened 2015-03-19
Added overlayfs backend type to the lxc_container module

cloudnull:
This commit adds the overlayfs type to the lxc_container module. In
Adding the overlayfs type the commit adds the ability to clone a
container. While cloning is not locked down to only the overlayfs
container backend it is of particular interest when using the overlayfs
backend as it provides for amazingly fast snapshots/clones.

Changes to the resource types and documentation have been added on how
the new backend type can be used along with the clone operation.

This PR addresses a question asked by @fghaas on the original merged pull 
request "https://github.com/ansible/ansible-modules-extras/pull/123".

### DIFF
--- a/cloud/lxc/lxc_container.py
+++ b/cloud/lxc/lxc_container.py
@@ -38,6 +38,7 @@ options:
           - lvm
           - loop
           - btrfs
+          - overlayfs
         description:
           - Backend storage type for the container.
         required: false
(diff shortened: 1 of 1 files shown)


cloud/lxc/lxc_container.py
```diff
@@ -351,6 +396,15 @@
...
         }
+    },
+    'clone': {
+        'variables': {
+            'backing_store': '--backingstore',
+            'lxc_path': '--lxcpath',
+            'fs_size': '--fssize',
+            'name': '--orig',
+            'clone_name': '--new'
+        }
```
hughsaunders, line 425: need clone_snapshot here?

cloudnull:
@hughsaunders - no the snapshot option is parsed within the clone command. Those variables are only the "required" variables when using that action.

cloudnull:
The commits here have been squashed.

cloudnull:
@hughsaunders the latest squashed commit makes the module more declarative while also providing snapshot/archive functionality within the clone method.

bcoca:
So i would like to change the semantics, no new 'clone' state, just a clone_from: field, if the name already exists it should not clone again when re-running the playbook.

cloudnull:
@bcoca I will get right on that.

PR #331, opened 2015-03-20
Fix patch action plugin

### LINKED ISSUE #338: patch module does not support local files
Given the following task:

```
- name: Apply patches
  patch: >
    basedir="/home/user/virtualenv/lib64/python2.7/site-packages/"
    src="patches/some.patch"
    remote_src=False
```

plus an accompanying file in `files/patches/some.patch` in the same role, I get:

```
failed: [remotehost] => {"failed": true}
msg: src patches/some.patch doesn't exist or not readable
```

I ran Ansible through an strace and found it does not even attempt to stat the local file, so this does not look like a typo/configuration error to me.

renard:
This patch requires the one provided in PR: https://github.com/ansible/ansible/pull/10507

Fixes #338

### DIFF
--- a/files/patch.py
+++ b/files/patch.py
@@ -145,6 +145,9 @@ def main():
         module.fail_json(msg="patch command not found")
     patch_func = lambda opts: module.run_command("%s %s" % (patch_bin, ' '.join(opts)))
 
+    # patch need an absolute file name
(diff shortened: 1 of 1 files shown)


renard:
Thanks.
Please do not forget ansible/ansible#10507 as given in the description for this fix to work.

PR #334, opened 2015-03-23
Add `verify` param; set false to ignore SSL cert

msabramo:
Ignoring SSL cert verification may be necessary when testing with a server that has a self-signed certificate.

See https://github.com/ansible/ansible-modules-extras/pull/288#issuecomment-85196736

Cc: @mhite, @sudarkoff

### DIFF
--- a/network/f5/bigip_facts.py
+++ b/network/f5/bigip_facts.py
@@ -56,6 +56,14 @@ options:
         default: null
         choices: []
         aliases: []
+    validate_certs:
+        description:
+            - If C(no), SSL certificates will not be validated. This should only be used
+              on personally controlled sites using self-signed certificates.
+        required: false
+        default: 'yes'
+        choices: ['yes', 'no']
+        version_added: 1.9.1
     session:
         description:
             - BIG-IP session support; may be useful to avoid concurrency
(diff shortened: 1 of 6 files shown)


bcoca:
can you make it validate_certs? as to not confuse it with other types of verification and make it the same as other modules already use.

msabramo:
@bcoca: Yep, that makes sense. Will do.

msabramo:
I just added `validate_certs` to the `DOCUMENTATION` strings.

bcoca:
fyi, there won't be a 1.8.5, the soonest this will be released will be 1.9.1 or 2.0, we are in the middle of releasing 1.9.

msabramo:
Updated `version_added` to `1.9.1`.

msabramo:
It's too bad that these modules are bundled with ansible, as that means waiting a long time for updates. Probably this and a lot of other modules would be better off if they were shipped separately from ansible core.

https://groups.google.com/forum/#!topic/ansible-devel/4gm047a0JT0

msabramo:
@bcoca: How does this look now?

bcoca:
good, will merge in soon

PR #44, opened 2014-10-02
add zabbix modules

cove:
add zabbix modules for groups, hosts screens and webchecks

PR #343, opened 2015-03-28
add zabbix_screen module

cove:
Split up of PR: #44


monitoring/zabbix_screen.py
```diff
@@ -0,0 +1,423 @@
...
+    login_user:
+        description:
+            - Zabbix user name.
+        required: true
+        default: null
+    login_password:
+        description:
+            - Zabbix user password.
+        required: true
+        default: null
```
bcoca, line 53: required fields do not have defaults

PR #345, opened 2015-03-28
add zabbix_host module

cove:
Split up of PR #44

### DIFF
--- a/monitoring/zabbix_host.py
+++ b/monitoring/zabbix_host.py
@@ -1,0 +1,458 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2013-2014, Epic Games, Inc.
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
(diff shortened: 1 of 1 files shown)


monitoring/zabbix_host.py
```diff
@@ -0,0 +1,458 @@
...
+#
+
+DOCUMENTATION = '''
+---
+module: zabbix_host
+short_description: Zabbix host creates/updates/deletes
+description:
+   - When the host does not exists, a new host will be created, added to any host groups and linked to any templates.
+   - When the host already exists, the host group membership will be updated, along with the template links and interfaces.
+   - Delete a host from Zabbix if the host exists.
```
bcoca, line 29: this seems to describe what present/absent do

monitoring/zabbix_host.py
```diff
@@ -0,0 +1,458 @@
...
+options:
+    server_url:
+        description:
+            - Url of Zabbix server, with protocol (http or https).
+              C(url) is an alias for C(server_url).
+        required: true
+        default: null
+        aliases: [ "url" ]
+    login_user:
+        description:
```
bcoca, line 43: auth information and docs can be refactored into module_utils as shared code (look at ec2 modules).

monitoring/zabbix_host.py
cove, line 29: i think it's being explicit that it's not just a dumb add/remove of the host, but it links and unlinks it properly.

PR #346, opened 2015-03-28
add zabbix_group module

cove:
Split up of PR #44

### DIFF
--- a/monitoring/zabbix_group.py
+++ b/monitoring/zabbix_group.py
@@ -1,0 +1,208 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2013-2014, Epic Games, Inc.
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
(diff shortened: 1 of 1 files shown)


bcoca:
this looks like a rewrite of the current zabbix_group module, aside from timeout what other advantages does this one have over the previous?

cove:
it predates the module in the repo, it's not based on it. this one handles the quirks returned by the zabbix server, such as creating host groups, and the args are consistent with the other zabbix modules.

bcoca:
can you elaborate on the 'quirks'? 

Also the module in the repo was submitted well before this one, which is why it probably made it in first. The parameters also do not change with respect to your modules (with the exception of timeout).

cove:
looking back at email threads from the time this was written in 2013, it looks like there were a few issues we ran into when creating a zabbix groups.

1) the error returned by the server was too generic to determine if the creation failed due to the group already existing or another error, it would have required checking the error message. additionally if the group did exist, the server would sometimes create duplicate groups . so we ended up having to just test if the group exists, and that's been stable for the past year under heavy use. Update: it looks like we eventually found we could catch already exists exceptions, and needed to catch it in addition to checking if the group already exists.
2) running the create group command in parallel can also cause other generic API failures from the server, so there's an explanation on how to work around the issue in the doc section.

as for the arguments, zabbix_host and zabbix_group take a list of groups. the current module only takes one group.

Yeah this previous module made it in a few days before this one after the modules were split out from ansible core, it had been in the PR queue for a while before that as it was being iterated on.

let me know if you'd like more info. :)

gregdek:
Tagging as "owner_pr" because @cove is an owner of other Extras modules -- but it can't be merged. So this is probably a good test case for policy of "do we automerge new modules by experienced contributors".

resmo:
There is a similar module from me already in extras. 

But this one looks like more maintainable and improved. It would like to replace the existing one in favour of this one. The only thing that needs to be done is to add an alias `host_group` for argument `host_groups` for backward compatibly.

resmo:
@bcoca can be closed as #524 has been merged.

bcoca:
closing as per merge above

PR #354, opened 2015-03-30
Add docker_compose module

ghost:
This adds support for using docker-compose (formerly fig) as a module.
For more information, see here: http://docs.docker.com/compose/


bcoca:
The exception handling is too generic, errors will depend on
underlying api being clear, which is not always the case, we prefer
more fine grained messages on our side, always adding the message from
the underlying library.

cloud/docker/docker_compose
```diff
@@ -0,0 +1,319 @@
...
+version_added: 1.8.4
+short_description: Manage docker containers with docker-compose
+description:
+     - Manage the lifecycle of groups of docker containers with docker-compose
+options:
+  command:
+    description:
+      - Select the compose action to perform
+    required: true
+    choices: ['build', 'kill', 'pull', 'rm', 'scale', 'start', 'stop', 'restart', 'up']
```
bcoca: command is not a pattern we like to follow, we prefer a more imperative 'state', in this case some I would need to know more about some of the options  which don't seem to fit

cloud/docker/docker_compose
```diff
@@ -0,0 +1,319 @@
...
+  compose_file:
+    description:
+      - Specify the compose file to build from
+    required: true
+    type: bool
+  insecure_registry:
```
bcoca: validate_certs is the parameter that we use for other modules and we like keeping that consitent

cloud/docker/docker_compose
```diff
@@ -0,0 +1,319 @@
...
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
```
bcoca: in general missing required: false and the default

cloud/docker/docker_compose
```diff
@@ -0,0 +1,319 @@
...
+    default: SIGKILL
+  no_build:
+    description:
+      - Do not build an image, even if it's missing
+    type: bool
+  no_deps:
+    description:
+      - Don't start linked services
+    type: bool
+  no_recreate:
```
bcoca: if using a state=present this should just be the default, unless the other parameters do not match, in which case a force=yes/no seems more appropriate

cloud/docker/docker_compose
abadger: insecure_registry follows the parameter from the docker module.

cloud/docker/docker_compose
abadger: You mean s/imperative/declarative/ right?

cloud/docker/docker_compose
bcoca: it can still be an alias, I would prefer to keep ansible consistent and allow people to use the same option name for the same thing

cloud/docker/docker_compose
bcoca: ^ what he said.

cloud/docker/docker_compose
abadger: I don't think it is the same thing... The docker module actually has several other parameters that would come into play if  insecure_registry==False to actually validate certs.  In the docker module, those are named tls_*... we might want to alias one of them to validate_certs if the functionality is actually equivalent there.

insecure_registry really just controls whether docker requires tls or not.  It doesn't do anything about verifying whether the client or server certificates are valid.

abadger:
I noticed that this module lacks the .py extension.  In current ansible, the ansible modules are python modules as well so they need to be named like docker_compose.py

cloud/docker/docker_compose
bcoca: ok, then we need a better description, as not knowing the actual docker equivalent it mislead me to equate them.

cloud/docker/docker_compose
ghost: In this case though, 'state' isn't really accurate. state=scale doesn't make much sense to me. state=up and state=start are two different commands and I think it would lead to confusion.

cloud/docker/docker_compose
ghost: I can use the docker modules description of "Use insecure private registry by HTTP instead of HTTPS", would that suffice? The current description is taken from docker-compose itself.

cloud/docker/docker_compose
ghost: By default, docker-compose recreates containers when rerun (stops/destroys/starts). This isn't really the desired behavior for ansible modules as I understand since it actually forces a change each time.

Thoughts?

cloud/docker/docker_compose
ghost: Aren't required: false and default=None the defaults here? Or do they need to be explicitly set?

I see no best practice for this and many modules have missing "required" and "default" fields if it is a requirement.

ghost:
To be clear, for error handling the expectation is to catch errors from docker-compose and have an explict message in the module instead of passing up the exceptions message?

bcoca:
no, normally the APIs give generic messages and are not very helpful to map to the ansible input, we prefer to use exception handling on 'connect/create/destroy' and return a message with the action attempted + the underlying exception.

gregdek:
closing and reopening to kick travis, since there are a bunch of unrelated failures here

gregdek:
OK, travis issues are cleared. @SamYaple if you can address @bcoca issues about exception handling and then let us know, we can re-review for inclusion.

ghost:
I am going to have to abandon this for now. Each new revision of docker-compose is changing key parts of the structure. Almost like it wasn't meant to be a library to be imported ;)

I am going to wait for that structure to stabilize before attempting. If someone else wants to continue the work they can go for it

gregdek:
Thanks @SamYaple.

PR #357, opened 2015-03-31
Lxc container home fix

### LINKED ISSUE #337: Bug: Unsafe call in lxc_container
The grep below may not work as intended. The grep will blindly match any line is the passwd file that contains the result of $(whoami) and is unsafe. Consider matching the user id, using getent, or something safer that what is used here.

```
# This is used to attach to a running container and execute commands from
# within the container on the host.  This will provide local access to a
# container without using SSH.  The template will attempt to work within the
# home directory of the user that was attached to the container and source
# that users environment variables by default.
ATTACH_TEMPLATE = """#!/usr/bin/env bash
pushd "$(grep $(whoami) /etc/passwd | awk -F':' '{print $6}')"
    if [[ -f ".bashrc" ]];then
        source .bashrc
    fi
popd
```

bcoca:
fixes #337

### DIFF
--- a/cloud/lxc/lxc_container.py
+++ b/cloud/lxc/lxc_container.py
@@ -398,7 +398,7 @@ LXC_ANSIBLE_STATES = {
 # home directory of the user that was attached to the container and source
 # that users environment variables by default.
 ATTACH_TEMPLATE = """#!/usr/bin/env bash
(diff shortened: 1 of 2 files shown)


sivel:
Looks like an extra commit for gluster may have snuck its way in?

bcoca:
damn ....

PR #358, opened 2015-03-31
Pacman module add upgrade & force features

k0ste:
add upgrade feature, patch by n0vember- (https://github.com/ansible/ansible-modules-extras/pull/41)


packaging/os/pacman.py
```diff
@@ -54,6 +54,14 @@
         choices: ["yes", "no"]
         version_added: "1.3"
 
+    force:
+        description:
+            - Force remove package, without any checks.
+        required: false
+        default: "no"
+        choices: ["yes", "no"]
+        version_added: "1.3"
```
bcoca: version added should be 2.0

packaging/os/pacman.py
```diff
@@ -61,6 +69,13 @@
         required: false
         default: "no"
         choices: ["yes", "no"]
+
+    upgrade
+        description:
+            - Whether or not to upgrade whole system
+        required: false
+        default: "no"
+        choices: ["yes", "no"]
```
bcoca, line 78: missing version added

packaging/os/pacman.py
```diff
@@ -130,6 +151,27 @@ def update_package_db(module):
     else:
         module.fail_json(msg="could not update package db")
 
+def upgrade(module):
+    cmdupgrade = "pacman -Suq --noconfirm"
```
bcoca, line 156: you should use module's find_bin vs expecting it to be in path

packaging/os/pacman.py
resmo, line 78: missing `:` after `upgrade`

k0ste, line 78: added

packaging/os/pacman.py
RubenKelevra: Well, before upgrading you should always sync with "y", so the command should be "pacman -Syuq --noconfirm". Doing a update_cache=yes is not necessary because an "-Syy" takes longer than needed.

RubenKelevra:
You should add yourself as author. :)

k0ste:
U think? :-[

RubenKelevra:
Sure, you added code, so you should add yourself to the header I think. :)

k0ste:
I just want working module for best package manager.

gregdek:
@elasticdog Please review patch; if you +1 we will merge

packaging/os/pacman.py
```diff
@@ -130,6 +152,22 @@ def update_package_db(module):
...
+    cmdneedrefresh = "pacman -Supq"
```
elasticdog, line 157: This is somewhat a minor implementation point, but using `pacman -Qu` will tell you via the exit code if there are any outdated packages (rather than trying to counting newlines from the parsed output of `pacman -Supq`). It might make the logic a bit cleaner to use that here?

elasticdog:
I've not tested locally, but aside from my comment above, this looks sane to me :+1:

gregdek:
So @elasticdog does that mean the PR needs to be updated to adhere to your comments, or are you good with it as-is? 

Also: this seems like a pretty big patch not to test...

elasticdog:
@gregdek you do realize that I'm just a random guy who submitted a patch to this module almost two years ago, right? I'm happy to offer suggestions, but I don't have an Ansible dev environment set up right now and don't really have the time to do that in order to test this PR. I think the code looks good, but you'll probably want to get someone else to verify that it's working as intended.

gregdek:
@elasticdog will do. The issue is that the original author, "afterburner", seems to be nowhere to be found, so this module is essentially without a maintainer currently.

@k0ste if we get this patch reviewed by someone on the core team at Ansible, are you willing to take over as maintainer of the module?

gregdek:
Putting into core_review as it seems we have no maintainer.

bcoca:
items from previous review still to be addressed

bcoca:
closing this through merged #762 which added the feedback on top of these changes and kept original attribution.

Issue #353, opened 2015-03-29
known_hosts module fails on attempt to remove non-present entry


mkoshelev:
##### Issue Type:

Bug report
##### Ansible Version:

ansible 1.9.0.1
  configured module search path = None
##### Environment:

Fedora 21
##### Summary:

Attempt to remove host entry via known_hosts module fails the task if the host entry in question is not present in .ssh/known_hosts.
##### Steps To Reproduce:

Example playbook:

```
- name: Test known_hosts module
  hosts: localhost

  tasks:
  - name: Try to remove entry from .ssh/known_hosts
    known_hosts: host="no.such.entry" state=absent
```

If for some reason you have 'no.such.entry' in .ssh/known_hosts - just run the playbook twice :)
##### Expected Results:

Task completes successfully. No changes reported.
##### Actual Results:

Task fails:

```
failed: [localhost] => {"cmd": "/usr/bin/ssh-keygen -F no.such.entry -f /home/user/.ssh/known_hosts", "failed": true, "rc": 1}
```

mcv21:
This is a problem with newer ssh-keygens - upstream changed its behaviour in 6.4 such that ssh-keygen -F returns 1 when it doesn't find a host; prior to that it returned 0 in this case.

This tripped us up when we moved our client machines from Debian wheezy to Debian jessie. I've been working on a fix, which is now complete, so I'll submit a pull request.

mcv21:
OOI, why is this still listed as Open? I think we fixed this...

bcoca:
yes, github only closes automatically when issues is references a 'fixes #issue' in descriptions, not in titles. closing this as per fix merged above

PR #359, opened 2015-03-31
Fix for issue #353 (handle change in ssh-keygen behaviour)

mcv21:
Prior to openssh 6.4, ssh-keygen -F returned 0 (and no output) when no
host was found. After then, it instead returned 1 and no output. This
revised code behaves correctly with either behaviour. There is
currently no other code path that results in exit(1) and no output.


mcv21:
This commit fixes #353

PR #368, opened 2015-04-01
Add type property to zypper

greghaynes:
To install patterns and other package types a type parameter needs to be
specified to zypper.

### DIFF
--- a/packaging/os/zypper.py
+++ b/packaging/os/zypper.py
@@ -50,6 +50,12 @@ options:
         required: false
         choices: [ present, latest, absent ]
         default: "present"
+    type:
+        description:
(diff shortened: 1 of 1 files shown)


alxgu:
You forgot the documentation. For example - zypper: name=kvm_server type=pattern state=present

I would suggest to change the variable name from install_type to type_of_pkg or only in the package_absent function.

:+1:

greghaynes:
bumping to get this merged

alxgu:
version_added: (2.0?) is missing in the documentation block.

robinro:
Thanks @greghaynes  :+1:

PR #376, opened 2015-04-04
Add an accept_licenses option to pkg5 module

mavit:
This accepts any software licences required by the package.

### DIFF
--- a/packaging/os/pkg5.py
+++ b/packaging/os/pkg5.py
@@ -39,6 +39,13 @@ options:
     required: false
     default: present
     choices: [ present, latest, absent ]
+  accept_licenses:
+    description:
+      - Accept any licences.
+    required: false
+    default: false
+    choices: [ true, false ]
+    aliases: [ accept_licences, accept ]
 '''
 EXAMPLES = '''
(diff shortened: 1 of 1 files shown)


packaging/os/pkg5.py
```diff
@@ -39,6 +39,13 @@
     required: false
     default: present
     choices: [ present, latest, absent ]
+  accept:
+    description:
+      - Accept any licences.
+    default: false
+    required: false
+    default: false
```
bcoca, line 46: extra default

packaging/os/pkg5.py
```diff
@@ -39,6 +39,13 @@
     required: false
     default: present
     choices: [ present, latest, absent ]
+  accept:
```
bcoca: can we rename to accept_licenses? you can keep accept as an alias

resmo:
and there is a typo in the commit msg

mavit:
- Option renamed to `accept_licenses`.
- Duplicate `default` removed from docs.
- Typo corrected in commit message.
- Replaced changes missing due to a bad merge.

PR #382, opened 2015-04-08
Adding openstack load balancer management

sky-shiny:
Adds load balancer support to openstack module.

### DIFF
--- a/cloud/openstack/os_load_balancer.py
+++ b/cloud/openstack/os_load_balancer.py
@@ -1,0 +1,284 @@
+#!/usr/bin/env python
+#THIS SOFTWARE CONTRIBUTION IS PROVIDED ON BEHALF OF BSKYB LTD.
+#BY THE CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES,
+#INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+#MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED
+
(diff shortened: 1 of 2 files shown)


gregdek:
Thanks for this PR. 

On our end, we need to figure out the workflow for OpenStack modules in particular, since they all seem to be living in core nowadays. We keep adding this directory in Extras and then removing it. 

So let's sort out some policy here. @emonty what should we do with this and other OpenStack modules in Extras?  My sense:
1. Have the openstack module directory in Extras, since there will be some modules that are "non-mainstream". (Not judging whether this is one of them.)
2. On a case-by-case basis, decide whether each new module request belongs in Extras or Core -- a decision I will happily defer.

For new modules that belong here, we can put them through the regular process.

Thoughts on this approach?

emonty:
@gregdek I think that's a wonderful process and I fully support it!

emonty:
@sky-shiny Thanks for the contribution! I think it would be great to have lb support. We need to get this aligned with the other modules though, as this one does not consume and of the same auth support or standard parameters. I'd love to get the logic into shade, but it's not strictly necessary as you can totally get the neutron client from shade and we can refactor later.

I recommend checking out the README.md in cloud/openstack in ansible-modules-core. It's got some info on how we're structuring these.

@gregdek per the previous process, I think load-balancers are a normal enough thing that they should likely target core, although we'll need to clean up the interface first. I'm going to leave it open and see if @sky-shiny would like to rework it over there, or if we should just put it on our list of things to get to.

gregdek:
OK. @sky-shiny, ball is in your court. Tell us how you'd like to proceed.

Issue #378, opened 2015-04-07
win_chocolatey doesn't work with 0.9.9 chocolatey release (which it now installs by default)


mkoshelev:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.9.0.1
  configured module search path = None
##### Environment:

Fedora 21
##### Summary:

The recent 0.9.9 release of chocolatey broke compatibility with the previous versions in a way which prevents win_chocolatey module to function properly:
1) install/uninstall via win_chocolatey hangs as choco command now asks for action confirmation - can be fixed by adding '-y' parameter to choco command line in the module;
2) choco command reports successful installation slightly differently which breaks result detection and makes win_chocolatey module report failure on successful install. Could be probably fixed by adding the new pattern to check - "has been installed successfully".

Note, win_chocolatey is installing incompatible chocolatey release 0.9.9 by default if there's no previous version installed on target system.
##### Steps To Reproduce:

Just try to install any package via win_chocolatey on Windows system where no chocolatey is installed yet. 
##### Expected Results:

chocolatey is installed and then requested package is installed.
##### Actual Results:

Playbook execution hangs. After stopped by Ctrl+C - chocolatey appears to be installed, the requested package is listed as installed by choco but no real application is present. Chocolatey log shows command was waiting on confirmation dialog.

elventear:
Please check https://github.com/ansible/ansible-modules-extras/pull/398 for a possible solution

anthonygreen:
This issue persists in Ansible 1.9.3. Is there a fix available?

elventear:
I think you need to use the develop branch in order to take advantage of that fix.

nartamonov:
When it will be integrated to stable Ansible version, which is now unusable with Windows because Chocolatey is very important part of configuration management? Fix was provided several months ago...

gregdek:
Extras updates are, as a matter of policy, not backported to the stable branch of Ansible.

bdgscotland:
Hey folks, I've worked around this by running this command as part of the play: 
`choco feature enable -n allowGlobalConfirmation`

nitzmahone:
closing, fix released in 2.0

ferventcoder:
@nitzmahone Thanks for the update Matt! Good seeing you (again) recently on the PowerScripting podcast!

PR #391, opened 2015-04-12
fixed chocolatey prompting for confirmation

arBmind:
chocolatey requires the `-y` option to install without opening a dialog since the release of 0.9.0


arBmind:
The text output is now different as well. I hope there is another way to check for success and changes.

I am not sure if this requires a version check. Can someone with an older version confirm that it works?

arBmind:
This is related to issue #378

elventear:
Please check https://github.com/ansible/ansible-modules-extras/pull/398

PR #392, opened 2015-04-13
Sendgrid API notifications module

mattmakai:
This module uses the Sendgrid API (as opposed to SMTP) to send notification emails to one or more email addresses.

### DIFF
--- a/notification/sendgrid.py
+++ b/notification/sendgrid.py
@@ -1,0 +1,143 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2015, Matt Makai <matthew.makai@gmail.com>
+#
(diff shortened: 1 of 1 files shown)


abadger:
Reviewed this at the pycon sprints.  Looks like a good new module!

Some things we discussed as we were working on this:
- Currently we return changed=False even though an action is performed.  This follows what the other notification modules do.  The thought is that these are not changes to the host running the command but external side-effects.  So we do not consider it a change.  This might change in the future but if so we'll change all the notification modules at the same time.
- We currently don't call out the idempotency of these notification modules either as they all follow the same pattern.  But it seemed good to document that running a task with this module would send notifications every time it's run.

# Merged

Hi!

issue is resolved.

Thank you!

PR #395, opened 2015-04-14
Added ceph custom modules

aisrael:
Hi. We've written a few custom modules for Ceph that we are now starting to use in our own playbooks.

Notably, they add the ability to:

``` yaml
# Create a pool
- ceph_osd_pool: state=present name=data pgnum=128 pgpnum=128
# Delete a pool
- ceph_osd_pool: state=absent name=pool0
# Create an RBD
- ceph_rbd: name=rbd1 state=present size=32 image_format=2 image_shared=true
# Resize (enlarge) an RBD, if it exists (or create if it doesn't)
- ceph_rbd: name=rbd1 state=present size=64
# Resize (shrink) an RBD, if it exists (or create if it doesn't)
- ceph_rbd: name=rbd1 state=present size=32 allow_shrink=true
# Map an RBD
- ceph_rbd_map: name=rbd1 read_only=true
# Unmap an RBD
- ceph_rbd_map: name=rbd1 state=absent
# Delete an RBD
- ceph_rbd: name=rbd1 state=absent
```

I've taken steps to see to it that all the above operations are idempotent, following the Ansible philosophy, and that they set `result['changed']` accordingly.


srgvg:
Perhaps someting @leseb would like to look at?

gregdek:
@srvg good idea. @leseb isn't listed as an official reviewer, but in this case we might make an exception. 

Still, for new modules, especially a collection like this, we need our standard two +1s. Boilerplate follows:

---

leseb:
@srvg  I'm going to have a look at this thanks!

cloud/ceph/ceph_osd_pool.py
```diff
@@ -0,0 +1,163 @@
...
+        result['cmd'] = cmd
+        result['stdout'] = out.rstrip("\r\n")
+        result['stderr'] = err.rstrip("\r\n")
+        result['start'] = str(startd)
+        result['end'] = str(endd)
+        result['delta'] = str(delta)
+
+    changed = False
+
+    with rados.Rados(conffile='/etc/ceph/ceph.conf') as cluster:
```
leseb, line 114: What about multi-cluster support?
Shouldn't we expose a ceph_config variable that will take `/etc/ceph/ceph.conf` for default so the user can override it if necessary?

cloud/ceph/ceph_osd_pool.py
```diff
@@ -0,0 +1,163 @@
...
+        pools = cluster.list_pools()
```
leseb, line 115: It looks like you're only using librados to list the pools, why don't use librados to perform all the operations instead of using Ceph cli commands?

cloud/ceph/ceph_rbd.py
```diff
@@ -0,0 +1,201 @@
...
+
+EXAMPLES = '''
+# Create an rbd image (or enlarge an existing one)
+- ceph_rbd: name=rbd1 size=128
+# Create an rbd image with a specified format and sharing (needed for mapping in v0.92)
+- ceph_rbd: name=rbd2 image_format=2 image_shared=true
+# Shrink an existing image, if it exists (otherwise, create one)
+- ceph_rbd: name=rbd1 size=32 allow_shrink=true
+# Remove an rbd
+- ceph_rbd: name=rbd1
```
leseb, line 89: shouldn't this be `ceph_rbd: name=rbd1 state=absent`?

leseb:
I'm generally delighted to see some modules for Ceph in Ansible.
One thing that I've found a big curious is that there is a mixed combinaison of libraries and CLI usage.
Is there any reason behind? For instance, you use librados to list the pools where you could simply use `ceph osd pool ls` if you want to stick with the CLI.
As far as I know most of the things that we do here are doable with both librados and librbd.

One last thing is that we should probably support clusters with different names than ceph.
We should also be careful and not assume that the user used to connect is `client.admin` (same for the key), which seems to be the case at the moment.

Thanks the work, I think it's a first good step.

gregdek:
Thanks @aisrael, and thanks @leseb for the review!

I think the next step for you is to make the changes that @leseb recommended. If we can get his sign-off, we'll move on to reviewing for Ansible-specific issues. You can double-check those yourself by looking at the following list:

Let us know when you're ready for a re-review. Thanks!

aisrael:
Thank you, @leseb and @gregdek for the review.

I realise I've taken some, ahem, shortcuts in my implementation (specifically, calling the `ceph` CLI)—we just wanted something up and usable in the shortest time possible.

Also, I wasn't so comfortable possibly duplicating code from the `ceph` CLI (itself a Python script), for both licensing and efficiency reasons (have to maintain two separate codebases if the `ceph` CLI behaviour changes).

However, for lack of a higher-level ceph Python library, I'll go ahead and try to implement everything using `librados` and `librbd` and get rid of CLI calls.

gregdek:
Thanks @aisrael. Looking forward to the next iteration.

leseb:
@aisrael thanks I look forward to seeing the next iteration. :)

robynbergeron:
Hi @aisrael -- just checking in to see how this is going. Let us know! :)

robynbergeron:
Hi @aisrael -- checking in again, as it's been awhile. Going to mark this as pending action since it's been more than a month -- please let us know what your plans are, or if you're just going to create a new PR when you have time and if we can close this. Thanks!

robynbergeron:

aisrael:
Hi. I've actually managed to get the Ceph `osd` commands to use direct `rados` library calls. However, I couldn't do the same for the `rbd` commands since the `rbd` executable was written in C++ (though, upon closer inspection it seems we can at least add an RBD mapping by `echo`-ing to `/sys/bus/rbd`).

I think at this point I'll just raise another pull request when I'm finished with all the above changes.

gregdek:
Thanks @aisrael -- we'll close this one now and await your new PR. Looking forward to it!

PR #402, opened 2015-04-15
sys.exit used, but not imported.

ndonegan:
My vim setup also removed some trailing white space.

### DIFF
--- a/network/dnsimple.py
+++ b/network/dnsimple.py
@@ -130,13 +130,12 @@ EXAMPLES = '''
 '''
 
 import os
-from sys import exit
 try:
     from dnsimple import DNSimple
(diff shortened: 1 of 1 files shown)


network/dnsimple.py
```diff
@@ -130,12 +130,13 @@
...
 
 import os
+from sys import exit
 try:
     from dnsimple import DNSimple
     from dnsimple.dnsimple import DNSimpleException
 except ImportError:
     print "failed=True msg='dnsimple required for this module'"
-    sys.exit(1)
+    exit(1)
```
bcoca: actually, this should not be an exit at all, it should set a var so later in main we can exist with fail_json

PR #405, opened 2015-04-15
Filtering out messages from MySQL server

banyek:
You can decide if MySQL server complains should be shown or not. 
You can filter out nogthing/warnings/errors.


bcoca:
still have the issue with the name as 'all' also seems to filter out errors

banyek:
Oh, now I get what you meant. 
Ok, I'll fix this soon

gregdek:
@banyek According to our new policy, we will auto-merge any changes you request, since you are the author of this module. I'm assuming that you don't want to merge this PR yet because it's not quite right, yes?  In which case, could we close it until you're ready with the new PR?

banyek:
sweet <3

PR #410, opened 2015-04-19
Add ec2_eni module (aws network interfaces)

wimnat:
Adds eni support for AWS.  Can be used to create, destroy, attach, detach and list ENIs.

Tested in my own environment and seems good but would appreciate some more testing by anyone else.

### DIFF
--- a/cloud/amazon/ec2_eni.py
+++ b/cloud/amazon/ec2_eni.py
@@ -1,0 +1,404 @@
+#!/usr/bin/python
+#
+# This is a free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This Ansible library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this library.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: ec2_eni
+short_description: Create and optionally attach an Elastic Network Interface (ENI) to an instance
+description:
+    - Create and optionally attach an Elastic Network Interface (ENI) to an instance. If an ENI ID is provided, an attempt is made to update the existing ENI. By passing 'None' as the instance_id, an ENI can be detached from an instance.
+version_added: "2.0"
+author: Rob White, wimnat [at] gmail.com, @wimnat
+options:
+  eni_id:
+    description:
+      - The ID of the ENI
+    required = false
+    default = null
+  instance_id:
+    description:
+      - Instance ID that you wish to attach ENI to. To detach an ENI from an instance, use 'None'.
+    required: false
+    default: null 
+  private_ip_address:
+    description:
+      - Private IP address.
+    required: false
+    default: null
+  subnet_id:
+    description:
+      - ID of subnet in which to create the ENI. Only required when state=present.
+    required: true
+  description:
+    description:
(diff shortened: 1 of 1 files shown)


evanccnyc:
@wimnat You should add the region option so that can be set there.

I'd love to have this checked in. Ill see who I can push to get it done.

wimnat:
@evanccnyc specifying region does work but it's not documented explicitly as i assume "extends_documentation_fragment: aws" will include it.

evanccnyc:
@wimnat Sorry about that I should have read more closely.

evanccnyc:
@bcoca anything else that needs to be done for this to get merged?

cloud/amazon/ec2_eni.py
```diff
@@ -0,0 +1,406 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this library.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: ec2_eni
+short_description: Create and optionally attach an Elastic Network Interface (ENI) to an instance
+description:
+    - Create and optionally attach an Elastic Network Interface (ENI) to an instance. If an ENI ID is provided, an attempt is made to update the existing ENI. By passing 'None' as the instance_id, an ENI can be detached from an instance.
+version_added: "1.9"
```
bcoca: 2.0

cloud/amazon/ec2_eni.py
```diff
@@ -0,0 +1,406 @@
...
+author: Rob White, wimnat [at] gmail.com, @wimnat
+options:
+  instance_id:
+    description:
+      - Instance ID that you wish to attach ENI to 
+    required: false
+    default: null 
+    aliases: []
```
bcoca: no need for empty aliases

cloud/amazon/ec2_eni.py
```diff
@@ -0,0 +1,406 @@
...
+    description:
+      - Private IP address.
+    required: false
+    default: null
+    aliases: []
+  subnet_id:
+    description:
+      - Subnet in which to create the ENI
+    required: true
+    default: null
```
bcoca: if required, default does not make sense

cloud/amazon/ec2_eni.py
```diff
@@ -0,0 +1,406 @@
...
+  delete_on_termination:
+    description:
+      - Delete the interface when the instance it is attached to is terminated. You can only specify this flag when the interface is being modified, not on creation.
+    required: false
+  source_dest_check:
+    description:
+      - By default, interfaces perform source/destination checks. NAT instances however need this check to be disabled. You can only specify this flag when the interface is being modified, not on creation.
+    required: false  
+extends_documentation_fragment: aws
+requirements: [ "boto" ]
```
bcoca: duplicates requirements already in doc fragments

cloud/amazon/ec2_eni.py
wimnat: @bcoca subnet_id is required for state: present but not for state: absent.  What's the best way to document?

bcoca: in the description

evanccnyc:
Any idea when this can get checked in?

cloud/amazon/ec2_eni.py
```diff
@@ -0,0 +1,396 @@
...
+    description:
+      - Optional description of the ENI.
+    required: false
+    default: null
+  security_groups:
+    description:
+      - List of security groups associated with the interface. Only used when state=present.
+    required: false
+    default: null
+  state:
```
bcoca, line 54: you should document possible states here with a choices: [ ...] entry, also the 'list' state is indicative of code belonging in a facts module.

wimnat:
The list state has been moved out to a separate facts module and I've added in the choices for 'state'.  I also squashed the commits.

gregdek:
@bcoca can you re-review?

PR #411, opened 2015-04-20
Added yum_repository module

binrush:
This module provides the ability to create yum repositories. In my opinion, it's more convinient to have separate module for this then to use copy/template modules.

```
# this is more readable then template/copy way
yum_repositoty: id=myrepo baseurl=http://mysite.com/myrepo/
```

Second, this module can be used with ansible command, for example to disable repositories on all host one can run:

```
ansible -m yum_repository -a "enabled=true"
```

I've tested this module on CentOS 6.5


packaging/os/yum_repository.py
```diff
@@ -0,0 +1,365 @@
...
+        if changed and not self.check_mode:
+            cp.write(open(repo_filename, 'w'))
+        return changed
+
+
+                
+
+def main():
+    module = AnsibleModule(
+        argument_spec = dict(
```
detiber, line 305: It would be nice to be able to specify additional repo attributes as well (for example, if using yum-plugin-priorities, being able to specify the repo priority would be nice).

possibly an additional_attributes param that is a list of key/values.

Also, you could probably move most of the less frequently used attributes to additional_attributes.

Overall, I can't wait until I can use this instead of the custom role I'm currently using.

packaging/os/yum_repository.py
```diff
@@ -0,0 +1,365 @@
...
+module: yum_repository
+short_description: Add and remove yum repositories
+description:
+    - Add or remove an yum repositories.
+notes:
+    - This module works on any distribution with yum installed.
+options:
+    id:
+        required: true
+        default: none
```
resmo, line 33: there is no default value if param is required.

packaging/os/yum_repository.py
```diff
@@ -0,0 +1,365 @@
...
+        description:
+            - Must be a unique name for each repository, one word.
+    state:
+        required: false
+        choices: [ "absent", "present" ]
+        default: "present"
+        description:
+            - A repository state.
+    name:
+        required: false
```
resmo, line 43: there should be a default if not requiref

packaging/os/yum_repository.py
```diff
@@ -0,0 +1,365 @@
...
+        description:
+            - A human readable string describing the repository.
+    baseurl:
+        required: false
+        description:
+            - URL to the directory where the yum repository’s ‘repodata’ directory lives.
+    enabled:
+        required: false
+        choices: [ "yes", "no" ]
```
resmo, line 52: default?

packaging/os/yum_repository.py
```diff
@@ -0,0 +1,365 @@
...
+        description:
+            - Relative cost of accessing this repository. 
+    skip_if_unavailable:
+        required: false
+        choices: [ "yes", "no" ]
+        description:
+            - If set to yes yum will continue running if this repository cannot be contacted for any reason.
+
+author: Rushan Shaymardanov
+version_added: "0.1"
```
resmo, line 178: this should be the next ansible major version. e.g. "2.0"

resmo:
i am missing a lot of defaults, in doc as well in the param dict.

packaging/os/yum_repository.py
binrush, line 43: If parameter is not specified, it will not be added to repository configuration and yum will use it's own default for it. If I add defaults for all non-required parameters, module will put them all in config , that is not very good I think.

resmo, line 43: if I understand the logic correctly, you check for `None` right? So the default would be `default: null`in this case.

binrush, line 43: I looked at existing yum module https://github.com/ansible/ansible-modules-core/blob/devel/packaging/os/yum.py . Parameter 'enablerepo' is not required and in documentation it has default: null, but in argument_spec it doesn't have default parameter. So, should I just add default: null in documentation to all parameters?

binrush:
As suggested by detiber, I think to rewrite this module using check_invalid_arguments=False, it will allow user to pass any parameters (including custom) to repository config and will reduce argument spec. Is it normal practice for ansible modules?

binrush:
As suggested by @detiber, I',m going to rewrite this module to use check_invalid_arguments=False to allow user to pass any parameter (including custom) to repository config and will reduce argument list. Is this allowed practice for ansible modules?

detiber:
I think having a dictionary argument would be cleaner than disabling check_invalid_arguments.

robynbergeron:
Hi @binrush -- just checking in to see how this is going. Can you let us know? :)

gregdek:

binrush:
@gregdek , I just need to decide, should I keep this module as is (with explicit list of parameters) or it will be better to rewrite it using check_invalid_arguments=False ?

robynbergeron:
Hi @binrush -- I can't answer your last question (or your question above from May 7 re: allowed practices -- sorry I missed that!) -- maybe @resmo or @detiber or @gregdek have an answer, also pinging @ansible/core folks in case they have a moment to answer. Sorry we kept you waiting on an answer, glad to see you're still interested in moving forward. :D

gregdek:
From what I can tell of @detiber comments on May 7, he said:

"I think having a dictionary argument would be cleaner than disabling check_invalid_arguments."

That's the only answer I have without more context. Maybe @detiber or @resmo can weigh in further.

packaging/os/yum_repository.py
detiber, line 305: @binrush personally, I would add an new param called additional_attributes that takes a dict to allow for setting repo attributes for plugins.

I'd avoid disabling param checking just because it seems like it could create a confusing user experience.

An example showing what I'm thinking:

```
yum_repository: 
  id: example
  baseurl='http://example.com/repo/'
  additional_attributes:
    priority: 1
```

DavidWittman:
This is a duplicate of https://github.com/ansible/ansible-modules-extras/pull/1016, which is now merged.

bcoca:
closing as per merged #1016 above, if there are any features this module adds that the merged one is missing, feel free to submit as an update.

PR #414, opened 2015-04-21
Update mail.py to allow html content

adematte:
Adding parameter subtype to allow html based content. The default behavior of text/plain has been preserved.

### DIFF
--- a/notification/mail.py
+++ b/notification/mail.py
@@ -110,6 +110,11 @@ options:
       - The character set of email being sent
     default: 'us-ascii'
     required: false
+  subtype:
+    description:
+      - The minor mime type, can be either text or html. The major type is always text.
+    default: 'plain'
+    required: false
 """
 
 EXAMPLES = '''
@@ -183,7 +188,8 @@ def main():
             body = dict(default=None),
             attach = dict(default=None),
             headers = dict(default=None),
-            charset = dict(default='us-ascii')
+            charset = dict(default='us-ascii'),
+            subtype = dict(default='plain')
         )
     )
 
@@ -200,6 +206,7 @@ def main():
     attach_files = module.params.get('attach')
     headers = module.params.get('headers')
     charset = module.params.get('charset')
+    subtype = module.params.get('subtype')
     sender_phrase, sender_addr = parseaddr(sender)
 
     if not body:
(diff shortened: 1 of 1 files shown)


gregdek:
@dagwieers Please review; if you give +1 we will merge

bcoca:
needs rebase

notification/mail.py
```diff
@@ -111,12 +111,16 @@
       - The character set of email being sent
     default: 'us-ascii'
     required: false
+  subtype:
```
bcoca, line 113: mimetype might be a better name for this option

adematte, line 113: true! i'll rebase and make the necessary changes when I get the chance to.

dagwieers, line 113: Well, it is not really a mimetype you have to provide, maybe mimesubtype would be preferable ?
I prefer subtype over mimetype (as it is more correct), cfr. https://docs.python.org/2/library/mimetypes.html

Whatever the decision is, I support this change. +1

bcoca:
we can split the alias? I'm just guessing people will not be as familiar with the subtype as with the mimetype name for this feature.

dagwieers:
Why not just make it clear in the documentation that subtype, is the "minor" part of a "text" MIME type ? (which isn't very clear now from the description)

At least that way it does not pretend to expect a full mimetype, and anyone who does not know what is to be expected, will read from the documentation what it is. Besides, now only 'plain' and 'html' are allowed, or are any of the other "text" MIME types a possibility as well ? e.g. text/richtext, text/sgml, text/xml or text/x-uuencode

bcoca:
@dagwieers agreed, documenting it clearly seems to be the best solution

adematte:
rebased and improved description of the parameter. Good to go?

gregdek:
Back to @dagwieers for re-review.

dagwieers:
I support this change. Simple and concise. +1

gregdek:
Merging. Thanks @dagwieers for the review, and thanks @adematte for the PR.

resmo:
version_added of the new arg is missing...

abadger:
Good spot @resmo!  Added version_added: 955bf92ff892a7359a045e1ddb3b29b7809a230b

PR #417, opened 2015-04-22
Document mail module's 'msg' param and fix examples

nanobeep:
The 'msg' alias for 'subject' isn't in the documentation, so adding it.

In the gmail example, it uses both the 'subject' and 'msg' params, but 'msg' is an alias of 'subject', so you are essentially declaring the same param twice. If you use this example, then no subject is sent (I tested with gmail). Documentation example is updated to use 'body' as intended.

Also, updated the simple example to use 'subject' instead of the 'msg' alias since it is more explicit.

### DIFF
--- a/notification/mail.py
+++ b/notification/mail.py
@@ -61,7 +61,7 @@ options:
     required: false
   subject:
     description:
-      - The subject of the email being sent.
+      - The subject of the email being sent. Alias: I(msg)
     aliases: [ msg ]
     required: true
   body:
@@ -115,7 +115,7 @@ options:
 
 EXAMPLES = '''
 # Example playbook sending mail to root
-- local_action: mail msg='System {{ ansible_hostname }} has been successfully provisioned.'
+- local_action: mail subject='System {{ ansible_hostname }} has been successfully provisioned.'
 
 # Sending an e-mail using Gmail SMTP servers
 - local_action: mail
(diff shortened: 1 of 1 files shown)


bcoca:
So msg was documented already:
     aliases: [ msg ]

But you are correct about the example being misleading, even though it seems msg was intended as an alias for body and got incorrectly assigned to subject.????

nanobeep:
Hmm, the alias isn't showing on the actual documentation page, so that's why I added it explicitly: http://docs.ansible.com/mail_module.html (search for 'msg' and it only shows the mention in the examples)

As far as the original intention of 'msg', I'm not sure what the author intended, but I agree it is likely a mis-aliased param.

nanobeep:
For the aliases documentation, I followed the example in the [files module](https://github.com/ansible/ansible-modules-core/blob/devel/files/file.py) which explicitly documents the aliases since they aren't automatically documented (I'm assuming). Perhaps they should be auto-documented and not explicitly documented?

bcoca:
ah .. im too used just looking at the 'documentation source' and did not realize the html style sheet did not include aliases

nanobeep:
Cool :) My main concern with exploring that 'msg' may have been intended to be for 'body' is that this has been in the wild for awhile and there are probably many people using the examples with 'msg' as an alias for the subject. Changing that alias could break 'mail' for a lot of people. So perhaps 'msg' should stay aliased to subject?

The 'msg' param is a bit confusing all around, so maybe it makes sense to remove it from the documentation altogether and consider it deprecated?

bcoca:
agreed on all points, remove it from the example and don't add it to the description.

nanobeep:
Cool, I've updated the pull-request accordingly.

PR #423, opened 2015-04-25
Add check_mode support to HipChat module

JeanFred:
The HipChat module declares to support check_mode,
but the message is sent in any case.

With this, if executed in check mode, the module will exit
before actually sending the message to HipChat.

I based this on the flowdock behaviour, except that the module returns things as changed, since things would have changed (as a message would have been sent). This is my understanding of changed and check mode, but happy to change it one way or the other.

(Also fix a couple of PEP8 issues a bit further down).

### DIFF
--- a/notification/hipchat.py
+++ b/notification/hipchat.py
@@ -96,6 +96,11 @@ def send_msg(module, token, room, msg_from, msg, msg_format='text',
 
     url = api + "?auth_token=%s" % (token)
     data = urllib.urlencode(params)
+
+    if module.check_mode:
+        # In check mode, exit before actually sending the message
+        module.exit_json(changed=False)
+
     response, info = fetch_url(module, url, data=data)
     if info['status'] == 200:
         return response.read()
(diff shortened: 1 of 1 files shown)


bcoca:
notification modules are a weird beast, we don't consider sending messages 'changed' as they should always do this and 'changed' is reserved for 'system modifications' which a notification does not really do.

JeanFred:
Thanks for reviewing this @bcoca. I’m happy to have the module report Unchanged in check mode. One thing though, right now this module always returns Changed=True in normal run. Should that be changed too?

Your position that sending e message is not changed sounds reasonable to me − I think this should be documented somewhere (or I missed it). I’d be happy to submit a couple of pull requests if this needs to be harmonized. (Right now my issue is rather that some notification modules do send the message in check mode).

bcoca:
The rule might not be in the docs, but I do remember being part of the online discussion that led to it (I was writing the jabber module at the time). 

So your implementation is consistent with how the module currently works, it is just not consistent with the 'undocumented rule', we can change this here or in a diff PR if you wish.

JeanFred:
@bcoca I changed to match the undocumented rule, squashed and rebased. 

The Travis build is failing but if I understand correctly it is because other modules have syntax not compatible with Python 2.4 (and not even the HipChat module).

PR #438, opened 2015-04-30
win_webpicmd module for IIS module installation

petemounce:
Chocolatey 0.9.9+ deprecated support for the `webpi` custom source, so I needed to write this.

[Windows Web Platform Installer](http://www.microsoft.com/web/downloads/platform.aspx) is a way of installing products and applications for Microsoft IIS on Windows. It has a [command line](http://www.iis.net/learn/install/web-platform-installer/web-platform-installer-v4-command-line-webpicmdexe-rtw-release); this ansible module allows IIS modules to be installed via this means.

To find out names of modules, use `webpicmd /list /listoption:available`.

Notes:
- `webpicmd` must be installed and on `PATH` first (see `win_chocolatey` module; package is `webpicmd`)
- `webpicmd` does not allow modules to be uninstalled
- IIS must be installed first (see `win_feature` module; package is `Web-Server`)
- Installations will
  - accept EULA (which otherwise requires user input)
  - suppress reboots (so you have to manage those; see `win_reboot` module)

### DIFF
--- a/windows/win_webpicmd.ps1
+++ b/windows/win_webpicmd.ps1
@@ -1,0 +1,135 @@
+#!powershell
+# This file is part of Ansible
+#
+# Copyright 2015, Peter Mounce <public@neverrunwithscissors.com>
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+$ErrorActionPreference = "Stop"
+
+# WANT_JSON
+# POWERSHELL_COMMON
+
+$params = Parse-Args $args;
+$result = New-Object PSObject;
+Set-Attr $result "changed" $false;
+
+If ($params.name)
+{
+    $package = $params.name
+}
+Else
+{
+    Fail-Json $result "missing required argument: name"
+}
+
+Function Find-Command
+{
+    [CmdletBinding()]
+    param(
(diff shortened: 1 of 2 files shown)


gregdek:
Adding @owner_pr flag because @petemounce is co-owner of other Extras modules

bcoca:
i would require a state=present| absent so not only can we install but remove packages

petemounce:
@bcoca webpicmd itself [does not support removing packages](http://www.iis.net/learn/install/web-platform-installer/web-platform-installer-v4-command-line-webpicmdexe-rtw-release) as far as I can tell. [It was requested in their uservoice](http://webmatrix.uservoice.com/forums/128313-webmatrix-suggestions/suggestions/2164851-uninstall-feature-in-webpi) but declined with the suggested workaround of using add/remove programs.

gregdek:
Hey @petemounce -- what's the status on this one? Ready for a re-review?

petemounce:
@gregdek yes, I think so.

petemounce:
@trondhindenes @elventear ?

trondhindenes:
On it - I'll get it tested tomorrow.

trondhindenes:
Yup, looks good - +1. 
I noticed that the module returns a success even if you give it a non-existing package, but webpicmd also does that, so it's no problem per se. An option would be to validate the package against output from webpicmd /list but that probably involves a bit of nasty parsing since the output is pure text as opposed to xml/json/whatever. That may be something for the next version of this. 
Another enhancement is probably to test %programfiles%\microsoft\webxxxx\webpicmd.exe if the exe isn't available thru $env:path. But I guess it should be based on the way webpi installs itself.

These are both optional enhancements imho.

In any case, +1 from me.

trondhindenes:
btw @petemounce you could probably just parse the entire output of the "webpicmd /list /listoption:all" command and do a select-string on it and see if there's anything returned (good) or not (bad). Rudimentary but it should pick up most typos at least. Maybe add a "verify=off" for running without this verification (as it slows down the script with a couple of extra seconds).

petemounce:
@trondhindenes given that custom sources are probably coming _back_ to chocolatey soonish, I think I won't invest effort for the non-existing package filter. We'd probably ditch this module in favour of supporting installing webpi modules via chocolatey again.

Can I please get another +1?

gregdek:
Actually, @petemounce, since you're an approved reviewer, you get an extra implicit +1. So... merging. Thanks!

PR #440, opened 2015-05-01
Create win_dotnet_ngen module

petemounce:
When .NET is installed or updated, ngen is triggered to optimise the installation. This triggers high CPU while it's happening, and usually happens at an inconvenient time.

This allows you to trigger it when you like. Full details and background in doc.
[Reference](http://blogs.msdn.com/b/dotnet/archive/2013/08/06/wondering-why-mscorsvw-exe-has-high-cpu-usage-you-can-speed-it-up.aspx)

(I don't know a way to figure out whether this is required without actually running it.)

### DIFF
--- a/windows/win_dotnet_ngen.py
+++ b/windows/win_dotnet_ngen.py
@@ -31,9 +31,10 @@ description:
     - This happens via scheduled task, usually at some inopportune time.
     - This module allows you to run this task on your own schedule, so you incur the CPU hit at some more convenient and controlled time.
     - "http://blogs.msdn.com/b/dotnet/archive/2013/08/06/wondering-why-mscorsvw-exe-has-high-cpu-usage-you-can-speed-it-up.aspx"
-    - "Note: there are in fact two scheduled tasks for ngen but they have no triggers so aren't a problem"
-    - "Note: there's no way to test if they've been completed (?)"
-    - Note: the stdout is quite likely to be several megabytes
+notes:
+    - there are in fact two scheduled tasks for ngen but they have no triggers so aren't a problem
+    - there's no way to test if they've been completed (?)
+    - the stdout is quite likely to be several megabytes
 options:
 author: Peter Mounce
 '''
--- a/windows/win_dotnet_ngen.py
+++ b/windows/win_dotnet_ngen.py
@@ -33,6 +33,7 @@ description:
     - http://blogs.msdn.com/b/dotnet/archive/2013/08/06/wondering-why-mscorsvw-exe-has-high-cpu-usage-you-can-speed-it-up.aspx
     - Note: there are in fact two scheduled tasks for ngen but they have no triggers so aren't a problem
     - Note: there's no way to test if they've been completed (?)
+    - Note: the stdout is quite likely to be several megabytes
 options:
 author: Peter Mounce
 '''
--- a/windows/win_dotnet_ngen.py
+++ b/windows/win_dotnet_ngen.py
@@ -1,0 +1,43 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2015, Peter Mounce <public@neverrunwithscissors.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+# this is a windows documentation stub.  actual code lives in the .ps1
+# file of the same name
+
+DOCUMENTATION = '''
+---
+module: win_dotnet_ngen
+version_added: "1.9"
+short_description: Runs ngen to recompile DLLs after .NET  updates
+description:
+    - After .NET framework is installed/updated, Windows will probably want to recompile things to optimise for the host.
+    - This happens via scheduled task, usually at some inopportune time.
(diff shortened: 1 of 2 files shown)


windows/win_dotnet_ngen.py
```diff
@@ -0,0 +1,44 @@
...
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+# this is a windows documentation stub.  actual code lives in the .ps1
+# file of the same name
+
+DOCUMENTATION = '''
+---
+module: win_dotnet_ngen
+version_added: "1.9"
```
resmo: should be "2.0"

windows/win_dotnet_ngen.py
```diff
@@ -0,0 +1,44 @@
...
+short_description: Runs ngen to recompile DLLs after .NET  updates
+description:
+    - After .NET framework is installed/updated, Windows will probably want to recompile things to optimise for the host.
+    - This happens via scheduled task, usually at some inopportune time.
+    - This module allows you to run this task on your own schedule, so you incur the CPU hit at some more convenient and controlled time.
+    - http://blogs.msdn.com/b/dotnet/archive/2013/08/06/wondering-why-mscorsvw-exe-has-high-cpu-usage-you-can-speed-it-up.aspx
```
resmo: make sure you qoute any string containing : and is not yaml.

windows/win_dotnet_ngen.py
```diff
@@ -0,0 +1,44 @@
...
+    - Note: there are in fact two scheduled tasks for ngen but they have no triggers so aren't a problem
```
resmo: quote this line

windows/win_dotnet_ngen.py
```diff
@@ -0,0 +1,44 @@
...
+    - Note: there's no way to test if they've been completed (?)
```
resmo: quote this line

windows/win_dotnet_ngen.py
```diff
@@ -0,0 +1,44 @@
...
+    - Note: the stdout is quite likely to be several megabytes
```
resmo: and this as well

windows/win_dotnet_ngen.py
```diff
@@ -0,0 +1,44 @@
...
+options:
+author: Peter Mounce
+'''
+
+EXAMPLES = '''
+  # Run ngen tasks
+  win_dotnet_ngen:
```
resmo, line 44: imho this would not work. shouldn't it be `action: win_dotnet_ngen`?

bcoca, line 44: these 3 are equivalent:

```
- action: module=<modulename>
- action: <modulename>
- <modulename>:
```

petemounce:
@resmo and @bcoca thanks for the feedback; updated.

windows/win_dotnet_ngen.py
```diff
@@ -0,0 +1,44 @@
...
+module: win_dotnet_ngen
+version_added: "2.0"
+short_description: Runs ngen to recompile DLLs after .NET  updates
+description:
+    - After .NET framework is installed/updated, Windows will probably want to recompile things to optimise for the host.
+    - This happens via scheduled task, usually at some inopportune time.
+    - This module allows you to run this task on your own schedule, so you incur the CPU hit at some more convenient and controlled time.
+    - "http://blogs.msdn.com/b/dotnet/archive/2013/08/06/wondering-why-mscorsvw-exe-has-high-cpu-usage-you-can-speed-it-up.aspx"
+    - "Note: there are in fact two scheduled tasks for ngen but they have no triggers so aren't a problem"
+    - "Note: there's no way to test if they've been completed (?)"
```
bcoca: use notes field instead, at same level as options/descriptions:

petemounce: Thanks; done in 598d458569b2f6f24a236e740f0d81d72f12670d

petemounce:
The build fails in Travis, but #466 looks to address the failures (python 2.4 syntax).

gregdek:
@petemounce can I assume this is ready for re-review?

petemounce:
@gregdek yes

gregdek:
@ansible/windows friends, please review. Boilerplate follows.

---

smadam813:
@petemounce I will test this out after lunch and let you know how it works for me.

smadam813:
I tested locally and it does everything it is supposed to.  What is the meaning of the variable arity?  I get that it's specifying the architecture, but needed to look at the code to understand what it was.  Maybe better variable name?

I would suggest instead of calling Invoke-NGen twice, one with cmdlet specified and one without, to only call it once.  Then you can make a list within the function for supported architectures and loop over that list.  All you need to do if you want to add an architecture is to append to the list if done this way.  This is completely not necessary, just may be more clear.

petemounce:
@smadam813 hm... 
- `bitness`? I reckon just as obscure
- `x86orx64`? Misleading because of the confusing way Windows calls 64-bit-related folders
- `processor_architecture` - mirrors the env variable but will resolve to a different value
- `dotnet_framework_folder` - but it's not, it's a fragment of the path

Or, how about a variable `dotnet_framework_folder` with a set of allowed values

```
framework\v4.0.30319
framework64\v4.0.30319
...
```

but then that requires the module to be changed every time a new .NET version is released that goes into a new folder.

Or, perhaps, `path_to_ngen_exe`? But that requires the user to know where it is (which would go in the doc, but require the doc to be read).

smadam813:
@petemounce I think the way you are doing it will work just fine!  Like I said, this is completely not necessary.  My only thoughts were handling an arbitrary number of frameworks based off of information you can query rather than specifying them.  You could run the risk of having to update the script in the future.  Hope that makes sense.

I give this a +1 to be merged.

petemounce:
@smadam813 cool; thanks. I don't mind updating the module if new ngen executables appear - that's a reasonably rare occurrence (4x in 13 years) ;-)

Another passing review required before merge - @trondhindenes @elventear ?

elventear:
:+1:

gregdek:
Two +1s, no -1s -- merging. Thanks @petemounce, and thanks to the reviewers!

PR #441, opened 2015-05-01
win_scheduled_task module for windows

petemounce:
Fledgling module to allow scheduled tasks to be managed.

Note: At present, I only need enabled/disabled support. I'm using it to disable windows update scheduled tasks, since I'm managing that process myself. 

There's lots of scope for more features. See `Get-Help *Scheduled*` for inspiration :)

### DIFF
--- a/windows/win_scheduled_task.py
+++ b/windows/win_scheduled_task.py
@@ -34,18 +34,18 @@ options:
       - Name of the scheduled task
       - Supports * as wildcard
     required: true
-  state:
+  enabled:
     description:
       - State that the task should become
     required: false
     choices:
-      - Disabled
-      - Enabled
-    default: Enabled
+      - yes
+      - no
+    default: yes
 author: Peter Mounce
 '''
 
 EXAMPLES = '''
   # Disable the scheduled tasks with "WindowsUpdate" in their name
-  win_scheduled_task: name="*WindowsUpdate*" state=disabled
+  win_scheduled_task: name="*WindowsUpdate*" enabled=no
 '''
--- a/windows/win_scheduled_task.py
+++ b/windows/win_scheduled_task.py
@@ -34,7 +34,6 @@ options:
       - Name of the scheduled task
       - Supports * as wildcard
     required: true
-    default: null
   state:
     description:
       - State that the task should become
--- a/windows/win_scheduled_task.py
+++ b/windows/win_scheduled_task.py
@@ -24,7 +24,7 @@
 DOCUMENTATION = '''
 ---
 module: win_scheduled_task
-version_added: "1.9"
+version_added: "2.0"
 short_description: Manage scheduled tasks
 description:
     - Manage scheduled tasks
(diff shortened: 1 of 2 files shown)


bcoca:
@cchurch

windows/win_scheduled_task.py
```diff
@@ -0,0 +1,54 @@
...
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+# this is a windows documentation stub.  actual code lives in the .ps1
+# file of the same name
+
+DOCUMENTATION = '''
+---
+module: win_scheduled_task
+version_added: "1.9"
```
resmo: update to "2.0"

windows/win_scheduled_task.py
```diff
@@ -0,0 +1,54 @@
...
+short_description: Manage scheduled tasks
+description:
+    - Manage scheduled tasks
+options:
+  name:
+    description:
+      - Name of the scheduled task
+      - Supports * as wildcard
+    required: true
+    default: null
```
resmo: if required, there is no default. should be removed.

windows/win_scheduled_task.py
```diff
@@ -0,0 +1,54 @@
...
+    aliases: []
```
resmo: if there aren't aliases, no need to add a empty list. could be removed

petemounce:
@resmo thanks for the feedback; done.

gregdek:
Adding @owner_pr flag because @petemounce is co-owner of other Extras modules

bcoca:
instead of state=enabled|disabled, a enabled=yes|no flag is more within ansible norm. Also, could the module create/delete scheduled tasks, then state=present|absent will also be nice to have.

petemounce:
@bcoca I agree that adding create/delete would be a nice to have - but I don't need it, currently. It is a significant chunk of work that I won't have time to do in the short term.

perzizzle:
Has anyone signed up to add create/delete to this module? I have the need and could spend a little time working on it.

petemounce:
@perzizzle I have not, but happy to test those changes.

My particular use case for this module is disabling Windows update related tasks before baking an image, so I don't get punched with them running later and interesting with production traffic. Those tasks afaik can only be disabled, not deleted.

gregdek:
@petemounce is this ready for re-review in your opinion?

petemounce:
@gregdek yes. I have addressed all the points. I am not planning on increasing the scope to allow create/delete of scheduled tasks as @bcoca suggested - if that's required, then this should either not get merged, or get merged and an issue opened to come back and add it (@perzizzle was interested to implement).

perzizzle:
I'm still interested in adding this feature it just turned out to be non-trivial. I can have something ready within in the next week or two.

gregdek:
Thanks @perzizzle.

bcoca:
Ok with merging as is now and trying to add more features later on, incremental is good.

petemounce:
Thanks for the merge! :)

PR #444, opened 2015-05-03
Add docker_login module

olsaki:
- Ansible version of "docker login" CLI command
- Persists Docker registry authentification in .dockercfg (only login once - no need to specify credentials over and over again anymore)
- Works for all other docker-py based modules (docker, docker_images) as well as the Docker CLI client

### DIFF
--- a/cloud/docker/docker_login.py
+++ b/cloud/docker/docker_login.py
@@ -46,9 +46,9 @@ options:
     required: true
   email:
     description:
-       - The email address for the registry account
+       - The email address for the registry account. Note that private registries usually don't need this, but if you want to log into your Docker Hub account (default behaviour) you need to specify this in order to be able to log in.
     required: false
-    default: anonymous@localhost.local
+    default: None
   reauth:
     description:
        - Whether refresh existing authentication on the Docker server (boolean)
@@ -218,7 +218,7 @@ def main():
             registry        = dict(required=False, default='https://index.docker.io/v1/'),
             username        = dict(required=True),
             password        = dict(required=True),
-            email           = dict(required=False, default='anonymous@localhost.local'),
+            email           = dict(required=False, default=None),
             reauth          = dict(required=False, default=False, type='bool'),
             dockercfg_path  = dict(required=False, default='~/.dockercfg'),
             docker_url      = dict(default='unix://var/run/docker.sock'),
--- a/cloud/docker/docker_login.py
+++ b/cloud/docker/docker_login.py
@@ -84,7 +84,7 @@ Login to a Docker registry without performing any other action. Make sure that t
 
 - name: login to private Docker remote registry and force reauthentification
   docker_login:
-    registry: https://your.private.registry.io/v1/
+    registry: your.private.registry.io
     username: yourself
     password: secrets3
     reauth: yes
@@ -139,10 +139,6 @@ class DockerLoginManager:
 
         if self.reauth:
             self.log.append("Enforcing reauthentification")
-
-        # Extract hostname part from self.registry if url was specified.
-        registry_url = urlparse(self.registry)
-        self.registry = registry_url.netloc or registry_url.path
 
         # Connect to registry and login if not already logged in or reauth is enforced.
         try:
--- a/cloud/docker/docker_login.py
+++ b/cloud/docker/docker_login.py
@@ -229,7 +229,7 @@ def main():
     try:
         manager = DockerLoginManager(module)
         manager.login()
-        module.exit_json(changed=manager.has_changed(), msg=manager.get_msg(), registry=manager.registry)
+        module.exit_json(changed=manager.has_changed(), msg=manager.get_msg())
 
     except Exception as e:
         module.fail_json(msg="Module execution has failed due to an unexpected error", error=repr(e))
--- a/cloud/docker/docker_login.py
+++ b/cloud/docker/docker_login.py
@@ -153,6 +153,8 @@ class DockerLoginManager:
                 reauth=self.reauth,
                 dockercfg_path=self.dockercfg_path
             )
+        except DockerAPIError as e:
+            self.module.fail_json(msg="Docker API Error: %s" % e.explanation)
         except Exception as e:
             self.module.fail_json(msg="failed to login to the remote registry", error=repr(e))
 
--- a/cloud/docker/docker_login.py
+++ b/cloud/docker/docker_login.py
@@ -244,4 +244,5 @@ def main():
 # import module snippets
 from ansible.module_utils.basic import *
 
-main()
+if __name__ == '__main__':
+    main()
--- a/cloud/docker/docker_login.py
+++ b/cloud/docker/docker_login.py
@@ -220,7 +220,7 @@ def main():
         argument_spec = dict(
             registry        = dict(required=False, default='https://index.docker.io/v1/'),
             username        = dict(required=True),
-            password        = dict(required=True),
+            password        = dict(required=True, no_log=True),
             email           = dict(required=False, default=None),
             reauth          = dict(required=False, default=False, type='bool'),
             dockercfg_path  = dict(required=False, default='~/.docker/config.json'),
--- a/cloud/docker/docker_login.py
+++ b/cloud/docker/docker_login.py
@@ -139,6 +139,10 @@ class DockerLoginManager:
         if self.reauth:
             self.log.append("Enforcing reauthentification")
 
+        # Extract hostname part from self.registry if url was specified.
+        registry_url = urlparse(self.registry)
+        self.registry = registry_url.netloc or registry_url.path
+
         # Connect to registry and login if not already logged in or reauth is enforced.
         try:
             self.response = self.client.login(
--- a/cloud/docker/docker_login.py
+++ b/cloud/docker/docker_login.py
@@ -97,7 +97,6 @@ Login to a Docker registry without performing any other action. Make sure that t
 '''
 
 import os.path
-import sys
 import json
 import base64
 from urlparse import urlparse
--- a/cloud/docker/docker_login.py
+++ b/cloud/docker/docker_login.py
@@ -58,7 +58,7 @@ options:
     description:
        - Use a custom path for the .dockercfg file
     required: false
-    default: ~/.dockercfg
+    default: ~/.docker/config.json
   docker_url:
     descriptions:
        - Refers to the protocol+hostname+port where the Docker server is hosted
@@ -176,6 +176,9 @@ class DockerLoginManager:
 
         # Create dockercfg file if it does not exist.
         if not os.path.exists(self.dockercfg_path):
+            dockercfg_path_dir = os.path.dirname(self.dockercfg_path)
+            if not os.path.exists(dockercfg_path_dir):
+                os.makedirs(dockercfg_path_dir)
             open(self.dockercfg_path, "w")
             self.log.append("Created new Docker config file at %s" % self.dockercfg_path)
         else:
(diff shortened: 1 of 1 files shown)


cloud/docker/docker_login.py
```diff
@@ -0,0 +1,243 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this software.  If not, see <http://www.gnu.org/licenses/>.
+
+######################################################################
+
+DOCUMENTATION = '''
+---
+module: docker_login
+author: Olaf Kilian
+version_added: "1.9"
```
resmo: should be 2.0

cloud/docker/docker_login.py
```diff
@@ -0,0 +1,243 @@
...
+description:
+     - Ansible version of the "docker login" CLI command.
+     - This module allows you to login to a Docker registry without directly pulling an image or performing any other actions.
+     - It will write your login credentials to your local .dockercfg file that is compatible to the Docker CLI client as well as docker-py and all other Docker related modules that are based on docker-py.
+options:
+  registry:
+    description:
+       - URL of the registry, for example: https://index.docker.io/v1/
+    required: true
+    default: null
```
resmo: there is no default if required. can be removed.

cloud/docker/docker_login.py
```diff
@@ -0,0 +1,243 @@
...
+    aliases: []
```
resmo: No aliases is the default. can be removed.

cloud/docker/docker_login.py
```diff
@@ -0,0 +1,243 @@
...
+    import os.path
+    import sys
+    import json
+    import base64
+    import docker.client
+    from requests.exceptions import *
+    from urlparse import urlparse
+except ImportError, e:
+    print "failed=True msg='failed to import python module: %s'" % e
+    sys.exit(1)
```
resmo: Do not use print and sys.exit in modules. Look at this example here  https://github.com/ansible/ansible-modules-extras/blob/devel/cloud/cloudstack/cs_affinitygroup.py#L95
and here https://github.com/ansible/ansible-modules-extras/blob/devel/cloud/cloudstack/cs_affinitygroup.py#L209
how you fail with module.fail_json if a python module is missing

cloud/docker/docker_login.py
```diff
@@ -0,0 +1,243 @@
...
+        return ". ".join(self.log)
+
+    def has_changed(self):
+        return self.changed
+
+def main():
+
+    module = AnsibleModule(
+        argument_spec = dict(
+            registry        = dict(required=True, default=None),
```
resmo: There is no default if `required=True`

cloud/docker/docker_login.py
```diff
@@ -0,0 +1,243 @@
...
+        if self.response.has_key("Status"):
+            self.log.append(self.response["Status"])
+            if self.response["Status"] == "Login Succeeded":
+                self.changed = True
+        else:
+            self.log.append("Already Authentificated")
+
+        # Update the dockercfg if changed but not failed.
+        if self.has_changed():
+            self.update_dockercfg()
```
resmo, line 169: If I understand this right, It would be easy to implement check mode here if you add 

``` python
if self.has_changed() and not self.module.check_mode: 
     self.update_dockercfg()
```

and don't forget to add `supports_check_mode=True` just like this exmaple. https://github.com/ansible/ansible-modules-extras/blob/devel/cloud/cloudstack/cs_affinitygroup.py#L206

olsaki:
@resmo thanks for the review! Will update the PR asap.

olsaki:
@resmo thanks for your suggestions. I have updated the PR - can you have a look again please.

cloud/docker/docker_login.py
```diff
@@ -0,0 +1,244 @@
...
+       - The username for the registry account
+    required: true
+  password:
+    description:
+       - The plaintext password for the registry account
+    required: true
+  email:
+    description:
+       - The email address for the registry account
+    required: false
```
resmo, line 50: there should be a default if not required

cloud/docker/docker_login.py
```diff
@@ -0,0 +1,244 @@
...
+  docker_login:
+    username: docker
+    password: rekcod
+    email: docker@docker.io
+    dockercfg_path: /tmp/.mydockercfg
+
+'''
+
+import os.path
+import sys
```
resmo: imho this is not used. can be renoved.

resmo:
2 minor things, otherwise looks good to me!

cloud/docker/docker_login.py
olsaki: Removed "sys" import, but "os.path" is used here https://github.com/olsaki/ansible-modules-extras/blob/devel/cloud/docker/docker_login.py#L128 and here https://github.com/olsaki/ansible-modules-extras/blob/devel/cloud/docker/docker_login.py#L177

olsaki:
@resmo thanks again, good points, corrected those! Also added more meaningful error messages on registry login errors. I would appreciate a final review.

olsaki:
Anybody else? @bcoca is this clean to merge or any other fixes or rework needed?

gregdek:
ping @bcoca / @jimi-c / @abadger for mergey-ness

abadger:
Earlier I had wondered if this was the way we wanted authentication to work
(as separate modules) or if we wanted it to be integrated.  @bcoca
mentioned that there was precedent in another cloud module but we didn't
settle on whether it's a good architecture or not. We didn't discuss this
in the context of the new extras policies, though.  Is this some thing that
should be discussed amongst the current module authors on the mailing
list?  Our do we just want to have one current module author ok the
strategy?  Our something else?

bcoca:
I think it is a good question to raise in ansible-devel

olsaki:
Any news about this? Where can I follow the discussion?

gregdek:
@bcoca @jimi-c @abadger might I recommend that this be the topic of our next hangout?

gregdek:
@olsaki Sorry for the delay here, we'll get a conversation started about this soon.

cloud/docker/docker_login.py
```diff
@@ -0,0 +1,246 @@
...
+short_description: Manage Docker registry logins
+description:
+     - Ansible version of the "docker login" CLI command.
+     - This module allows you to login to a Docker registry without directly pulling an image or performing any other actions.
+     - It will write your login credentials to your local .dockercfg file that is compatible to the Docker CLI client as well as docker-py and all other Docker related modules that are based on docker-py.
+options:
+  registry:
+    description:
+       - URL of the registry, for example: https://index.docker.io/v1/
+    required: true
```
abadger: We probably want this to be optional, defaulting to docker hub

cloud/docker/docker_login.py
```diff
@@ -0,0 +1,246 @@
...
+        return ". ".join(self.log)
+
+    def has_changed(self):
+        return self.changed
+
+def main():
+
+    module = AnsibleModule(
+        argument_spec = dict(
+            registry        = dict(required=True),
```
abadger: Same as above -- make this optional and default to docker hub.

cloud/docker/docker_login.py
```diff
@@ -0,0 +1,246 @@
...
+       - Refers to the protocol+hostname+port where the Docker server is hosted
+    required: false
+    default: unix://var/run/docker.sock
+  timeout:
+    description:
+       - The HTTP request timeout in seconds
+    required: false
+    default: 600
+
+requirements: [ "docker-py" ]
```
abadger: Add a requirements "python >= 2.6"

Since docker-py requires newer python (that lets us know that we don't have to make this module compatible with pyhton-2.4.)

cloud/docker/docker_login.py
```diff
@@ -0,0 +1,246 @@
...
+        return self.changed
+
+def main():
+
+    module = AnsibleModule(
+        argument_spec = dict(
+            registry        = dict(required=True),
+            username        = dict(required=True),
+            password        = dict(required=True),
+            email           = dict(required=False, default='anonymous@localhost.local'),
```
abadger: Is this a sensible default?  I don't know what really happens to the email address inside of docker-py so I'm not sure if we must pass this into the API at all?

abadger: Ah -- from the commit history and comments, I see that you added this in response to the prior review.  I think that may have been a misunderstanding.  If the docker-py api is fine taking None here, we probably want the default to be None and document that the default is None in the documentation.

cloud/docker/docker_login.py
```diff
@@ -0,0 +1,246 @@
...
+        manager.login()
+        module.exit_json(changed=manager.has_changed(), msg=manager.get_msg())
+
+    except Exception as e:
+        module.fail_json(msg="Module execution has failed due to an unexpected error", error=repr(e))
+
+# import module snippets
+from ansible.module_utils.basic import *
+
+main()
```
abadger: Stylistically it's nice to use this:

``` python
if __name__ == '__main__':
    main()
```

That way if we ever need unittests for this module we can import it without running the module.

abadger:
Talked about this at today's community meeting and decided we're fine with the idea.  it is similar in function to openstack/os_auth.py.

I made some comments to fix up in the module then we can merge.

Is there anything about the docker modules in the core repository that should change because of this module?

gregdek:
@olsaki back to you. :)

memelet:
Would be nice if 'owner' and 'group' were arguments. When creating creds for a specific user (which I would guess is a common case) the .dockercfg file created is owned by root.

gregdek:

# 

gregdek:
@olsaki a friendly ping to remind you that we're waiting on your changes based on @abadger comments.

olsaki:
pong - back on it, sorry for the delay.

olsaki:
@gregdek Made the requested changes. Please give me a day to test the module again before updating the branch.

olsaki:
@abadger from my point of view there is no change required in any core module. This module integrates with all other docker-py based modules (docker, docker_images) as well as the Docker CLI client, by simply persisting the required login credentials.

olsaki:
There seem to be some changes in the underlaying docker-py now since the initial submission of this module. I have already adapted to some of those (like new location and structure of the auth file) but need some more time to look into others (like handling of private registries). Working on it.

olsaki:
works_for_me

olsaki:
@memelet nice idea, from my perspective we easily can do that. But I would like to wait with that until this module got initially accepted.

abadger:
Reviewed changes and they look good.

Do we need to specify a minimum version of docker-py to work with this?  Some of the changes look like the API may be backwards incompatible.

abadger:
olaski - if you're on IRC.freenode.net, I'm abadger1999 in #ansible-devel.  Feel free to ping me if you want to.  This is very close to being done and I don't want it to fall off my radar by accident ;-)

olsaki:
@abadger thanks for the fast reply.

I had some issues with docker-py < 1.1.0 (auth not working properly).

The module has been successfully tested with docker-py 1.1.0 and from what I have seen should be compatible with all recent versions upwards from there. However, there can be incompatibility with older versions of the docker CLI client that expects the auth config to be in a different location and format.

Will give #ansible-devel a visit on the next chance!

abadger:
@olsaki Cool.  So we should probably list minimum versions if we know them in the documentation (the requirements: field.  It's free form but we try do something like ["libraryname", "library2 >= 1.1.0", "application >= 1.0"].  And we should also check the docker-py version at the same time as we check that we're able to import it.  I think that will be all.

olsaki:
@abadger ok, will add that info as advised. Just need some time to clarify the min. version of the docker cli tool to work with this and will get this back to you asap.

abadger:
@olsaki Cool thanks.  Note that 2.0 is going to go from beta to rc pretty soon.  We're going to make the cutoff for new modules then so it would be good to get this in quickly.  If that's going to be difficult to verify the minimum version, we could merge this with a version that's somewhat too stringent (the minimum version that you've tested with) and then relax it later when you find out more information.

(And please feel free to ping me on IRC if you update but I don't seem to notice here... release is always a hectic time and it's easy to miss something.)

cloud/docker/docker_login.py
```diff
@@ -0,0 +1,251 @@
...
+    def has_changed(self):
+        return self.changed
+
+def main():
+
+    module = AnsibleModule(
+        argument_spec = dict(
+            registry        = dict(required=False, default='https://index.docker.io/v1/'),
+            username        = dict(required=True),
+            password        = dict(required=True),
```
abadger: Noticed one more thing:  we should mark this as no_log like this:

``` python
    password    = dict(required=True, no_log=True),
```

That way ansible will apply some heuristics to keep that parameter from being displayed in output, logged, etc.

olsaki:
@abadger thanks for your suggestions and support. Much appreciated! Implemented it and back to you. Sorry for the delay, hope I'm not just too late with it.

PR #445, opened 2015-05-03
Add webfaction modules

quentinsf:
These modules, which provide a wrapper around the webfaction API and allow you to create sites, apps, databases, domains and mailboxes, have proven fairly popular in my own github space, so I thought I should try and move them somewhere a bit more official!

Hope they're useful?

All the best, 
Quentin

P.S. I wasn't sure what to put in the version_added documentation field, so at present they say 1.99.

### DIFF
--- a/cloud/webfaction/webfaction_app.py
+++ b/cloud/webfaction/webfaction_app.py
@@ -72,7 +72,7 @@ def main():
     module = AnsibleModule(
         argument_spec = dict(
             name = dict(required=True),
-            state = dict(required=False, default='present'),
+            state = dict(required=False, choices=['present', 'absent'], default='present'),
             type = dict(required=True),
             autostart = dict(required=False, choices=BOOLEANS, default=False),
             extra_info = dict(required=False, default=""),
--- a/cloud/webfaction/webfaction_app.py
+++ b/cloud/webfaction/webfaction_app.py
@@ -13,7 +13,7 @@ short_description: Add or remove applications on a Webfaction host
 description:
     - Add or remove applications on a Webfaction host.  Further documentation at http://github.com/quentinsf/ansible-webfaction.
 author: Quentin Stafford-Fraser
-version_added: 1.99
+version_added: 2.0
 notes:
     - "You can run playbooks that use this on a local machine, or on a Webfaction host, or elsewhere, since the scripts use the remote webfaction API - the location is not important. However, running them on multiple hosts I(simultaneously) is best avoided. If you don't specify I(localhost) as your host, you may want to add C(serial: 1) to the plays."
     - See `the webfaction API <http://docs.webfaction.com/xmlrpc-api/>`_ for more info.
@@ -23,7 +23,6 @@ options:
         description:
             - The name of the application
         required: true
-        default: null
 
     state:
         description:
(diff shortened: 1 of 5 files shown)


cloud/webfaction/webfaction_app.py
```diff
@@ -0,0 +1,153 @@
...
+        required: true
+
+    login_password:
+        description:
+            - The webfaction password to use
+        required: true
+'''
+
+import xmlrpclib
+from ansible.module_utils.basic import *
```
sivel: The module_utils import should be located just before the call to `main()`

quentinsf:
Thanks guys - have made changes as per your suggestions..

cloud/webfaction/webfaction_app.py
```diff
@@ -0,0 +1,151 @@
...
+import xmlrpclib
+
+webfaction = xmlrpclib.ServerProxy('https://api.webfaction.com/')
+
+def main():
+
+    module = AnsibleModule(
+        argument_spec = dict(
+            name = dict(required=True),
+            state = dict(required=False, default='present'),
```
resmo: add choises=['present', 'absent'] here like you did in doc.

cloud/webfaction/webfaction_app.py
```diff
@@ -0,0 +1,151 @@
...
+# Quentin Stafford-Fraser 2015
+
+DOCUMENTATION = '''
+---
+module: webfaction_app
+short_description: Add or remove applications on a Webfaction host
+description:
+    - Add or remove applications on a Webfaction host.  Further documentation at http://github.com/quentinsf/ansible-webfaction.
+author: Quentin Stafford-Fraser
+version_added: 2.0
```
resmo: use "2.0" for versions

cloud/webfaction/webfaction_domain.py
```diff
@@ -0,0 +1,132 @@
...
+        description:
+            - The webfaction account to use
+        required: true
+
+    login_password:
+        description:
+            - The webfaction password to use
+        required: true
+'''
+
```
resmo, line 68: missing examples?

cloud/webfaction/webfaction_app.py
quentinsf: Thanks resmo - now added

cloud/webfaction/webfaction_domain.py
quentinsf, line 68: Added in latest commit.

cloud/webfaction/webfaction_app.py
quentinsf: Now changed to strings.

gregdek:
Did last checkins break something? It fails Travis build, but I don't have enough context to know why...

quentinsf:
I don't think it's an issue with the webfaction stuff - the limited build errors I can see are all elsewhere.

I obviously don't have a way to run the same tests here, but the modules do at least compile cleanly for me, and if I copy them into an ansible/library/cloud directory with appropriate renaming, the docs also build correctly.

bcoca:
so we are just missing the GPL header at this point, everything else looks OK

quentinsf:
Have added GPL headers, fixed a missing param in the documentation, and made one other change which makes database creation/deletion more symmetric/repeatable.

bcoca:
@quentinsf 
the login/password functionality and documentation can centralized, for examples look at 
https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/ec2.py
https://github.com/ansible/ansible/blob/devel/lib/ansible/utils/module_docs_fragments/aws.py

PR #447, opened 2015-05-04
Allow for specifying name instead of host as per the documentation

swindmill:
The documentation at http://docs.ansible.com/known_hosts_module.html indicates that this module needs only a "name" parameter but the code is looking for a "host" parameter and fails with a "KeyError: 'host'" if no "host" parameter is given.


bcoca:
host is an alias for name, but the code should look for name specifically.

PR #448, opened 2015-05-04
Add ec2_win_password module

ghost:
This is a simple module to fetch the randomly generated passwords from EC2 Windows instances. This scratches an itch in our shop because our playbooks deal with lots of Windows instances configured to use EC2Config Service random passwords. Maybe it will also help others.

### DIFF
--- a/cloud/amazon/ec2_win_password.py
+++ b/cloud/amazon/ec2_win_password.py
@@ -13,14 +13,10 @@ options:
     description:
       - The instance id to get the password data from. 
     required: true
-    default: null 
-    aliases: []
   key_file:
     description:
       - path to the file containing the key pair used on the instance
     required: true
-    default: null
-    aliases: []
   region:
     description:
       - The AWS region to use.  Must be specified if ec2_url is not used. If not specified then the value of the EC2_REGION environment variable, if any, is used.
@@ -42,7 +38,6 @@ tasks:
     key_file: "~/aws-creds/my_test_key.pem"
 '''
 
-import sys
 from base64 import b64decode
 from os.path import expanduser
 from Crypto.Cipher import PKCS1_v1_5
--- a/cloud/amazon/ec2_win_password.py
+++ b/cloud/amazon/ec2_win_password.py
@@ -1,0 +1,96 @@
+#!/usr/bin/python
+
+DOCUMENTATION = '''
+---
+module: ec2_win_password
+short_description: gets the default administrator password for ec2 windows instances
+description:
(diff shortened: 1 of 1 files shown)


petemounce:
Excellent - I just implemented this in the scripts that I have orchestrating my windows provisioning. One of the issues I ran into is that when ec2config is set to randomise the admin password, it will do that on each instance launch. A reboot triggered via the AWS API appears to count as that (but not a reboot triggered from the OS).

The EC2::GetPasswordData operation thankfully has a timestamp on it which is the last time that it was updated (presumably via ec2config itself), so I could poll and wait after the reboot for this timestamp to be later than the time of the reboot itself.

cchurch:
I'd implemented something similar in https://github.com/ansible/ansible-modules-core/pull/378, but I like your implementation better.

I've tested and it works for me.  :+1:

ghost:
@petemounce @cchurch Thanks for taking a look and testing it.

cloud/amazon/ec2_win_password.py
```diff
@@ -0,0 +1,96 @@
...
+description:
+    - Gets the default administrator password from any EC2 Windows instance.  The instance is referenced by its id (e.g. i-XXXXXXX). This module has a dependency on python-boto.
+version_added: "2.0"
+author: Rick Mendes
+options:
+  instance_id:
+    description:
+      - The instance id to get the password data from. 
+    required: true
+    default: null
```
resmo: if required, there is no default

cloud/amazon/ec2_win_password.py
```diff
@@ -0,0 +1,96 @@
...
+tasks:
+- name: get the Administrator password
+  ec2_win_password:
+    profile: my-boto-profile
+    instance_id: i-XXXXXX
+    region: us-east-1
+    key_file: "~/aws-creds/my_test_key.pem"
+'''
+
+import sys
```
resmo: is this needed?

cloud/amazon/ec2_win_password.py
```diff
@@ -0,0 +1,96 @@
...
+      - The instance id to get the password data from. 
+    required: true
+    default: null 
+    aliases: []
+  key_file:
+    description:
+      - path to the file containing the key pair used on the instance
+    required: true
+    default: null
+    aliases: []
```
resmo: no need for empty aliases

gregdek:
Looks like we have a +1 from an Ansi-bull -- @bcoca / @jimi-c / @abadger should that be sufficient for a merge?

PR #457, opened 2015-05-07
win_service

t-v:
a custom module to manage windows services.


windows/win_service.py
```diff
@@ -0,0 +1,63 @@
...
+    - Configures the service startup type
+options:
+  name:
+    description:
+      - the service name
+    required: true
+  state:
+    description:
+      - the service state
+    required: true (when enabled is undefined)
```
bcoca, line 39: required should be false and a default should be noted (same for enabled below), in the description of both you can explain that either one is needed for the module to work.

windows/win_service.py
```diff
@@ -0,0 +1,63 @@
...
+    choices:
+      - running
+      - stopped
+      - paused
+  enabled:
+    description:
+      - the service startuptype
+    required: true (when state is undefined)
+    choices:
+      - automatic (true)
```
bcoca, line 49: instead of a 'trinary boolean' make this a list with the 3 explicit choices

windows/win_service.ps1
```diff
@@ -0,0 +1,111 @@
...
+            };
+        } else {
+            $msg+=@("Service state was already set to $state");
+        };
+    };
+};
+
+
+if ($enabled -ne "") {
+    if (($enabled -eq $true) -or ($enabled -eq "true")) {
```
bcoca, line 67: this if/else is skippable if we go to specific values an not a trinary boolean

t-v, line 67: The main reason I did add the Boolean option, was to keep it consistent with the Linux service module.
But if it would be better to lose that, I'll do that.

bcoca, line 67: don't bother with making it like the unix one, it already has a win_ prefix and people should know this is 'different'.

cchurch:
There is already a win_service module, why do we need another one?

gregdek:
So @cchurch is right, we already have a module in Ansible Core called win_service:

https://github.com/ansible/ansible-modules-core/blob/devel/windows/win_service.ps1

@TimothyVandenbrande, what's the difference between this one and that one?

t-v:
yes, this one is obsolete.
I originally created it because it was missing in the 1st windows-supported devel, but since than the other one was already merged (and I never noticed).
I'll close this one already.

PR #458, opened 2015-05-07
New module for creating annotations with Circonus SaaS service

NickatEpic:
This module requires requests, but otherwise has no external dependencies. It creates annotations in the Circonus service using their v2 API.

### DIFF
--- a/monitoring/circonus_annotation.py
+++ b/monitoring/circonus_annotation.py
@@ -128,5 +128,5 @@ def main():
         module.fail_json(msg='Request Failed', reason=e)
     module.exit_json(changed=True, annotation=resp.json())
 
-from ansible.module_utils.basic import AnsibleModule
+from ansible.module_utils.basic import *
 main()
--- a/monitoring/circonus_annotation.py
+++ b/monitoring/circonus_annotation.py
@@ -1,5 +1,7 @@
 #!/usr/bin/env python
 
+# (c) 2014-2015, Epic Games, Inc.
+
 import requests
 import time
 import json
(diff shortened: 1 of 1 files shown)


monitoring/circonus_annotation.py
```diff
@@ -0,0 +1,131 @@
+#!/usr/bin/env python
```
bcoca: shebang should be /usr/bin/python

monitoring/circonus_annotation.py
```diff
@@ -0,0 +1,131 @@
...
+
+import requests
+import time
+import json
+
+DOCUMENTATION = '''
+---
+module: circonus_annotation
+short_description: create an annotation in circonus
+description:
```
bcoca, line 13: version_added: "2.0"

monitoring/circonus_annotation.py
```diff
@@ -0,0 +1,131 @@
...
+    description: 'This is a detailed description of the config change'
+    category: 'This category groups like annotations'
+    start_time: 1395940006
+    end_time: 1395954407
+'''
+def post_annotation(annotation, api_key):
+    ''' Takes annotation dict and api_key string'''
+    base_url = 'https://api.circonus.com/v2'
+    anootate_post_endpoint = '/annotation'
+    resp = requests.post(base_url + anootate_post_endpoint,
```
bcoca: no exception handling?

monitoring/circonus_annotation.py
```diff
@@ -0,0 +1,131 @@
...
+            stop=dict(required=False, type='int'),
+            category=dict(required=True),
+            title=dict(required=True),
+            description=dict(required=True),
+            duration=dict(required=False, type='int'),
+            api_key=dict(required=True)
+            )
+        )
+    annotation = create_annotation(module)
+    try:
```
bcoca, line 125: normally we shy away from 'catchall' exception handling, but in this case the module is simple enough that this might sufice.

NickatEpic, line 125: I can make it catch more specific exceptions, but the handling for all of them is the same as the API doesn't provide any errors that make sense to try and intelligently respond to. Bubbling up the error payload and letting the playbook optionally retry seems the only reasonable choice. Would it be better if I reformat how I pass the error into the reason keyword?

bcoca:
That is normally our point, capture the exception at points where you can return a meaningful messages  + the exception. Since I only see one possible point of exception here, I'm not going to press the issue here.

monitoring/circonus_annotation.py
NickatEpic, line 13: Fixed in 910761b.

monitoring/circonus_annotation.py
NickatEpic: Fixed in 910761b

PR #461, opened 2015-05-08
messaging: rabbitmq_binding, rabbitmq_exchange, rabbitmq_queue

manuel-sousa:
These modules allow for creation/deletion of exchanges, queues and bindings.

They use rabbitMQ REST API.


bcoca:
could you please open a PR per module? makes it easier for us to revise and merge 1 by 1

messaging/rabbitmq_binding.py
```diff
@@ -0,0 +1,188 @@
+#!/usr/bin/python
+
+DOCUMENTATION = '''
+module: rabbitmq_binding
+author: Manuel Sousa
+version_added: 1.5.4
```
resmo, line 6: should be '2.0'

messaging/rabbitmq_binding.py
```diff
@@ -0,0 +1,188 @@
...
+                        "arguments": module.params['arguments']
+                    })
+                )
+        elif module.params['state'] == 'absent':
+            r = requests.delete( url, auth = (module.params['login_user'],module.params['login_password']))
+
+        if r.status_code == 204 or r.status_code == 201:
+            module.exit_json(
+                changed = True,
+                result = "Success",
```
resmo, line 168: exit_json is always success afaik.

resmo:
good job. quite useful for me. Will test them.

manuel-sousa:
I was about to do the 3 seperate pull rquests, is it still required?

manuel-sousa:
As for the version, sorry i'm still using 1.5.4 that's why I used that as reference cause i knew it worked there;)

resmo:
I do not work for ansible inc but @bcoca does. That's a yes. ☺

manuel-sousa:
Ok, i'm making the change to the version and removing the success as commented and then submit the three of them separately. Thanks

resmo:
@manuel-sousa @bcoca can be closed, modules merged in separete PRs.

PR #462, opened 2015-05-08
Rabbitmq queue

manuel-sousa:
This module allows creation/deletion of rabbitmq queues. It uses rabbitMQ REST API.

### DIFF
--- a/messaging/rabbitmq_queue.py
+++ b/messaging/rabbitmq_queue.py
@@ -1,0 +1,235 @@
+#!/usr/bin/python
+
+DOCUMENTATION = '''
+module: rabbitmq_queue
+author: Manuel Sousa
+version_added: 1.5.4
+
+short_description: This module manages rabbitMQ queues
+description:
+  - This module uses rabbitMQ Rest API to create/delete queues
+requirements: [ python requests ]
+options:
+    name:
+        description:
+            - Name of the queue to create
+        required: true
+    state:
+        description:
+            - Whether the queue should be present or absent
+            - Only present implemented atm
+        choices: [ "present", "absent" ]
+        required: false
+        default: present
+    login_user:
+        description:
+            - rabbitMQ user for connection
+        required: false
+        default: guest
+    login_password:
+        description:
+            - rabbitMQ password for connection
+        required: false
+        default: false
+    login_host:
+        description:
+            - rabbitMQ host for connection
+        required: false
+        default: localhost
+    login_port:
+        description:
+            - rabbitMQ management api port
+        required: false
+        default: 15672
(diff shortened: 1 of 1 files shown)


messaging/rabbitmq_queue.py
```diff
@@ -0,0 +1,257 @@
...
+        argument_spec = dict(
+            state = dict(default='present', choices=['present', 'absent'], type='str'),
+            name = dict(required=True, type='str'),
+            login_user = dict(default='guest', type='str'),
+            login_password = dict(default='guest', type='str', no_log=True),
+            login_host = dict(default='localhost', type='str'),
+            login_port = dict(default='15672', type='str'),
+            vhost = dict(default='/', type='str'),
+            durable = dict(default=True, choices=BOOLEANS, type='bool'),
+            autoDelete = dict(default=False, choices=BOOLEANS, type='bool'),
```
resmo: CamelCase is not really a problem, but it is not very common in the python world. Not sure if there are any conventions of ansible inc (@bcoca?) about it. Since you already have snake_case for `login_user` you should not mix it, would you mind changing all your CamelCase arguments to snake_case style?

manuel-sousa: I actually noticed the same thing when I copied the login_\* syntax from the mysql modules. I've done this change later when i figured this might actually be useful to share to other people . Since there were already other modules using this syntax i wanted to preserve it as it was instead of changing to camel case.

I've done a quick search now on some modules and it seems they all follow that syntax instead of camel case (which i personally prefer). Consistency in behavior makes all sense and as such I'll be making the changes later and submit.

bcoca: We really don't care either way, but I would suggest that you keep at least your code consistent withing the same module so the next person to try to modify the code does not get confused.

resmo:
+1

messaging/rabbitmq_queue.py
```diff
@@ -0,0 +1,257 @@
...
+        choices: [ "yes", "no" ]
+        default: no
+    message_ttl:
+        description:
+            - How long a message can live in queue before it is discarded (milliseconds)
+        required: False
+    auto_expires:
+        description:
+            - How long a queue can be unused before it is automatically deleted (milliseconds)
+        required: false
```
bcoca, line 88: non required arguments should state their default

manuel-sousa, line 88: I'll add the defaults, one question. When the default is an empty dictionary is it ok to say default: {} or is it better to say None ?

bcoca, line 88: yes

PR #463, opened 2015-05-08
Rabbitmq exchange

manuel-sousa:
This module allows creation/deletion of rabbitmq exchanges. It uses rabbitMQ REST API.


resmo:
+1

PR #464, opened 2015-05-08
Rabbitmq binding

manuel-sousa:
This module allows creation/deletion of rabbitmq bindings. It uses rabbitMQ REST API.

### DIFF
--- a/messaging/rabbitmq_binding.py
+++ b/messaging/rabbitmq_binding.py
@@ -73,13 +73,13 @@ options:
             - destination exchange or queue for the binding
         required: true
         aliases: [ "dst", "dest" ]
-    destinationType:
+    destination_type:
         description:
             - Either queue or exchange
         required: true
         choices: [ "queue", "exchange" ]
-        aliases: [ "type", "destType" ]
-    routingKey:
+        aliases: [ "type", "dest_type" ]
+    routing_key:
         description:
             - routing key for the binding
             - default is #
@@ -93,10 +93,10 @@ options:
 
 EXAMPLES = '''
 # Bind myQueue to directExchange with routing key info
-- rabbitmq_binding: name=directExchange destination=myQueue type=queue routingKey=info
+- rabbitmq_binding: name=directExchange destination=myQueue type=queue routing_key=info
 
 # Bind directExchange to topicExchange with routing key *.info
-- rabbitmq_binding: name=topicExchange destination=topicExchange type=exchange routingKey="*.info"
+- rabbitmq_binding: name=topicExchange destination=topicExchange type=exchange routing_key="*.info"
 '''
 
 import requests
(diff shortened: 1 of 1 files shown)


messaging/rabbitmq_binding.py
```diff
@@ -0,0 +1,185 @@
+#!/usr/bin/python
+
+DOCUMENTATION = '''
```
resmo, line 22: MINOR: add a `---` below this line.

messaging/rabbitmq_binding.py
```diff
@@ -0,0 +1,185 @@
...
+module: rabbitmq_binding
+author: Manuel Sousa
+version_added: 2.0
```
resmo: `version_added` value must be quoted because it is not a float number, so add change it to `"2.0"`

messaging/rabbitmq_binding.py
```diff
@@ -0,0 +1,185 @@
+#!/usr/bin/python
+
```
resmo, line 21: Miissing GPL header.

messaging/rabbitmq_binding.py
```diff
@@ -0,0 +1,185 @@
...
+import urllib
+import json
+
+def main():
+    module = AnsibleModule(
+        argument_spec = dict(
+            state = dict(default='present', choices=['present', 'absent'], type='str'),
+            name = dict(required=True, aliases=[ "src", "source" ], type='str'),
+            login_user = dict(default='guest', type='str'),
+            login_password = dict(default='guest', type='str'),
```
resmo: logging sensible data can be turned off by adding a `no_log=True`. I would generally recommend it for passwords. 

```
login_password = dict(default='guest', type='str', no_log=True),
```

resmo:
a few minor things as mentioned, but generally looks good. I just wanted to let you know that there is some helper code around for dealing with http and urls in here https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/urls.py. afaik it helps for global configuration using http, e.g. http proxies, etc, so you don't have to revinvent the wheel. It might help or not, just for your info, you can use it later if you feel it would help.

manuel-sousa:
@resmo thanks for the review and tips. I'll make the changes now and check out the URLS library later.

messaging/rabbitmq_binding.py
manuel-sousa, line 21: Should I had an header like the other modules that state it's part of ansible-modules or a generic one like MIT?

resmo, line 21: I didn't see any module using MIT. I used the GPL but I guess MIT would also be fine.

bcoca:
ansible and it's modules are GPL, the shared module code in ansible/module_utils/\* is BSD licensed to allow module authors to use it and distribute their own modules w/o restriction, but that is for those independent from ansible itself.

resmo:
@bcoca thx for clearification.

manuel-sousa:
I noticed the failed test, i'll check it out tomorrow on python 2.4.

resmo:
+1

PR #466, opened 2015-05-08
Fix up modules that have python24 syntax error

sivel:
There was quite a lot of python24 incompatible syntax in the extras repository.

This PR changes all of the python24 incompatible syntax over to now be compatible.

The travis compileall task completes without issue now.


resmo:
+1

bcoca:
certain modules were not 2.4 compatible cause they depend on libs that are not compatible, we really don't need to fix that syntax.

sivel:
I don't know that we have a great way to easily know if a module requires a newer version of Python.

We can add more files to the exclude list. There aren't any problems there.

However none of these changes reduce functionality functionality of the modules.

This will give us a pretty good baseline to add exclusions to as things progress.

For a later time: We may also want to standardize on a DOCUMENTATION value that indicates minimum required Python versions.  I think I'll see if I can make some proposal for that potentially.

bcoca:
we have the existing requires: field, in which we should add the python version, just need to start doing the bookkeeping right. I'm going to close this request in favor of documenting each module, we normally don't do sweeping updates like this .

PR #469, opened 2015-05-09
Interpolate container name into error message

sigmavirus24:
##### Ansible Version:

```
# ansible --version
ansible 1.9.0.1
```
##### Ansible Configuration:

None
##### Environment:

Ubuntu 14.04 / Ubuntu 14.04
##### Summary:

If you have a task that creates and starts a container and you run it with -vv, if the container fails to start then you'll see the error message without the container name interpolated.
##### Steps To Reproduce:

Create a task which will cause the created lxc container to fail to start.
##### Expected Results:

```
msg: The container [ infra1_glance_api ] failed to start. Check to lxc is available and that the container is in a functional state.
```
##### Actual Results:

```
msg: The container [ %s ] failed to start. Check to lxc is available and that the container is in a functional state.
```

### DIFF
--- a/cloud/lxc/lxc_container.py
+++ b/cloud/lxc/lxc_container.py
@@ -794,7 +794,7 @@ class LxcContainerManagement(object):
                 rc=1,
                 msg='The container [ %s ] failed to start. Check to lxc is'
                     ' available and that the container is in a functional'
-                    ' state.'
+                    ' state.' % self.container_name
             )
 
     def _check_archive(self):


cloudnull:
:+1:  - this should be fixed.

sigmavirus24:
FWIW, I looked at the test failures on [Travis](https://travis-ci.org/ansible/ansible-modules-extras/builds/61846629) and none of them are due to this patch.

gregdek:
Since cloudnull gives a thumbs-up here, I'm tagging this with the new owner_pr label.

PR #332, opened 2015-03-23
system/lvol: Suppress prompts from lvcreate

AdmiralNemo:
Occasionally, `lvcreate` will prompt on stdin for confirmation. In
particular, this may happen when the volume is being created close to
the location on disk where another volume existed previously. When this
happens, Ansible will hang indefinitely with no indication of the
problem. To work prevent this problem, the `--yes` command-line argument
can be passed to `lvcreate`, which will instruct it not to prompt.

Signed-off-by: Dustin C. Hatch dustin@hatch.name

PR #470, opened 2015-05-09
Suppress prompts from lvcreate using --yes when LVM supports this option

fdupoux:
Occasionally, "lvcreate" will prompt on stdin for confirmation. In particular, this may happen when the volume is being created close to the location on disk where another volume existed previously. When this happens, Ansible will hang indefinitely with no indication of the problem. To work prevent this problem, the "--yes" command-line argument can be passed to "lvcreate", which will instruct it not to prompt.

A simple patch which adds "--yes" to the lvcreate command was already proposed and merged before (PR 332):
https://github.com/ansible/ansible-modules-extras/pull/332
The problem with this fix is it introduced a regression as old LVM versions (before 2.02.99) did not support the "--yes" option. Hence lvcreate just rejected the option and failed with old LVM versions. This patch has then been reverted.

With this new fix the lvol module will first get the LVM version number and it will only add "--yes" to the lvcreate command if the LVM version is 2.02.99 or more recent. This fix has been tested on CentOS 7.1 (which comes with LVM 2.02.115) and Ubuntu-12.04.5 (which comes with LVM 2.02.66) in order to test both situations.

### DIFF
--- a/system/lvol.py
+++ b/system/lvol.py
@@ -125,7 +125,10 @@ def main():
     if version_found == None:
         module.fail_json(msg="Failed to get LVM version number")
     version_yesopt = mkversion(2, 2, 99) # First LVM with the "--yes" option
-    yesopt = "--yes" if version_found >= version_yesopt else ""
+    if version_found >= version_yesopt:
+        yesopt = "--yes"
+    else:
+        yesopt = ""
 
     vg = module.params['vg']
     lv = module.params['lv']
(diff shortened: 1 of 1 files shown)


gregdek:
@jhoekx Please review this PR; if you give +1, we will merge

jhoekx:
Looks good! I checked on RHEL 5 also.

+1

gregdek:
Merged. Thanks @fdupoux and thanks @rosmo for the review!

PR #474, opened 2015-05-11
Changing from v1 to v2 hipchat api format.

pb8226:
I wanted to use v2 of the hip chat api.  I made the necessary changes to use v2 and replaced http with requests for posting JSON to simplify the code.  I have tested this against my hip chat account and all the parameters were working.


sivel:
Generally speaking all URL interaction should be handled by the functionality provided by `ansible.module_utils.urls` such as using `fetch_url`.  Adding a dependency on `requests` makes this module diverge from standards and also adds a new dependency that will break backwards compatibility.

pb8226:
thanks @sivel , I'll change the code to remove the requests dependency

bcoca:
i'm not currently using hipchat, what happens to those still using v1 of the api? some of these changes seem to indicate users will have to change their plays.

pb8226:
@bcoca I was actually curious how Ansible likes to handle multiple versions of an external API.  Anyone using HipChat via Atlassian will have the v2 of API available to them.  Is it preferred that I make a separate module for the v2 of the HipChat API and then users can choose whichever one they want to use?

pb8226:
@bcoca Another option would be to add a property to the module to state whether to use the v1 or v2 of the hipchat api.

bcoca:
we normally try to keep backwards compatibility to avoid breaking people's playbook, adding an api_version='v2' (defaults to v1) is something that would work.

bcoca:
or if module can figure out which one to use with the currently supplied data, that is also acceptable.

pb8226:
Makes sense, I can probably figure out the version of the api from the "/v2/" in the URL.  So even if the user supplies his own URL I can do a regex on it to see if it's a v1 or v2 format.

pb8226:
Also it seems like it's preferable to not add any new dependencies.  The reason I'm asking is there is a Python HipChat client, but I have no information on how much benefit it would provide.

pb8226:
Changes have been made to use only Ansible libraries for REST calls.  Module changes add support for the V2 of the HipChat API while also being compatible with the V1 HipChat API.  The module defaults to the V1 HipChat API.  If a user explicitly sets the HipChat server URL the module will check if it's a v1 or v2 HipChat URL and use the appropriate format.

hostmaster:
api v1 is market as 'deprecated'. It probably would be better to switch to v2 by default.

pb8226:
@hostmaster The concern with switching to v2 by default was that existing play books might break.

hostmaster:
@pb8226 it will break for sure. v2 has a diferent auth_token.

pb8226:
@hostmaster yep, that's why I added v2 support so I could use the new tokens.  I was discouraged, further up this thread, to make it the default though.

hostmaster:
@pb8226 please have a look at https://github.com/ansible/ansible-modules-extras/pull/674 when you have time

PR #476, opened 2015-05-11
Fix up modules that have python24 syntax errors

sivel:
This PR makes numerous modules python2.4 compatible.

In the case where a module has an external dependency that relies on a newer version of python, excludes have been added.

In addition, there were a few modules without `.py` file extensions

If additional changes are needed, please let me know and I will revise this PR.


system/crypttab.py
```diff
@@ -155,8 +155,9 @@ def main():
 
 
     if changed and not module.check_mode:
-        with open(path, 'wb') as f:
-            f.write(str(crypttab))
+        f = open(path, 'wb')
+        f.write(str(crypttab))
+        f.close()
```
abadger: with statement should be fixed with a try: finally: something like this:

```
try:
    f = open(path, 'wb')
    f.write(str(crypttab))
finally:
    f.close()
```

abadger:
Thanks sivel!

# Merged

Hi!

issue is resolved.

Thank you!

PR #480, opened 2015-05-11
pacman: change -Syy to -Sy for "update-cache"

RubenKelevra:
-Syy do always update, which is not needed, since the caches might already be up to date.

### DIFF
--- a/packaging/os/pacman.py
+++ b/packaging/os/pacman.py
@@ -76,7 +76,7 @@ EXAMPLES = '''
 # Recursively remove package baz
 - pacman: name=baz state=absent recurse=yes
 
-# Run the equivalent of "pacman -Syy" as a separate step
+# Run the equivalent of "pacman -Sy" as a separate step
 - pacman: update_cache=yes
 '''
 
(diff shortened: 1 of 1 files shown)


bcoca:
this leaves us with no way to force the sync, which is rarely needed but it does happen

RubenKelevra:
I thought about using the force=yes on update_cache=yes as well.

bcoca:
that overloads force a bit too much, i might want to force install or cache update but not both?

RubenKelevra:
Sure ... but doing "-Syy" everytime is insane, especially when you want to install more than one package.

RubenKelevra:
Sure ... but doing "-Syy" everytime is insane, especially when you want to install more than one package.

The name "cache update" sounds completly wrong to me. The man-page of pacman names this option "refresh":

SYNC OPTIONS
-y, --refresh
    Download a fresh copy of the master package list from the server(s) defined in pacman.conf(5). This should typically be used each time you use --sysupgrade or -u. Passing two --refresh or -y flags will force a refresh of all package lists, even if they appear to be up- to-date.

I know other package-mangers syntaxes are different, but since there are also possible options with the package cache this option sounds completly wrong to me at the moment.

A cache-update would be equivalent to "pacman -Syuw --noconfirm" which refreshes and update all packages to the cache.

So the option should be named as "refresh" and "force-refresh" or "complete-refresh".

PR #481, opened 2015-05-12
2 bug fixes for monit module

LorenzoLuconi:
Two bug fixes for monit module
- status returned by monit has capitol letter but status are compared to lowercase strings, so I added a lower function before return status string in status function. Without this fix monit module cannot really detect the process status returned by monit.
- after an unmonitor command monit status could be "not monitored - unmonitor pending"

There is also a little optimization for status function called twice

### DIFF
--- a/monitoring/monit.py
+++ b/monitoring/monit.py
@@ -77,7 +77,7 @@ def main():
             # Process 'name'    Running - restart pending
             parts = line.split()
             if len(parts) > 2 and parts[0].lower() == 'process' and parts[1] == "'%s'" % name:
-                return ' '.join(parts[2:])
+                return ' '.join(parts[2:]).lower()
         else:
             return ''
 
--- a/monitoring/monit.py
+++ b/monitoring/monit.py
@@ -119,7 +119,7 @@ def main():
         if module.check_mode:
             module.exit_json(changed=True)
         status = run_command('unmonitor')
-        if status in ['not monitored']:
+        if status in ['not monitored'] or 'unmonitor pending' in status:
             module.exit_json(changed=True, name=name, state=state)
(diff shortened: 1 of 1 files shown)


monitoring/monit.py
```diff
@@ -119,7 +120,7 @@ def run_command(command):
         if module.check_mode:
             module.exit_json(changed=True)
         status = run_command('unmonitor')
-        if status in ['not monitored']:
+        if status in ['not monitored'] or 'unmonitor pending' in status:
```
abadger, line 123: What's the string that you're seeing in status?  It seems like this test should be something more like one of these:

```
# We're able to give the exact string for everything:
if status in ('not monitored', 'unmonitor pernding'):
# We're not able to give the exact string for anything:
if [s for s in ('not monitored', 'unmonitor pending') if s in status]:
# We're able to give the exact string for one but not the other:
if status == 'not monitored' or 'unmonitor pending' in status:
```

A comment with examples of both strings that we're looking for might clarify the conditional for future people working on this.

```
```

abadger, line 123: Ah, saw the example in your origina comment... so maybe a test like this:

``` python
if status.startswith('not monitored'):
```

?

abadger, line 123: And a comment about the two strings it could be.

LorenzoLuconi, line 123: I think "unmonitor pending" can be combined with other status not only "not monitored", so "startWith" could work but I don't know if it works always.
I choosed this kind of approach because is the same used in the other commands in monit module: a list of statuses or the action is pending.

gregdek:
@dstoflet Please review; if you give +1, we will merge

dstoflet:
+1

PR #482, opened 2015-05-12
Initial commit of nmcli: NetworkManager module. Currently supports:

alcamie101:
Create, modify, remove of - team, team-slave, bond, bond-slave, ethernet
TODO:
vlan, bridge, wireless related connections.
more docs here: https://github.com/alcamie101/ansible-nmcli
Why did I write an Ansible module for NetworkManager?
The main reason I started writing this module was because I was building or should I say rebuilding an OpenStack undercloud the N'th time and I was sick of having to write Jinja templates for configuring the interfaces and then write all of the supporting playbook guff to manage deployment and consequent network management shell cmds for one type of interface and then rewrite it again for another type of interface.

This makes things simple, quick and extremely flexible. E.g. You can build multiple bond interfaces with assorted NIC configurations and deploy them within a couple of minutes. If you're not happy with bonding, you can, in a few seconds, change the configuration to team and team-slaves and redeploy that in a couple of minutes just by changing the 'type'. At the moment, the playbook process here would be:

Create bond
create bond-slave
deploy 'playbook-add.yml'
decide you want teams not bonds
Remove above connection 'playbook-del.yml'
in 'playbook-add.yml' change bond to team and bond-slave to team-slave
deploy 'playbook-add.yml' and voila, you now have teams rather than bonds.

### DIFF
--- a/network/nmcli.py
+++ b/network/nmcli.py
@@ -31,25 +31,24 @@ options:
     state:
         required: True
         choices: [ present, absent ]
-    description:
-        - Whether the device should exist or not, taking action if the state is different from what is stated.
-    enabled:
+        description:
+            - Whether the device should exist or not, taking action if the state is different from what is stated.
+    autoconnect:
         required: False
         default: "yes"
         choices: [ "yes", "no" ]
         description:
-            - Whether the service should start on boot. B(At least one of state and enabled are required.)
+            - Whether the connection should start on boot.
             - Whether the connection profile can be automatically activated ( default: yes)
     conn_name:
         required: True
-        default: None
         description:
             - Where conn_name will be the name used to call the connection. when not provided a default name is generated: <type>[-<ifname>][-<num>]
     ifname:
         required: False
         default: conn_name
         description:
-            - Where INAME will be the what we call the interface name. Required with 'up', 'down' modifiers.
+            - Where IFNAME will be the what we call the interface name.
             - interface to bind the connection to. The connection will only be applicable to this interface name.
             - A special value of "*" can be used for interface-independent connections.
             - The ifname argument is mandatory for all connection types except bond, team, bridge and vlan.
@@ -72,14 +71,17 @@ options:
     ip4:
         required: False
         default: None
-        description: The IPv4 address to this interface using this format ie: "192.168.1.24/24"
+        description:
+            - The IPv4 address to this interface using this format ie: "192.168.1.24/24"
     gw4:
         required: False
-        description: The IPv4 gateway for this interface using this format ie: "192.168.100.1"
+        description:
+            - The IPv4 gateway for this interface using this format ie: "192.168.100.1"
     dns4:
         required: False
         default: None
-        description: A list of upto 3 dns servers, ipv4 format e.g. To add two IPv4 DNS server addresses: ['"8.8.8.8 8.8.4.4"']
+        description:
+            - A list of upto 3 dns servers, ipv4 format e.g. To add two IPv4 DNS server addresses: ['"8.8.8.8 8.8.4.4"']
     ip6:
         required: False
         default: None
@@ -88,10 +90,12 @@ options:
     gw6:
         required: False
         default: None
-        description: The IPv6 gateway for this interface using this format ie: "2001:db8::1"
+        description:
+            - The IPv6 gateway for this interface using this format ie: "2001:db8::1"
     dns6:
         required: False
-        description: A list of upto 3 dns servers, ipv6 format e.g. To add two IPv6 DNS server addresses: ['"2001:4860:4860::8888 2001:4860:4860::8844"']
+        description:
+            - A list of upto 3 dns servers, ipv6 format e.g. To add two IPv6 DNS server addresses: ['"2001:4860:4860::8888 2001:4860:4860::8844"']
     mtu:
         required: False
         default: None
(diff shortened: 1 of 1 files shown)


network/nmcli.py
```diff
@@ -0,0 +1,1089 @@
...
+module: nmcli
+author: Chris Long
+short_description: Manage Networking
+requirements: [ nmcli, dbus ]
+description:
+    - Manage the network devices. Create, modify, and manage, ethernet, teams, bonds, vlans etc.
+options:
+    state:
+        required: True
+        default: "present"
```
bcoca: when required is true, there is no default

network/nmcli.py
```diff
@@ -0,0 +1,1089 @@
...
+    description:
+        - Whether the device should exist or not, taking action if the state is different from what is stated.
+    enabled:
+        required: False
+        default: "yes"
+        choices: [ "yes", "no" ]
+        description:
+            - Whether the service should start on boot. B(At least one of state and enabled are required.)
+            - Whether the connection profile can be automatically activated ( default: yes)
+    action:
```
bcoca: we shy away from action/command type of interfaces as they are imperative and not declarative

add/modify seem to overlap with state present
delte with state absent

up/down with enabled yes/no or it's own flag up: yes/no

show looks like an attempt to make this a facts plugin

if you shift enabled to 'autoconnect true/false' then we can use it as i described above

network/nmcli.py
```diff
@@ -0,0 +1,1089 @@
...
+                 70: "IP Config",
+                 80: "IP Check",
+                 90: "Secondaries",
+                 100: "Activated",
+                 110: "Deactivating",
+                 120: "Failed"
+            }
+
+    def __new__(cls, *args, **kwargs):
+        return load_platform_subclass(Nmcli, args, kwargs)
```
bcoca: where is load_platform_subclass defined?

network/nmcli.py
```diff
@@ -0,0 +1,1089 @@
...
+            flags=dict(required=False, default=None, type='str'),
+            ingress=dict(required=False, default=None, type='str'),
+            egress=dict(required=False, default=None, type='str'),
+        ),
+        supports_check_mode=True
+    )
+
+    nmcli=Nmcli(module)
+
+    if nmcli.syslogging:
```
bcoca: ansible already outputs to syslog, no need to do yourself

network/nmcli.py
```diff
@@ -0,0 +1,1089 @@
...
+        default: None
+        choices: [ add, modify, show, up, down ]
+        description:
+            - Set to 'add' if you want to add a connection.
+            - Set to 'modify' if you want to modify a connection. Modify one or more properties in the connection profile.
+            - Set to 'delete' if you want to delete a connection. Delete a configured connection. The connection to be deleted is identified by its name 'cfname'.
+            - Set to 'show' if you want to show a connection. Will show all devices unless 'cfname' is set.
+            - Set to 'up' if you want to bring a connection up. Requires 'cfname' to be set.
+            - Set to 'down' if you want to bring a connection down. Requires 'cfname' to be set.
+    cname:
```
bcoca: description contradicts required/default

cname might be confusing with dns aliases, conn_name or something more explicit might be better

alcamie101: cname is the name used in NetworkManagers own docs. I can change to conn_name if that's less confusing.

network/nmcli.py
alcamie101: will fix that. Thanks

network/nmcli.py
alcamie101: Ahh ok. I pulled this from the user module. Thanks for the insight.

network/nmcli.py
```diff
@@ -0,0 +1,1089 @@
...
+            - This is the type of device or network connection that you wish to create for a bond, team or bridge. (NetworkManager default: balance-rr)
+    master:
+        required: False
+        default: None
+        description:
+            - master <master (ifname, or connection UUID or cname) of bridge, team, bond master connection profile.
+    ip4:
+        required: False
+        default: None
+        description: The IPv4 address to this interface using this format ie: "192.168.1.24/24"
```
bcoca: descriptions should always be lists ( - prefix)

network/nmcli.py
alcamie101: Old reference, thanks for pointing it out.

network/nmcli.py
bcoca: you can always keep as an alias, but a less confusing default name is a better interface imho

network/nmcli.py
```diff
@@ -0,0 +1,1089 @@
...
+description:
+    - Manage the network devices. Create, modify, and manage, ethernet, teams, bonds, vlans etc.
+options:
+    state:
+        required: True
+        default: "present"
+        choices: [ present, absent ]
+    description:
+        - Whether the device should exist or not, taking action if the state is different from what is stated.
+    enabled:
```
bcoca: saying that the service is enabled or not is confusing and sounds like it belongs in the service module, after looking at code this sets the connection to autoconnect or not.

network/nmcli.py
alcamie101: have removed as I had added functions to defunct the need for actions

network/nmcli.py
alcamie101: correcting.

alcamie101:
bcoca. I've corrected all of the action points and tested.

alcamie101:
+1

PR #486, opened 2015-05-13
Adding module to manage Ruby Gem dependencies via Bundler

thoiberg:
Fairly simple module that wraps around the Bundler gem to make managing Ruby dependencies during deployment easier.


packaging/language/bundler.py
```diff
@@ -0,0 +1,123 @@
...
+module: bundler
+short_description: Manage Ruby Gem dependencies with Bundler
+description:
+  - Manage installation and Gem version dependencies for Ruby using the Bundler gem
+version_added: "2.0.0"
+options:
+  executable:
+    description:
+      - The path to the bundler executable
+    required: false
```
bcoca, line 33: default? required: false fields should state a default (even if its just null)

packaging/language/bundler.py
```diff
@@ -0,0 +1,123 @@
...
+  state:
+    description:
+      - The desired state of the Gem bundle. C(latest) updates gems to the most recent, acceptable version
+    required: false
+    choices: [present, latest]
+    default: present
+  chdir:
+    description:
+      - The directory to execute the bundler commands from. This directoy needs to contain a valid Gemfile or .bundle/ directory
+    required: false
```
bcoca, line 44: deafult: temporary working directory

packaging/language/bundler.py
```diff
@@ -0,0 +1,123 @@
...
+  exclude_groups:
+    description:
+      - A list of Gemfile groups to exclude during operations. This only applies when state is C(present). Bundler considers this a 'remembered' 
+      property for the Gemfile and will automatically exclude groups in future operations even if C(exclude_groups) is not set
+    required: False
+  extra_args:
```
bcoca, line 52: Keep this, but we prefer to have the actual options as parameters in the module that translate into the args to the tool

thoiberg:
@bcoca Bundler is annoyingly inconsistent in the arguments it allows on its commands so I added extra_args to avoid filling the module with arguments that may or may not have an effect base on the `state` variable. Not sure what the Ansible stance on situations like this is, but I'm happy to convert as many Bundle arguments to module arguments and just leave `extra_args` for future proofing if that's preferable.

bcoca:
I have mixed feelings on this, i understand how annoying bundler can be, but would also like to make ansible users have a better experience. I'm leaving it up to you as it is quite a bit of work.

thoiberg:
I agree with providing a better user experience, and by adding in additional modules it will at least provide more documentation in a single location. I'll close the PR and add in some of the (in my experience) more commonly used arguments to the module and leave the `extra_args` for any additional args, and then open a new PR in a couple of days. 

Thanks for your time.

PR #494, opened 2015-05-14
Set windows environment variables

jhawkesworth:
A module for setting environment variables on windows hosts.


gregdek:
Adding "owner_pr" flag because @jhawkesworth already owns modules in Extras.

PR #504, opened 2015-05-18
Add ec2_healthcheck module

eyalroth:
This module waits for an EC2 instance to become reachable.

### DIFF
--- a/cloud/amazon/ec2_healthcheck.py
+++ b/cloud/amazon/ec2_healthcheck.py
@@ -1,0 +1,174 @@
+#!/usr/bin/python
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+module: ec2_healthcheck
+short_description: Waits for an ec2 instance to become reachable.
+description:
+     - Performs a health check to an ec2 instance by checking its' status
+       several times in a given interval, until the instance becomes reachable.
+       The module assumes the instance already exists.
+requirements:
+    - "boto"
+options:
+  instance_id:
+    description:
+      - The ID of the ec2 instance.
+    required: true
+  region:
+    description:
+      - The ec2 region.
+    required: true
+  interval:
+    description:
+      - The time interval (in seconds) between each health check.
+    required: true
+  repeat:
+    description:
+      - The number of times to perform an aditional health check after the
+        previous one (0 = only one check will be performed).
+    required: false
+    default: 1
+  fail_unreachable:
+    description:
+      - Determines whether to fail or not when the instance is unreachable.
+    required: false
+    default: true
+  aws_access_key:
+    description:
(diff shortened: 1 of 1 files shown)


wimnat:
What's the use case for this module?

If you started an ec2 instance with the ec2 module you can just specify wait=yes to wait until it's ready.

Or you could just use the wait_for option to wait for SSH or RDP to be available.

er-net:
Me and my team used both of these options (ec2 wait + wait_for ssh), but they proved not to be enough; meaning, the rest of our Ansible playbook had issues relating to the instance's status even after both of these steps have passed successfully.

eyalroth:
Me and my team used both of these options (ec2 wait + wait_for ssh), but they proved not to be enough; meaning, the rest of our Ansible playbook had issues relating to the instance's status even after both of these steps have passed successfully.

gregdek:
@eyalroth do you have a sample playbook where this works and ec2 module with wait / wait_for ssh options don't work?

eyalroth:
@gregdek Here's a [role](http://pastebin.com/CrcThkJN) from of our playback responsible for creating ec2 instances.

gregdek:
@eyalroth sorry if I'm confused -- but it looks like you're just using the wait parameter built into the ec2 module? I don't see the ec2_healthcheck referenced in this role anywhere.

er-net:
@gregdek Oh, I thought you wanted to see the playbook without the ec2_healthcheck module. Here's the full version of the [role](http://pastebin.com/Rh6aHR9r).

gregdek:
Deferring to the AWS reviewers here. Still not sure if this isn't best accomplished than just using waits in the other modules, but if they say it is, then we'll include it.

cc: @jsmartin @scicoin-project @tombamford @garethr @lorin @jarv @jsdalton @silviud @adq @zbal @zeekin @willthames @lwade @carsongee @defionscode
@tastychutney @bpennypacker

scottanderson42:
I'm not sure if this is necessary or not. While we definitely have this problem as well, we were able to work around the issue like this:

``` yaml
- name: Wait for SSH on creation host
  local_action:
    module: wait_for
    host: "{{ wait_ssh_host }}"
    port: 22
    # The check fails if the port is open but the ssh daemon isn't yet ready to accept
    # actual traffic, right after the instance is started.
    #delay: 10
    timeout: 640
    state: started

- name: Really wait for SSH on host
  local_action: command ssh -o StrictHostKeyChecking=no -i {{ environment_public_ssh_private_key_file }} {{ aws_instance_user }}@{{ wait_ssh_host }} echo Rhubarb
  register: result
  until: result.rc  == 0
  retries: 60
  delay: 10
```

I suspect I might be more interested in a generic module that reported the current status of an instance. This could be used with `until` to do whatever checking the user wished. The current module seems a bit specific to a particular use-case.

jarv:
I have code (outside of Ansible) that does this as well. The wait=true is typically not enough because it will only block until the instance becomes available which is before ssh is up. Waiting for the status checks to pass is a better indication that the instance is fully booted. I'm not 100% sure whether this _always_ means that ssh is available but it does appear to be true.

I think the better thing to do is to add a new parameter to the existing ec2 module to block for system status checks, something like "wait_for_system_status".

My reason for not using ssh is I had an ansible play that launches an instance, creates an ami and then terminates it. In that case there is no reason to allow ssh access (if it's in a vpc, or by security group). I can't think of any other use cases though.

carsongee:
I agree with @jarv that this seems to be something better to do within the existing ec2 module, and would be very helpful to have there.

gregdek:
@eyalroth what do you think of submitting a patch to the ec2 module itself?

eyalroth:
@gregdek Sounds good to me. I shall see about adding this module as a new option in the ec2 module.

gregdek:
Thanks @eyalroth. Appreciate your patience in getting us to the right answer for everyone. I'll close this PR and we'll look forward to the new one (which should also be a quicker review, btw; changes to existing modules are easier to review than new modules.)

bcoca:
even ec2_facts might be a good place for this if it does not fit in ec2

PR #512, opened 2015-05-19
Added eval for pasting tag lists

skornehl:
We faced some problems while adding tags which included Variables. The reason was that inside of Ansible some tags were unicode formatted. 

We got a list like this:
['tag1', u'tag1-production', u'host:host1', 'tag2']

Which resulted in tags with leading "u_" in Datadog. I added a 'best hope' if-check because the tags are a string.

### DIFF
--- a/monitoring/datadog_event.py
+++ b/monitoring/datadog_event.py
@@ -116,7 +116,10 @@ def post_event(module):
     if module.params['date_happened'] != None:
         body['date_happened'] = module.params['date_happened']
     if module.params['tags'] != None:
-        body['tags'] = module.params['tags'].split(",")
+        if module.params['tags'].startswith("[") and module.params['tags'].endswith("]"):
+            body['tags'] = eval(module.params['tags'])
+        else:
+            body['tags'] = module.params['tags'].split(",")
     if module.params['aggregation_key'] != None:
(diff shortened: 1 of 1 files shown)


bcoca:
why not just make tags a list type in the argument spec?

skornehl:
Tried this, but than the u was part of the string instead of marking it as unicode

gregdek:
@arturaz Please review; if you +1, we will merge.

arturaz:
Um. I'm not sure why I'm mentioned here 😃

skornehl:
Author: Artūras 'arturaz' Šlajus x11@arturaz.net
Thats why :)

arturaz:
Oh, lol. Yeah, seems good.

gregdek:
@arturaz, we're moving to the "owner==maintainer" model, but it's clear that we haven't yet succeeded in notifying everyone of the change. :)  Thanks for the review, and thanks @skornehl for the patch.

gregdek:
So, reopening this -- a naked eval is risky, so we reverted the patch. Please resubmit a new PR for this, and consider using ast.literal_eval or  safe_eval here. Sorry for the reversion. :(

(We will be adding eval to a list of things to review for.)

skornehl:
You are absolutely right! Sorry, thatI did not think about that.

PR #525, opened 2015-05-22
Use expanduser to deal with path in bower module

QuantumGhost:
There's a bug in the bower module. 
When using the bower module with a path containing  `~` (`~/foo` for exmple), it would report a failure saying that  `No such file or directory: '~/foo'` even though the path actually exists.

This patch make sure that `expanduser` is always called so bower module can correctly deal with paths containing `~`.

### DIFF
--- a/packaging/language/bower.py
+++ b/packaging/language/bower.py
@@ -154,7 +154,7 @@ def main():
 
     name = module.params['name']
     offline = module.params['offline']
-    path = module.params['path']
+    path = os.path.expanduser(module.params['path'])
     state = module.params['state']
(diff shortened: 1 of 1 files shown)


abadger:
The logic looks fine but I think we probably want to apply it when we get the path out of the module parameters around line 158 where we do:

```
     path = module.params['path']
```

QuantumGhost:
Yep, you're right. I've changed it.

abadger:
@QuantumGhost  Thanks much for the fix!

# Merged

Hi!

issue is resolved.

Thank you!

PR #529, opened 2015-05-24
Fix alternatives module in non-English locale

cthiemann:
The alternatives module parses the output of update-alternatives, but the expected English phrases may not show up if the system locale is not English. Setting LC_ALL=C when invoking update-alternatives fixes this problem.

This is the symptom:

TASK: [laptop | Debian: update alternatives to use X11-enabled emacs] ********\* 
failed: [_hostname_] => {"failed": true, "parsed": false}
OpenSSH_6.7p1-hpn14v5, OpenSSL 1.0.1m 19 Mar 2015
debug1: Reading configuration data /etc/ssh/ssh_config
debug1: auto-mux: Trying existing master
debug1: mux_client_request_session: master session id: 2
  File "<stdin>", line 1742, in <module>
  File "<stdin>", line 96, in main
AttributeError: 'NoneType' object has no attribute 'group'

### DIFF
--- a/system/alternatives.py
+++ b/system/alternatives.py
@@ -85,7 +85,7 @@ def main():
 
     # Run `update-alternatives --display <name>` to find existing alternatives
     (rc, display_output, _) = module.run_command(
-        [UPDATE_ALTERNATIVES, '--display', name]
+        ['env', 'LC_ALL=C', UPDATE_ALTERNATIVES, '--display', name]
     )
 
     if rc == 0:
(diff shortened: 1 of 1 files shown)


gregdek:
@DavidWittman Please look at this PR; if you give it a +1, we will merge.

DavidWittman:
Thanks for the fix @cthiemann! @gregdek These changes look good to me, but I think ansible/ansible#11076 may be more appropriate.

gregdek:
@DavidWittman I'm a simple man, so explain things to me. :)

I'm presuming that the fix offered here is a localized fix for this module, and the fix in ansible/ansible#11076 is a more generalized fix?

If we incorporate this fix now, does it conflict with 11076 later? If so, then maybe we should close this in favor of 11076. If not, happy to merge.

cthiemann:
@gregdek @DavidWittman Just my two cents: Yes, ansible/ansible#11076 is more general and I initially thought about proposing something like that instead of the localized fix.

However, no matter what the global policy on locale will be in the future, the alternatives module explicitly relies on English output, thus I think it's appropriate to set LC_ALL=C locally to decouple it from the global policy.

Haven't tested it, but I don't think this would conflict with 11076.

PR #535, opened 2015-05-26
Add restore_backup module

yhekma:
I was missing a module to easily revert files from a previous backup. It is ofcourse easily done with a script, but I thought a module would be easier to integrate in a readable fashion in playbooks.


sivel:
New module submissions should be made against the 'devel' branch and should also contain a `version_added` key in the `DOCUMENTATION`, which at this point should be `2.0`.

yhekma:
will do, thanks

bcoca:
you don't need to close the ticket, you can force push on the branch and the PR would auto update

PR #537, opened 2015-05-26
Add restore_backup module

yhekma:
I was missing a module to easily revert files from a previous backup. It is ofcourse easily done with a script, but I thought a module would be easier to integrate in a readable fashion in playbooks.
I did a pull request earlier, but it was against the wrong branch and I needed to change some thins as well (turns out strptime is a more sane way of dealing with stuff than only a regex).


bcoca:
thanks this is one that has been on my todo list for a looong time.

files/restore_backup.py
```diff
@@ -0,0 +1,108 @@
...
+DOCUMENTATION = '''
+---
+module: restore_backup
+short_description: restores a backup created with 'backup=true' using, for instance, the lineinfile module
+description:
+  - looks for a backup file in the same dir as the file to be restored and renames it to the original.
+author: Yoram Hekma
+version_added: 2.0
+requirements:
+  - python > 2.6
```
sivel: This code does not actually require `python > 2.6`.  It appears to work on py24 as well.

files/restore_backup.py
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env python
+
+import os
+import re
+import json
```
sivel: There is no need to import `json` here.  You actually get it from `ansible.module_utils.basic` and it uses try/except logic to pull in simplejson if available.

files/restore_backup.py
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+    backup_path = os.path.join(dirpath, latest_backup)
+
+    try:
+        os.rename(backup_path, dest)
+    except Exception, e:
+        return {
+            'changed': False,
+            'stderr': json.dumps("%s (renaming %s to %s)" % (e, backup_path, dest)),
```
sivel: It looks like sometimes you get a `stderr` and some times you don't.  In general a standardized output would be good.  Also wouldn't this probably be better as a `fail_json` in this specific location?

yhekma:
I think this is somewhat more clean, don't you think?

files/restore_backup.py
```diff
@@ -0,0 +1,101 @@
...
+    choices: ["yes", "no"]
+    default: "no"
+'''
+
+EXAMPLES = '''
+  - action: restore_backup dest=/etc/samba/smb.conf
+'''
+
+
+def restore_backup(m, dest):
```
sivel: I think I would recommend, still calling this `module` instead of `m`.  Short names like that make code more difficult to understand.  

This is a common approach to pass `module` in so you have access to it, so +1 there.

files/restore_backup.py
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env python
```
bcoca: should be #!/usr/bin/python

sivel: This module is also missing the standard GPLv3 license header normally found in modules.

yhekma: Uhm, in the guidelines it says I should use #!/usr/bin/env
https://github.com/ansible/ansible/blob/devel/CODING_GUIDELINES.md

sivel: it does, but that is for bin scripts like `ansible-playbook`.  It also says:

> module code should still use '/usr/bin/python' as this is replaced automatically by 'ansible_python_interpreter', see the FAQ in the docs for more info

yhekma:
Trying to keep up:)

bcoca:
so we have brand new doc that might make this an easier process

bcoca:
might need formatting fix

yhekma:
@bcoca I already had a look at the page you mentioned. Any extra pointers wrt the formatting fix you are talking about?

files/restore_backup.py
```diff
@@ -0,0 +1,118 @@
...
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: restore_backup
+short_description: restores a backup created with 'backup=true' using, for instance, the lineinfile module
+description:
+  - looks for a backup file in the same dir as the file to be restored and renames it to the original.
+author: Yoram Hekma
+version_added: 2.0
```
resmo: should not be a float number, use `version_added: "2.0"`

files/restore_backup.py
```diff
@@ -0,0 +1,118 @@
...
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: restore_backup
+short_description: restores a backup created with 'backup=true' using, for instance, the lineinfile module
+description:
+  - looks for a backup file in the same dir as the file to be restored and renames it to the original.
+author: Yoram Hekma
```
resmo: use new format e.g. `author: "Yoram Hekma @yhekma"

files/restore_backup.py
```diff
@@ -0,0 +1,118 @@
...
+    description:
+      - Create a backup file including the timestamp information so you can get
+        the original file back if you somehow clobbered it incorrectly.
+    required: false
+    choices: ["yes", "no"]
+    default: "no"
+'''
+
+EXAMPLES = '''
+  - action: restore_backup dest=/etc/samba/smb.conf
```
resmo: explain what this does, what about an example for restore?

resmo:
The idea is nice, but I think it does not handle every case. What would be restored if there are more then one backup?

My suggestion is to add 3 more arguments: 
- `version` argument which has `default=latest` `choices=['earliest', 'latest']`
- `date`, `default=None` where you set the date from today to the past e.g. `date=2015-05-26`
- `time`, `default=None` so you can set a time as well e.g. `date=11:00`

Examples:

```
# restore latest of beginning from a particular date (looking back in the past)
restore_backup: dest=/etc/samba/smb.conf backup=yes date=2015-04-01

# restore latest of beginning from a particular date and time (looking back in the past)
restore_backup: dest=/etc/samba/smb.conf backup=yes date=2015-04-01 time=12:00

# restore earliest of beginning from a particular date and time (looking forward in the future)
restore_backup: dest=/etc/samba/smb.conf backup=yes date=2015-04-01 time=12:00 version=earliest
```

With these arguments you can pretty exact specify from which file you want to have restored.

Any thoughts?

yhekma:
I thought about the ability to restore more more than only the latest, but decided against that for sake of both simplicity and usability. The latter is admittedly only on a personal level.
In practice, when I want a restore, I just made a mistake and want a quick rollback. For the specific definition of dates to be practical for use with multiple targets/systems you would have to have a good idea as to what backup files are available on which systems. In my case I would want more control and maybe use a script.

The options earliest/latest are nice I think:)

Just my 2 cents

resmo:
That's fine. Your use case would be the default with no date and no time and `version=latest` which is also the default. So won't need to add a date or time or version at all. e.g. 

```
restore_backup: dest=/etc/samba/smb.conf backup=yes 
```

I just thought it would be nice to have more options. E.g. you test something and run a task lineinfile 10 times with backup before you realise it will not work, so you can use this module with date argument. But you are right for the moment, the current functionality is absolutely sufficient.

files/restore_backup.py
```diff
@@ -0,0 +1,136 @@
...
+            backup=dict(default=False, type='bool'),
+            version=dict(default='latest', choices=['latest', 'earliest']),
+        )
+    )
+
+    dest = module.params['dest']
+    version = module.params['version']
+
+    if module.params['backup']:
+        ext = time.strftime("%Y-%m-%d@%H:%M:%S~", time.localtime(time.time()))
```
bcoca: use the module.backup_local function instead

files/restore_backup.py
```diff
@@ -0,0 +1,136 @@
...
+        else:
+            backup_file = file_dict[reversed(mtime_list[-1])]
+
+    except (IndexError, KeyError):
+        module.exit_json(stderr='No backup found', changed=False)
+
+    backup_path = os.path.join(dirpath, backup_file)
+
+    try:
+        os.rename(backup_path, dest)
```
bcoca: use atomic move, rename can fail, also selinux needs to be considered, atomic move does all of this already

yhekma:
@bcoca You are quite right and I will change it so that it uses the atomic move. The reason I didn't do it in the first place was because then the restore would immediately restore that version so it would kind of defeat the purpose. Now that you mention the safety and selinux I will have a second look at it.

yhekma:
Thinking about it some more, I only need to consider selinux to be an issue when I am creating the backup, not when restoring it, right? I mean, the file will retain the contexts it was created with, which in turn should be the same as when it was backed up. Any thoughts?
Regarding the backup functionality you are absolutely right, but am still wondering how to implement this cleanly considering the backup_local function does not take an optional target filename...

yhekma:
@bcoca Looking at the code, I see that backup_local does use shutil.copy2 as well, so I am not convinced this would be a safer way to approach things. If I would use backup_local I would have to first rename the source which would, as far as I can see, needlessly complicate things. Thoughts?
Regarding the atomic move I was wondering if you could shed some light on your theory. The way I understand it is that a rename is atomic by definition. And since we are dealing with files in the same directory we do not have to worry about crossing filesystem bounderies (which would prevent the os.rename from working).

bcoca:
backup_local would keep it as a single place for the same code, in case a security issue or a change comes down the pike we only need to change it in one place. 

atomic move still makes sense because you are not guaranteed that all the backup kept the correct context.

Both make even more sense when you consider that the idea of a /var/cache/ansible_backups dir to centralize backups on a machine or making backups go to a central repository on the 'master', then modifying backup_local and atomic_move will fix it for all modules involved while in this case we would have to do extra coding.

yhekma:
Hmz, thats a good point on both accounts. Ill update the code.

gregdek:
Wow, @yhekma, nice work here. @bcoca / @resmo -- time for a re-review?

files/restore_backup.py
```diff
@@ -0,0 +1,135 @@
...
+    dest = module.params['dest']
+    version = module.params['version']
+
+    if module.params['backup']:
+        ext = time.strftime("%Y-%m-%d@%H:%M:%S~", time.localtime(time.time()))
+        backup_dest = "%s.%s-tmp" % (dest, ext)
+        try:
+            shutil.copy2(dest, backup_dest)
+        except (shutil.Error, IOError), e:
+            module.fail_json(msg='Could not make backup of %s to %s: %s' % (dest, backup_dest, e))
```
bcoca: still using it's own backup logic

yhekma:
Seems like I forgot a shutil instance and since I was away for a few days it took some more time.

gregdek:
Sorry @yhekma -- looks like you should probably rebase as well.

yhekma:
@gregdek Uhm, could you elaborate? I just did a rebase on upstream/devel, what am I missing?

yhekma:
oh oops, it seems like I did a merge out of habit a while ago.... If you want I can close this pull request and open a new one without the merge and only with rebases?

bcoca:
Just fix your local branch and force push, no need to close pr

yhekma:
Ok, I take it it is no problem if I lose the history? The easiest is just to do 1 commit on top of the HEAD of devel :)

bcoca:
No problem for us, we just want the history of who added the feature, the
intermediate steps are not needed.

---

Brian Coca

## 

Brian Coca

yhekma:
Ok, I just put it on top of devel:HEAD

yhekma:
@bcoca, @gregdek  is there anything else you think should be changed?

bcoca:
you still never changed it to use backup_local

yhekma:
@bcoca hmz, I thought you meant I should be using the means provided by the module instead of shutil and since I now use atomic_move it should be ok. Let me see what I can do. The problem (well, complicating factor really) is that when I use backup_local I cannot specify the destination name

bcoca:
actually that is part of the reason, currently the backup module uses a slightly different timestamp than what you are using, which will create incompatible backup file names

gregdek:
Looks like @yhekma made the change in fe652ed -- putting back into review

gregdek:
Back into community review, with a new process:

# 

gregdek:
Thanks @yhekma for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

PR #539, opened 2015-05-26
Add datadog_monitor, a module to configure and manage monitor statements

apenney:
within datadog.

### DIFF
--- a/monitoring/datadog_monitor.py
+++ b/monitoring/datadog_monitor.py
@@ -1,0 +1,193 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+#
+# Author: Ashley Penney <apenney@ntoggle.com>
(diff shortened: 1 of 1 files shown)


skornehl:
Hi,

good idea, we actually thought about doing the same.

What about modifying/deleting/silencing a monitor? For is this would be a common case to silence monitors before deploying - or changing thresholds in some cases.

Why don't you use the datadog API plugin: http://docs.datadoghq.com/integrations/python/ I think this is more reliable than building own REST calls

skornehl:
To Answer partially myself :)

> Why don't you use the datadog API plugin: http://docs.datadoghq.com/integrations/python/ I think this is more reliable than building own REST calls

Because it is not supported to create Monitors with it.

apenney:
Closed in favor of #575

PR #548, opened 2015-05-28
Add module to run puppet

emonty:
There is a growing pattern for using ansible to orchestrate runs of
existing puppet code. For instance, the OpenStack Infrastructure team
started using ansible for this very reason. It also turns out that
successfully running puppet and interpreting success or failure is
harder than you'd expect, thus warranting a module and not just a shell
command.

This is ported in from
http://git.openstack.org/cgit/openstack-infra/ansible-puppet

### DIFF
--- a/system/puppet.py
+++ b/system/puppet.py
@@ -1,0 +1,186 @@
+#!/usr/bin/python
+
+# Copyright (c) 2015 Hewlett-Packard Development Company, L.P.
+#
+# This module is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This software is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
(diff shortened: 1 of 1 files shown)


resmo:
what about adding check_mode support by adding the --no-op argument?

emonty:
Fantastic suggestion!
On May 29, 2015 1:32 AM, René Moser notifications@github.com wrote:what about adding check_mode support by adding the --no-op argument?

resmo:
some more thoughts...

maybe it would help to add a `--no-noop` per _default_, in case of `noop` is specified as default behaviour in `puppet.conf`, otherwise you won't have any possibilities to run it in `--no-noop`

and environment support would be helpful :)

perzizzle:
Is there a reason you don't use /etc/puppetlabs/puppet/puppet.conf to determine the default puppet master?

bcoca:
merged, @perzizzle we can not improve upon the baseline, just send a PR

emonty:
@perzizzle that would be a great thing to search for if puppetmaster is None - there are some cases (rolling upgrades to a new puppet master) where having the orchestration override the configured puppetmaster is important. Also - /etc/puppetlabs/puppet.conf is the location of the fancy paid-for file - /etc/puppet/puppet.conf is where the Open Source file defaults to - so we should probably check for both.

That said - I totally agree with @bcoca - sounds like a great PR

PR #549, opened 2015-05-29
Add force option to alternatives module

jamespharaoh:
This allows you to pass the --force option to update-alternatives, from the manpage:

``
       --force
              Let update-alternatives replace or drop any real file that is installed where an alternative link has to be installed or removed.

```
```

### DIFF
--- a/system/alternatives.py
+++ b/system/alternatives.py
@@ -33,6 +33,7 @@ version_added: "1.6"
 author:
     - '"David Wittman (@DavidWittman)" <dwittman@gmail.com>'
     - '"Gabe Mulley (@mulby)" <gabe.mulley@gmail.com>'
+    - '"James Pharaoh" <james@pharaoh.uk>'
 options:
   name:
     description:
@@ -47,6 +48,13 @@ options:
       - The path to the symbolic link that should point to the real executable.
       - This option is required on RHEL-based distributions
     required: false
+  force:
+    description:
+      - Replace or drop any real file that is installed where an alternative
+        link has to be installed or removed.
+    choices: [ "yes", "no" ]
+    required: false
+    default: "yes"
 requirements: [ update-alternatives ]
 '''
 
@@ -69,6 +77,7 @@ def main():
             name = dict(required=True),
             path = dict(required=True),
             link = dict(required=False),
+            force = dict(default=False, type='bool'),
         ),
         supports_check_mode=True,
     )
@@ -77,6 +86,7 @@ def main():
     name = params['name']
     path = params['path']
     link = params['link']
+    force = params['force']
 
     UPDATE_ALTERNATIVES = module.get_bin_path('update-alternatives',True)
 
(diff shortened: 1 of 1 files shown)


gregdek:
@DavidWittman @mulby Please look at this PR; if you give it a +1, we will merge.

system/alternatives.py
```diff
@@ -114,7 +124,16 @@ def main():
                         link = line.split()[1]
                         break
 
-    if current_path != path:
+    if not force:
+        need_to_force = False
+    elif not os.path.islink(link):
```
mulby, line 129: Won't this result in "--force" being passed in even if the user of the task didn't pass in a "force=yes" parameter to the module?

I guess I would expect it to only pass "--force" on to update-alternatives if it was passed into the module...

jamespharaoh, line 129: No, it shouldn't do that and the first branch of the if handles that specifically and with the highest precedence of all the options here.

mulby, line 129: Ah yes, you are correct, I misread this earlier.

What I was trying to get at is really the question of whether this
complexity is necessary. If "force" is passed into the module, why not just
always pass "--force" into update-alternatives?

I suspect there will be cases where you would want --force that aren't
listed here... Because we don't know about them or because the tool changes
over time.

DavidWittman, line 129: I haven't done any testing against this, but I think I'm with @mulby here. Rather than try to be intelligent about whether or not force should be used, we should trust the user's suggestion.

@jamespharaoh are there any problems with that approach?

system/alternatives.py
```diff
@@ -123,10 +142,17 @@ def main():
                 if not link:
                     module.fail_json(msg="Needed to install the alternative, but unable to do so as we are missing the link")
 
-                module.run_command(
-                    [UPDATE_ALTERNATIVES, '--install', link, name, path, str(DEFAULT_LINK_PRIORITY)],
-                    check_rc=True
-                )
+                if need_to_force:
```
DavidWittman, line 145: Rather than splitting these commands up with a conditional, I think it's best to DRY and create a separate variable for the command to be run.

E.g.

``` python
cmd =  [UPDATE_ALTERNATIVES, '--install', link, name, path, str(DEFAULT_LINK_PRIORITY)]
if need_to_force:
    cmd.append('--force')
module.run_command(cmd, check_rc=True)
```

Not sure off the top of my head if you can just append the `--force`, but if that doesn't work you can just do `cmd.insert(1, '--force')`.

system/alternatives.py
wislta-github, line 129: well arguably we could always provide the --force flag, although it makes no difference. the point of this logic was more to do with detecting if a change was required, which requires extra checks (the islink and readlink calls here), but only if --force is specified.

system/alternatives.py
wislta-github, line 145: i actually did something like this originally... i'll come up with a new pull request when i have a chance...

DavidWittman, line 145: @wislta-github You can (and should) use the same PR. You can just push new commits to your branch and they'll show up here. The preferred way (so you don't pollute the commit history) is to amend your commit with `git commit --amend` and then force-push to your branch with `git push -f <remote> alternatives-force`.

gregdek:
Looks like this is now failing merge checks; @jamespharaoh can you rebase?

gregdek:
@jamespharaoh A friendly reminder: this pull request has been marked as needing your action, and it hasn't been updated in the last month. 

gregdek:
Closed due to inactivity.

PR #551, opened 2015-05-29
Modules for modifying a F5s GTM

perzizzle:
There is plenty of room for more features but thought I'd put this out there for comment since I've implemented all the features I need (changing facts module to use classes, flushing out add/remove functionality in virtual servers etc).

I have tests I can run to validate the functionality but they are obviously dependent on having an F5 available.

### DIFF
--- a/windows/unzip.ps1
+++ b/windows/unzip.ps1
@@ -1,0 +1,84 @@
+#!powershell
+# Michael Perzel 1/22/2014
+# WANT_JSON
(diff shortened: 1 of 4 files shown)


bcoca:
could you submit each module in it's own PR? its a lot easier to review and accept one at a time

perzizzle:
Sure, I can do this shortly. Although to do interesting things you need to use at least a couple of the modules.

PR #553, opened 2015-05-29
Windows unzip

perzizzle:
Windows module for unzipping files.

This is identical to #277 but has been moved to a dedicated branch. Please see discussion at previous pull request.

### DIFF
--- a/windows/unzip.ps1
+++ b/windows/unzip.ps1
@@ -1,0 +1,84 @@
+#!powershell
+# Michael Perzel 1/22/2014
+# WANT_JSON
+# POWERSHELL_COMMON
+
+$params = Parse-Args $args;
+
+$result = New-Object PSObject -Property @{
+    changed = $false
+    success = $false
+    output = ""
+}
+
+If ($params.source) {
+    $source = $params.source
+}
+Else {
+    Fail-Json (New-Object PSObject $fail) "missing required argument: source"
+}
+
+If ($params.destination) {
+    $destination = $params.destination
+}
+Else {
+    Fail-Json (New-Object PSObject $fail) "missing required argument: destination"
+}
+
+If ($params.include) {
+    $include = $params.include
+}
+If ($params.exclude) {
+    $exclude = $params.exclude
+}
+
+If ($params.creates) {
+    $creates = $params.creates
+}
+
+If( $creates -and (Test-Path $destination/$creates)) {
+    $result.output += "File already exists, not unzipping"
+    $result.success = $true
+    Exit-Json $result
+}
+
+If(!(Test-Path $source)) {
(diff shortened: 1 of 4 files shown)


wenottingham:
How does this compare to #174 and #177?

perzizzle:
Those pull requests require you to install PSCX but supports more file types. This solution only requires powershell.

schwartzmx:
There are many many threads pertaining to problems with using PowerShell's built-in `Shell.Application`, there was a reason I used PSCX in my PRs for `win_zip` and `win_unzip`, it's well tested and runs 7zip under the hood (also well tested).   It also has more functionality for other forms of compressed files.

PowerShell 5 comes with a builtin library for extracting compressed files. :pray:

jhawkesworth:
There's
[System.IO.Compression.ZipFile]::ExtractToDirectory($zipfile, $dest)
available in PS4 at least, possibly PS3 too.
Its not the most fully featured implementation though unfortunately.

bcoca:
split the diff ? use PSCX if it exists, fallback on the powershell builtins otherwise and show message pointing at 7zip install if needed?

perzizzle:
Seems like a reasonable solution to me. I personally am not going to install pscx and haven't had any issues unzipping files using powershell but I can see why others would be willing to do so. If powershell adds a commandlet for unzip/zipping files the module could be updated then.

schwartzmx:
I can make that happen in https://github.com/ansible/ansible-modules-extras/pull/177 and https://github.com/ansible/ansible-modules-extras/pull/174.   We should figure out what will work for everyone.  After all, we are working together to bring more windows modules to ansible.  And we really need all the windows help/support we can get :+1:

gregdek:
So we need to make a decision. 

@bcoca has proposed a solution. "split the diff? use PSCX if it exists, fallback on the powershell builtins otherwise and show message pointing at 7zip install if needed?"  It sounds like there's broad agreement that it's a sensible approach.

So do we implement that solution in this PR, which appears to be for unzipping only, or do we implement the solution in PRs #174 and #177, which are for unzip and zip, respectively? @schwartzmx and @perzizzle looking for your guidance.

perzizzle:
Not sure it makes a difference where the features are merged. Guess whoever volunteers? I could do it but it'll take me a week or so.

schwartzmx:
I've pushed an update to https://github.com/ansible/ansible-modules-extras/pull/174, pertaining to our conversation if anyone wants to take a look.  I will make very similar changes to https://github.com/ansible/ansible-modules-extras/pull/177 if this checks out okay.

thanks

gregdek:
Since @schwartzmx has volunteered near-term, we'll go with #174 and #177 -- @perzizzle if you can review them that would be helpful. Closing this one. Thanks!

PR #559, opened 2015-05-30
Add backup_copy option to the patch module

thread13:
Hi Guys,

I'm sure there's a better way to do this -- but as backups are a generally desirable feature, having at least numbered backups provided by the patch utility itself would be nice. 

Kind regards,
/t13

PS. I'm sure making "timestamped" backups would be better, I'm just not that well acquainted with ansible code yet ( much less than an hour ) to make it the right way. However, if you're also busy, I wonder if we can have this for now, since IMHO it's much better than nothing.

### DIFF
--- a/files/patch.py
+++ b/files/patch.py
@@ -65,7 +65,7 @@ options:
     required: false
     type: "int"
     default: "0"
-  backup_copy:
+  backup:
     description:
       - passes --backup --version-control=numbered to patch, 
         producing numbered backup copies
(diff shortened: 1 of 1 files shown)


gregdek:
@jirutka @luisperlaz Please look at this PR; if you give it a +1, we will merge.

bcoca:
Just name the option 'backup' as other modules do. Also not all version of patch support this form of backup so using the shared backup_local function might be a better way to keep it cross OS/Distros.

thread13:
I have renamed the backup parameter as requested.

However, as for backup_local, it turns out it may be better to keep it the current way:
-- first, there may be more than one target file and the dest parameter not set: GNU patch then takes the file names from the patch file, and we'll have to parse it as well ; this is just extra logic which is always prone to error ;
-- second, the docs explicitly state that the module targets GNU patch utility ; I have looked up the source code at http://ftp.gnu.org/gnu/patch/?C=M;O=A , and the earliest available version of 31-Aug-1997 ( patch-2.5 ) does have this option for backup. 

So for now I suggest we let patch do the job .

bcoca:
good points, specially the 2nd, on the first we will probably get push back down the line to make it 'non-gnu' patch compatible. For now I'll just remove my objections.

thread13:
Great )

And thank you

Issue #455, opened 2015-05-06
lvg is not idempotent -- attempts to create VG that already exists


gtirloni:
**Issue Type:** Bug Report

**Ansible Version:** 1.9.1

**Environment:** CentOS 7.1 x86_64

**Summary:** lvg module attempts to create volume group that already exists

**Steps to Reproduce:**

```
- name: Create LVM VolumeGroup
  lvg:
    vg: "something"
    state: present
    pvs: "/dev/md/something"
```

Run it twice.

**Expected Result:** lvg module would detect VG already exists and be happy that "state=present" is satisfied.

**Actual Results:** lvg module attempts to create VG and gets an error from vgcreate.

```
TASK: [lvm | Create LVM VolumeGroup] ****************************************** 
failed: [prd] => {"err": "  Can't initialize physical volume \"/dev/md/something\" of volume group \"something\" without -ff\n", "failed": true, "rc": 5}
msg: Creating physical volume '/dev/md/something' failed
```

alxgu:
@gtirloni does https://github.com/ansible/ansible-modules-extras/pull/562 fix your issue?
So maybe that issue can be closed.

gtirloni:
Yep, thanks!

xelor81:
I have the very same issue here, all mentioned fixes are in place however the system still trying either to create the vg on the physical device or (when devices are not defined in the pvs array) reports that none of the physical volumes has been given.

```
[support@us000832 playbooks]$ ansible us000832 -s -m lvg -a "vg=other state=present"
us000832.eu.tieto.com | FAILED! => {
    "changed": false,
    "failed": true,
    "msg": "No physical volumes given."
}

[support@us000832 playbooks]$ ansible us000832 -v -s -m lvg -a "vg=other state=present pvs=/dev/md/something"
Using /etc/ansible/ansible.cfg as config file
us000832.eu.tieto.com | FAILED! => {
    "changed": false,
    "failed": true,
    "msg": "Device /dev/md/something not found."
```

PR #562, opened 2015-05-31
Devices in the current_devs list must also be converted to absolute d…

fdupoux:
…evice paths so comparison with dev_list works

This fixes issue #455 as all lists of devices must contains real paths so they can be compared. The code compares two lists of devices (current_devs and dev_list) which are the lists of PVs in the VG (the list of PV devices we currently have in the VG and the list of PV devices we want to have in the VG).

At the beginning of the module os.path.realpath is used to make sure devices in dev_list are converted to a real path if necessary. But this operation is not being done in current_devs, hence this list can contain a symbolic link to a device and this device will be considered as different in the following comparisons even though they refer to the same device.
        devs_to_remove = list(set(current_devs) - set(dev_list))
        devs_to_add = list(set(dev_list) - set(current_devs))

The fix simply uses os.path.realpath to convert all devices in current_devs to real paths.

### DIFF
--- a/system/lvg.py
+++ b/system/lvg.py
@@ -211,7 +211,7 @@ def main():
                     module.fail_json(msg="Refuse to remove non-empty volume group %s without force=yes"%(vg))
 
         ### resize VG
-        current_devs = [ pv['name'] for pv in pvs if pv['vg_name'] == vg ]
+        current_devs = [ os.path.realpath(pv['name']) for pv in pvs if pv['vg_name'] == vg ]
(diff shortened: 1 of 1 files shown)


gregdek:
@abulimov Please take a look at this pull request. If you give it a +1, we will merge it.

abulimov:
Looks good to me, thanks for this fix!

gregdek:
Merged! Thanks @fdupoux for the patch, and thanks @abulimov for the review.

Issue #550, opened 2015-05-29
Bundle python2-lxc into the distribution


leonty:
##### ISSUE TYPE

Feature Idea
##### COMPONENT NAME

lxc_container module
##### ANSIBLE VERSION

N/A
##### SUMMARY

Please bundle python2-lxc into the distribution or cooperate with its developers to make it available as a ready-to-use module in public repos (pypi, apt). For now the module lxc_container demands manual administering the target machine to build & install python2-lxc from git.
Thanks.

bcoca:
i just used pip to install it, but it is named lxc-python2

sivel:
FWIW the pip module can install a package directly from a git repo, such as:

```
pip:
    name: git+https://github.com/lxc/python2-lxc.git#egg=lxc
```

leonty:
sivel, thanks.

bcoca:
I have uploaded this module to pypi (https://pypi.python.org/pypi/lxc-python2/0.1). Please update the page http://docs.ansible.com/lxc_container_module.html with the module name - lxc-python2 instead of python2-lxc.

The module author also indicated that: 
"Feel free to upload it to pypi yourself, as far as upstream is concerned the python2 module is an unsupported binding which we intend to deprecate in the very near future (it was a one time hack to get OpenStack support, but we've since moved to LXD for that).".
(full discussion https://github.com/lxc/python2-lxc/issues/5)

So probably it's reasonable to choose another lxc python2 binding that is actively maintained.

cloudnull:
Just as a note here. the Python2 LXC binding that is being used is stable for LXC version 1 and >. As soon as Ansible supports Python3 we'll be able to drop the requirement for this library and move to the proper LXC python3 bindings which are what powers the user space tools in modern LXC and is being supported by the greater LXC community. Moving to a new library would be problematic in the future as we'd like to stay inline with what the LXC community is creating as their supported platform.  

Now I don't know timelines as to when Ansible might be able to support Python3 however we could look to using only userspace tools instead of the bindings if that would help spur adoption and limit confusion with how the lxc-container-module is supposed to be used.

bcoca:
inventory scripts really don't need ansible to support anything, they get executed as an external script/binary and are never imported into ansible

cloudnull:
[...]

PR #567, opened 2015-06-01
Updates the doc information for the python2-lxc dep

cloudnull:
The python2-lxc library has been uploaded to pypi as such this commit
updates the requirements and doc information for the module such that
it instructs the user to install the pip package "lxc-python2" while
also noting that the package could be gotten from source as well. In
the update comments have been added to the requirements list which
notes where the package should come from,

Closes-Bug: https://github.com/ansible/ansible-modules-extras/issues/550

### DIFF
--- a/cloud/lxc/lxc_container.py
+++ b/cloud/lxc/lxc_container.py
@@ -173,9 +173,9 @@ options:
           - list of 'key=value' options to use when configuring a container.
         required: false
 requirements:
-  - 'lxc >= 1.0'
-  - 'python >= 2.6'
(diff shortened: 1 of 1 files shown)


gregdek:
@cloudnull considering the comments on #550, what's the status of this PR?

cloudnull:
@gregdek its good to go. I added this pr in response to issue #550 to better document how to install `lxc-python2` libs and where they can be installed from. In that issue they ask for the libs to be bundled in the upstream Ansible packages however I don't think that's something that Ansible should do as the external dependency is easy enough to install in a variety of ways.

PR #568, opened 2015-06-02
Added eval for pasting tag lists

skornehl:
Follow up for #512 - switched to ast.literal_eval


gregdek:
Hm, looks like this can't be auto-merged -- probably due to the reversion changing head. Please rebase to head and resubmit; we've basically got approval, since the only difference between this and the patch to #512 is the change from eval to ast.literal_eval.

gregdek:
Ahhh, ok. So looks like @bcoca was in this corner of the code and committed a fix; that's why the PR now conflicts. @skornehl, thanks for your help; if you can check out latest and make sure that the fix suits your needs, we can close this.

skornehl:
@gregdek, @bcoca : No as I already mentioned in #512 that sadly does not fix anything. As an example:

``` yaml
---
- hosts: localhost
  gather_facts: false
  tasks:
    - name: test DD
      datadog_event:
        api_key=123456
        title="test"
        text="test"
        tags="['tag1', u'tag1-production', u'host:host1', 'tag2']"
```

This results in a Message:

bildschirmfoto 2015-06-03 um 07 11 00

The Problem ist that my Tag List is auto generated and contains unicode parts. The next step would be looping over the list and encoding utf to ascii - but this seems like a very bad hack. I will reach out for the Datadog Support and see if they can help on their side and tell you afterwards.

bcoca:
> The next step would be looping over the list and encoding utf to ascii - but this seems like a very bad hack.
> Isn't this what your PR was counting on? as the assembled '[ ] ' string having coerced the unicode parts into basestrings.

something like this would be simple enough:
`b = [str(x) for x in a]`

bcoca:
Actually the strange thing is that this should all be converted to utf8 as you are converting the whole payload to json for sending to the datadog API.

skornehl:
> Isn't this what your PR was counting on? as the assembled '[ ] ' string having coerced the unicode parts into basestrings.

Yes it was, in first time I thought about a quick fix which will work so I can use it soon. But since I already invested some time on this I believe there should be a solution which isn't a hack.

> Actually the strange thing is that this should all be converted to utf8 as you are converting the whole payload to json for sending to the datadog API.

I will try to debug here again later today to understand whats happening, but maybe it is better to use the datadogpy package instead of sending directly to the REST API.

skornehl:
Datadog Support responded

> It looks like this is due to the way the list is parsed, since it is already a string, so the u'host:host1' would show as "u'host:host1'". Would it be possible to submit the tag list as a unicode string (something like tags=u'["tag1", "tag1-production", "host:host1", "tag2"]')?

I will close this PR for now and reopen it once I have a proper solution. Thanks for involve!

bcoca:
the more i go over this ... the more this looks like an issue with the json dumper, u'string' should not be in the json, it is all already supposed to be utf8, this looks like a bug on our side

PR #570, opened 2015-06-02
Added proxmox_template module

UnderGreen:
Module for managing OpenVZ or ISO templates on ProxmoxVE via API.

### DIFF
--- a/cloud/misc/proxmox_template.py
+++ b/cloud/misc/proxmox_template.py
@@ -129,10 +129,10 @@ def get_template(proxmox, node, storage, content_type, template):
 def get_content(proxmox, node, storage):
   return proxmox.nodes(node).storage(storage).content.get()
 
-def upload_template(module, proxmox, node, storage, content_type, realpath, timeout):
+def upload_template(module, proxmox, api_host, node, storage, content_type, realpath, timeout):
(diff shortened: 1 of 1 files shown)


cloud/misc/proxmox_template.py
```diff
@@ -0,0 +1,245 @@
...
+    description:
+      - the user to authenticate with
+    required: true
+  api_password:
+    description:
+      - the password to authenticate with
+      - you can use PROXMOX_PASSWORD environment variable
+    default: null
+    required: false
+  https_verify_ssl:
```
bcoca: you can keep this as an alias, but we would prefer to have the same name here as most other modules 'validate_certs'

cloud/misc/proxmox_template.py
```diff
@@ -0,0 +1,245 @@
...
+  force:
+    description:
+      - can be used only with C(state=present), exists template will be overwritten
+    default: false
+    required: false
+    type: boolean
+  state:
+    description:
+     - Indicate desired state of the template
+    choices: ['present', 'absent', 'list']
```
bcoca: list seems to indicate you want a prmox_facts kind of module also, there are some modules that do this in one place but we prefer to separate information from management modules

UnderGreen:
Thank you, Brian!
I fixed this small remarks.

Issue #244, opened 2015-02-11
Monit doesn't take pending state into consideration for whether a process is running


gaqzi:
When determining whether a process is running through monit the module only checks for the presence of the string "running", which can lead to false positives when the process is waiting to be stopped.

The module should be updated to ensure that "stop pending" is not present as well.

Now monit doesn't support waiting for an action to finish, so if we wanted to then ensure the process was started then we would have to wait until the pending goes away and then re-evaluate. With some kind of timeout to determine whether monit has hung. 
Initializing is also a state that is not covered and also probably should we waited for.

I'm willing to write a patch for this if the approach of adding a loop if pending/initialized is part of the status string would be acceptable. If either is in status then sleep for 1s and try status again. If it reaches a max retries/timeout without a state changed then it fails. If either string is not part of status then continue processing as today.

mpeters:
I've been hit by this bug several times. It's super frustrating because it's a race condition that can pop up at any time and fail on some nodes but not others. I'll probably be working on a patch in the not too distant future if nothing is fixed in the meantime.

gaqzi:
@mpeters I have a patch prepared, I just need to get an approval to release it from my client. Will be bugging them about it this week!

mpeters:
That would be great. Please keep bugging them and keep us informed :)

mpeters:
Just to verify that it covers all the cases that I'm interested in, I just had this failure happen (on 1 node of 4, so seems like a race-condition):

Tried to tell monit to restart processes

`monit: name=foo state=restarted`
and received the following failure:
`{"failed": true, "status": "not monitored"}`

If I'm restarting the process then monit shouldn't care that it was previously not monitored, but should restart it and monit it, correct?

gaqzi:
@mpeters hm, I don't think I've seen that particular state before. Was monit monitoring before you started the run?

Either way I've made a PR now and if you could please try it out in your environment as well to see if it covers all the use cases that would be awesome :)

mpeters:
@gaqzi While I can't prove a negative (and race conditions are such negatives), this is working for me. Using your module I haven't experienced a single failure that I used to. And without it I still see random failures.

I do have two comments though, on a possible improvements/extensions of your work:
1. When `state="reloaded"` the module issues a reload command to monit, but then doesn't stick around looking at the status to make sure name given is present. Should it?
[...]

PR #572, opened 2015-06-02
monit: Add retry for pending/initializing services

gaqzi:
If there are already ongoing actions for a process managed by monit, the
module would exit unsuccessfully. It could also give off false positives
because it did not determine whether the service was started/stopped
when it was in a pending state. Which might be turning the service off,
but the action was to start it.

For example "Running - pending stop" would be regarded as the service
running and "state=enabled" would do nothing.

This will make Ansible wait for the state to finalize, or a timeout decided
by the new `max_retries` option, before it decides what to do.

This patch is related to issue #244 and aims to fix it. I'm a bit unsure of 
the procedure to deal with patches between version 2 and 1.9-stable, 
should I create another pull request for 1.9-stable since this is a bug?


bcoca:
cc @dstoflet

gaqzi:
@mpeters [raised a good point](https://github.com/ansible/ansible-modules-extras/issues/244#issuecomment-108910990) in #244 that we're not waiting around when reloaded is being issued. I'll have a look at it and work at incorporating that with this PR as well.

mpeters:
Currently we don't allow the "pending" status to be valid when doing an "unmonitor" like we do for everything else. That's not a flaw in your patch, but a flaw in the original that can be fixed by this work.

Line 152 should be changed from:

```
if status in ['not monitored']:
```

to

```
if status in ['not monitored'] or 'unmonitor pending' in status:
```

gregdek:
Looks like this needs to be rebased -- also @dstoflet are you available to review?

gregdek:
@gaqzi A friendly reminder: this pull request has been marked as needing your action, and it hasn't been updated in the last month. 

gregdek:
Closed due to inaction.

gaqzi:
@gregdek sorry for the late reply. :)

I have now rebased my branch to the current latest. 
Can this be re-opened or should I do a new pull request?

bcoca:
hmm, github is not allowing us to reopen, so I guess you need to resubmit

PR #574, opened 2015-06-03
Factor common functions for F5 modules

Etienne-Carriere:
Linked to ansible PR : https://github.com/ansible/ansible/pull/11142


gregdek:
Because this relies on changes to Ansible core, we cannot merge until we review the parent change, which may take us a while with the 2.0 release cycle going on. Thanks for the PR, and we appreciate your patience.

PR #575, opened 2015-06-03
datadog_monitor module

skornehl:
I'm sorry for creating a concurrent PR to #539 but we needed this module and the existing one didn’t have all features we needed. I also decided to use the datadogpy instead of plain REST, I think this is more reliable than reimplementing the REST calls.
Short description:
The options are all from the datadog API Page except state which can be set to 'present', 'absent', 'muted' and 'unmuted' 
I'm catching Exception which is not the best choice but datadog itself just raises http exceptions without any explanation. If I try to mute a monitor which is already muted it raises for example a 403. Due to that fact I also decided to fail if this happens (but with a real error message).

Sample playbook:

``` yaml

---
- hosts: localhost
  gather_facts: false
  tasks:
    - name: Test Monitor
      datadog_monitor:
        type: "metric alert"
        name: "Test monitor"
        state: "present"
        query: "datadog.agent.up".over("host:host1").last(2).count_by_status()"
        message: "Some message."
        api_key: "9775a026f1ca7d1c6c5af9d94d9595a4"
        app_key: "87ce4a24b5553d2e482ea8a8500e71b8ad4554ff"
```

### DIFF
--- a/monitoring/datadog_monitor.py
+++ b/monitoring/datadog_monitor.py
@@ -41,15 +41,12 @@ options:
     api_key:
         description: ["Your DataDog API key."]
         required: true
-        default: null
     app_key:
         description: ["Your DataDog app key."]
         required: true
-        default: null
     state:
         description: ["The designated state of the monitor."]
         required: true
-        default: null
         choices: ['present', 'absent', 'muted', 'unmuted']
     type:
         description: ["The type of the monitor."]
@@ -63,7 +60,6 @@ options:
     name:
         description: ["The name of the alert."]
         required: true
-        default: null
     message:
         description: ["A message to include with notifications for this monitor. Email notifications can be sent to specific users by using the same '@username' notation as events."]
         required: false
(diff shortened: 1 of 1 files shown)


monitoring/datadog_monitor.py
```diff
@@ -0,0 +1,278 @@
...
+- "Options like described on http://docs.datadoghq.com/api/"
+version_added: "2.0"
+author: '"Sebastian Kornehl" <sebastian.kornehl@asideas.de>'
+notes: []
+requirements: [datadog]
+options:
+    api_key:
+        description: ["Your DataDog API key."]
+        required: true
+        default: null
```
bcoca: default is irrelevant for require: true fields

skornehl:
@bcoca Thanks four your input! Patched that parts away.

apenney:
Looks awesome, I'll close my PR in favor of this one!

apenney:
Looks awesome, I'll close my PR in favor of this one!

PR #579, opened 2015-06-03
now handles keys protected with a passphrase

ghost:
This is an update to the ec2_win_password module to make it work with passphrase protected keys.


gregdek:
@rickmendes can you add version added in the doc section (2.0)? Once you do that we'll happily merge.

ghost:
@gregdek Thanks for the reminder. Done.

ghost:
@gregdek As the owner of this module, is there anything I can do to get these updates merged? Thanks.

gregdek:
Gah, looks like a process breakdown; we should have auto-merged the second you fixed version_added as an owner_pr. My fault; I missed it.

And unfortunately, looks like it now needs rebasing. So if you can take care of that, we should be able to merge this basically immediately.

ghost:
@gregdek Rebased as requested. Thanks.

PR #582, opened 2015-06-04
Creation of F5 virtual_server module

Etienne-Carriere:


network/f5/bigip_virtual_server.py
```diff
@@ -0,0 +1,414 @@
...
+    - "Best run as a local_action in your playbook"
+requirements:
+    - bigsuds
+options:
+    server:
+        description:
+            - BIG-IP host
+        required: true
+        default: null
+        choices: []
```
bcoca: no need for empty choices/aliases, nor default when option is required

network/f5/bigip_virtual_server.py
```diff
@@ -0,0 +1,414 @@
...
+        aliases: []
+    validate_certs:
+        description:
+            - If C(no), SSL certificates will not be validated. This should only be used
+              on personally controlled sites using self-signed certificates.
+        required: false
+        default: 'yes'
+        choices: ['yes', 'no']
+        version_added: 2.0
```
bcoca: no need for version_added for option when its a new module

network/f5/bigip_virtual_server.py
```diff
@@ -0,0 +1,414 @@
...
+            if name is None:
+                module.fail_json(msg="name parameter required when " \
+                                     "state=enabled/disabled")
+            if not module.check_mode:
+                pass
+            else:
+                # check-mode return value
+                result = {'changed': True}
+
+    except Exception, e:
```
bcoca, line 455: we prefer having localized exception handlers so you can give meaningful messages on top of the exception, normally the api libraries give very cryptic responses and it is nicer to give the user something relative to the action being attempted.

gwagner:
This module looks fantastic and would be a big help to our org.  What do we need to do to get it into the modules extras?

gregdek:
Looks like we're waiting on the author to localize exception code here.

Etienne-Carriere:
Exact, I have to localize exceptions .  
@gwagner : On the feature part, I don't have persistence settings which is important for the VS. I don't know currently how to present the default vs non default persistance methods . (differences between API and GUI).
On the planning : I come back from 2 weeks holidays . Will try to write it for the end of the month of June .

Etienne-Carriere:
@bcoca , I added some localized exceptions even I keep the global exception for having the module.fail_json . 
Could you review it ?

network/f5/bigip_virtual_server.py
```diff
@@ -0,0 +1,405 @@
...
+    try:
+        api.LocalLB.VirtualServer.create(
+            definitions = [{'name': [name], 'address': [destination], 'port': port, 'protocol': 'PROTOCOL_TCP'}],
+            wildmasks = ['255.255.255.255'],
+            resources = [{'type': 'RESOURCE_TYPE_POOL', 'default_pool_name': pool}],
+            profiles = _profiles)
+        result = True 
+        desc = 0
+    except Exception, e :
+        print e.args
```
bcoca: Don't use print in modules, use either exit or fail json functions.  Also the point of having localized exceptions is to add an error message on top that is more helpful than most library exception messages + traceback (but we also want to include those), if you are just going to pass the exception message, there is really no difference with using the general catchall exception.

network/f5/bigip_virtual_server.py
```diff
@@ -0,0 +1,405 @@
...
+            if (x not in profiles_list) and (x!= "/Common/tcp"):
+                to_del_profiles.append({'profile_context': 'PROFILE_CONTEXT_TYPE_ALL', 'profile_name': x})
+        if len(to_del_profiles)>0:
+            api.LocalLB.VirtualServer.remove_profile(virtual_servers = [name],profiles = [to_del_profiles])
+            updated=True
+        if len(to_add_profiles)>0:
+            api.LocalLB.VirtualServer.add_profile(virtual_servers = [name],profiles= [to_add_profiles])
+            updated=True
+        return updated
+    except bigsuds.OperationFailed, e:
```
bcoca, line 235: so this is more or less a good example of using the localized exceptions

bcoca:
module seems mostly OK, just the 1 localized exception using print and not adding anything meaningful, all the others are good as they give the user something more meaningful to start with.

gregdek:
Back into community review.

Etienne-Carriere:
@gwagner : do you have a feedback on the module ?

network/f5/bigip_virtual_server.py
```diff
@@ -0,0 +1,438 @@
...
+---
+module: bigip_virtual_server
+short_description: "Manages F5 BIG-IP LTM virtual servers"
+description:
+    - "Manages F5 BIG-IP LTM virtual servers via iControl SOAP API"
+version_added: "2.0"
+author: Etienne Carriere
+notes:
+    - "Requires BIG-IP software version >= 11"
+    - "F5 developed module 'bigsuds' required (see http://devcentral.f5.com)"
```
bcoca, line 31: ~~this should be in the 'requirements' documentation field~~
nvmd, just saw it there

gregdek:
@caphrim007 Please review this PR to make sure it adheres to the following guidelines:

network/f5/bigip_virtual_server.py
```diff
@@ -0,0 +1,438 @@
...
+        return updated
+    except bigsuds.OperationFailed, e:
+        raise Exception('Error on setting profiles : %s' % e)
+
+def set_snat(api,name,snat):
+    updated = False
+    try:
+        current_state=get_snat_type(api,name)
+        if snat is None:
+            return update
```
caphrim007: This variable should be "updated" instead of "update"

network/f5/bigip_virtual_server.py
```diff
@@ -0,0 +1,438 @@
...
+  - name: Add VS
+    local_action:
+        module: bigip_virtual_server
+        server: lb.mydomain.net
+        user: admin
+        password: secret
+        state: present
+        partition: MyPartition
+        name: myvirtualserver
+        destination: "{{ ansible_default_ipv4["address"] }}"
```
caphrim007: Just a nitpick that the double quotes around "address" here should be single quotes so that a copy/paste of the example works.

network/f5/bigip_virtual_server.py
```diff
@@ -0,0 +1,438 @@
...
+    # about it!
+    try:
+        api.LocalLB.VirtualServer.create(
+            definitions = [{'name': [name], 'address': [destination], 'port': port, 'protocol': 'PROTOCOL_TCP'}],
+            wildmasks = ['255.255.255.255'],
+            resources = [{'type': 'RESOURCE_TYPE_POOL', 'default_pool_name': pool}],
+            profiles = _profiles)
+        created = True
+        return created
+    except bigsudsOperationFailed, e :
```
caphrim007: This needs to be "bigsuds.OperationFailed"

network/f5/bigip_virtual_server.py
```diff
@@ -0,0 +1,438 @@
...
+        destination: "{{ ansible_default_ipv4["address"] }}"
+        port: 443
+        pool: "{{ mypool }}"
+        snat: Automap
+        description: Test Virtual Server
+        all_profiles:
+            - http
+            - clientssl
+
+  - name: Modify Port of the Virtual Server
```
caphrim007, line 138: Does this example require a "destination" to work correctly? Line 271 seems to require a destination. If I use the module without a destination, then the port is not changed.

network/f5/bigip_virtual_server.py
```diff
@@ -0,0 +1,438 @@
...
+        required: false
+        default: 'Common'
+    name:
+        description:
+            - "Virtual server name."
+        required: true
+        aliases: ['vs']
+    destination:
+        description:
+            - "Destination IP of the virtual server (only host is currently supported) . Required when state=present and vs does not exist. Error when state=absent."
```
caphrim007: I do not get an error when I set the state=absent and include this parameter. I think just the documentation need be updated in this case.

network/f5/bigip_virtual_server.py
```diff
@@ -0,0 +1,438 @@
...
+                result = {'changed': True}
+
+    except Exception, e:
+        module.fail_json(msg="received exception: %s" % e)
+
+    module.exit_json(**result)
+# import module snippets
+from ansible.module_utils.basic import *
+from ansible.module_utils.f5 import *
+main()
```
caphrim007: Please include the module execution in a conditional per the checklist http://docs.ansible.com/ansible/developing_modules.html#module-checklist

```
if __name__ == '__main__':
    main()
```

caphrim007:
needs_revision

@gregdek I've tested this module against 11.6.0 version of BIG-IP. Some changes need to be made which I've noted in comments. @Etienne-Carriere I'd be happy to test again with an updated PR. Thanks!

gregdek:
Thanks @caphrim007 for your testing and review.

@Etienne-Carriere Please make the suggested revisions, and when you're done, please add a comment with the text "ready_for_review" and we will continue with the review process.  Thanks!

network/f5/bigip_virtual_server.py
Etienne-Carriere: Corrected the documentation . It is right it is not an issue to have destination with state=absent . It is just ignored.

network/f5/bigip_virtual_server.py
Etienne-Carriere, line 138: I changed the code so that when the VS is updated : 
- If destination is present and port absent => destination is updated
- if port is present and destination is absent => port is updated
- if both are present => both are updated

It is more logic

Etienne-Carriere:
I made the corrections . It is now "ready_for_review" .

gregdek:
Thanks @Etienne-Carriere -- @caphrim007 back to you for review!

network/f5/bigip_virtual_server.py
```diff
@@ -0,0 +1,454 @@
...
+                        result['changed']|=set_profiles(api,name,all_profiles)
+                        result['changed']|=set_default_persistence_profiles(api,name,default_persistence_profile)
+                        api.System.Session.submit_transaction()
+                    except Exception,e:
+                        raise Exception("Error on updating Virtual Server : %s" % e)
+                else:
+                    # check-mode return value
+                    result = {'changed': True}
+
+        elif state in ('disabled', 'enabled'):
```
caphrim007: Doesn't look like these states do anything. Do you want to add them? If not, we can remove them for now and patch them in later

network/f5/bigip_virtual_server.py
```diff
@@ -0,0 +1,454 @@
...
+def set_snat(api,name,snat):
+    updated = False
+    try:
+        current_state=get_snat_type(api,name)
+        if snat is None:
+            return updated
+        if snat == 'None' and current_state != 'SRC_TRANS_NONE':
+            api.LocalLB.VirtualServer.set_source_address_translation_none(virtual_servers = [name])
+            updated = True
+        if snat == 'Automap' and current_state != 'SRC_TRANS_AUTOMAP':
```
caphrim007: This and the above `if` statement can be changed to `if...elif...elif...`

network/f5/bigip_virtual_server.py
```diff
@@ -0,0 +1,454 @@
...
+                    # about it!
+                    try:
+                        vs_create(api,name,destination,port,pool)
+                        set_profiles(api,name,all_profiles)
+                        set_snat(api,name,snat)
+                        set_description(api,name,description)
+                        set_default_persistence_profiles(api,name,default_persistence_profile)
+                        result = {'changed': True}
+                    except bigsuds.OperationFailed, e:
+                            raise Exception('Error on creating Virtual Server : %s' % e)
```
caphrim007: Can we space this over so it is aligned under the exception like the rest of the code?

network/f5/bigip_virtual_server.py
```diff
@@ -0,0 +1,454 @@
...
+
+    if 1 > port > 65535:
+        module.fail_json(msg="valid ports must be in range 1 - 65535")
+  
+    try:
+        api = bigip_api(server, user, password)
+        result = {'changed': False}  # default
+
+        if state == 'absent':
+                if not module.check_mode:
```
caphrim007: Can this block (to line 387) be spaced over to conform with the rest of the code?

caphrim007:
needs_review

@Etienne-Carriere @gregdek Finally had a chance to look over this and a couple of small things. After addressing inline comments it should be ready to go.

network/f5/bigip_virtual_server.py
Etienne-Carriere: Finally, I added the code for enabled/disabled state

Etienne-Carriere:
I committed changes according to @caphrim007 review .

"ready_for_review"

caphrim007:
@Etienne-Carriere looks good! Thanks for your work on this!

works_for_me
passes_guidelines
shipit

gregdek:
Since @caphrim007 is the most knowledgeable F5 guy out there, I think his "shipit" is sufficient.

Thanks @Etienne-Carriere for the new module!

abadger:
Merged to devel.  Thanks for all the work guys!  New modules should head for the next major release (2.1.x).  The 2.0.x process dragged on for much longer than planned (major rewrite for a lot of the core code so in hindsight it's understandable but it's also very unfortunate) but 2.1.x should be back to a shorter release cycle so it won't take forever to get this module into a release.

caphrim007:
Thanks @abadger! Release cycle isn't a problem. I'm guilty of letting some of this stuff slip too. I will be addressing that going forward so that its less of a manual process on my end. Pulled in a number of directions, but it'll just get better going forward. Thanks!

PR #583, opened 2015-06-04
Ovirt add tagging

joshainglis:
Added the ability to tag VMs using the oVirt module
This pull is dependent on pull #581

PR #581, opened 2015-06-04
Added auth key insertion to ovirt module

joshainglis:
Added functionality to ovirt module to allow insertion of authorised key on vm instantiation.
Minor formatting changes where code has been touched.

PR #388, opened 2015-04-12
Refactor of the Ovirt module to remove duplicate code and cleanup

Firstyear:
I am cleaning up the ovirt module before I work to add support for new features. This change includes:
- Move functions into class.
- Fix up various components that relied on these functions.
- Remove duplicate code in create_vm.
- Refactor create_vm.

PR #589, opened 2015-06-06
make ovirt module idempotent (optionally)

joshainglis:
- The ovirt module can now be idempotent (using `async=no`)
- Removed all `sys.exit` calls and replaced with `fail_json`
- Removed all `print` statements
- Returns instance ip if it is available
- Builds on #583

### DIFF
--- a/cloud/misc/ovirt.py
+++ b/cloud/misc/ovirt.py
@@ -160,7 +160,7 @@ options:
   authorised_key_user:
     description:
      - insert user key into authorised keys on instantiation
-    default: null
+    default: 'root'
     required: false
     aliases: []
 
@@ -202,6 +202,7 @@ action: ovirt >
     disk_int=virtio
     authorised_key_file=/home/user/.ssh/id_rsa.pub
     authorised_key_user=root
+    tags=test,ansible,jenkins
 
 # stopping an instance
 action: ovirt >
@@ -281,9 +282,38 @@ def get_cloud_init(auth_key, auth_key_user):
             sys.exit(1)
     return cloud_init
 
+
+def add_tags(conn, tags):
+    """
+
+    :type conn: ovirtsdk.api.API
+    """
+    new_tags = {tag.strip().lower() for tag in tags.strip().replace(" ", "_").split(',')}
+
+    try:
+        existing_tags = {tag.get_name() for tag in conn.tags.list()}
+    except Exception:
+        print ("Could not get existing tags from server")
+        sys.exit(1)
+    else:
+        tags_to_add = []
+        for tag in new_tags:
+            if tag in existing_tags:
+                tags_to_add.append(conn.tags.get(name=tag))
+            else:
+                try:
+                    tags_to_add.append(conn.tags.add(params.Tag(name=tag)))
+                except:
+                    print("Failed to add tag to ovirt")
+                    sys.exit(1)
+        return params.Tags(tag=tags_to_add)
+
+
+
+
 # ------------------------------------------------------------------- #
 # Create VM from scratch
-def create_vm(conn, vmtype, vmname, zone, vmdisk_size, vmcpus, vmnic, vmnetwork, vmmem, vmdisk_alloc, sdomain, vmcores, vmos, vmdisk_int, auth_key, auth_key_user):
+def create_vm(conn, vmtype, vmname, zone, vmdisk_size, vmcpus, vmnic, vmnetwork, vmmem, vmdisk_alloc, sdomain, vmcores, vmos, vmdisk_int, auth_key, auth_key_user, tags):
     if vmdisk_alloc == 'thin':
         # define VM params
         vmparams = params.VM(
@@ -295,6 +325,7 @@ def create_vm(conn, vmtype, vmname, zone, vmdisk_size, vmcpus, vmnic, vmnetwork,
             cpu=params.CPU(topology=params.CpuTopology(cores=int(vmcores))),
             type_=vmtype,
             initialization=get_cloud_init(auth_key, auth_key_user),
+            tags=add_tags(conn, tags)
         )
 
         # define disk params
@@ -305,7 +336,7 @@ def create_vm(conn, vmtype, vmname, zone, vmdisk_size, vmcpus, vmnic, vmnetwork,
             interface=vmdisk_int,
             type_="System",
             format='cow',
-            storage_domains=params.StorageDomains(storage_domain=[conn.storagedomains.get(name=sdomain)])
+            storage_domains=params.StorageDomains(storage_domain=[conn.storagedomains.get(name=sdomain)]),
         )
         # define network parameters
         network_net = params.Network(name=vmnetwork)
@@ -321,6 +352,7 @@ def create_vm(conn, vmtype, vmname, zone, vmdisk_size, vmcpus, vmnic, vmnetwork,
             cpu=params.CPU(topology=params.CpuTopology(cores=int(vmcores))),
             type_=vmtype,
             initialization=get_cloud_init(auth_key, auth_key_user),
+            tags=add_tags(conn, tags)
         )
 
         # define disk params
@@ -358,13 +390,14 @@ def create_vm(conn, vmtype, vmname, zone, vmdisk_size, vmcpus, vmnic, vmnetwork,
 
 
 # create an instance from a template
-def create_vm_template(conn, vmname, image, zone, auth_key, auth_key_user):
+def create_vm_template(conn, vmname, image, zone, auth_key, auth_key_user, tags):
     vmparams = params.VM(
         name=vmname,
         cluster=conn.clusters.get(name=zone),
         template=conn.templates.get(name=image),
         disks=params.Disks(clone=True),
         initialization=get_cloud_init(auth_key, auth_key_user),
+        tags=add_tags(conn, tags)
     )
     try:
         conn.vms.add(vmparams)
@@ -450,7 +483,8 @@ def main():
             sdomain             = dict(),
             region              = dict(),
             authorised_key_file = dict(),
-            authorised_key_user = dict(default='root')
+            authorised_key_user = dict(default='root'),
+            tags                = dict()
         ),
     )
 
@@ -476,6 +510,7 @@ def main():
     region        = module.params['region']              # oVirt Datacenter
     auth_key_file = module.params['authorised_key_file'] # Instantiate with given pubkey in authorised hosts
     auth_key_user = module.params['authorised_key_user'] # user account associated with the auth_key
+    tags          = module.params['tags']                # tags to be associated with the VM
 
     #initialize connection
     c = conn(url+"/api", user, password)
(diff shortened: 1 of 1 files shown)


joshainglis:
I've added idempotency, but left it defaulted to off so as not to break backwards compatibility.

gregdek:
Pretty hefty patch! @vincentvdk can you review?

gregdek:
Also kinda looks like, combined with #581 and #583, this is nearly a complete rewrite of this module... wondering if it should be reviewed as a new module instead of as a set of PRs to this module.

vincentvdk:
Hi, 
That are some nice additions. To me it looks like this can be merged into the existing module. Not sure why the idempotency was added as an option. The initial version of this module has always been idempotent as far as I know.

joshainglis:
Hi @vincentvdk,
Just wanted to clarify the idempotency thing.
The issue I was trying to solve was centred on being able to easily bring up a new VM (or a large group of them), then provision.
Say I wanted to create a new VM then provision it. Previously the process would have involved:
- Creating VM (using `state='present'`)
- Writing a `do-until` loop to wait for the VM to reach the `down` state (and I was not aware of a clear way of doing this from a playbook level, as trying to start the VM prior to it reaching the `down` state throws an error)
- Starting the VM (using `state='started'`)
- Writing a `do-until` loop to wait for the VM to publish its IP addresses (and I couldn't see a way of retrieving this info from the module)

_(please correct me if the above process is wrong)_

What I wanted to be able to do is just create a new VM, start it, and return its IP in one fell swoop. So now we create a VM with `state='started' async='no' wait_for_ip='yes'`, and register the returned value (which we can then add to hosts and provision).
And if we want to minimise wait time to create lots of VMs, it's a 3 step process:
- Create the VMs with `state='present' async='yes'`
- Start the VMs with `state='started' async='yes'` (this will still wait for VMs to reach the `'down'` state before sending the `start` command, but wont wait for them to actually start)
- Wait for the VMs to actually start, then the IPs to be published, and register them with `state='started' async='no' wait_for_ip='yes'`

Sorry for any confusion, and I hope this clarifies my meaning.

vincentvdk:
@joshainglis That makes sense. Looks good for me.

gregdek:
Need to change the shebang back to #!/usr/bin/python, per module guidelines here:

joshainglis:
@vincentvdk: Thanks!
@gregdek: shebang line fixed

Firstyear:
Hi,

I have a really similar pull request here:

https://github.com/ansible/ansible-modules-extras/pull/388

I think I cover some of the same group, but I also:
- Rewrite to a class instead of functions
- Add support for fact gathering from machines
- Add support for cloud init scripting
- Add support for blocking on actions until VM's reach a certain state.

I'm wondering how we can merge our work since I think we are both trying to achieve many of the same goals.

joshainglis:
Hi @Firstyear,
It does indeed look like we're aiming for very similar goals!
- I really like your idea of making it a class
- Looks like we're both gathering facts (just a different set of facts)
  - We can definitely union these
- I like your implementation of cloud-init (using params.Action)
  - I think we should be able to union our two use-cases here, too
- We have different implementations of waiting for state. (recursion vs while-loop)
  - We can further discuss which is better suited
- We can bring in tags from my implementation
- Would be good if we can do a `create-->cloud-init-->boot-->wait for ip-->gather facts` in one hit

Look forward to working with you :)

Firstyear:
Since I have the class structure there do we want to work on my branch then? Probably least resistance.

What I'll do is later today I'll:
- Look at what facts you gather and union in my collect facts module. I'm certainly willing to add as many facts as we want to gather!
- I'll put your version of waiting for state in too so we can compare them. I'm certainly not tied to my version of it at all.
- Bring in tags, because that would be a good to have.
- Add a state "wait for ip", which is a combination of up, and facts.

The only concern I have with "create-->cloud-init-->boot-->wait for ip-->gather facts" is that it relies on there being a reboot stanza in the cloud init script, or hardcoding one at the tail of the cloud init section. Some people may not want that, which is why in my systems I have this setup as create --> cloud-init --> wait for down. My custom script provides the shutdown section.

This way my work flow is:
Check vm exists. 
- If doesn't exist -> deploy template, cloud init and wait for down

Check vm power state
- if not running -> poweron

But it would also mean you could create:
Check vm exists:
- If doesn't exist, deploy from template

Check vm power
- if not running -> cloud init, wait for up

Which would let you cloud init and boot every time rather than just using state=up. Could be useful in some use cases. So you can create more flexible work flows with the cloud init action not being tied to wait for ip, but wait for any state.

Finally, there are still some issues with create from template where it doesn't respect the CPU and Memory parameters. That should probably be solved to have this "really polished".

Thoughts?

joshainglis:
@Firstyear 
Sorry, missed your above message before I jumped into it :/. Hope we didn't overlap!

Firstyear:
Let me look at what you have and we'll see :)

wimnat:
Just a comment from my AWS experience - if you're gathering facts, should that be a separate module, similar to ec2_facts module?

Firstyear:
Okay, so looking at your work, I have a few comments:
- With waiting for state, have a look what I've done. Instead of having the wait for state in each actions function, there is a generic "wait for state" at the end that is triggered if desired. That way the action function and the wait for state code are smaller and "correct". If you want wait for IP, you can always just poll that one fact of the VM, and then have gather_facts trigger after wait_for state is complete. Additionally, you only need to wait for state if changed == true. 
- Can you explain the benefit of the OVIRT_STATE_MAP?
- My cloud init takes a custom script from variables, which means you can do variable substitution and all kinds of fun. Plus you don't need to hardcode filepaths. Consider I launch two machines at once. In this case /tmp/_vmcustomscript will be clobbered for one or the other machine potentially. 
- In cloud init, authorized key should be optional, not mandatory. There are plenty of use cases where it isn't wanted. 
- Create_vm the logic can be condensed because the difference between thin and thick for the disk isn't that much. Again, take a look at my create_vm function
- create_vm_template in mine can alter cpu and memory limits of the machine. Again, take a look there
- in get_instance_data we shoud prefix all ovirt facts with ovirt_guest_ . 
- module.params(async) isn't needed: If there is no "wait for state" then you imply async. Makes for a simpler state check.

I'm happy to make these changes myself unless you want to jump in and go for it yourself? Your call.

@wimnat No reason it couldn't be seperate, no. Probably would keep the interface a little cleaner.

wimnat:
@Firstyear my personal preference would be a separate module. Nothing worse that trying to break down a monolithic module down the track

Firstyear:
I was agreeing it should be separate to keep things cleaner and to help reduce playbook option sprawl.

gregdek:
Hmmmm.

@vincentvdk approved a patch 3 days ago, which I did not merge in a timely fashion -- and since then, we've added a bunch of stuff to this PR.

So. I guess I'll ask @vincentvdk to take a quick look at the subsequent changes. Until then, let's not make any more patches to this PR, so we can settle it down.

Firstyear:
If we let it settle and merge it in, I'll apply my changes in a new PR after then. Does that sound reasonable?

joshainglis:
@Firstyear 
- **Wait for state**: Good point! My reasoning for doing it this way is to ensure that the VM is in the correct state to receive a command prior to sending it. Think of it as preconditions, postconditions and error conditions. For example: If you had just created a VM, it's state goes to `'image_locked'` (which maps to `'creating'`), and if you send the `vm.start()` command while it is in that state, it will generate an error (as the precondition for the start command is that the VM is in the `'down'` state). So instead, the play will wait for the precondition before sending the command. Then we make the postcondition optional with the `async` option (I'm not sold that `async` is necessarily the best term here, I think I like your `'wait_for_state'` better). Where the postcondition is the expected state after the command. And if the VM goes in to an error condition at any point, we can `fail_json` out as required. It is a bit more cumbersome, but IMHO stronger in terms of idempotency. As in, if you declare that you want a VM in a 'started' state, then the module should find the correct path for it to reach that state. If it doesn't exist: Create it. If it's in some transition state, wait for it to reach a stable state. If it's stopped: start it, etc.
- **`OVIRT_STATE_MAP`**: It's essentially there the compress the number of states that need to be handled in terms of the (pre|post|error) conditions. It maps all of the possible states that ovirt provides into a smaller set of (effectively equivalent) states.
- **cloud-init**: Could you further explain what you mean here? AFAIK I have merged our two cloud-init implementations, just adding the condition that the content for customscript comes from a file rather than a string. Happy to put string back in, though. Or is this not what you're referring to?
- **cloud-init key**: Agreed!
- **create_vm**: Agreed!
- **create_template**: I was under the impression (from the comments in your code) the these did not work as expected? Happy to add, though :)
- **get_instance_data**: Sounds like we should move this to a `_facts` module. But yes, also agree with this.
- **async**: Yeah, we have different names representing the same thing here (though inverse). Happy to go with your `'wait_for_state'` as I think it is clearer.

Also, I've created a new branch we can work on: https://github.com/joshainglis/ansible-modules-extras/tree/ovirt-merge-with-firstyear so that we can let this one settle.

joshainglis:
@Firstyear 
Just noticed your `'wait_for_state'` isn't boolean. I don't know that I agree with this method. That requires the user give redundant info IMHO. Unless I'm missing something? What would be an instance where `'wait_for_state'` would be anything other than waiting for the state implied by `state=`?

Firstyear:
I'll open a new PR on your new branch, and we can discuss there.

vincentvdk:
Hi, 
What is all this activity :-) Anyhow, I had a brief look at the changes and looks fine for me except one thing. In  0cc761b the zone is renamed to cluster. Is the "zone" option also added as an option to the params (could really find this change)? Otherwise people's playbooks might break.

bcoca:
options can be renamed, but you should at least keep the old name as an alias to prevent playbooks from breaking

vincentvdk:
@bcoca That's actually what i meant. thanks for making it more clear

Firstyear:
Hi @joshainglis ,

Can we fix up the option names / aliases so that we can merge this work?

Firstyear:
We should probably close this, given I have a new request from Josh and My work at https://github.com/ansible/ansible-modules-extras/pull/707 .

lpenz:
I've tested this, and there is an issue: authorized_key_user_groups is used in the code, even though it is not declared as a available parameter

cloud/misc/ovirt.py
```diff
@@ -199,231 +285,944 @@
...
+                authorized_key=[
+                    params.AuthorizedKey(
+                        user=params.User(user_name=authorized_key_user),
+                        key=key
+                    ),
+                ]
+            )
+
+        cloud_init = params.CloudInit(
+            host=instance_host_name,
```
lpenz, line 584: You have to create a host instance with the attribute "address" for that, like so:

``` python
        if instance_host_name is not None:
            host = params.Host(
                address=instance_host_name,
            )
```

(...)

``` python
        cloud_init = params.CloudInit(
            host=host,
            files=files,
```

cloud/misc/ovirt.py
```diff
@@ -199,231 +285,944 @@
...
+                        try:
+                            self.conn.tags.add(params.Tag(name=tag))
+                        except:
+                            self.module.fail_json(msg=u"Failed to add tag '{}' to ovirt".format(tag))
+                    try:
+                        infra_tag = self.conn.tags.get(name=tag)
+                        tags_to_add.append(infra_tag)
+                    except:
+                        self.module.fail_json(msg=u"Failed to get tag '{}' from ovirt".format(tag))
+                return tags_to_add
```
lpenz, line 626: You should `return []` outside the if, to cover the case where `tags = ''` (the default, that evaluates to False in python)

cloud/misc/ovirt.py
```diff
@@ -199,231 +285,944 @@
...
+                type='str',
+            ),
+            authorized_key_file=dict(
+                type='str',
+            ),
+            authorized_key_user=dict(
+                type='str',
+                default='root'
+            ),
+            tags=dict(
```
lpenz, line 1086: Missing here:

``` python
            authorized_key_user_groups=dict(
                type='str',
                default=None
            ),
```

robynbergeron:
Closing this issue as #707 is a new improved version of this PR, as requested by @Firstyear who is working with @joshainglis.  (If I'm interpreting this wrong, let me know and we can reopen it :D)

@lpenz -- you may want to add your notes about the authorized_key_user_groups issue to #707, though it looks like #978 (same issue, reported by @nutter007 ) is referenced in #707.

Thanks, everyone! :)

PR #591, opened 2015-06-08
Adds handler for error responses

everson:
This handles error messages when they came in the api response. e.g. malformed query.

/cc @skornehl


everson:
Closing this since there are other changes that I want to propose together.

PR #593, opened 2015-06-09
Bug fixes on error handling and update handling

everson:
- Adds handler for error responses
- Fixes the bug where it was using only the keys to determine whether a change was made, i.e. values changes for existing keys was reported incorrectly.


everson:
hey @skornehl, please take a look at this.

skornehl:
Looks good, thanks for this!

PR #594, opened 2015-06-09
Adding support for setting the topic of a channel

davidsiefert:
I currently use IRC to notify a channel of progress in a playbook, and I would like to sometimes set the topic of the channel as updates progress.  This change allows one to optionally set a "topic" argument to do so.

### DIFF
--- a/notification/irc.py
+++ b/notification/irc.py
@@ -47,6 +47,12 @@ options:
       - The message body.
     required: true
     default: null
+  topic:
+    description:
+      - Set the channel topic
+    required: false
+    default: null
+    version_added: 2.0
   color:
     description:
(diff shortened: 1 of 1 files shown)


notification/irc.py
```diff
@@ -47,6 +47,11 @@
       - The message body.
     required: true
     default: null
+  topic:
```
bcoca, line 50: missing version added

davidsiefert, line 50: What number would I put here? 1.9?

davidsiefert, line 50: Ah, based on recent commit history it looks like it should be 2.0

davidsiefert:
Hey @bcoca added "version-added" to my topic parameter--I'm hoping this is what you were looking for.  Thanks!

PR #597, opened 2015-06-11
Feature/ec2 vpc route table

wimnat:
Module by @erydo rebased to latest ansible-modules-extras

For review by @defionscode 

Obsoletes PR https://github.com/ansible/ansible-modules-core/pull/322

### DIFF
--- a/cloud/amazon/ec2_vpc_route_table.py
+++ b/cloud/amazon/ec2_vpc_route_table.py
@@ -1,121 +1,75 @@
 #!/usr/bin/python
-# This file is part of Ansible
 #
-# Ansible is free software: you can redistribute it and/or modify
+# This is a free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
 # the Free Software Foundation, either version 3 of the License, or
 # (at your option) any later version.
 #
-# Ansible is distributed in the hope that it will be useful,
+# This Ansible library is distributed in the hope that it will be useful,
 # but WITHOUT ANY WARRANTY; without even the implied warranty of
 # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 # GNU General Public License for more details.
 #
 # You should have received a copy of the GNU General Public License
-# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+# along with this library.  If not, see <http://www.gnu.org/licenses/>.
 
 DOCUMENTATION = '''
 ---
 module: ec2_vpc_route_table
-short_description: Configure route tables for AWS virtual private clouds
+short_description: Manage route tables for AWS virtual private clouds
 description:
-    - Create or removes route tables from AWS virtual private clouds.'''
-'''This module has a dependency on python-boto.
-version_added: "1.8"
+    - Manage route tables for AWS virtual private clouds
+version_added: "2.0"
+author: Robert Estelle, @erydo
 options:
   vpc_id:
     description:
-      - "The VPC in which to create the route table."
+      - VPC ID of the VPC in which to create the route table.
     required: true
   route_table_id:
     description:
-      - "The ID of the route table to update or delete."
+      - The ID of the route table to update or delete.
     required: false
     default: null
   resource_tags:
     description:
-      - 'A dictionary array of resource tags of the form: { tag1: value1,'''
-''' tag2: value2 }. Tags in this list are used to uniquely identify route'''
-''' tables within a VPC when the route_table_id is not supplied.
+      - A dictionary array of resource tags of the form: { tag1: value1, tag2: value2 }. Tags in this list are used to uniquely identify route tables within a VPC when the route_table_id is not supplied.
     required: false
     default: null
-    aliases: []
-    version_added: "1.6"
   routes:
     description:
-      - List of routes in the route table. Routes are specified'''
-''' as dicts containing the keys 'dest' and one of 'gateway_id','''
-''' 'instance_id', 'interface_id', or 'vpc_peering_connection'. '''
-''' If 'gateway_id' is specified, you can refer to the VPC's IGW '''
-''' by using the value "igw".
+      - List of routes in the route table. Routes are specified as dicts containing the keys 'dest' and one of 'gateway_id', 'instance_id', 'interface_id', or 'vpc_peering_connection'. If 'gateway_id' is specified, you can refer to the VPC's IGW by using the value 'igw'.
     required: true
     aliases: []
   subnets:
     description:
-      - An array of subnets to add to this route table. Subnets may either'''
-''' be specified by subnet ID, Name tag, or by a CIDR such as '10.0.0.0/24'.
+      - An array of subnets to add to this route table. Subnets may be specified by either subnet ID, Name tag, or by a CIDR such as '10.0.0.0/24'.
     required: true
-    aliases: []
   propagating_vgw_ids:
     description:
-      - Enables route propagation from virtual gateways specified by ID.
+      - Enable route propagation from virtual gateways specified by ID.
     required: false
-    aliases: []
   wait:
     description:
-      - wait for the VPC to be in state 'available' before returning
+      - Wait for the VPC to be in state 'available' before returning.
     required: false
     default: "no"
     choices: [ "yes", "no" ]
-    aliases: []
   wait_timeout:
     description:
-      - how long before wait gives up, in seconds
+      - How long before wait gives up, in seconds.
     default: 300
-    aliases: []
   state:
     description:
-      - Create or terminate the VPC
-    required: true
+      - Create or destroy the VPC route table
+    required: false
     default: present
-    aliases: []
-  region:
-    description:
-      - region in which the resource exists.
-    required: false
-    default: null
-    aliases: ['aws_region', 'ec2_region']
-  aws_secret_key:
-    description:
-      - AWS secret key. If not set then the value of the AWS_SECRET_KEY'''
-''' environment variable is used.
-    required: false
-    default: None
-    aliases: ['ec2_secret_key', 'secret_key' ]
-  aws_access_key:
-    description:
-      - AWS access key. If not set then the value of the AWS_ACCESS_KEY'''
-''' environment variable is used.
-    required: false
-    default: None
-    aliases: ['ec2_access_key', 'access_key' ]
-  validate_certs:
-    description:
-      - When set to "no", SSL certificates will not be validated for boto'''
-''' versions >= 2.6.0.
-    required: false
-    default: "yes"
-    choices: ["yes", "no"]
-    aliases: []
-    version_added: "1.5"
-
-requirements: [ "boto" ]
-author: Robert Estelle
+    choices: [ 'present', 'absent' ]
+extends_documentation_fragment: aws
 '''
 
 EXAMPLES = '''
-# Note: None of these examples set aws_access_key, aws_secret_key, or region.
-# It is assumed that their matching environment variables are set.
+# Note: These examples do not set authentication details, see the AWS Guide for details.
 
 # Basic creation example:
 - name: Set up public subnet route table
@@ -583,3 +537,4 @@ from ansible.module_utils.ec2 import *  # noqa
 
 if __name__ == '__main__':
     main()
+    
--- a/cloud/amazon/ec2_vpc_route_table.py
+++ b/cloud/amazon/ec2_vpc_route_table.py
@@ -369,17 +369,22 @@ def ensure_subnet_associations(vpc_conn, vpc_id, route_table, subnets,
     return {'changed': changed}
 
 
-def ensure_propagation(vpc_conn, route_table_id, propagating_vgw_ids,
+def ensure_propagation(vpc_conn, route_table, propagating_vgw_ids,
                        check_mode):
 
-    # NOTE: As of boto==2.15.0, it is not yet possible to query the existing
-    # propagating gateways. However, EC2 does support this as evidenced by
-    # the describe-route-tables tool. For now, just enable the given VGWs
-    # and do not disable any others.
+    # NOTE: As of boto==2.38.0, it is not yet possible to query the existing
+    # propagating gateways. However, EC2 does support this as shown in its API
+    # documentation. For now, a reasonable proxy for this is the presence of
+    # propagated routes using the gateway in the route table. If such a route
+    # is found, propagation is almost certainly enabled.
     changed = False
     for vgw_id in propagating_vgw_ids:
+        for r in list(route_table.routes):
+            if r.gateway_id == vgw_id:
+                return {'changed': False}
+
         changed = True
-        vpc_conn.enable_vgw_route_propagation(route_table_id,
+        vpc_conn.enable_vgw_route_propagation(route_table.id,
                                               vgw_id,
                                               dry_run=check_mode)
 
@@ -430,7 +435,7 @@ def ensure_route_table_present(vpc_conn, vpc_id, route_table_id, resource_tags,
             )
 
     if propagating_vgw_ids is not None:
-        result = ensure_propagation(vpc_conn, route_table.id,
+        result = ensure_propagation(vpc_conn, route_table,
                                     propagating_vgw_ids,
                                     check_mode=check_mode)
         changed = changed or result['changed']
--- a/cloud/amazon/ec2_vpc_route_table.py
+++ b/cloud/amazon/ec2_vpc_route_table.py
@@ -50,8 +50,8 @@ options:
     aliases: []
   subnets:
     description:
-      - An array of subnets to add to this route table. Subnets may either be'''
-''' specified by subnet ID or by a CIDR such as '10.0.0.0/24'.
+      - An array of subnets to add to this route table. Subnets may either'''
+''' be specified by subnet ID or by a CIDR such as '10.0.0.0/24'.
     required: true
     aliases: []
   wait:
--- a/cloud/amazon/ec2_vpc_route_table.py
+++ b/cloud/amazon/ec2_vpc_route_table.py
@@ -373,7 +373,7 @@ def ensure_propagation(vpc_conn, route_table_id, propagating_vgw_ids,
             changed = True
             vpc_conn.enable_vgw_route_propagation(route_table_id,
                                                   vgw_id,
-                                                  test_run=check_mode)
+                                                  dry_run=check_mode)
 
     return {'changed': changed}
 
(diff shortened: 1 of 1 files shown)


cloud/amazon/ec2_vpc_route_table.py
```diff
@@ -0,0 +1,540 @@
...
+        'state': {'choices': ['present', 'absent'], 'default': 'present'},
+    })
+    module = AnsibleModule(
+        argument_spec=argument_spec,
+        supports_check_mode=True,
+    )
+    if not HAS_BOTO:
+        module.fail_json(msg='boto is required for this module')
+
+    ec2_url, aws_access_key, aws_secret_key, region = get_ec2_creds(module)
```
willthames: Recommend using get_aws_connection_info and then connect_to_aws(boto.vpc,...)

As that handles a lot of things like e.g security tokens, boto profiles etc, so that you don't have to.

cloud/amazon/ec2_vpc_route_table.py
```diff
@@ -0,0 +1,538 @@
...
+      - VPC ID of the VPC in which to create the route table.
+    required: true
+  route_table_id:
+    description:
+      - The ID of the route table to update or delete.
+    required: false
+    default: null
+  resource_tags:
+    description:
+      - A dictionary array of resource tags of the form: { tag1: value1, tag2: value2 }. Tags in this list are used to uniquely identify route tables within a VPC when the route_table_id is not supplied.
```
juliedavila: This should be quoted or ansible-doc is unable to parse the curly brackets

juliedavila:
After creating a route table with:

``` yaml
    - name: Create subnet
      ec2_vpc_route_table:
        vpc_id: vpc-efbcb08a
        region: us-east-1
        routes:
         - dest: 0.0.0.0/0
           gateway_id: igw
        subnets:
          - subnet-f1e17bda
        resource_tags:
          Environment: Testing
      register: new_subnet
```

Then removing it by specifying `state: absent`, I get the following traceback:

``` shell
<localhost> REMOTE_MODULE ec2_vpc_route_table vpc_id=vpc-efbcb08a state=absent region=us-east-1
failed: [localhost] => {"failed": true, "parsed": false}
  File "/Users/jondavila/.ansible/tmp/ansible-tmp-1435087657.06-111020801215805/ec2_vpc_route_table", line 2341, in <module>
    main()
  File "/Users/jondavila/.ansible/tmp/ansible-tmp-1435087657.06-111020801215805/ec2_vpc_route_table", line 526, in main
    module.check_mode
  File "/Users/jondavila/.ansible/tmp/ansible-tmp-1435087657.06-111020801215805/ec2_vpc_route_table", line 394, in ensure_route_table_absent
    vpc_conn.delete_route_table(route_table.id, dry_run=check_mode)
  File "/usr/local/lib/python2.7/site-packages/boto/vpc/__init__.py", line 307, in delete_route_table
    return self.get_status('DeleteRouteTable', params)
  File "/usr/local/lib/python2.7/site-packages/boto/connection.py", line 1223, in get_status
    raise self.ResponseError(response.status, response.reason, body)
boto.exception.EC2ResponseError: EC2ResponseError: 400 Bad Request
<?xml version="1.0" encoding="UTF-8"?>
<Response><Errors><Error><Code>DependencyViolation</Code><Message>The routeTable 'rtb-32b85056' has dependencies and cannot be deleted.</Message></Error></Errors><RequestID>be9f2658-a9ab-4db8-bf9a-4a415d47e2da</RequestID></Response>
```

This exception should be captured. Perhaps the module should remove the dependencies automatically.

juliedavila:
@wimnat friendly ping, I'm not traveling this week and would love to get this merged in

wimnat:
@defionscode @bcoca what are your thoughts on splitting out route functionality to another module e.g. ec2_vpc_route?

You would use this module to first create the route table and then the other module to add routes.

I don't like the specification of routes in this module as a list of dicts - i would prefer to define module attributes.

juliedavila:
@wimnat my main concern is from a UX/principle of least surprise perspective. My initial thought is that not many would want to create a route table w/o routes; now this argument could of course be applied to our current VPC split efforts (VPCs imply the desire for subnets, etc) but I think the main argument is/was maintainability of the code. The original VPC module has crossed the point (imo) of being unsustainable and becoming too fragile.

In general, I agree that first class module attributes are the _most_ desired type. However, in this particular case I see the split causing additional maintenance overhead with minimal advantage (if any) to the overall UX perspective.

Of course I'm not dogmatic, counter-arguments are welcomed.

wimnat:
@erydo mention as i thought he may have thoughts...

My only counter-argument is the usual argument for using first class module attributes.

I feel that for someone perhaps unfamiliar with VPCs, the fact that a route is defined using a dict and a dict key could be gateway_id OR instance_id OR interface_id OR vpc_peering_connection is a little more confusing that having these arguments defined as mutually exclusive in a separate module with their own documentation description.

Separately, i'm not sure i like this module using tags to look up route tables either.  A call to get_route_table_by_tags returns the first route table it finds that matches the tags.  This may not necessarily be the correct table and changing its routes is going to have pretty serious consequences.  Using only route_table_id seems like the safer option.

juliedavila:
```
I feel that for someone perhaps unfamiliar with VPCs, the fact that a route is defined using a dict and a dict key could be gateway_id OR instance_id OR interface_id OR vpc_peering_connection is a little more confusing that having these arguments defined as mutually exclusive in a separate module with their own documentation description.
```

Yep, I agree. Though, still not sure it's a good enough tradeoff, I'm curious to hear how others feel. 

```
Separately, i'm not sure i like this module using tags to look up route tables either. A call to get_route_table_by_tags returns the first route table it finds that matches the tags. This may not necessarily be the correct table and changing its routes is going to have pretty serious consequences. Using only route_table_id seems like the safer option.
```

Yeah, good point, boto is lacking in many areas, which, on a side note, @wimnat, and everyone else, I'm in the process of building boto3 compatibility into module_utils/ec2.py so AWS modules can start being built in boto3 and older ones can get boto3 functionality added in.

wimnat:
@defionscode code let's wait and see what others think.

@gregdek can you please ping aws reviewers?

gregdek:
@ansible/aws reviewers assemble!

halberom:
I really like these new modules, but I'm struggling with this one.  My use case is a multi-az VPC with an auto scaling NAT group.  Each NAT instance runs a script that associates itself with the relevant route table.

Using the new modules I can
- spin up a vpc
- create 4 subnets in 2 az's (public_0, public_1, private_0 and private_1)
- create an internet gateway

so far so good, now the problem.  I want to create a public route table and 2 private route tables.  However at this point I've not got any instances, so I need the private route tables to be created with only associated subnets

In other words, I need the destination to not be mandatory, as it'll be populated later.

I noticed you've commented above on splitting this out into 2 modules - one to create, and one to assign, - not sure if splitting is necessary.

Thoughts?

halberom:
my current work around is to have an igw task, and then a private routes task (set to igw, overridden later - so I have to use ignore_errors in future runs) - tbh i probably need the task separation anyway so as to distinguish easily with the return results, but having to ignore_errors on subsequent runs is a bit annoying.

however this has led to a bug popping up - with multiple subnets, the tag Name matching doesn't work

```
      - resource_tags:
           Name: "{{ env_stack + '_public' }}"
           Environment: "{{ env_stack }}"
         subnets:
           - "{{ env_stack + '_public_0' }}"
           - "{{ env_stack + '_public_1' }}"
         routes:
           - dest: 0.0.0.0/0
             gateway_id: igw
```

gives

```
TASK [create the igw route table] ***********************************************
{u'msg': u'Multiple subnets named "prod_public_0"', u'failed': True, 'item': {u'routes': [{u'dest': u'0.0.0.0/0', u'gateway_id': u'igw'}], u'resource_tags': {u'Environment': u'prod', u'Name': u'prod_public'}, u'subnets': [u'prod_public_0', u'prod_public_1']}, 'invocation': {'module_name': u'ec2_vpc_route_table', 'module_args': {u'routes': [{u'dest': u'0.0.0.0/0', u'gateway_id': u'igw'}], u'subnets': [u'prod_public_0', u'prod_public_1'], u'region': u'us-east-1', u'vpc_id': u'vpc-efdccb8a', u'resource_tags': {u'Environment': u'prod', u'Name': u'prod_public'}}}}
```

_I don't have 2 subnets with that tag Name_

whereas this works

```
      - resource_tags:
           Name: "{{ env_stack + '_public' }}"
           Environment: "{{ env_stack }}"
         subnets:
           - "10.30.0.0/23"
           - "10.30.2.0/23"
         routes:
           - dest: 0.0.0.0/0
             gateway_id: igw
```

halberom:
would it be possible to add the associated subnet_ids in the return?  not everything takes a name, and without it I'm going to need to figure out a way of mapping my subnet return results to my route return results to get id to id  (maybe a custom filter)

e.g.

```
        "results": [
            {
                "route_table_id": "rtb-1234567", 
                "invocation": {
                  ...
                }, 
                "changed": false, 
                "item": {
                  ...
                    "subnets": [
                       { 
                           name: "prod_private_0", 
                           id: "subnet-xyz"
                       }
                    ]
                }
            }, 
```

gregdek:
Thanks for the feedback @halberom -- still waiting for some official AWS reviewers:

ping @jsmartin @scicoin-project @tombamford @garethr @lorin @jarv @jsdalton @silviud @adq @zbal @zeekin @willthames @lwade @carsongee @defionscode

wimnat:
@halberom thanks for the feedback.  I will try and get my headspace back in to this module but i was kinda waiting for some core feedback regarding the discussion a little further up and also for https://github.com/ansible/ansible-modules-core/pull/1628 to get merged

Do you think the module split would help out in your scenario?  Why/why not?

@gregdek off topic, is the REVIEWERS.md file updated automatically? Just wondered, as I'm missing! :(

juliedavila:
@wimnat given the lack of other opinions on the matter I think I'm against splitting this up further as I don't see any advantage other than being an easier ux for ppl new to aws/ansible with potentially added maintenance overhead.

On the other note, for looking up by tags, have you explored boto3? I currently have a PR to add a boto3 connection feature to module_utils/ec2.py. Perhaps boto3 is a bit more smooth for the lookup stuff, it certainly can coexist with the boto code.

PR for that is: https://github.com/ansible/ansible/pull/11591

halberom:
I don't see any need to split it, but I would like the mandatory destination non-mandatorized (if that's a word) as mentioned above.  This is because I'm going and creating/modifying later.  A re-run now errors with the route already existing but with a different destination.

wimnat:
OK i will work to make necessary changes this week.

First step - new module! I've written a route table facts module so people can easily get route table id if they need to make changes - this allows me to remove the lookup by tag functionality in this module which is currently unreliable (as tags can be duplicated).

Hopefully should have some commits here shortly.  In the mean time, please check out the other module - https://github.com/wimnat/ansible-modules/commit/ce389ecdbfacc1f356e85fa8990dedff8b8384be.

halberom:
@wimnat nice!  I took a quick look at your facts module.   Would be good to get a PR open for it so comments can be associated.  Just a couple of thoughts for it - maybe include the mandatory `region` param in examples.  Also, it would be nice to be able to filter on vpc id.  `get_all_route_tables([],{'vpc-id':'vpc-1234ab56'})`

wimnat:
@halberom PR done - https://github.com/ansible/ansible-modules-extras/pull/778.  Now allows you to use any filter as per the amazon api, including vpc-id.

juliedavila:
@wimnat do you have other commits in the works for this?

wimnat:
@defionscode yep, more commits pending to try and address some of the comments here. Just trying to find the time to finish and test.  I'll try and push to finish by the end of the weekend.

bretmartin:
@wimnat Thank you for all your work on this. I have tested and gotten things working for my route table configuration, after making some changes for propagating gateways. I have that work over at https://github.com/h3biomed/ansible-modules-extras/tree/ec2_vpc_route_table. I created that branch from a later version of ansible:devel though. What's the proper/preferred way for me to provide these changes?

copumpkin:
I have a question on the underlying principles of these convergent modules, and how it applies to this one. It seems like in general, to have any hope of making a module convergent, you need a stable/unique identifier. That's why this one either requires the `route_table_id` (which is guaranteed unique) or `resource_tags` (which we cross our fingers to be unique).

But stable/unique identifiers aren't really enough to make a module "right". You also want predictable identifiers so that you can specify them at creation time and have them stay the same across runs.

In that sense, `resource_tags` is predictable, but only stable/unique if nobody touches it. `route_table_id` is unique/stable, but completely unpredictable. We don't have any other desirable identifiers of a route table, so are we screwed?

Here's one thought: identify the route table by the subnet it's associated with, and don't allow people to create unassociated route tables (or if you do, make that the less commonly use codepath). This matches my intuition at least: I create a subnet, and want to give it a route table. I don't care about the route table ID, or necessarily even about tagging it (using tags for deep functionality feels somehow wrong), but I do care who's using the route table. Doing it this way might lead to a proliferation of route tables that could be shared across subnets, but Ansible is managing them for me anyway so I don't really care. For all I care, it could detect that a route table with the same routes already exists and reuse that behind the scenes.

Is that too big of a departure from the current design, too late in the game?

juliedavila:
@copumpkin 

I like the idea of adding some kind of `unique_in_subnet` type of parameter where the module would do a comparison of routes defined in the task and routes that currently exist and acting appropriately. 

However, I would be more open to it as a PR to this module once it has been merged. Once the latest changes by @wimnat and @bretmartin have been added and tests are ran it will be merged in. 

And to clarify I wouldn't want to remove the tag comparison but would see the addition of the subnet comparison as an additional option this way it is up to the end user how they want to determine route table existence, if at all.

wimnat:
Previous discussion in this PR has led me to take away resource_tags as an identifier because, as you pointed out, they are not guaranteed to be unique.

But of course we don't want to specify route table id if we're running this for the first time because we don't have an ID yet.  A subsequent run of the playbook will just generate another route table.  I can see where you're coming from with using subnet's.  I'll take a look when i finish this module off.  I know i keep saying this but real work is is just in my way - should have time from this weekend onwards.

lukehoersten:
@copumpkin The problem with "keying" off subnets is that this too can be messed with outside of Ansible's view. For example, we have 3 subnets using routeTableA and we use the console to remove a subnet. Will Ansible create a new route table with all 3 subnets or will it add the subnet to the existing route table. The semantics are unintuitive. At least with tags, which Amazon intended for this use, you almost have to expect this behavior when changing tags.

We could do something similar to the `ec2` module's `count_tag` directive where the user specifies what tag and value should be used to uniquely ID route tables.

In general it seems there's a unique ID issue with AWS and EC2 modules in general. I think we should unify how this is handled. A lot of extra YAML structure must be built up to support the concept of returned AWS IDs that aren't user defined. User defined IDs have the benefit of being able to establish relationships without having to grab "runtime" data.

juliedavila:
`In general it seems there's a unique ID issue with AWS and EC2 modules in general. I think we should unify how this is handled.`

I agree, the challenge is increased when you have an AWS resource type like ec2 where you can create any N number of instances with all the exact same attributes except for maybe IP address related attributes and then you want to create a unified mechanism that also works with AWS resources that are easier to uniquely identify because you can't create N amounts of that resource with the same attributes, such as S3.

lukehoersten:
In that case you can just add more tag specifiers. For example, a `count_tag: {Env: myGroup}` (a group) vs. `route_table_tag: {Name: myTable01}` (a single instance).

I'm having an issue with subnets for example. I want to generate a few subnets using `with_items` but really don't have a good way to reference them because the `ec2` module requires `vpc_subnet_id` (generated by AWS) instead of something like `vpc_subnet_tag` (created by me).

lukehoersten:
Here's another case where it's a problem: https://github.com/ansible/ansible-modules-core/issues/1068 - the `ec2_group` is not idempotent because it doesn't have a stable ID. Using the AWS-generated IDs makes it so the module can't tell if a change meant "make a new one" or "modify the current one". Too ambiguous.

bretmartin:
I appreciate all the thought that's being put in here -- I just wanted to chime in and say this module is still very useful in its current state. I can grab things like subnet IDs using the AWS CLI and I guess my route table situation is simple enough that I don't run into problems with the possibility of non-unique tag arrays.

If it's possible for this to be merged and these other concerns addressed in the future without major changes or breakage, I hope it will be.

lukehoersten:
Agreed. I'd like to start using the module :)

juliedavila:
@wimnat your help with this is greatly appreciated. Let's try and get this merged in sooner rather than later. For now just fix the docstring (re: line comment) then I'll run my final battery of tests and hopefully we can get this through this week.

As far as idempotency I think, for now, I prefer to see the EC2 module way of doing it where you can specify a specific tag as the thing to check for the pre-existence of the route table.

kalefranz:
+1 for getting this merged

lukehoersten:
It should be noted that when using the new `ec2_vpc_net` (replacing `ec2_vpc`) and `ec2_vpc_subnet` module group there's no way to configure route tables without this module. So until this module is merged into extras we can't fully test the new `vpc_net` group.

bcoca:
it still needs to be fixed as per comments

juliedavila:
@wimnat friendly ping, if you have a brief moment can you just patch the docstring? I'll then test. There is a level of urgency with this as I'd like to have it merged before the official 2.0 release.

wimnat:
Currently away but will be done by the end of the weekend - 30th

wimnat:
@defionscode here we go. 

Biggest change is addition of lookup parameter to specify to lookup by tag or id and it'll throw an error if lookup by tag returns more than one table as the previous operation of working on the first table returned was dangerous. Perhaps lookup by subnets could be added later.

The boto error when table has dependencies is now handled properly but i'll probably add a force option to remove these automatically at a later date.

Docstring quoted.

Rebased with latest devel.

Hopefully this is now ready for merge but if there are any issues let me know asap as i currently have the capacity to fix quickly.

lukehoersten:
Wow excellent work. Thanks for putting this together.

bcoca:
cc @defionscode @bennojoy @carsongee @erydo @wimnat

juliedavila:
I'll make my final (hopefully) review of this in the evening today.

juliedavila:
+1

bretmartin:
+1 for merging this _after_ the two lines I commented on above are corrected.

@wimnat The changes needed are trivial, but if you'd like me to PR against your branch like I did last time, just let me know.

wimnat:
@bretmartin good spot.  Fixed.

bretmartin:
@wimnat Thanks. I'm +1 on merging now!

PR #598, opened 2015-06-11
Feature/ec2 vpc subnet

wimnat:
Module by @erydo rebased to latest ansible-modules-extras

For review by @defionscode 

Obsoletes PR https://github.com/ansible/ansible-modules-core/pull/322

### DIFF
--- a/cloud/amazon/ec2_vpc_subnet.py
+++ b/cloud/amazon/ec2_vpc_subnet.py
@@ -111,16 +111,18 @@ EXAMPLES = '''
 '''
 
 
-import sys
+import sys  # noqa
 import time
 
 try:
     import boto.ec2
     import boto.vpc
     from boto.exception import EC2ResponseError
+    HAS_BOTO = True
 except ImportError:
-    print "failed=True msg='boto required for this module'"
-    sys.exit(1)
+    HAS_BOTO = False
+    if __name__ != '__main__':
+        raise
 
 
 class AnsibleVPCSubnetException(Exception):
@@ -249,6 +251,8 @@ def main():
         argument_spec=argument_spec,
         supports_check_mode=True,
     )
+    if not HAS_BOTO:
+        module.fail_json(msg='boto is required for this module')
 
     ec2_url, aws_access_key, aws_secret_key, region = get_ec2_creds(module)
     if not region:
--- a/cloud/amazon/ec2_vpc_subnet.py
+++ b/cloud/amazon/ec2_vpc_subnet.py
@@ -164,36 +164,17 @@ def get_resource_tags(vpc_conn, resource_id):
             vpc_conn.get_all_tags(filters={'resource-id': resource_id})}
 
 
-def dict_diff(old, new):
-    x = {}
-    old_keys = set(old.keys())
-    new_keys = set(new.keys())
-
-    for k in old_keys.difference(new_keys):
-        x[k] = {'old': old[k]}
-
-    for k in new_keys.difference(old_keys):
-        x[k] = {'new': new[k]}
-
-    for k in new_keys.intersection(old_keys):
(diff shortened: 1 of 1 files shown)


cloud/amazon/ec2_vpc_subnet.py
```diff
@@ -0,0 +1,249 @@
...
+    )
+    if not HAS_BOTO:
+        module.fail_json(msg='boto is required for this module')
+
+    ec2_url, aws_access_key, aws_secret_key, region = get_ec2_creds(module)
+    if not region:
+        module.fail_json(msg='Region must be specified')
+
+    try:
+        vpc_conn = boto.vpc.connect_to_region(
```
willthames: Use get_aws_connection_info in conjunction with connect_to_aws(boto.vpc,...)

cloud/amazon/ec2_vpc_subnet.py
```diff
@@ -0,0 +1,249 @@
...
+author: Robert Estelle (@erydo)
+options:
+  vpc_id:
+    description:
+      - VPC ID of the VPC in which to create the subnet.
+    required: false
+    default: null
+  resource_tags:
+    description:
+      - A dictionary array of resource tags of the form: { tag1: value1, tag2: value2 }. This module identifies a subnet by CIDR and will update the subnet's tags to match. Tags not in this list will be ignored.
```
juliedavila: This should be quoted or ansible-doc cannot parse it b/c of the curlies

```
    description:
      - "A dictionary array of resource tags of the form: { tag1: value1,
        tag2: value2 }. This module identifies a subnet by CIDR and will update
the subnet's tags to match. Tags not in this list will be ignored."
```

juliedavila:
`region` should be added to the doc string as it is required with "normal" usage.

Overall, this looks great, but a couple of style suggestions.

Currently the returned value of creating a subnet is:

```
changed: [localhost] => {"changed": true, "subnet": {"az": "us-east-1b", "cidr": "10.0.1.0/28", "id": "subnet-0aea7121", "tags": {"Name": "Frank", "delete": "me"}}, "subnet_id": "subnet-0aea7121"}
```

As you can see the keys `id`and `subnet_id` are redundant. This should probably be patched so it only returns one or the other (my preference is for `subnet_id`).  

Lastly, I prefer to see a standalone `name` parameter. This way the end user doesn't have to remember to create a `Name: $value` under the `resource_tags` parameter. Like this:

``` yml
- name: Create subnet for database servers
  ec2_vpc_subnet:
    state: present
    name: database_subnet
    vpc_id: vpc-123456
    cidr: 10.0.1.16/28
    resource_tags:
      Name: Database Subnet
  register: database_subnet
```

Not a huge thing, would not hold off on merging from this alone, just a style thing.

halberom:
I'm getting `changed: True` for repeated runs. Actual json output appears to be identical.

wimnat:
@halberom can you share your playbook please?

halberom:
https://github.com/HighOps/ansible_ec2_vpc_nat_asg

wimnat:
@halberom i think i got to the bottom of your changed=true problem. This module was maintaining existing tags when comparing to what was defined under 'tags' in the module. The module would maintain the old tags so the comparison would always be unequal and force a change.

I've removed this - now only the tags defined here are applied.  Tags that exist already but not defined in the module will be removed. I know some people may not like this destructive nature but Ansible is about 'Desired State Configuration' not a script to add some tags and preserve what already exists.

@defionscode I've added more detail to what is returned and eliminated the duplicate ID, replicating what exists in my facts module https://github.com/ansible/ansible-modules-extras/pull/783.  If both these modules pass muster then i'd appreciate the +1 to finally get this merged!

juliedavila:
+1

PR #606, opened 2015-06-13
Fix the lxc container restart state

cloudnull:
The lxc container restart state does not ensure that the container
is in fact started unless another config or command is passed into
the task. to fix this the module simply needs to have the function
call added `self._container_startup()` after the container is
put into a stopped state.

Signed-off By: Kevin Carter kevin.carter@rackspace.com


gregdek:
Merging as owner_pr.

PR #608, opened 2015-06-14
Add option to send a private message in irc module

amuraru:


amuraru:
@jpmens @sivel  please review

jpmens:
Untested, but the code looks ok to me.

notification/irc.py
```diff
@@ -63,6 +63,10 @@
     description:
       - Channel name
     required: true
+  nick_to:
+    description:
+      - Nick to send the message to
+    required: false
```
abadger, line 69: - Need to add version_added: 2.0
- Probably should update the nick: documentation as well so that people know that nick is the irc nickname being sent from.
- Depending on what you decide below about the relationship between nick and channel, need to add information about that to the documentation here.

notification/irc.py
```diff
@@ -173,7 +177,10 @@ def send_msg(channel, msg, server='localhost', port='6667', key=None, topic=None
         irc.send('TOPIC %s :%s\r\n' % (channel, topic))
         sleep(1)
 
-    irc.send('PRIVMSG %s :%s\r\n' % (channel, message))
+    if nick_to:
+        irc.send('PRIVMSG %s :%s\r\n' % (nick_to, message))
+    else:
+        irc.send('PRIVMSG %s :%s\r\n' % (channel, message))
```
abadger, line 194: Here, the code is deciding to ignore channel if nick_to is set.  I'm not sure that's correct behaviour.  If channel is set then the user probably intends for the message to be sent to the channel and to the nick_to user.  So probably you should do:

``` python
+    if nick_to:
+        irc.send('PRIVMSG %s :%s\r\n' % (nick_to, message))
+    if channel:
+        irc.send('PRIVMSG %s :%s\r\n' % (channel, message))
```

notification/irc.py
```diff
@@ -113,7 +117,7 @@
 
 
 def send_msg(channel, msg, server='localhost', port='6667', key=None, topic=None,
```
abadger: May need to make channel an optional keyword argument.  See comment below.

notification/irc.py
```diff
@@ -191,6 +198,7 @@ def main():
             server=dict(default='localhost'),
             port=dict(default=6667),
             nick=dict(default='ansible'),
+            nick_to=dict(),
```
abadger: I'm thinking about the relationship between channel and nick_to here.  I think that in most cases the user will want to either send to a channel or to one or more irc nicks.  A few might want to send to both a channel and one or more IRC nicks. So perhaps the following would be appropriate:
- Make channel optional
- Use required_one_of() to make either nick_to or channel be set.
- Check that if topic is set, then channel is set.  (I don't think any of the pre-existing AnsibleModule checks will handle this so you'll just have to test it below when you get the values out of module.params)
- Make nick_to a list so that more than one irc nick can be specified

(And then update documentation and code appropriately to note the new behaviour and when it changed).

abadger:
@jpmens Any thoughts from you on the suggestions I made?  I can merge as if you think it's okay without the suggestions or wait for @amuraru to make them.  Up to you :-)

abadger:
@amuraru, To move this forward, I'm willing to merge this if you make the changes I noted.  If @jpmens comments that the changes are unnecessary I'll merge sooner.  Thanks!

notification/irc.py
bcoca: at this point, why not make channel also a list?

I think nick_to doesn't need to be one, they can always use with_items, either way works for me.

notification/irc.py
amuraru, line 69: @abadger Done

notification/irc.py
amuraru, line 194: @abadger Good call, done.

notification/irc.py
amuraru: @abadger Done

notification/irc.py
amuraru: @abadger - Made nick_to a list

amuraru:
@abadger Please check the last version of this patch - and let me know what you think

abadger:
Need to document the default of nick_to but that's something I can do after merging.  Looks good now!  Thanks!

# Merged

Hi!

issue is resolved.

Thank you!

PR #609, opened 2015-06-15
Add the ability to growfs from the filesystem module optionally.

Firstyear:
Add the flag "growfs" to the filesystem module, allowing a filesystem to be grown. This works really nicely with lvol, to create a work flow such as:

```
- name: Create lvol
  sudo: yes
  action: lvol lv=test vg=vg size=1G state=present

-name: lvol filesystem
  sudo: yes
  action: filesystem fstype=xfs dev=/dev/mapper/vg-test growfs=yes
```

Changing the lvol size, automatically resizes the FS from, and when you initially create the lvol, creates it too.


gregdek:
@abulimov please review for functionality and adherence to Ansible module guidelines:

If you give +1, we will merge the changes.

system/filesystem.py
```diff
@@ -41,6 +41,12 @@
     description:
     - If yes, allows to create new filesystem on devices that already has filesystem.
     required: false
+  growfs:
+    choices: [ "yes", "no" ]
+    default: "no"
+    description:
+    - If yes, if the block device and filessytem size differ, grow the filesystem into the space. This DOES NOT support shrinking. Will only grow if the FS is mounted.
```
abulimov: The FS should be mounted only in the case of XFS filesystem. We could check if fstype=XFS and volume is mounted in this module, and just resize FS in other cases. Also, while XFS cannot be shrinked, we can rename growfs to resizefs and support shrinking via resize2fs. In this case, we should check that FS is not mounted, and run `e2fsck -f /dev/some_disk` before shrinking. Option to shrink FS makes sense, because `lvol` module supports shrinking.

abulimov:
Current version is good, but it would be better to improve functionality before merging. Or at least fix description about mounted FS.

system/filesystem.py
Firstyear: The comment about mounted FS should really be a warning about XFS specifically. It doesn't actually check if it's mounted or not, so the warning would make more sense.

I don't think you can express shrinking in a playbook. Consider this playbook structure.

Create VG
Create Lvol
Create Filesystem with growfs

So, now you change the Lvol size, and it grows. All good. 

So what if you reduce the lvol size? You would reduce the block size BEFORE you shrink the FS. So you would actually need to re-organise to:

Create VG
Create Filesystem with growfs
Create Lvol

Now it will shrink. But at the same time you have removed the ability to grow unless you ran the play twice ... 

I think that shrinking will probably need more consideration, and would likely need direct integration with the lvol module.

abulimov: Ansible usage is not limited to playbooks. It is often used as orchestration engine, for massive actions on remote hosts. For me, shrinking makes sense in orchestration.

As for comment about mounted FS, we could change it to mention XFS, or (better) add check to module, so running ansible in check mode could warn us about not mounted XFS.

Firstyear: True, it is. I'm just thinking about the general case. It seems like a lot of extra work and potential states to deal with in a way, but I can see in the long term it's the right solution. I'll work on shrink as well then, but for now, only ext*.

Yeah, I'll add the check then for xfs, and word the warning differently.

abulimov: Thanks, it would be really great!

gregdek:
Looks like at the very least the description needs to be changed, so marking this "needs revision" for now. How much of a revision is up to @Firstyear I guess. :)

Firstyear:
Okay, I've done another push. I've fixed some of the doc comments about XFS, changed some command mapping bits to clear up a few things and to provide extensibility later.

Right now, resizing for shrink is just too hard.
You need to potentially unmount, fsck, resize. There are so many more edge cases, potential issues and complexities. Right now, btrfs is the easiest, with an online shrink capability.

So I've made it able to support shrink in the future, but I think that it is just a really hard thing to achieve at this time without some better thought and time.

system/filesystem.py
```diff
@@ -63,17 +66,68 @@ def main():
             dev=dict(required=True, aliases=['device']),
             opts=dict(),
             force=dict(type='bool', default='no'),
+            resizefs=dict(type='bool', default='no'),
         ),
         supports_check_mode=True,
     )
 
+    # There is no "single command" to manipulate filesystems, so we map them all out and their options
```
alxgu, line 77: You can use fsadm for fs-check and resize.
It supports ext2, ext3, ext4, ReiserFS and XFS and in near feature btrfs.

fsadm -y resize /dev/vg/test

gregdek:
@abulimov can you re-review?

system/filesystem.py
```diff
@@ -82,9 +136,18 @@ def main():
...
     fs = raw_fs.strip()
 
-
-    if fs == fstype:
+    if fs == fstype and resizefs == False:
         module.exit_json(changed=False)
+    elif fs == fstype and resizefs == True:
+        cmd = module.get_bin_path(growcmd, required=True)
+        if module.check_mode:
+            changed = True
```
abulimov: shouldn't we just exit now with changed=True? Now, if we are in check mode, execution will continue till next  check for module.check_mode on line 156. This flow seems not very obvious and error-prone in future.

system/filesystem.py
```diff
@@ -41,9 +41,12 @@
     description:
     - If yes, allows to create new filesystem on devices that already has filesystem.
     required: false
-  opts:
+  resizefs:
```
abulimov, line 44: It seems that you have accidentally removed "opts" options from docs

system/filesystem.py
```diff
@@ -82,9 +136,18 @@ def main():
...
-
-    if fs == fstype:
+    if fs == fstype and resizefs == False:
         module.exit_json(changed=False)
+    elif fs == fstype and resizefs == True:
+        cmd = module.get_bin_path(growcmd, required=True)
+        if module.check_mode:
+            changed = True
+        else:
+            rc,out,err = module.run_command("%s %s" % (cmd, dev))
```
abulimov, line 149: shouldn't we check rc?

gregdek:
Back to you, @Firstyear.

Firstyear:
Hi,

Fixes for your comments have been provided.

system/filesystem.py
```diff
@@ -41,6 +41,12 @@
     description:
     - If yes, allows to create new filesystem on devices that already has filesystem.
     required: false
+  resizefs:
+    choices: [ "yes", "no" ]
+    default: "no"
+    description:
+    - If yes, if the block device and filessytem size differ, grow the filesystem into the space. Note: XFS Will only grow if mounted.
```
abulimov: The build has failed because of this line. As you use ': ' symbol in description string, it becomes dict after YAML parsing in the process of generating documentation. You can check it in any yaml parses, for example [here](https://yaml-online-parser.appspot.com/). To fix this you can just put description line in quotes, or remove ': ' symbol.

abulimov:
Except one small issue with documentation YAML, everything looks good to me, thanks!

Firstyear:
Changed from : to ,

gregdek:
Hmmm... fails travis?

abulimov:
Strange, but Travis now fails on rendering docs for bundler plugin, which is untouched by this request.

gregdek:
Yep, looking at it.

gregdek:
So it's likely to be breaking because of an unquoted colon in the doc strings; our yaml parser doesn't like that. We just introduced the doc string parser into Travis, so we're probably picking up bleedover from other docstrings that are broken now. We'll get back to this soon...

gregdek:
Closed and reopened to kick Travis.

gregdek:
Kicking Travis

gregdek:
@Firstyear so looks like you got caught in a bug in documentation elsewhere; we committed a new Travis check, and your code contains old code in bundler that's breaking the commit. Could you please rebase? Sorry for the confusion.

Firstyear:
I had some issues with this rebase. Please review and make sure the pull request only changes filesystem.py.

Firstyear:
Yeah, it's accidentally caught my ovirt changes because I did those on devel, and I needed to rebase onto devel. Any ideas on how to remove that from the pull request?

gregdek:
Hmm... cache the changes to ovirt, blow it away, rebase again?

Firstyear:
Hi,

Git just confuses and infuriates me, so it was easier to just make a new clean branch, and re-submit the pull req. See https://github.com/ansible/ansible-modules-extras/pull/672

PR #620, opened 2015-06-17
Fix TravisCI failure for system/osx_defaults.py on python 2.4

dagwieers:
Previous commit failed TravisCI checks because of this issue.

You cannot use ternary expressions (x if C else y) in python 2.4.
https://www.python.org/dev/peps/pep-0308/

And another issue is related to using 'except Exception as e'.

Now Travis continues happily on python 2.4


gregdek:
Thanks @dagwieers -- also seeing this Travis failure in a seemingly unrelated failed build, which confused me. We'll look to get another reviewer on this one soon.

PR #621, opened 2015-06-17
New module to copy (push) files to a vCenter datastore

dagwieers:
New pull request from proper branch.


gregdek:
Yeah, this seems like a Travis issue today -- we'll look into it.

dagwieers:
I send pull request #620 with a fix for the Travis issue.
You cannot use ternary expressions (x if C else y) in python 2.4.
https://www.python.org/dev/peps/pep-0308/

PR #622, opened 2015-06-17
haproxy module wait

ggermis:
I was testing rolling updates with haproxy and found that the haproxy module simply does an enable/disable without checking if the service actually was successfully started / reachable. This caused all services to enter the DOWN state after running the ansible playbook.

I added a few options to the haproxy module to allow it to wait for a specific status before continuing and raising an error when it could not reach the expected state within a set number of retries

The options added were:
- wait (default = false): When state=enabled, wait until the service gets to the status 'UP'. When state=disabled, wait until the service gets to the status 'MAINT'
- wait_retries (default = 20): How many times to retry (show stat)
- wait_interval (default = 1): How many seconds to wait before the next retry

This ensures that our rolling updates never bring down all services. It will try to enable 'serial' number of apps at a time and if that block fails to reach the 'UP' status, will fail the ansible-playbook command, leaving the other services running and allowing us to intervene

### DIFF
--- a/network/haproxy.py
+++ b/network/haproxy.py
@@ -73,16 +73,19 @@ options:
       - Wait until the server reports a status of 'UP' when state=enabled, or status of 'MAINT' when state=disabled
     required: false
     default: false
+    version_added: "2.0"
   wait_retries:
     description:
       - number of times to check for status after changing the state
     required: false
     default: 20
+    version_added: "2.0"
   wait_interval:
     description:
       - number of seconds to wait between retries
     required: false
     default: 1
+    version_added: "2.0"
 '''
 
 EXAMPLES = '''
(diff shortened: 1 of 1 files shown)


resmo:
LGTM
+1

gregdek:
@ravibhure please review this patch -- if you give it +1, we will merge.

network/haproxy.py
```diff
@@ -68,6 +68,21 @@
       - When disabling server, immediately terminate all the sessions attached to the specified server. This can be used to terminate long-running sessions after a server is put into maintenance mode, for instance.
     required: false
     default: false
+  wait:
```
bcoca, line 71: you want to add version_added for new parameters for an existing module

ggermis, line 71: added version_added: "2.0" to new parameters

ravibhure:
+1 :+1:

gregdek:
Merged. Thanks @ggermis for the PR, and thanks @resmo and @ravibhure for the reviews!

PR #625, opened 2015-06-19
Add GUIDELINES for AWS module development

willthames:
Starting point for a reference when doing pull request reviews.
If something doesn't meet the guidelines we can point people
at them. If something is bad but is not mentioned in the
guidelines, we should add it here.


gregdek:
Thanks @willthames -- a great start.

PR #636, opened 2015-06-22
specify int parameter types for wait_interval and wait_retries

ggermis:
The wait_retries and wait_interval parameters for the haproxy module should be defined as type integer otherwise python will raise an exception when the values are used

### DIFF
--- a/network/haproxy.py
+++ b/network/haproxy.py
@@ -78,13 +78,13 @@ options:
     description:
       - number of times to check for status after changing the state
     required: false
-    default: 20
+    default: 25
     version_added: "2.0"
   wait_interval:
     description:
       - number of seconds to wait between retries
     required: false
-    default: 1
+    default: 5
     version_added: "2.0"
 '''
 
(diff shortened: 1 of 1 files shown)


gregdek:
@ravibhure please review; if you give +1, we will merge.

abadger:
@ggermis Why the change to the value of WAIT_INTERVAL and WAIT_RETRIES?

The parameter types portion of this change looks good, though.

ravibhure:
@ggermis, I will take look.

ggermis:
@abadger I made a pull request a few days ago (https://github.com/ansible/ansible-modules-extras/pull/622) to add the new parameters. During our tests it soon got clear that the defaults that I set earlier (20 seconds - being 20 retries with 1 second intervals) were too short for some services (eg: a dropwizard app took much longer to start up before becoming healthy), so I decided to change the defaults to 125 seconds (25 retries with 5 second intervals) so in most cases you won't have to change the defaults

abadger:
@ggermis Sounds good to me.  Thanks for the explanation!

abadger:
@ravibhure did you want to review and +1/ask for changes?  I'm not a user of this module but codewise the changes look okay to me.

ravibhure:
+1 @ggermis  :+1:

gregdek:
Merged. Thanks @ravibhure for the review, and thanks @ggermis for the PR!

PR #641, opened 2015-06-24
apache2 reload module added

qmaxquique:


gregdek:
Thanks for submitting this module to Ansible Extras. 

resmo:
-1
the module functionality already covered by a core module (service).

qmaxquique:
Thank you @resmo . I'm closing this pull request.
How do you do to reload apache gracefully with the service module ? (apachectl graceful)

resmo:
@qmaxquique sorry for not explaining why I gave a -1. no need for closing the PR because of that.

IMHO it is a special case and the service module might not have a graceful but as you mentioned it, you can use `command` as well in a handler. I do not see any benefit to have a separate module for it. Or did I miss anything? Where do you see the benefit?

bcoca:
I would not add a module for every service X special option that service module does not have coded, I would actually look at adding a 'execute with this freeform option' to the service module.

bcoca:
https://github.com/ansible/ansible-modules-core/issues/1616

PR #643, opened 2015-06-24
route53_health_check: new module

zimbatm:
Allows to define and update Route53 health-checks

Create and update actions are defined in the module because boto is
broken in the first case and doesn't implement the second-one.

Continued from https://github.com/ansible/ansible-modules-core/pull/1276

### DIFF
--- a/cloud/amazon/route53_health_check.py
+++ b/cloud/amazon/route53_health_check.py
@@ -1,0 +1,357 @@
+#!/usr/bin/python
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: route53_health_check
+short_description: add or delete health-checks in Amazons Route53 DNS service
+description:
+  - Creates and deletes DNS Health checks in Amazons Route53 service
+  - Only the port, resource_path, string_match and request_interval are
+    considered when updating existing health-checks.
+version_added: "2.0"
+options:
+  state:
+    description:
+      - Specifies the action to take.
+    required: true
+    choices: [ 'present', 'absent' ]
+  ip_address:
+    description:
+      - IP address of the end-point to check. Either this or `fqdn` has to be
+        provided.
+    required: false
+    default: null
+  port:
+    description:
+      - The port on the endpoint on which you want Amazon Route 53 to perform
+        health checks. Required for TCP checks.
+    required: false
+    default: null
+  type:
+    description:
+      - The type of health check that you want to create, which indicates how
+        Amazon Route 53 determines whether an endpoint is healthy.
+    required: true
+    choices: [ 'HTTP', 'HTTPS', 'HTTP_STR_MATCH', 'HTTPS_STR_MATCH', 'TCP' ]
+  resource_path:
+    description:
+      - The path that you want Amazon Route 53 to request when performing
+        health checks. The path can be any value for which your endpoint will
+        return an HTTP status code of 2xx or 3xx when the endpoint is healthy,
+        for example the file /docs/route53-health-check.html.
+
+        * Required for all checks except TCP.
+        * The path must begin with a /
+        * Maximum 255 characters.
+    required: false
+    default: null
+  fqdn:
+    description:
+      - Domain name of the endpoint to check. Either this or `ip_address` has
+        to be provided. When both are given the `fqdn` is used in the `Host:`
+        header of the HTTP request.
+    required: false
+  string_match:
+    description:
+      - If the check type is HTTP_STR_MATCH or HTTP_STR_MATCH, the string
+        that you want Amazon Route 53 to search for in the response body from
+        the specified resource. If the string appears in the first 5120 bytes
+        of the response body, Amazon Route 53 considers the resource healthy.
+    required: false
+    default: null
+  request_interval:
+    description:
+      - The number of seconds between the time that Amazon Route 53 gets a
+        response from your endpoint and the time that it sends the next
+        health-check request.
+    required: true
+    default: 30
+    choices: [ 10, 30 ]
+  failure_threshold:
+    description:
+      - The number of consecutive health checks that an endpoint must pass or
+        fail for Amazon Route 53 to change the current status of the endpoint
+        from unhealthy to healthy or vice versa.
+    required: true
+    default: 3
+    choices: [ 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 ]
+author: "zimbatm (@zimbatm)"
+extends_documentation_fragment: aws
+'''
+
+EXAMPLES = '''
+# Create a health-check for host1.example.com and use it in record
+- route53_health_check:
+    state: present
+    fqdn: host1.example.com
+    type: HTTP_STR_MATCH
+    resource_path: /
+    string_match: "Hello"
+    request_interval: 10
+    failure_threshold: 2
+  record: my_health_check
+
+- route53:
+    action: create
+    zone: "example.com"
+    type: CNAME
+    record: "www.example.com"
+    value: host1.example.com
+    ttl: 30
+    # Routing policy
+    identifier: "host1@www"
+    weight: 100
+    health_check: "{{ my_health_check.health_check.id }}"
+
+# Delete health-check
+- route53_health_check:
+    state: absent
+    fqdn: host1.example.com
+
+'''
+
+import time
+import uuid
+
+try:
+    import boto
+    import boto.ec2
+    from boto import route53
+    from boto.route53 import Route53Connection, exception
+    from boto.route53.healthcheck import HealthCheck
+    HAS_BOTO = True
+except ImportError:
+    HAS_BOTO = False
+
+# Things that can't get changed:
+#  protocol
+#  ip_address or domain
+#  request_interval
+#  string_match if not previously enabled
+def find_health_check(conn, wanted):
+    """Searches for health checks that have the exact same set of immutable values"""
+    for check in conn.get_list_health_checks().HealthChecks:
+        config = check.HealthCheckConfig
+        if config.get('IPAddress') == wanted.ip_addr and config.get('FullyQualifiedDomainName') == wanted.fqdn and config.get('Type') == wanted.hc_type and config.get('RequestInterval') == str(wanted.request_interval):
+            return check
+    return None
+
+def to_health_check(config):
+    return HealthCheck(
+        config.get('IPAddress'),
+        config.get('Port'),
+        config.get('Type'),
+        config.get('ResourcePath'),
+        fqdn=config.get('FullyQualifiedDomainName'),
+        string_match=config.get('SearchString'),
+        request_interval=int(config.get('RequestInterval')),
+        failure_threshold=int(config.get('FailureThreshold')),
+    )
+
+def health_check_diff(a, b):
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @zimbatm. Also a good idea to make sure that you adhere to the AWS-specific standards for new modules here: https://github.com/ansible/ansible-modules-extras/blob/devel/cloud/amazon/GUIDELINES.md

AWS reviewers: @jsmartin @scicoin-project @tombamford @garethr @lorin @jarv @jsdalton @silviud @adq @zbal @zeekin @willthames @lwade @carsongee @defionscode

cloud/amazon/route53_health_check.py
```diff
@@ -0,0 +1,351 @@
...
+    e = boto.jsonresponse.Element()
+    h = boto.jsonresponse.XmlHandler(e, None)
+    h.parse(body)
+    return e
+
+def main():
+    argument_spec = ec2_argument_spec()
+    argument_spec.update(dict(
+            state               = dict(required=False, choices=['present', 'absent'], default='present'),
+            ip_address          = dict(required=False),
```
willthames: this (and other similar lines) can just be `dict()` - `required=False` is implied.

cloud/amazon/route53_health_check.py
```diff
@@ -0,0 +1,351 @@
...
+            module.fail_json(msg="parameter 'string_match' is required for the HTTP(S)_STR_MATCH types")
+        elif len(string_match_in) > 255:
+            module.fail_json(msg="parameter 'string_match' is limited to 255 characters max")
+    elif string_match_in:
+        module.fail_json(msg="parameter 'string_match' argument is only for the HTTP(S)_STR_MATCH types")
+
+    region, ec2_url, aws_connect_kwargs = get_aws_connection_info(module)
+    # connect to the route53 endpoint
+    try:
+        conn = Route53Connection(**aws_connect_kwargs)
```
willthames, line 319: normally I'd object to this but as boto.route53 doesn't have a `connect_to_region` method this is the best approach.

cloud/amazon/route53_health_check.py
```diff
@@ -0,0 +1,351 @@
...
+     - Creates and deletes DNS Health checks in Amazons Route53 service
+     - Only the port, resource_path, string_match and request_interval can be updated
+version_added: "2.0"
+options:
+  state:
+    description:
+      - Specifies the action to take.
+    required: true
+    choices: [ 'present', 'absent' ]
+  ip_address:
```
carsongee, line 32: From the form to create health checks, the ip address field is required and the fqdn is optional, are you sure these are correct as they are?
image

cloud/amazon/route53_health_check.py
```diff
@@ -0,0 +1,351 @@
...
+    import boto
+    import boto.ec2
+    from boto import route53
+    from boto.route53 import Route53Connection, exception
+    from boto.route53.healthcheck import HealthCheck
+    HAS_BOTO = True
+except ImportError:
+    HAS_BOTO = False
+
+# Things that can't get changed:
```
carsongee, line 140: It seems like this would be worth putting in the docs so people know what defines the health check for idempotence.

willthames:
There's lots of stylistic things I don't like (such as the heavy reliance on XML) but I think it's all due to limitations in boto.route53 (and also the fact that route53 isn't region based means it's genuinely different to most AWS modules)

cloud/amazon/route53_health_check.py
zimbatm, line 140: Is this enough?  e5bfeb29da820c36624f378d43327d4a78888f02

zimbatm:
I agree on the style. boto doesn't make things easy :/ I'm also not super fluent with python. I'm glad to fix any specific issues. I'll be squashing all the commits once all the points have been addressed.

cloud/amazon/route53_health_check.py
zimbatm, line 32: It depends. If you click on "Specify Endpoint By" -> "Domain name" then the IP address is forbidden.

cloud/amazon/route53_health_check.py
zimbatm, line 319: ~~Like this ? a05a4f66886c1c3a6e6137c6e2b8c0c14cf03dbe~~ EDIT: oops wrong comment. Agree with this one but don't have a better idea.

cloud/amazon/route53_health_check.py
zimbatm: Like this ? a05a4f66886c1c3a6e6137c6e2b8c0c14cf03dbe

zimbatm:
Alright, all comments have been addressed. Let me know if there is anything else.

A bit OT but I would love to have a framework to build unit tests along with the module. Right now it's not possible to find regressions in code without running it by hand.

gregdek:
@willthames @carsongee can you re-review? Thanks!

cloud/amazon/route53_health_check.py
```diff
@@ -0,0 +1,352 @@
...
+
+EXAMPLES = '''
+# Create a health-check for host1.foo.com and use it in record
+- route53_health_check:
+      state: present
+      fqdn: host1.foo.com
+      type: HTTP
+      resource_path: /
+      string_match: "Hello"
+      request_interval: 5
```
carsongee: This isn't a valid interval

cloud/amazon/route53_health_check.py
```diff
@@ -0,0 +1,352 @@
...
+    choices: [ 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 ]
+author: "zimbatm (@zimbatm)"
+extends_documentation_fragment: aws
+'''
+
+EXAMPLES = '''
+# Create a health-check for host1.foo.com and use it in record
+- route53_health_check:
+      state: present
+      fqdn: host1.foo.com
```
carsongee: you should use example.com for these since it is for just this sort of thing

cloud/amazon/route53_health_check.py
```diff
@@ -0,0 +1,352 @@
...
+    default: 3
+    choices: [ 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 ]
+author: "zimbatm (@zimbatm)"
+extends_documentation_fragment: aws
+'''
+
+EXAMPLES = '''
+# Create a health-check for host1.foo.com and use it in record
+- route53_health_check:
+      state: present
```
carsongee: nitpicky, but these should be two spaces instead of four

cloud/amazon/route53_health_check.py
```diff
@@ -0,0 +1,352 @@
...
+      fqdn: host1.foo.com
+      type: HTTP
+      resource_path: /
+      string_match: "Hello"
```
carsongee: Using this example with the fixed request_interval, I am getting:

```
msg: parameter 'string_match' argument is only for the HTTP(S)_STR_MATCH types
```

So it looks like it also needs to set `type`

cloud/amazon/route53_health_check.py
zimbatm: Thanks good catch, fixed in d1cddd5eca7ef5facbb8c4995b1b061af06dccbf

cloud/amazon/route53_health_check.py
zimbatm: yes totally agree. fixed in f71b55c948f5720c4e3f65ee3c357476182f1256

cloud/amazon/route53_health_check.py
zimbatm: fixed in 13fdaab32d1298fc8d5fbf18f2e746370d866eda

cloud/amazon/route53_health_check.py
zimbatm: Yes that's correct. Fixed in 814d059c678457262587d3f3d90b1ddf028694d5

zimbatm:
@willthames @carsongee @gregdek all comments addressed. Let me know if you want to have another round of review or if I should squash those commits

cloud/amazon/route53_health_check.py
```diff
@@ -0,0 +1,352 @@
...
+
+import time
+import uuid
+
+try:
+    import boto
+    import boto.ec2
+    from boto import route53
+    from boto.route53 import Route53Connection, exception
+    from boto.route53.healthcheck import HealthCheck
```
carsongee, line 135: What version was this added in?  It wasn't available in 2.21.0 that I had on my system, and doesn't seem to be documented

zimbatm, line 135: I am using 2.38.0 at the moment. According to the git logs it should be in 2.27.0+

Should I try to detect and notify accordingly the missing dependency ?

cloud/amazon/route53_health_check.py
carsongee, line 32: I can't actually get this module to work since I can't seem to get auth working with boto and anything over version 2.21.0, so I can't verify this, but doesn't that mean they both need to be optional, and then you just need a check that at least one of them is specified and fail if it isn't.  As it is now, since fqdn is required and you say it is forbidden if you use IP address, then wouldn't this simply not work with an IP address?

cloud/amazon/route53_health_check.py
zimbatm, line 135: Added minimum boto version in af8d0a9656fd532a0b8f9bd1ed97c5de5e055736

cloud/amazon/route53_health_check.py
zimbatm, line 32: Actually this is an error in the documentation again. Fixed in f7cfa00d6f9b2aeed70bfcfb5df6b6ef2870de4e .
If neither the fqdn nor the ip_address are provided then AWS will return a descriptive error. I suppose I could add a check and build my own error response.

I am currently using the module in the following forms, so I am sure those are working. Again I wish we could build unit tests or integration tests:

``` yaml
- local_action:
    module: route53_health_check
    state: "{{ binding_state }}"
    ip_address: "{{ ansible_ec2_public_ipv4 }}"
    type: TCP
    port: 5443
    request_interval: 10
    failure_threshold: 2

- local_action:
    module: route53_health_check
    state: "{{ binding_state }}"
    fqdn: "api-{{ cluster_name }}.{{ cluster_domain }}"
    ip_address: "{{ ansible_ec2_public_ipv4 }}"
    type: HTTP
    resource_path: /internal/info
    request_interval: 10
    failure_threshold: 2
```

For auth I am using the AWS_SECRET_ACCESS_KEY and AWS_ACCESS_KEY_ID environment variables. What is your method of auth ?

zimbatm, line 32: Added a check that at least one of those params is provided: 8f0e7142831d5daf65f7f2749ad4b18e1d7f53a4

gregdek:
Thanks @zimbatm -- looks like a lot of progress. @willthames @carsongee all concerns addressed here?

cloud/amazon/route53_health_check.py
```diff
@@ -0,0 +1,354 @@
...
+        health checks. The path can be any value for which your endpoint will
+        return an HTTP status code of 2xx or 3xx when the endpoint is healthy,
+        for example the file /docs/route53-health-check.html.
+
+        * Required for all checks except TCP.
+        * The path must begin with a /
+        * Maximum 255 characters.
+    required: false
+    default: null
+  fqdn:
```
carsongee, line 62: I think this probably makes more sense as hostname since you can set it to something like aws-health for ip_address based health checks.  I also think it needs a note here, something like: "If ip_address is set, this becomes the string set in the request `HOST` header during health checks". or similar.

cloud/amazon/route53_health_check.py
carsongee, line 32: The auth problem was local, on a fresh VM this all worked fine.

cloud/amazon/route53_health_check.py
zimbatm, line 62: Fixed in 1ca8d73650e761a3202d9324de1db7a15aa6bfca

carsongee:
This all looks good to me, :+1:

zimbatm:
Alright. All changes have been squashed into one commit for history clarity.

gregdek:
Waiting on @willthames final review -- if he gives thumbs-up will merge.

gregdek:
Satisfied with @carsongee's review here, plus @willthames's semi-review. Marking for inclusion.

gregdek:
cc: @ansible/core team

PR #644, opened 2015-06-25
Added zone.py module to manage Solaris zones

pmarkham:
Manage Solaris zones.

### DIFF
--- a/system/solaris_zone.py
+++ b/system/solaris_zone.py
@@ -1,6 +1,6 @@
 #!/usr/bin/python
 
-# (c) 2013, Paul Markham <pmarkham@netrefinery.com>
+# (c) 2015, Paul Markham <pmarkham@netrefinery.com>
 #
 # This file is part of Ansible
 #
@@ -37,13 +37,18 @@ options:
   state:
     required: true
     description:
-      - C(present), create the zone.
-      - C(running), if the zone already exists, boot it, otherwise, create the zone
-          first, then boot it.
+      - C(present), configure and install the zone.
+      - C(installed), synonym for C(present).
+      - C(running), if the zone already exists, boot it, otherwise, configure and install
+          the zone first, then boot it.
       - C(started), synonym for C(running).
       - C(stopped), shutdown a zone.
       - C(absent), destroy the zone.
-    choices: ['present', 'started', 'running', 'stopped', 'absent']
+      - C(configured), configure the ready so that it's to be attached.
+      - C(attached), attach a zone, but do not boot it.
+      - C(detach), stop and detach a zone
+    choices: ['present', 'installed', 'started', 'running', 'stopped', 'absent', 'configured', 'attached', 'detached']
+    default: present
   name:
     description:
       - Zone name.
@@ -68,19 +73,25 @@ options:
   config:
     required: false
     description:
-      - 'The zonecfg configuration commands for this zone, separated by commas, e.g.
-        "set auto-boot=true,add net,set physical=bge0,set address=10.1.1.1,end"
-        See the Solaris Systems Administrator guide for a list of all configuration commands
-        that can be used.'
+      - 'The zonecfg configuration commands for this zone. See zonecfg(1M) for the valid options
+        and syntax. Typically this is a list of options separated by semi-colons or new lines, e.g.
+        "set auto-boot=true;add net;set physical=bge0;set address=10.1.1.1;end"'
     required: false
-    default: null
+    default: empty string
   create_options:
     required: false
     description:
-      - 'Extra options to the zonecfg create command. For example, this can be used to create a
-        Solaris 11 kernel zone'
+      - 'Extra options to the zonecfg(1M) create command.'
     required: false
-    default: null
+    default: empty string
+  attach_options:
+    required: false
+    description:
+      - 'Extra options to the zoneadm attach command. For example, this can be used to specify
+        whether a minimum or full update of packages is required and if any packages need to
+        be deleted. For valid values, see zoneadm(1M)'
+    required: false
+    default: empty string
   timeout:
     description:
       - Timeout, in seconds, for zone to boot.
@@ -89,15 +100,15 @@ options:
 '''
 
 EXAMPLES = '''
-# Create a zone, but don't boot it
+# Create and install a zone, but don't boot it
 solaris_zone: name=zone1 state=present path=/zones/zone1 sparse=true root_password="Be9oX7OSwWoU."
-      config='set autoboot=true, add net, set physical=bge0, set address=10.1.1.1, end'
+      config='set autoboot=true; add net; set physical=bge0; set address=10.1.1.1; end'
 
-# Create a zone and boot it
+# Create and install a zone and boot it
 solaris_zone: name=zone1 state=running path=/zones/zone1 root_password="Be9oX7OSwWoU."
-      config='set autoboot=true, add net, set physical=bge0, set address=10.1.1.1, end'
+      config='set autoboot=true; add net; set physical=bge0; set address=10.1.1.1; end'
 
-# Boot an already created zone
+# Boot an already installed zone
 solaris_zone: name=zone1 state=running
 
 # Stop a zone
@@ -105,6 +116,16 @@ solaris_zone: name=zone1 state=stopped
 
 # Destroy a zone
 solaris_zone: name=zone1 state=absent
+
+# Detach a zone
+solaris_zone: name=zone1 state=detached
+
+# Configure a zone, ready to be attached
+solaris_zone: name=zone1 state=configured path=/zones/zone1 root_password="Be9oX7OSwWoU."
+      config='set autoboot=true; add net; set physical=bge0; set address=10.1.1.1; end'
+
+# Attach a zone
+solaris_zone: name=zone1 state=attached attach_options='-u'
 '''
 
 class Zone(object):
@@ -120,45 +141,60 @@ class Zone(object):
         self.timeout        = self.module.params['timeout']
         self.config         = self.module.params['config']
         self.create_options = self.module.params['create_options']
+        self.attach_options = self.module.params['attach_options']
 
         self.zoneadm_cmd    = self.module.get_bin_path('zoneadm', True)
         self.zonecfg_cmd    = self.module.get_bin_path('zonecfg', True)
         self.ssh_keygen_cmd = self.module.get_bin_path('ssh-keygen', True)
 
-    def create(self):
+    def configure(self):
         if not self.path:
             self.module.fail_json(msg='Missing required argument: path')
 
-        t = tempfile.NamedTemporaryFile(delete = False)
+        if not self.module.check_mode:
+            t = tempfile.NamedTemporaryFile(delete = False)
 
-        if self.sparse:
-            t.write('create %s\n' % self.create_options)
-            self.msg.append('creating sparse root zone')
-        else:
-            t.write('create -b %s\n' % self.create_options)
-            self.msg.append('creating whole root zone')
+            if self.sparse:
+                t.write('create %s\n' % self.create_options)
+                self.msg.append('creating sparse-root zone')
+            else:
+                t.write('create -b %s\n' % self.create_options)
+                self.msg.append('creating whole-root zone')
 
-        t.write('set zonepath=%s\n' % self.path)
+            t.write('set zonepath=%s\n' % self.path)
+            t.write('%s\n' % self.config)
+            t.close()
 
-        if self.config:
-            for line in self.config:
-                t.write('%s\n' % line)
-        t.close()
+            cmd = '%s -z %s -f %s' % (self.zonecfg_cmd, self.name, t.name)
+            (rc, out, err) = self.module.run_command(cmd)
+            if rc != 0:
+                self.module.fail_json(msg='Failed to create zone. %s' % (out + err))
+            os.unlink(t.name)
 
-        cmd = '%s -z %s -f %s' % (self.zonecfg_cmd, self.name, t.name)
-        (rc, out, err) = self.module.run_command(cmd)
-        if rc != 0:
-            self.module.fail_json(msg='Failed to create zone. %s' % (out + err))
-        os.unlink(t.name)
+        self.changed = True
+        self.msg.append('zone configured')
 
-        cmd = '%s -z %s install' % (self.zoneadm_cmd, self.name)
-        (rc, out, err) = self.module.run_command(cmd)
-        if rc != 0:
-            self.module.fail_json(msg='Failed to install zone. %s' % (out + err))
+    def install(self):
+        if not self.module.check_mode:
+            cmd = '%s -z %s install' % (self.zoneadm_cmd, self.name)
+            (rc, out, err) = self.module.run_command(cmd)
+            if rc != 0:
+                self.module.fail_json(msg='Failed to install zone. %s' % (out + err))
+            self.configure_sysid()
+            self.configure_password()
+            self.configure_ssh_keys()
+        self.changed = True
+        self.msg.append('zone installed')
 
-        self.configure_sysid()
-        self.configure_password()
-        self.configure_ssh_keys()
+    def uninstall(self):
+        if self.is_installed():
+            if not self.module.check_mode:
+                cmd = '%s -z %s uninstall -F' % (self.zoneadm_cmd, self.name)
+                (rc, out, err) = self.module.run_command(cmd)
+                if rc != 0:
+                    self.module.fail_json(msg='Failed to uninstall zone. %s' % (out + err))
+            self.changed = True
+            self.msg.append('zone uninstalled')
 
     def configure_sysid(self):
         if os.path.isfile('%s/root/etc/.UNCONFIGURED' % self.path):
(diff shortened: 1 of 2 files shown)


gregdek:
Thanks for submitting this module to Ansible Extras.

system/zone.py
```diff
@@ -0,0 +1,353 @@
...
+import tempfile
+
+DOCUMENTATION = '''
+---
+module: zone
+short_description: Manage Solaris zones
+description:
+   - Create, start, stop and delete Solaris zones. This module doesn't currently allow
+     changing of options for a zone that's already been created.
+version_added: "1.5"
```
resmo: should be "2.0" at this time point.

system/zone.py
```diff
@@ -0,0 +1,353 @@
...
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+import sys
+import os
+import platform
+import tempfile
+
+DOCUMENTATION = '''
+---
+module: zone
```
resmo: IMHO the name is possibly a bit too generic. What about solaris_zone?

system/zone.py
```diff
@@ -0,0 +1,353 @@
...
+version_added: "1.5"
+author: Paul Markham
+requirements:
+  - Solaris 10 or later
+options:
+  state:
+    required: true
+    description:
+      - C(present), create the zone.
+        C(running), if the zone already exists, boot it, otherwise, create the zone
```
resmo: you can make a list for every newline here. The convention in many modules is to make `state=present` the default. Consider this as well if it makes sense.

system/zone.py
```diff
@@ -0,0 +1,353 @@
...
+'''
+
+EXAMPLES = '''
+# Create a zone, but don't boot it
+zone: name=zone1 state=present path=/zones/zone1 whole_root=true root_password="Be9oX7OSwWoU."
+      config='set autoboot=true, add net, set physical=bge0, set address=10.1.1.1, end'
+
+# Create a zone and boot it
+zone: name=zone1 state=running path=/zones/zone1 whole_root=true root_password="Be9oX7OSwWoU."
+      config='set autoboot=true, add net, set physical=bge0, set address=10.1.1.1, end'
```
resmo: wouldn't it make sense to add separate arguments for this config string elements? e.g.
`autoboot=true`, `add_net=true`, `physical_ifce=bge0`, `address=10.1.1.1`? Any thoughts?

resmo:
I looked over the module, but can not test it nor do I fully understand what the code does but looks generally good.  I would suggest to change the name to make clear this is for Solaris, eventually check in the module if the OS is Solaris and fail if it's not.

pmarkham:
Thanks for the great feedback. I've updated and committed the following changes:

```
- Renamed module to solaris_zone.py
- Updated 'version_added'
- Updated description of 'state' to make each line a list item
- Check that OS is Solaris
```

Regarding your comment:

  wouldn't it make sense to add separate arguments for this config string elements? e.g.
  autoboot=true, add_net=true, physical_ifce=bge0, address=10.1.1.1?

Ideally, yes, these would be separate module arguments. However, there's a large number of options and sub-options to the underlying OS command, which would make this a maintenance nightmare, if it was even possible to do with key=value pairs.

resmo:
ok. in this case it makes no sense to have separate args. that's ok for me.

system/solaris_zone.py
```diff
@@ -0,0 +1,360 @@
...
+
+    def exit_with_msg(self):
+        msg = ', '.join(self.msg)
+        self.module.exit_json(changed=self.changed, msg=msg)
+
+def main():
+    module = AnsibleModule(
+        argument_spec      = dict(
+            name           = dict(required=True),
+            state          = dict(required=True, choices=['running', 'present', 'stopped', 'absent']),
```
resmo: I've got one more little thing. Make a default for `state` and use `present` to be the default by convention.

system/solaris_zone.py
```diff
@@ -0,0 +1,360 @@
...
+    description:
+      - Zone name.
+    required: true
+  path:
+    description:
+      - The path where the zone will be created. This is required when the zone is created, but not
+        used otherwise.
+    required: false
+    default: null
+  whole_root:
```
mavit: I'd agree with @dago that `sparse` seems like a better name for this option.

system/solaris_zone.py
```diff
@@ -0,0 +1,360 @@
...
+        id.write('1       # Network has subnets?\n')
+        id.write('1       # root password prompted for?\n')
+        id.write('1       # locale and term prompted for?\n')
+        id.write('1       # security policy in place\n')
+        id.write('1       # NFSv4 domain configured\n')
+        id.write('0       # Auto Registration Configured\n')
+        id.write('vt100')
+        id.close()
+
+    def configure_ssh_keys(self):
```
mavit, line 239: This will fail for Solaris 8 branded zones, because Solaris 8 does not include `ssh`.

resmo, line 239: @mavit There is a check to look for solaris 10 on line 338

dago, line 239: You can run Solaris 8 and 9 branded zones on a Solaris 10 system. However, I would consider this a corner case which IMHO can be ignored as 8/9 branded zones are discontinued anyway.

gregdek:
Lots of action. Looks popular. :) @pmarkham let us know when you've made all the revisions you intend to make so we can re-review.

system/solaris_zone.py
```diff
@@ -0,0 +1,360 @@
...
+        id.write('1       # System is on a network?\n')
+        id.write('1       # Extended network information gathered?\n')
+        id.write('0       # Autobinder succeeded?\n')
+        id.write('1       # Network has subnets?\n')
+        id.write('1       # root password prompted for?\n')
+        id.write('1       # locale and term prompted for?\n')
+        id.write('1       # security policy in place\n')
+        id.write('1       # NFSv4 domain configured\n')
+        id.write('0       # Auto Registration Configured\n')
+        id.write('vt100')
```
mavit, line 236: Should the terminal type be configurable?  Would `xterm` be a better default than `vt100`, nowadays?

system/solaris_zone.py
bcoca: state is required, so no default is needed, also 'started' would be preferable over running

system/solaris_zone.py
```diff
@@ -0,0 +1,360 @@
...
+  - Solaris 10 or later
+options:
+  state:
+    required: true
+    description:
+      - C(present), create the zone.
+      - C(running), if the zone already exists, boot it, otherwise, create the zone
+          first, then boot it.
+      - C(stopped), shutdown a zone.
+      - C(absent), destroy the zone.
```
bcoca, line 46: this was not strictly needed, works both ways, as long st it is a valid yaml list it can be one item or several.

system/solaris_zone.py
dago: "running" is modeled after the Solaris terminology from the status field as used in the zoneadm command.

system/solaris_zone.py
```diff
@@ -0,0 +1,360 @@
...
+        will not have a password.
+    required: false
+    default: null
+  config:
+    required: false
+    description:
+      - 'The zonecfg configuration commands for this zone, separated by commas, e.g.
+        "set auto-boot=true,add net,set physical=bge0,set address=10.1.1.1,end"
+        See the Solaris Systems Administrator guide for a list of all configuration commands
+        that can be used.'
```
bcoca: so instead of this format I would recommend adding complex args here, instead of key value pairs make it a freeform dictionary

```
config:
   auto-boot: True
   physical: bge0
....
```

This way you don't have to create your own 'subparser'

system/solaris_zone.py
dago, line 236: The terminal type is only for the console as TERM is inherited when you login via ssh. Using vt100 is good practice.

system/solaris_zone.py
dago: This opens a can of worms by introducing a different format for zone configuration. The comma separated arguments are directly transferable to zonecfg and there are lots of options which vary between Solaris releases. It may be possible to use a list of strings as lines to be passed to zonecfg.

system/solaris_zone.py
mavit, line 239: Come to think of it, extra options would be needed before this module would be able to [create](http://docs.oracle.com/cd/E22645_01/html/820-2914/gfjcg.html#scrolltoc) and [install](http://docs.oracle.com/cd/E22645_01/html/820-2914/gentextid-1181.html#gfjyv) branded zones in the first place, so this is never going to cause a failure as things stand.  Sorry for the noise!

system/solaris_zone.py
mavit: It's probably premature to be thinking about the following, but presumably we'd desire some hypothetical future version of this module to be able to modify the configuration of an existing zone.  It'd do this by walking the output of either `zonecfg export` (which has the same format as `zonecfg create`) or `zonecfg info` (which is almost-but-not-quite-YAML) and looking for differences.  Figuring out the right `zonecfg set` commands needed to make the changes would require understanding the syntax (if not the semantics) of the configuration.

bcoca: ok, we can leave as is for now, explore a better format in the future, probably using another option name to keep this one backwards compatible.

system/solaris_zone.py
```diff
@@ -0,0 +1,360 @@
...
+options:
+  state:
+    required: true
+    description:
+      - C(present), create the zone.
+      - C(running), if the zone already exists, boot it, otherwise, create the zone
+          first, then boot it.
+      - C(stopped), shutdown a zone.
+      - C(absent), destroy the zone.
+    choices: ['present', 'running', 'stopped', 'absent']
```
bcoca: ~~it would be better to have 'started' than 'running', as it is the obvious opposite of 'stopped', make the choices into nice complementary pairs~~ had already made this comment

system/solaris_zone.py
bcoca: lets split the diff and make 'running' and 'started' synonymous

pmarkham:
I've done the following updates:
- Changed 'whole_root' option to 'sparse'
- Added state='started' as synonym for state='running'.

That's all the changes I intend to do for this pull request.

Future updates:
- Allow zones to be detached (and attached?) so they can be moved between hosts. Talking to Dago separately about this.
- Look at being able to modify zone options
- Anything else that someone can think of.

pmarkham:
After rebooting my virtual machine, I'm getting an error trying to unlink the .UNCONFIGURED file. I wasn't getting this last week; is anyone else seeing this or is just me? 

I will probably update the module so that it doesn't barf if this file is missing.

pmarkham:
Ok, I've restored my Solaris VM to a snapshot from last week and it's all working. Maybe all the half-configured zones was causing an issue.

Regardless, I've updated the module to handle this situation just in case it re-appears.

So, now, the module is ready as far as I'm concerned.

gregdek:
Thanks @pmarkham! Back into community_review: @bcoca @resmo can you oblige us again?

resmo:
LGTM. although `state` should be defaulted to `present` (IMHO this should also be a convention) but can only give +0 because I was not able to test it.

gregdek:
Maybe that's a +1/2 -- an Extras module with new functionality has lower risk anyway, and I imagine users will report breakage pretty quickly. 

@dago @mavit @bcoca any additional thumbs up from you?

pmarkham:
I thought making the default state being 'present'. However, my personal preference is to always specify the state as it's clearer what you mean. Also, while 'present' is probably the safest default, 'running/started' is almost always what you'd want.

I'm happy to update the module to make 'present' the default.

system/solaris_zone.py
```diff
@@ -0,0 +1,362 @@
+#!/usr/bin/python
+
+# (c) 2013, Paul Markham <pmarkham@netrefinery.com>
```
mavit: 2015?

system/solaris_zone.py
```diff
@@ -0,0 +1,362 @@
...
+     changing of options for a zone that's already been created.
+version_added: "2.0"
+author: Paul Markham
+requirements:
+  - Solaris 10 or later
+options:
+  state:
+    required: true
+    description:
+      - C(present), create the zone.
```
mavit: Is it worth being explicit that this will both configure and install the zone?  It might be nice to have a synonym `installed` for `present`?

system/solaris_zone.py
```diff
@@ -0,0 +1,362 @@
...
+  root_password:
+    description:
+      - The password hash for the root account. If not specified, the zone's root account
+        will not have a password.
+    required: false
+    default: null
+  config:
+    required: false
+    description:
+      - 'The zonecfg configuration commands for this zone, separated by commas, e.g.
```
mavit: "separated by commas or newlines".

system/solaris_zone.py
```diff
@@ -0,0 +1,362 @@
...
+        cmd = '%s -z %s install' % (self.zoneadm_cmd, self.name)
+        (rc, out, err) = self.module.run_command(cmd)
+        if rc != 0:
+            self.module.fail_json(msg='Failed to install zone. %s' % (out + err))
+
+        self.configure_sysid()
+        self.configure_password()
+        self.configure_ssh_keys()
+
+    def configure_sysid(self):
```
mavit, line 214: Configuring the zone doesn't work like this with Solaris 11 zones, causing the zone to go into manual configuration and the boot timeout to trip.

On Solaris 11, one instead must pass `-c -c config_profile.xml` to `zoneadm install`.  The XML profile can be created with the `sysconfig create-profile -o ...` command.

For now, would it be best to document that this module doesn't work with Solaris 11 zones?

system/solaris_zone.py
```diff
@@ -0,0 +1,362 @@
...
+      - The password hash for the root account. If not specified, the zone's root account
+        will not have a password.
+    required: false
+    default: null
+  config:
+    required: false
+    description:
+      - 'The zonecfg configuration commands for this zone, separated by commas, e.g.
+        "set auto-boot=true,add net,set physical=bge0,set address=10.1.1.1,end"
+        See the Solaris Systems Administrator guide for a list of all configuration commands
```
mavit: I always struggle to find what I'm looking for on the Oracle website.  Worth adding a link to https://docs.oracle.com/cd/E26505_01/html/817-1592/z.config.ov-12.html#scrolltoc?

system/solaris_zone.py
```diff
@@ -0,0 +1,362 @@
...
+            if elapsed > self.timeout:
+                self.module.fail_json(msg='timed out waiting for zone to boot')
+            rc = os.system('ps -z %s -o args|grep "/usr/lib/saf/ttymon.*-d /dev/console" > /dev/null 2>/dev/null' % self.name)
+            if rc == 0:
+                break
+            time.sleep(10)
+            elapsed += 10
+
+    def destroy(self):
+        cmd = '%s -z %s delete -F' % (self.zonecfg_cmd, self.name)
```
mavit: Should the zone be `zoneadm uninstall`ed first, to reclaim the disk space used by the zone?  If not, it probably needs documenting that the user will need to do this themselves.

system/solaris_zone.py
dago: The best match for `present` is probably `installed`, but I find it non-intuitive how the module should reach the state as there are two possible transitions, either over `install` which populates the files for the zone or `attach` which merely checks a zone already on the filesystems and marks it as installed.

system/solaris_zone.py
pmarkham: Well, the module _was_ written in 2013. I'll update this.

system/solaris_zone.py
pmarkham: Agreed. I'll add `installed` as a synonym for `present` and expand on the documentation.

system/solaris_zone.py
pmarkham: The parameter to the module is defined as a list, which must be comma separated, I believe. The resulting file which is input to zoneadm can be either comma or newline separated.

system/solaris_zone.py
pmarkham, line 214: I'll have to build a Solaris 11 machine and look at this.

system/solaris_zone.py
pmarkham: Interesting point. I always create zones on their own zfs filesytems so the few times I've need to delete the zone, I just delete the zfs filesystem.

Anyone else got any thoughts on what the best option here is?

system/solaris_zone.py
pmarkham: Thinking about this a bit more, there's probably no reason this needs to be a list. It can probably just be a string so you use whatever is valid zonecfg syntax. I'll look at this as a future enhancement.

system/solaris_zone.py
dago: `uninstall` does just that when zoneroot is on a separate ZFS filesystem. On Solaris 11 the ZFS setup is much more complicated so uninstall is essential.

system/solaris_zone.py
dago: From `zonecfg(1m)`:

```
     zonecfg supports a semicolon-separated list of  subcommands.
     For example:

       # zonecfg -z myzone "add net; set physical=myvnic; end"
```

Accepting a string should be fine and most flexible.

pmarkham:
I've updated this based on the reviews by @mavit and @dago
- Changed `config` from a list to a string so any valid zonecfg(1M) syntax is accepted. Note, commas are no longer a valid separator.
- Uninstall the zone before destroying it so the file system is cleaned up.
- Made default state `present`
- Added `attached`, `detached` and `configured` states to allow zones to be moved between hosts.
- Updated documentation and examples.
- Code tidy up and refactoring.

@dago, I've added the attach and detach states you were after. Can you please verify it's working the way you wanted.

mavit:
The requirements say "Solaris 10 or later" and at present this module only works on Solaris 10.  Apart from that, it's +1 from me!

system/solaris_zone.py
```diff
@@ -0,0 +1,446 @@
...
+      - C(present), configure and install the zone.
+      - C(installed), synonym for C(present).
+      - C(running), if the zone already exists, boot it, otherwise, configure and install
+          the zone first, then boot it.
+      - C(started), synonym for C(running).
+      - C(stopped), shutdown a zone.
+      - C(absent), destroy the zone.
+      - C(configured), configure the ready so that it's to be attached.
+      - C(attached), attach a zone, but do not boot it.
+      - C(detach), stop and detach a zone
```
johanwiren: Should be C(detached) according to the line below.

pmarkham: Thanks, I'll fix this.

pmarkham:
@mavit, in my initial testing on Solaris 11, it did configure the zone. It failed to install it as I didn't have an IPS package repository available, but it seems like it'll work.

My impression was the @dago was using it on Solaris 11. Hopefully he can confirm this. If it doesn't work, Ill update the comment.

pmarkham:
Updated documentation. @mavit, I changed my mind and made the requirement "Solaris 10". The "or later" bit may not be accurate when they release Solaris 12 :-)

system/solaris_zone.py
```diff
@@ -0,0 +1,446 @@
...
+    required: true
+    description:
+      - C(present), configure and install the zone.
+      - C(installed), synonym for C(present).
+      - C(running), if the zone already exists, boot it, otherwise, configure and install
+          the zone first, then boot it.
+      - C(started), synonym for C(running).
+      - C(stopped), shutdown a zone.
+      - C(absent), destroy the zone.
+      - C(configured), configure the ready so that it's to be attached.
```
bcoca, line 47: configured should not be needed, present shoudl handle both creation and update

pmarkham, line 47: They're slightly different things.

There's three stages to setting up a zone. What you'd normally do is:
- configure the zone - defines the zone to the system -- it's not usable at this stage.
- install the zone - copy in all the operating system files the zone needs to run.
- boot the zone

The installed/present state does the first two; the running/started state does all three.

Attaching and detaching zones lets you moved them between physical servers.
When attaching a zone it's:
- configure the zone - optional if the zone has already been configured.
- attach the zone - point to existing file system with the operating system files.
- boot the zone

So we do need to differentiate between the states configured, installed and running.

mavit:
@pmarkham: See comment https://github.com/ansible/ansible-modules-extras/pull/644#discussion_r33668443 regarding Solaris 11.  The zone configures and installs fine, but then pauses during boot to prompt on the console for the zone's hostname, network settings, etc..  This trips the module's wait-for-boot timeout and returns a failed status.

magnus919:
@mavit this is why I think "present" should likely be the default over "running" or "started"; once a zone is configured & installed, it's customary to do some level of filesystem-level manipulation of the zone before booting it for the first time (thus avoiding getting prompted for configuration input on the first boot). Aside from the sysidcfg file that is normally populated before first boot, it's also common to manipulate /etc/nsswitch.conf and /etc/resolv.conf, etc, before booting.

pmarkham:
@mavit, I've managed to get a Solaris 11 test machine going and I can see the issue you're having. I don't want to add code to create the profile XML file as I think that's better suited to a template; fortunately Ansible has a good templating module :-)

What I have added is an `install_options` variable which will allow you to pass additional options to the `zoneadm install` command, so you use something like `install_options="-u sc_profile.xml`". In my testing, this successfully installed and booted the zone.

I'll will push the updated code tomorrow after I've done a bit more testing.

@magnus919, "present" is the default state.

pmarkham:
Latest code has been pushed.

pmarkham:
Fixed, thanks.

PR #646, opened 2015-06-26
CloudStack: omit hypervisor option if not set

atsaki:
cs_instance uses first found hypervisor when it deploys a VM if user does not specify hypervisor type.
This causes mismatch of hypervisor type between option and templates like ollowing.

```
msg: Failed: 'Hypervisor passed to the deployVm call, is different from the hypervisor type of the template'
```

This patch omits hypervisor option if user does not set hypervisor type because CloudStack uses hypervisor type of the template properly if it is not set.

### DIFF
--- a/cloud/cloudstack/cs_instance.py
+++ b/cloud/cloudstack/cs_instance.py
@@ -70,8 +70,6 @@ options:
   hypervisor:
     description:
       - Name the hypervisor to be used for creating the new instance.
-      - Relevant when using C(state=present) and option C(ISO) is used.
-      - If not set, first found hypervisor will be used.
     required: false
     default: null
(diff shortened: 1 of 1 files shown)


resmo:
hi @atsaki. Thanks for reporting!

If I understand you correctly you are the problem is when using template/iso, the hypervisor param should be omitted if param hypervisor is set on the template/iso, Because the first found hypervisor may be different, which causes the error. Makes sense!

The API doc says: "the hypervisor on which to deploy the virtual machine. The parameter is required and respected only when hypervisor info is not set on the ISO/Template passed to the call". 

So I will fix it to actually look at the template/iso if the hypervisor is set and omit the `hypervisor` for `deployVirtualMachine` if so.

PR #648, opened 2015-06-28
add jenkins module

kpabijanskas:
Adds basic functionality such as triggering builds, checking if jobs exist, and checking information about them


tersmitten:
Nice!

gregdek:

resmo:
Looks generally good, but I do not like much `actions`.

I do not see much benefit about getting list of jobs and the action `has_job`. 

I would rather propose adding a param `state` (and remove `action`) e.g. `state=build` and extend it with more options `state=enabled` `state=disabled`. Having a `state` would also allow us to implement check mode support.

The module should always return a dict of the build job data.

In future versions of the module, we could extend it with more features like `state=present` and `state=absent`.

kpabijanskas:
I have renamed action to state as per your suggestion.

Need to think about the best way of handling <code>enabled</code>/<code>disabled</code> states though.

There is no point using <code>enabled</code>/<code>disabled</code> for jenkins itself - service module already handles this. There is no point using <code>enabled</code>/<code>disabled</code> to add/remove jobs - template/copy can do this for you, since jenkins jobs are just files on disk. Nor there is much point in using it for building job, since <code>disabled</code> is redundant then.

Regarding <code>has_job</code> and <code>get_job_list</code>, you are right. I initially added <code>has_job</code>, and <code>get_job_list</code>, which was added later, made it redundant. <code>get_job_list</code> may be better, since that allows you to check if a job is present anyway when using <code>when</code>, and it also allows you to loop over all jobs if needed later on, etc. <code>has_job</code> returns a simple boolean, so is a good candidate for removal.

Regarding <code>build_job</code> returning a dict, this is something I had planned anyway, but the problem is that the underlying library does not return any data (other than a success notice) when triggering a build, so it is difficult to ensure that you are returning the data for the correct build. While you can immediately get the latest pending build information, this still leaves a very short window (some milliseconds) that another build could be triggered from elsewhere - and then you are getting information for the wrong build! Ideally, in the future I would like to add the possibility to wait for the build to finish before returning, but since currently I could not guarantee that the build data is for the correct build, I simply left it out. Being able to trigger a build is useful anyway.

There will be other things I would like to eventually add to this module, such as plugin management, so there will be additional states in the future. But it is already, I believe, useful, so would be good to have now rather than waiting for it to be perfect.

resmo:
Ok, this is not exactingly what I suggested. I try to clearify:

This module should be renamed to jenkins_job and only handle jenkins jobs

state vs action: action does an action every time. state does it only if the wanted state and the current state is different. So this needs to be implemented, a simple rename of the option is not enough. example:

if we say `jenkins_job: name=foo state=absent`, the module needs to get a list of the jobs and look if the job is in that list or not, if not no action needed. You can actually use the code for has_job to look if an action is needed, in this example remove the job. 

If create job is out of scope, that is ok for now.

service **must not** be handled. only jobs.

robynbergeron:

gregdek:
Closing due to contributor inaction. @kpabijanskas feel free to reopen if/when you're ready to make the recommended revisions.

PR #651, opened 2015-06-29
Fixed incorrect connect_to_aws call

wimnat:
switched to boto.vpc instead of boto.ec2

Module will always fail without this

### DIFF
--- a/cloud/amazon/ec2_vpc_igw.py
+++ b/cloud/amazon/ec2_vpc_igw.py
@@ -29,7 +29,7 @@ options:
     default: null
   region:
     description:
-      - The AWS region to use.  Must be specified if ec2_url is not used. If not specified then the value of the EC2_REGION environment variable, if any, is used. See U(http://docs.aws.amazon.com/general/latest/gr/rande.html#ec2_region)
+      - The AWS region to use. If not specified then the value of the EC2_REGION environment variable, if any, is used. See U(http://docs.aws.amazon.com/general/latest/gr/rande.html#ec2_region)
     required: false
     default: null
     aliases: [ 'aws_region', 'ec2_region' ]
@@ -47,7 +47,7 @@ EXAMPLES = '''
 
 # Ensure that the VPC has an Internet Gateway.
 # The Internet Gateway ID is can be accessed via {{igw.gateway_id}} for use in setting up NATs etc.
-module: ec2_vpc_igw
+ec2_vpc_igw:
   vpc_id: vpc-abcdefgh
   state: present
 register: igw
--- a/cloud/amazon/ec2_vpc_igw.py
+++ b/cloud/amazon/ec2_vpc_igw.py
@@ -20,18 +20,25 @@ short_description: Manage an AWS VPC Internet gateway
 description:
     - Manage an AWS VPC Internet gateway
 version_added: "2.0"
-author: Robert Estelle, @erydo
+author: Robert Estelle (@erydo)
 options:
   vpc_id:
     description:
       - The VPC ID for the VPC in which to manage the Internet Gateway.
     required: true
     default: null
+  region:
+    description:
+      - The AWS region to use.  Must be specified if ec2_url is not used. If not specified then the value of the EC2_REGION environment variable, if any, is used. See U(http://docs.aws.amazon.com/general/latest/gr/rande.html#ec2_region)
+    required: false
+    default: null
+    aliases: [ 'aws_region', 'ec2_region' ]
   state:
     description:
       - Create or terminate the IGW
     required: false
     default: present
+    choices: [ 'present', 'absent' ]
 extends_documentation_fragment: aws
 '''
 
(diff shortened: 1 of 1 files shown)


cloud/amazon/ec2_vpc_igw.py
```diff
@@ -116,8 +114,8 @@ def main():
     argument_spec = ec2_argument_spec()
     argument_spec.update(
         dict(
-            vpc_id = dict(required=True),
-            state = dict(choices=['present', 'absent'], default='present')
+            vpc_id = dict(required=True, default=None),
```
willthames, line 123: why have a `default` if `required`?

cloud/amazon/ec2_vpc_igw.py
```diff
@@ -116,8 +114,8 @@ def main():
...
+            state = dict(required=False, default='present', choices=['present', 'absent'])
```
willthames, line 124: Not sure of the reason behind this change either - `required=False` is default and everything else is just reordering

cloud/amazon/ec2_vpc_igw.py
```diff
@@ -39,16 +40,13 @@
...
-# in setting up NATs etc.
-      local_action:
-        module: ec2_vpc_igw
-        vpc_id: {{vpc.vpc_id}}
-        region: {{vpc.vpc.region}}
-        state: present
-      register: igw
-'''
+# The Internet Gateway ID is can be accessed via {{igw.gateway_id}} for use in setting up NATs etc.
+module: ec2_vpc_igw
```
willthames: If not using `local_action` (probably because of `connection: local` elsewhere), then just use

```
ec2_vpc_igw:
  vpc_id: vpc-abcdef12
  region: ap-southeast-2
register: igw
```

(added region back because it's a required argument)

cloud/amazon/ec2_vpc_igw.py
```diff
@@ -133,14 +131,14 @@ def main():
 
     if region:
```
willthames, line 138: `region` should be added to the documentation too (it's required for modules that use the `ec2_argument_spec` rather than `aws_argument_spec`)

cloud/amazon/ec2_vpc_igw.py
wimnat: Disagree - you can set EC2_REGION as an environment variable or as part of your boto profile. I haven't included region in other aws module examples.

cloud/amazon/ec2_vpc_igw.py
wimnat, line 138: Yep, done

cloud/amazon/ec2_vpc_igw.py
wimnat, line 124: The change here was adding choices for state

cloud/amazon/ec2_vpc_igw.py
wimnat, line 123: My coding style but i take your point

cloud/amazon/ec2_vpc_igw.py
willthames, line 124: It had choices for state in there previously.

cloud/amazon/ec2_vpc_igw.py
willthames: Agree with your point on region. I'd still get rid of `module:` though.

cloud/amazon/ec2_vpc_igw.py
```diff
@@ -20,18 +20,25 @@
...
+author: Robert Estelle (@erydo)
 options:
   vpc_id:
     description:
       - The VPC ID for the VPC in which to manage the Internet Gateway.
     required: true
     default: null
+  region:
+    description:
+      - The AWS region to use.  Must be specified if ec2_url is not used. If not specified then the value of the EC2_REGION environment variable, if any, is used. See U(http://docs.aws.amazon.com/general/latest/gr/rande.html#ec2_region)
```
willthames: " Must be specified if ec2_url is not used." should probably be removed - as you rightly point out, it could be in boto profile - I realise this is likely wrong in a lot of other modules too.

wimnat: Yep agree. I pulled it straight from the ec2 module -- never the best point of reference ;)

cloud/amazon/ec2_vpc_igw.py
wimnat: Done

cloud/amazon/ec2_vpc_igw.py
wimnat, line 124: You're right. Sorry, not sure what's happened here!

gregdek:
@erydo please review; if you +1 we will merge.

cloud/amazon/ec2_vpc_igw.py
```diff
@@ -133,14 +137,14 @@ def main():
...
-            connection = connect_to_aws(boto.ec2, region, **aws_connect_params)
+            connection = connect_to_aws(boto.vpc, region, **aws_connect_params)
         except (boto.exception.NoAuthHandlerFound, StandardError), e:
             module.fail_json(msg=str(e))
     else:
         module.fail_json(msg="region must be specified")
 
     vpc_id = module.params.get('vpc_id')
-    state = module.params.get('state', 'present')
+    state = module.params.get('state')
```
rwe, line 147: If this change is done because of the `default='present'` declaration above, it would be more clear to do `module.params['state']` to indicate that you are not expecting `None` back here.

rwe:
@gregdek I've added one additional comment but otherwise @willthames comments do not appear to have been addressed yet. I'd suggest reverting those specific changes (the unnecessary "required"/"default" changes) but otherwise the doc changes look fine.

wimnat:
OK.  All changes except the doc changes and the aws call fix have been discarded.

rwe:
Looks good to me :+1:

wimnat:
Ooops! Deleted this branch because i thought it was merged but it hadn't been so reopened.

@bcoca can you please merge before 2.0 release because without this bugfix the module doesn't work at all.

gregdek:
I'm presuming this is still good to be merged, since @erydo gave his original +1.

Marking for inclusion. cc: @ansible/core.

abadger:
I noticed that some of the changes that were requested to be reverted hadn't been.  So I just committed that follow on patch to do so.

Thanks for guiding this bugfix through a long process and getting it out the other side!

# Merged

Hi!

issue is resolved.

Thank you!

PR #653, opened 2015-06-29
New module - s3_bucket

wimnat:
New module for s3 buckets.  Could potentially deprecate the functionality of bucket create/destroy in current s3 module.

This module allows setting a policy, versioning, requester pays and tags on the bucket.

More PRs to come for modules to address s3 lifecycle, s3 website and s3 cors

### DIFF
--- a/cloud/amazon/s3_bucket.py
+++ b/cloud/amazon/s3_bucket.py
@@ -103,7 +103,7 @@ import xml.etree.ElementTree as ET
 
 try:
     import boto.ec2
-    from boto.s3.connection import OrdinaryCallingFormat
+    from boto.s3.connection import OrdinaryCallingFormat, Location
     from boto.s3.tagging import Tags, TagSet
     from boto.exception import BotoServerError, S3CreateError, S3ResponseError
     HAS_BOTO = True
@@ -248,7 +248,7 @@ def create_bucket(connection, module):
         else:
             current_tags_dict = dict((t.key, t.value) for t in current_tags[0])
 
-        if sorted(current_tags_dict) != sorted(tags):
+        if current_tags_dict != tags:
             try:
                 if tags:
                     bucket.set_tags(create_tags_container(tags))
@@ -386,7 +386,5 @@ def main():
 from ansible.module_utils.basic import *
 from ansible.module_utils.ec2 import *
 
-# this is magic, see lib/ansible/module_common.py
-#<<INCLUDE_ANSIBLE_MODULE_COMMON>>
-
-main()
+if __name__ == '__main__':
+    main()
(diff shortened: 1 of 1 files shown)


gregdek:

cc: @jsmartin @scicoin-project @tombamford @garethr @lorin @jarv @jsdalton @silviud @adq @zbal @zeekin @willthames @lwade @carsongee @defionscode @tastychutney @bpennypacker @loia

bpennypacker:
Why would we want to deprecate the existing bucket create/destroy functionality in the existing s3 module rather than simply extending the current s3 module to include optional support for versioning, requestor pays, etc?

wimnat:
Simplicity mainly.  I think one module should do one thing well.  In the case of 's3', that should be, IMO, getting and putting keys, setting metadata on keys, managing key acl etc.

Bucket management is a completely separate task so i feel there should be a separate module.

As an example, 'version' has recently been added to s3 module to allow getting of a specific version of a key.  How would we clearly add another module attribute for enabling versioning on the bucket?

gregdek:
So this could _potentially_ mean deprecating functionality and breaking users who are already using the current s3 module? But in the meantime, this functionality would exist in both?

If that's the case, I'm a bit hesitant to recommend it -- because either we'll carry both in the long term, or someone's going to choose the "wrong" one and be forced to refactor. 

Not necessarily against it, but it seems like a decision to be made with our eyes open. Calling in some backup here -- @ansible/core team, assemble!

wimnat:
Functionality duplication is not ideal but it already exists now with ec2_vpc and ec2_vpc_net. As long as deprecation is clearly documented then I think that's OK. I wouldn't envisage current s3 module create functionality being dropped for a long time.

bpennypacker:
I wasn't aware of the ec2_vpc_net module and after looking at it I'm a little surprised that it was added to 2.0 since as you indicated it seems to just replicate existing functionality in the ec2_vpc module (perhaps I'm missing something).  In any event, as you also said:

> As long as deprecation is clearly documented then I think that's OK.

I agree wholeheartedly with this statement, but I also think we shouldn't just simply start deprecating functionality because we can.  If we're going to start deprecating functionality in order to make the modules more, for lack of a better term, _modular_, then I think we need to first lay out a well documented roadmap that with all the modules in mind and not just the AWS ones. In order to do that I agree with @gregdek that we need some guidance from the @ansible/core team before we proceed.

juliedavila:
@bpennypacker This was added because ec2_vpc had become too monolithic and it was becoming difficult to maintain.

willthames:
@bpennypacker @gregdek that's pretty much how I feel regarding ansible/ansible-modules-core#1143

If there is a general preference for more testable modules, that's fine, but we should have a guidance document for how we wish this to be achieved (or at least point to a module doing things right, and extract best practices from that). There may also need to be some AWS specific testing documentation around testing AWS modules (e.g. mocking AWS?)

kalefranz:
+1 on the functionality this module adds. The current s3 module doesn't do this now, and probably never will. Merge this already.

PR #657, opened 2015-06-29
Adding win_regedit module.

smadam813:
I have written a module to Add, Edit, and Remove registry keys.  This is in a working state, but has some kinks that I still need to work out such as permissions to modify keys in HKLM and value types.  I figured I would get the ball rolling as see what everyone thinks.

### DIFF
--- a/windows/win_regedit.ps1
+++ b/windows/win_regedit.ps1
@@ -1,0 +1,173 @@
+#!powershell
+# This file is part of Ansible
+#
+# (c) 2015, Adam Keech <akeech@chathamfinancial.com>, Josh Ludwig <jludwig@chathamfinancial.com>
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+$ErrorActionPreference = "Stop"
+
+# WANT_JSON
+# POWERSHELL_COMMON
+
+$params = Parse-Args $args;
+$result = New-Object PSObject;
+Set-Attr $result "changed" $false;
+
+If ($params.name)
+{
+    $registryKeyName = $params.name
+}
+Else
+{
+    Fail-Json $result "missing required argument: name"
+}
+
+If ($params.state)
+{
+    $state = $params.state.ToString().ToLower()
+    If (($state -ne "present") -and ($state -ne "absent"))
+    {
+        Fail-Json $result "state is $state; must be present or absent"
+    }
+}
+Else
+{
+    $state = "present"
+}
+
+If ($params.value)
+{
+    $registryKeyValue = $params.value
+}
+ElseIf ($state -eq "present")
+{
+    Fail-Json $result "missing required argument: value"
+}
+
+If ($params.valuetype)
+{
+    $registryValueType = $params.valuetype.ToString().ToLower()
+    $validRegistryValueTypes = "binary", "dword", "expandstring", "multistring", "string", "qword"
+    If ($validRegistryValueTypes -notcontains $registryValueType)
+    {
+        Fail-Json $result "valuetype is $registryValueType; must be binary, dword, expandstring, multistring, string, or qword"
+    }
+}
+Else
+{
+    $registryValueType = "string"
+}
+
+If ($params.path)
+{
+    $registryKeyPath = $params.path
+}
+Else
+{
+    Fail-Json $result "missing required argument: path"
+}
+
+Function Test-RegistryValue {
+    Param (
+        [parameter(Mandatory=$true)]
+        [ValidateNotNullOrEmpty()]$Path,
+        [parameter(Mandatory=$true)]
+        [ValidateNotNullOrEmpty()]$Value
+    )
+    Try {
+        Get-ItemProperty -Path $Path -Name $Value
+        Return $true
+    }
+    Catch {
+        Return $false
+    }
+}
+
+if($state -eq "present") {
+    if (Test-Path $registryKeyPath) {
+        if (Test-RegistryValue -Path $registryKeyPath -Value $registryKeyName)
+        {
+            # Changes Type and Value
+            If ((Get-Item $registryKeyPath).GetValueKind($registryKeyName) -ne $registryValueType)
+            {
+                Try
+                {
+                    Remove-ItemProperty -Path $registryKeyPath -Name $registryKeyName
+                    New-ItemProperty -Path $registryKeyPath -Name $registryKeyName -Value $registryKeyValue -PropertyType $registryValueType
+                    $result.changed = $true
+                }
+                Catch
+                {
+                    Fail-Json $result $_.Exception.Message
+                }
+            }
+            # Only Changes Value
+            ElseIf ((Get-ItemProperty -Path $registryKeyPath | Select-Object -ExpandProperty $registryKeyName) -ne $registryKeyValue) 
+            {
+                Try {
+                    Set-ItemProperty -Path $registryKeyPath -Name $registryKeyName -Value $registryKeyValue
+                    $result.changed = $true
+                }
+                Catch
+                {
+                    Fail-Json $result $_.Exception.Message
+                }
+            }
+        }
+        else
+        {
+            Try
+            {
+                New-ItemProperty -Path $registryKeyPath -Name $registryKeyName -Value $registryKeyValue -PropertyType $registryValueType
+                $result.changed = $true
+            }
+            Catch
+            {
+                Fail-Json $result $_.Exception.Message
+            }
+        }
+    }
+    else
+    {
+        Try 
+        {
+            New-Item $registryKeyPath -Force | New-ItemProperty -Name $registryKeyName -Value $registryKeyValue -Force -PropertyType $registryValueType
+            $result.changed = $true
+        }
+        Catch
+        {
+            Fail-Json $result $_.Exception.Message
+        }
+    }
+}
+else
+{
+    if (Test-Path $registryKeyPath)
+    {
+        if (Test-RegistryValue -Path $registryKeyPath -Value $registryKeyName) {
+            Try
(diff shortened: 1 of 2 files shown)


gregdek:
LOL -- a teaser in the title!

Putting this is needs_revision state so we know not to review it yet. @smadam813 let us know when you're done with it and we'll move it into community_review.

smadam813:
@gregdek I added the functionality I wanted to and the permissions are actually a non-issue.  This is ready for community_review.  I squashed my commits together since this functionality should have been working at the pull requests time.

gregdek:
Thanks @smadam813.

Windows reviewers assemble! @trondhindenes @petemounce @elventear @jhawkesworth @angstwad @sivel @chrishoffman @cchurch 

(since @smadam813 is now a reviewer himself, we only need one +1 to merge)

jhawkesworth:
Looking very good, although I've not tested it yet.
Couple of minor things to think about, which may or may not actually need any change.

I'd consider retrieving the changed key/value after making a modification and comparing it as a way of guarding against silent failure (don't know if that is a real possibility in this case but I've had occasions in the past when merging into registry no error was returned but errors in one of the entries caused subsequent ones not to be merged).

Also, I'd consider sanitizing 
    Fail-Json $result $_.Exception.Message
primarily because that Exception message has to be converted into valid Json and I've had occasions where the exception message failed to convert due to a trailing backslash.  I think the trailing backslash case is handled now but I can imagine unbalanced quotes or commas upsetting the json conversion.

smadam813:
@jhawkesworth Thank you for the input!  As far as registry merging, my research and testing leads to it being handled behind the scenes with Powershell 3.0+.  The merge happens on the set and will produce an exception if the merge fails.  (I have yet to see a Set-ItemProperty failure)  I remember on earlier versions you would have to manage the hive and commit etc etc.  It was miserable :-(

Could you clarify what you mean with sanitizing the Fail-Json?  I see this method used a lot in the windows modules and have never experienced any issues with it.  I am new to this project though so clarification would help and I would be happy to fix :-)

elventear:
@smadam813 I think @jhawkesworth is referring to escaping a string according to JSON standards. I think you can use [`ConvertTo-Json`](https://technet.microsoft.com/en-us/library/hh849922.aspx) for that.

jhawkesworth:
Thanks for this.  Sounds like there's nothing to do for the registry
merging then.

I have hit problems in the past with passing exceptions straight to
Fail-Json, although I think these are handled now that we have switched to
calling ConvertTo-Json with the -Compress switch (which works around a bug
in ConvertTo-Json where it can't handle trailing backslashes).  Its a minor
thing really and just depends on whether the exception messages that are
likely to get thrown will convert into valid json or not.

gregdek:
@jhawkesworth does that constitute a +1 from you, or...?

smadam813:
@elventear and @jhawkesworth So something like?

```
Fail-Json $result $_.Exception.Message
```

to

```
Fail-Json $result ($_.Exception.Message | ConvertTo-Json -Compress)
```

smadam813:
@elventear and @jhawkesworth  After testing, the only difference appears to be how characters are escaped.  Please let me know if you would like this added and I will commit the change.

trondhindenes:
@smadam813 you shouldn't have to pipe fail-json to convertto-json.
`Fail-Json $result ($_.Exception.Message)` should be enough.

smadam813:
Sounds good @trondhindenes  I will leave it alone then.

jhawkesworth:
Hang on @gregdek - nearly there I think.  I've been testing this and this is probably me not knowing enough about the registry, but how could I create a new Key without a Data value?  (This often seems to be the case where the key is being used as container for multiple key-value pairs).   Since the module requires a value to be supplied, I guess this isn't possible.  Would be nice either to document this or figure out a way round it (perhaps have an optional keypath or keyname parameter which would allow create/removal of a key - in which case value would not always be mandatory).

So, if HKLM:\SOFTWARE\Wow6432Node\ didn't yet have a key called say Ansible, how could I create one without setting a value for it.

I tried 

win_regedit -a 'name=(Default) path=HKLM:\SOFTWARE\Wow6432Node\Ansible value=(value not set)'

The key got created but with a value of '(value', which is sort of understandable as I was trying to get the module to do something that it doesn't do, but raises the question, do values containing spaces need escaping or double quoting?  If so an example in the docs would be great.

Actually thinking about this this a bit further, I wonder if it would be an idea to keep the terminology the same as you see if you run regedit.  In regedit, Key only refers to the container type thing I'm trying to create above, and the other things are referred to as Values - 'String Value, Binary Value, DWORD (32-bit) Value, QWORD (64-bit) Value, Multi-String Value, Expandable String Value), and each of the Value types has a Name, Type and Data.

So for the module I suggest renaming what you've called 'value' to 'data', 'valuetype' to 'type' and in the module docs I suggest changing the comments above the examples to make it clear this is about manipulating the registry values, rather than the key names.
e.g change
 # Add Registry Key with Type DWord 
to
 # Add Registry Value with Type DWord

trondhindenes:
I agree with @jhawkesworth . A registry "key" is the "container" object, in which one can store KVPs constisting of "name" and "value".

smadam813:
@jhawkesworth Thank you for your indepth description!  I have made the changes you have suggested.  Please let me know what you think!

jhawkesworth:
Great, thanks for turning this round so quickly.  Will look as soon as I
can.

bcoca:
ping @trondhindenes @petemounce @elventear @jhawkesworth @angstwad @sivel @chrishoffman @cchurch

jhawkesworth:
Ok, did another test.  This is looking really good but... if you run 

win_regedit -a 'name=myFact path=HKLM:\SOFTWARE\Wow6432Node\Ansible data=version1 state=absent'

and the key Ansible doesn't exist under HKLM:\SOFTWARE\Wow6432Node\, the key is created, with the Name and Data value specified. but if you then run

win_regedit -a 'name=myFact path=HKLM:\SOFTWARE\Wow6432Node\Ansible data=version1 state=absent' 
the Name and Data Value are removed, but not the Key, so its a bit asymmetrical now.

I'm wondering if managing the Keys should either be a separate module or handled with explicit module parameters, rather than automagically.  I think creating registry keys much less common occurrence than wanting to edit Names and Data values, but I think we would need the capability to do both.

Otherwise, it looks really good though.

Hope this helps.

smadam813:
@jhawkesworth That is a good point.  I originally left that functionality out due to the extra logic to determine if a key really should be deleted or just the value / data of a key.

When I run win_regedit -a 'name=myFact path=HKLM:\SOFTWARE\Wow6432Node\Ansible data=version1 state=absent', the Ansible key is not created if it wasn't already there. If that key does exist, I can confirm it does not delete as it only currently deletes the data. What I will try to do is if you run with data=version1, it will delete the version1 data, otherwise if you leave data blank, it will delete the key Ansible. How does this sound?

smadam813:
@jhawkesworth When I run win_regedit -a 'name=myFact path=HKLM:\SOFTWARE\Wow6432Node\Ansible data=version1 state=absent', the Ansible folder is not created if it is not there.  What I will try to do is if you run with data=version1, it will delete the version1 data, otherwise if you leave data blank, it will delete the key Ansible.  How does this sound?

trondhindenes:
Hmm, we might have to revisit this one to provide functionality for manipulating both keys and values without confusing the users too much down the road. IMHO the optimal would be:
key=XXXX state=absent/present: forces the key to exist/not exist
key=XXXX value=XXXX state=absent/present forces the value to exist/not exist. Creates the key(s) if missing. Leaves the key alone if state=absent
key=XXXX value=XXXX data=XXXX state=present. Absent not valid, use the above to wipe value only. I don't know if that represents too much of a change from your current code for it to be feasable for this PR.

smadam813:
This is fine for this pull request.  I will make the changes this afternoon.

smadam813:
@trondhindenes @jhawkesworth I have added all the functionality that you guys have suggested.  I agree this is much clearer now with all the functionality intended.  The uses are as follows:

```
  # Creates Registry Key called MyCompany.
  win_regedit:
    key: HKCU:\Software\MyCompany

  # Creates Registry Key called MyCompany,
  # a value within MyCompany Key called "hello", and
  # data for the value "hello" containing "world".
  win_regedit:
    key: HKCU:\Software\MyCompany
    value: hello
    data: world

  # Creates Registry Key called MyCompany,
  # a value within MyCompany Key called "hello", and
  # data for the value "hello" containing "1337" as type "dword".
  win_regedit:
    key: HKCU:\Software\MyCompany
    value: hello
    data: 1337
    datatype: dword

  # Delete Registry Key MyCompany
  # NOTE: Not specifying a value will delete the root key which means
  # all values will be deleted
  win_regedit:
    key: HKCU:\Software\MyCompany
    state: absent

  # Delete Registry Value "hello" from MyCompany Key
  win_regedit:
    key: HKCU:\Software\MyCompany
    value: hello
    state: absent
```

trondhindenes:
Fantastic job @smadam813. I'll do some testing right away.

trondhindenes:
again, fantastic job @smadam813. As far as I can see this is ready to be merged. :+1:

smadam813:
Thanks @trondhindenes!  I appreciate all the feedback.

jhawkesworth:
+1
Just gave this another test and its working great.
This is going to be very handy addition, many thanks @smadam813 (and everyone else)
@gregdek - looks like you have 2 +1s now

smadam813:
Great!  Thanks for the review @jhawkesworth!

acaire:
Awesome module, thanks @smadam813.

Let me know if I should open a new issue, I just thought it'd be worthwhile voicing some quick concerns here - hopefully before 2.0 is released.

1) Is it worth renaming this module? Regedit refers to an application to edit the registry, win_registry would seem more appropriate.

2) I didn't see the earlier code from when @jhawkesworth suggested changing some param names - But If we were to trust Regedit and Wikipedia as a source of truth, Registry Keys are the containers to 'Value Name' and 'Value Data' entries.  IMHO the 'value' param should be renamed to 'name', and subsequently datatype may as well just be 'type'.

3) Following on from part of @jhawkesworth's comment, I hit similar confusion with default values but when needing to specifically set data(see 'txtfile' in the example below).  In a registry export, the '(Default)' value name is expressed as an unquoted at-sign. for example:

```
[HKEY_LOCAL_MACHINE\SOFTWARE\Classes\.log]
@="txtfile"
```

A small PR is required to correctly set a '(Default)' value as per http://tfl09.blogspot.com.au/2007/05/getting-registry-value-entry-with.htm, but I just wanted to confirm the desired behaviour.

Chef's Powershell module sets the Value Data entry of a key if an empty string representing the key's 'Value Name' is passed.  Should the default value behave the same if '@', '(Default)', or '' is passed?

PR #658, opened 2015-06-29
files/confparse - alternate ini parser

ritzk:
config parser for *nix configuration files, with minimal support for config files with sections. Allows one to read and write to config file.


gregdek:

PR #665, opened 2015-06-30
Ability to add/remove scheduled task

perzizzle:
Adds the ability to add/remove scheduled tasks while preserving the ability to enable/disable tasks. I've made some simplifying assumptions but thought I'd put this out here to validate the general approach I took since it doesn't have much similarity to the original module. It no longer supports wildcards in the name to disable many scheduled tasks. Also, the hack to determine if a scheduled task already exists works when run locally but fails when run via ansible. Finally, idemptotence isn't perfectly implemented. I can add these features if people agree with the approach.


perzizzle:
Check it out @bcoca @gregdek @petemounce 

These changes discussed at https://github.com/ansible/ansible-modules-extras/pull/441

gregdek:
@perzizzle OK, looks like there was a docs issue with bundler that was not related to this PR; please rebase to latest to get pass the Travis failure. Sorry about that.

perzizzle:
Looks like the build succeeded now, I wasn't paying attention and merged devel into my branch rather than rebasing. Hope that doesn't matter...

gregdek:
@petemounce please review; if you +1 we will merge.

perzizzle:
@petemounce to be clear I think it needs a bit more work to make it truly idempotent. It doesnt' currently handle just a change in a trigger (and only a subset of possible triggers). Just wanted to get this out here before spending more time on it in case there were other opinions.

gregdek:
So @perzizzle you're hoping for this patch in the short term, and then work on a fully idempotent solution in the longer term?

perzizzle:
Alright fixed a few issues. The only thing I know of outstanding is the fact changes to a trigger are not idempotent. This could be patched in later. The biggest difference (outside of the new features) between this module and @petemounce version is that one run can only impact 1 scheduled task at a time. You can no longer disable many scheduled tasks at a time.

perzizzle:
I also wasn't sure what the standard is for the author field in the documentation and comments. This module doesn't share much with the original so not sure if pete wants his name attached or not :)

gregdek:
Sorry for letting this sit so long, @perzizzle. Still some kinks in the workflow to be smoothed over.

@petemounce Please review this PR to make sure it adheres to the following guidelines:

windows/win_scheduled_task.py
```diff
@@ -32,20 +29,45 @@
   name:
     description:
       - Name of the scheduled task
-      - Supports * as wildcard
     required: true
   enabled:
```
petemounce, line 37: Should include `required: false`

windows/win_scheduled_task.py
```diff
@@ -32,20 +29,45 @@
...
     description:
-      - State that the task should become
-    required: false
+      - Enable/disable the task
     choices:
       - yes
       - no
     default: yes
-author: Peter Mounce
+  state:
```
petemounce, line 44: Should include `required: true` I think

windows/win_scheduled_task.py
```diff
@@ -32,20 +29,45 @@
...
+      - State that the task should become
+    choices:
+      - present
+      - absent
+  execute:
+    description:
+      - Command the scheduled task should execute
+    required: false
+  frequency:
+    description:
```
petemounce, line 64: Should include `required: false` I think

windows/win_scheduled_task.py
```diff
@@ -32,20 +29,45 @@
...
+      - weekly
+  time:
+    description:
+      - Time to execute scheduled task
+    required: false
+  daysOfWeek:
+    description:
+      - Days of the week to run a weekly task
+    required: false
+  path:
```
petemounce, line 78: Should include `required: false`

windows/win_scheduled_task.ps1
```diff
@@ -33,40 +34,161 @@ else
...
+    elseif ($measure.count -eq 0 -and $state -eq "absent" ){
+        Set-Attr $result "msg" "Task does not exist"
+        Exit-Json $result
+    }
+    elseif ($measure.count -eq 0){
+        $exists = $false
+    }
+    else {
+        # This should never occur
+        Fail-Json $result "$measure.count scheduled tasks found"
```
petemounce: I think this breaks interpolation - should be 

``` powershell
Fail-Json $result "$($measure.count) scheduled tasks found"
```

windows/win_scheduled_task.ps1
```diff
@@ -33,40 +34,161 @@ else
...
+        }
+        elseif ($frequency -eq "weekly"){
+            $trigger =  New-ScheduledTaskTrigger -Weekly -At $time -DaysOfWeek $daysOfWeek
+        }
+        else {
+            Fail-Json $result "frequency must be daily or weekly"
+        }
+    }
+
+    if ($state -eq "absent" -and $exists -eq $true) {
```
petemounce: I _think_ this needs to be 

``` powershell
if (($state -eq "absent") -and ($exists -eq $true)) {
```

?

windows/win_scheduled_task.ps1
```diff
@@ -33,40 +34,161 @@ else
...
+        Unregister-ScheduledTask -TaskName $name -Confirm:$false
+        $result.changed = $true
+        Set-Attr $result "msg" "Deleted task $name"
+        Exit-Json $result
+    }
+    elseif ($state -eq "absent" -and $exists -eq $false) {
```
petemounce: As above

windows/win_scheduled_task.ps1
perzizzle: It doesn't need to be but for readability it makes sense. I'll update.

petemounce:
I'm easy :-)

##

windows/win_scheduled_task.py
perzizzle, line 37: Confused by this comment, name is a required field.

```
$name = Get-Attr -obj $params -name name -failifempty $true -resultobj $result
```

windows/win_scheduled_task.py
perzizzle, line 44: I set this to default to true since creating disabled scheduled tasks seems unlikely.

```
if ($params.enabled)
{
  $enabled = $params.enabled | ConvertTo-Bool
}
else
{
  $enabled = $true #default
}
```

windows/win_scheduled_task.py
perzizzle, line 64: Updated

windows/win_scheduled_task.ps1
perzizzle: I'm not sure how this condition would ever occur (the UI and powershell won't let you create 2 tasks with the same name). That being said I updated the line.

perzizzle:
Alright @petemounce @gregdek I updated the documentation and the couple syntax suggestions pete had as well as added support for arguments and user of your scheduled task. I can post a gist of my tests if that'll help. I think I ran through ever possibility this morning without issue.

petemounce:
+1 shipit - thanks!

gregdek:
@petemounce thanks for the review, and @perzizzle thanks for the PR. 

Marking for inclusion.

gregdek:
cc: @ansible/core team.

gregdek:
OK, merged. @perzizzle may be worth double-checking those params to make sure all optional params have defaults, but I didn't want to block this PR for that.

PR #670, opened 2015-07-01
webfaction: Allow machine to be specified if account has more than one.

quentinsf:
If a webfaction account is associated with more than one server, this allows you to specify on which one an app/database should be created.


gregdek:
Merging as owner_pr.

PR #118, opened 2014-11-24
Hipchat v2 + legacy v1 support

mikemilano:
I updated the hipchat module to work with v2, as well as v1. There is a new module parameter, `version` which can be set to `1` or `2`.

I no longer have a v1 account so I am not able to test if this will still work for v1. If not, it should be minor fixes.

I did not get `notify` to work with the v2 API. It defaults to false.  The API specifies it wants a boolean value but does not accept anything. I tested it in with numeric, string, and boolean values from the plugin code, as well as tested with postman (chrome app).  All values I tested generate a bad request error stating the valid type should be boolean. `notify` will be ignored right now when `version` is set to 2.

The API changes are as follows:
- base changed from: `https://api.hipchat.com/v1` to `https://api.hipchat.com/v2`
- path changed from: `rooms/message` to `room/id_or_name/notification`
- method changed from `POST` or `GET` to `POST` only
- `room_id` is no longer a parameter since it is defined in the path
- `from` is no longer a parameter. This is now the user who owns the token, or the notification label setup with the token.
- `format` is no longer a parameter (not to be confused with `message_format`, which is still valid

V1: https://www.hipchat.com/docs/api/method/rooms/message
V2: https://www.hipchat.com/docs/apiv2/method/send_room_notification

PR #674, opened 2015-07-02
hipchat: Fix user notification for v2 api

hostmaster:
`notify` parameter is not working as expected for hipchat API v2. Parameter `notify: no` enables user notification and vise versa. In fact user notification is disabled by default. 

I think there is a little misunderstanding here. [Send message](https://www.hipchat.com/docs/apiv2/method/send_message) request is not supposed to be an equivalent of [room/message](https://www.hipchat.com/docs/api/method/rooms/message) from v1.It simply does not support all the parameters. [Send room notification](https://www.hipchat.com/docs/apiv2/method/send_room_notification) is a right one. It supports `notify` parameter to configure a user notification.

### DIFF
--- a/notification/hipchat.py
+++ b/notification/hipchat.py
@@ -5,7 +5,7 @@ DOCUMENTATION = '''
 ---
 module: hipchat
 version_added: "1.2"
-short_description: Send a message to hipchat
+short_description: Send a message to hipchat.
 description:
    - Send a message to hipchat
 options:
@@ -56,7 +56,7 @@ options:
     version_added: 1.5.1
   api:
     description:
-      - API url if using a self-hosted hipchat server
+      - API url if using a self-hosted hipchat server. For hipchat api version 2 use C(/v2) path in URI
     required: false
     default: 'https://api.hipchat.com/v1'
     version_added: 1.6.0
@@ -67,7 +67,15 @@ author: "WAKAYAMA Shirou (@shirou), BOURDEL Paul (@pb8226)"
 '''
 
 EXAMPLES = '''
-- hipchat: token=AAAAAA room=notify msg="Ansible task finished"
+- hipchat:  room=notify msg="Ansible task finished"
+
+# Use Hipchat API version 2
+
+- hipchat: 
+    api: "https://api.hipchat.com/v2/"
+    token: OAUTH2_TOKEN
+    room: notify 
+    msg: "Ansible task finished"
 '''
 
 # ===========================================
@@ -80,7 +88,6 @@ DEFAULT_URI = "https://api.hipchat.com/v1"
 
 MSG_URI_V1 = "/rooms/message"
 
-MSG_URI_V2 = "/room/{id_or_name}/message"
 NOTIFY_URI_V2 = "/room/{id_or_name}/notification"
 
 def send_msg_v1(module, token, room, msg_from, msg, msg_format='text',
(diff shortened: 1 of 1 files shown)


pb8226:
Thanks for the changes. Do you have a playbook that you used for testing? I want to make sure I test the scenario you're fixing.

hostmaster:
I don't think there is a scenaruio. I mean I used version from https://github.com/ansible/ansible-modules-extras/pull/118 before and then I just updated the module from upstream and found that `notify` is disabled now.

I'm using `api: https://api.hipchat.com/v2/`

hostmaster:
@pb8226 however I think i can create something like that if it helps.

pb8226:
@hostmaster Please if you can, no rush obviously.  I have a playbook I used for testing but it obviously didn't capture what you're testing so I wanted to run my playbook and then yours to see where the difference was.

hostmaster:
@pb8226 ok so just to clarify. you are using hipchat api v2 and user notification is working as expected right ?

pb8226:
This was the playbook I used for testing:

https://www.dropbox.com/s/sk2580g7j0iv56a/test-playbook.yml?dl=0

bcoca:
@pb8226 is this OK to merge? @shirou can you also review?

pb8226:
@bcoca I was waiting for a playbook from @hostmaster that showed the bug he was fixing.  Notifications were working for me with the previous code and I provided him the playbook I used for testing.  I haven't heard back after providing the playbook to him.

hostmaster:
@pb8226 ok I have  created  a slightly modified version of your playbook (I put hipchat settings in a separate file ) https://gist.github.com/hostmaster/3d554473be59bfc75894

hostmaster:
@pb8226 Here is my result

```
$ ansible-playbook -i localhost test_playbook.yml

PLAY [Test HipChat Module] ****************************************************

TASK: [Notify Room V2] ********************************************************
changed: [localhost]
```

Message is sent, **no notification**, color red 

```
TASK: [Notify Room V2] ********************************************************
changed: [localhost]
```

Message is sent, **no notification**, color purple

```
TASK: [Message Room V2] *******************************************************
changed: [localhost]
```

Message is sent, **user notification**, no color

```
TASK: [Notify Room V1] ********************************************************
changed: [localhost]

TASK: [Notify Room V1] ********************************************************
changed: [localhost]

TASK: [Message Room V1] *******************************************************
changed: [localhost]

localhost                  : ok=6    changed=6    unreachable=0    failed=0
```

hostmaster:
@pb8226 `notify` parameter is optional the case when it is ommited is not covered in your playbook. In my test it doesn't make user notification for API v2. (it should)

pb8226:
@hostmaster ok thank you, that makes sense. @bcoca this can be merged.

shirou:
FYI: I don't use hipchat anymore, I can not test. But it looks good for me. Thank you for updating!

gregdek:
Thanks @shirou.

@pb8226 would you be interested in taking over maintainership of this module?

hostmaster:
I've just received a clarification from Hipchat support regarding the difference among `Send Message` and `Send Room notification` API calls. 

>    Send Room Notification - Provides the user with more options to configure how the API is being sent out with different content-types, bigger request body, accessible by.
>    Send Message : Simple and straight forward. Only accessible by users and it only has a single request body which is message.
> 
> It depends really how you want to make use of for your integration. If you want more flexibility, you can se /notification. If you want something which is simple and lightweight, you can make use of /message.

PR #677, opened 2015-07-03
Add zfs cloning

wishdev:
Adds the ability to clone a snapshot when creating a zfs volume

### DIFF
--- a/system/zfs.py
+++ b/system/zfs.py
@@ -66,6 +66,10 @@ options:
       - The checksum property.
     required: False
     choices: ['on','off',fletcher2,fletcher4,sha256]
+  clone:
+    description:
+      - Name of the snapshot to clone
+   required: False
   compression:
     description:
       - The compression property.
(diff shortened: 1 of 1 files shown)


gregdek:
@johanwiren Please review this PR to make sure it adheres to the following guidelines:

If it passes these guidelines, and if you believe it’s a good PR otherwise, please give it a +1; at that point, we will merge.

johanwiren:
I think that the "clone" option should be named "origin" to better follow zfs naming.

Also, please add an example to the doc in how to create and destroy clones.

system/zfs.py
```diff
@@ -66,6 +66,10 @@
       - The checksum property.
     required: False
     choices: ['on','off',fletcher2,fletcher4,sha256]
+  clone:
+    description:
+      - Name of the snapshot to clone
+   required: False
```
johanwiren: Wrong indentation.

wishdev:
@johanwiren - Switched the property name, cleaned up the identing issue and added examples. I only added a generic example for destroying a file system because there is nothing special about removing a cloned file system.

johanwiren:
Great work!

@gregdek I approve.

PR #680, opened 2015-07-06
New module - s3_logging

wimnat:
New module to manage logging option of s3 buckets

### DIFF
--- a/cloud/amazon/s3_logging.py
+++ b/cloud/amazon/s3_logging.py
@@ -26,7 +26,6 @@ options:
     description:
       - "Name of the s3 bucket."
     required: true
-    default: null
   region:
     description:
      - "AWS region to create the bucket in. If not set then the value of the AWS_REGION and EC2_REGION environment variables are checked, followed by the aws_region and ec2_region settings in the Boto config file.  If none of those are set the region defaults to the S3 Location: US Standard."
@@ -47,7 +46,7 @@ options:
     description:
       - "The prefix that should be prepended to the generated log files written to the target_bucket."
     required: false
-    default: no
+    default: ""
 
 extends_documentation_fragment: aws
 '''
@@ -141,7 +140,7 @@ def main():
     argument_spec = ec2_argument_spec()
     argument_spec.update(
         dict(
-            name = dict(required=True, default=None),
+            name = dict(required=True),
             target_bucket = dict(required=False, default=None),
             target_prefix = dict(required=False, default=""),
             state = dict(required=False, default='present', choices=['present', 'absent'])
--- a/cloud/amazon/s3_logging.py
+++ b/cloud/amazon/s3_logging.py
@@ -40,8 +40,8 @@ options:
     choices: [ 'present', 'absent' ]
   target_bucket:
     description:
-      - "The bucket to log to."
-    required: true
+      - "The bucket to log to. Required when state=present."
+    required: false
     default: null
   target_prefix:
     description:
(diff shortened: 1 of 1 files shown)


bcoca:
cc @ansible/aws

bcoca:
cc @ansible/aws

gregdek:
cc @ansible/aws

cloud/amazon/s3_logging.py
```diff
@@ -0,0 +1,185 @@
...
+    description:
+      - "Enable or disable logging."
+    required: false
+    default: present
+    choices: [ 'present', 'absent' ]
+  target_bucket:
+    description:
+      - "The bucket to log to."
+    required: true
+    default: null
```
ghost, line 44: According to http://docs.ansible.com/developing_modules.html#module-checklist, default is not needed when required is true.

cloud/amazon/s3_logging.py
```diff
@@ -0,0 +1,185 @@
...
+- name: Enable logging of s3 bucket mywebsite.com to s3 bucket mylogs
+  s3_logging:
+    name: mywebsite.com
+    target_bucket: mylogs
+    target_prefix: logs/mywebsite.com
+    state: present
+
+- name: Remove logging on an s3 bucket
+  s3_logging:
+    name: mywebsite.com
```
ghost, line 66: missing: target_bucket: mylogs here

ghost:
@wimnat I did a some quick testing with this one and it works as advertised. Just a couple of documentation updates needed. The module is requiring target_bucket on absent so I think it should be added to the example.

wimnat:
@rickmendes I've removed the target_bucket requirement and updated the doc to mention it is required for state=present.  I didn't want to have state=absent dependent on target_bucket.

ghost:
@wimnat That works for me. It makes more sense to me to not have it required for state-absent. I give this a +1.

cloud/amazon/s3_logging.py
```diff
@@ -0,0 +1,185 @@
...
+  target_bucket:
+    description:
+      - "The bucket to log to. Required when state=present."
+    required: false
+    default: null
+  target_prefix:
+    description:
+      - "The prefix that should be prepended to the generated log files written to the target_bucket."
+    required: false
+    default: no
```
bcoca: i'm guessing the default here is null or is 'no' the prefix that gets added?

cloud/amazon/s3_logging.py
```diff
@@ -0,0 +1,185 @@
...
+description:
+    - Manage logging facility of an s3 bucket in AWS
+version_added: "2.0"
+author: Rob White (@wimnat)
+options:
+  name:
+    description:
+      - "Name of the s3 bucket."
+    required: true
+    default: null
```
bcoca: no default needed when required is true

cloud/amazon/s3_logging.py
```diff
@@ -0,0 +1,185 @@
...
+   
+    module.exit_json(changed=changed)
+    
+    
+def main():
+    
+    argument_spec = ec2_argument_spec()
+    argument_spec.update(
+        dict(
+            name = dict(required=True, default=None),
```
bcoca: default is never used here as you require the input

cloud/amazon/s3_logging.py
wimnat: Should be an empty string rather than null.

PR #683, opened 2015-07-06
ProfitBricks Compute Module

baldwinSPC:
@bcoca @gregdek Here's the first PR with the base compute module.

### DIFF
--- a/cloud/profitbricks/profitbricks.py
+++ b/cloud/profitbricks/profitbricks.py
@@ -119,7 +119,9 @@ options:
     default: 'present'
     choices: [ "running", "stopped", "absent", "present" ]
 
-requirements: [ "profitbricks" ]
+requirements:
+     - "profitbricks"
+     - "python >= 2.6"
 author: Matt Baldwin (baldwin@stackpointcloud.com)
 '''
 
@@ -178,8 +180,6 @@ EXAMPLES = '''
 import re
 import uuid
 import time
-import json
-import sys
 
 HAS_PB_SDK = True
 
@@ -232,7 +232,7 @@ def _create_machine(module, profitbricks, datacenter, name):
     image = module.params.get('image')
     assign_public_ip = module.boolean(module.params.get('assign_public_ip'))
     wait = module.params.get('wait')
-    wait_timeout = int(module.params.get('wait_timeout'))
+    wait_timeout = module.params.get('wait_timeout')
 
     try:
         # Generate name, but grab first 10 chars so we don't
@@ -298,8 +298,6 @@ def _create_machine(module, profitbricks, datacenter, name):
                                  wait_timeout, "create_virtual_machine")
 
 
-
-        # return (json.dumps(server_response))
         return (server_response)
     except Exception as e:
         module.fail_json(msg="failed to create the new server: %s" % str(e))
@@ -307,7 +305,7 @@ def _create_machine(module, profitbricks, datacenter, name):
 def _remove_machine(module, profitbricks, datacenter, name):
     remove_boot_volume = module.params.get('remove_boot_volume')
     wait = module.params.get('wait')
-    wait_timeout = int(module.params.get('wait_timeout'))
+    wait_timeout = module.params.get('wait_timeout')
     changed = False
 
     # User provided the actual UUID instead of the name.
@@ -349,7 +347,7 @@ def _startstop_machine(module, profitbricks, datacenter, name):
 def _create_datacenter(module, profitbricks):
     datacenter = module.params.get('datacenter')
     location = module.params.get('location')
-    wait_timeout = int(module.params.get('wait_timeout'))
+    wait_timeout = module.params.get('wait_timeout')
 
     i = Datacenter(
         name=datacenter,
@@ -381,7 +379,7 @@ def create_virtual_machine(module, profitbricks):
     auto_increment = module.params.get('auto_increment')
     count = module.params.get('count')
     lan = module.params.get('lan')
-    wait_timeout = int(module.params.get('wait_timeout'))
+    wait_timeout = module.params.get('wait_timeout')
     failed = True
     datacenter_found = False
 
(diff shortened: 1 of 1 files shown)


cloud/profitbricks/profitbricks.py
```diff
@@ -0,0 +1,632 @@
...
+import uuid
+import time
+import json
+import sys
+
+try:
+    from profitbricks.client import ProfitBricksService, Volume, Server, Datacenter, NIC, LAN
+except ImportError:
+    print "failed=True msg='profitbricks required for this module'"
+    sys.exit(1)
```
bcoca: modules should not print nor exist, use a variable to capture the import failure and then use fail_json to give out a the message, look at ec2 or openstack plugins as an example

cloud/profitbricks/profitbricks.py
```diff
@@ -0,0 +1,632 @@
...
+      - remove the bootVolume of the virtual machine you're destroying.
+    required: false
+    default: "yes"
+    choices: ["yes", "no"]
+  state:
+    description:
+      - create or terminate instances
+    required: false
+    default: 'present'
+    aliases: []
```
bcoca: we should show acceptable choices here:
'started', 'stopped', 'absent', 'present'

cloud/profitbricks/profitbricks.py
```diff
@@ -0,0 +1,632 @@
...
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: profitbricks
+short_description: Create, destroy, start, stop, and reboot a ProfitBricks virtual machine.
+description:
+     - Create, destroy, update, start, stop, and reboot a ProfitBricks virtual machine. When the virtual machine is created it can optionally wait for it to be 'running' before returning. This module has a dependency on profitbricks >= 1.0.0
+version_added: "1.9"
```
bcoca: 2.0 is current development version, 1.9 was already published so this won't make it there

cloud/profitbricks/profitbricks.py
```diff
@@ -0,0 +1,632 @@
...
+  auto_increment:
+    description:
+      - Whether or not to increment a single number in the name for created virtual machines.
+    default: yes
+    choices: ["yes", "no"]
+  name:
+    description:
+      - The name of the virtual machine.
+    required: true
+    default: null
```
bcoca: when required is true, default is irrelevant

cloud/profitbricks/profitbricks.py
```diff
@@ -0,0 +1,632 @@
...
+      - wait for the instance to be in state 'running' before returning
+    required: false
+    default: "yes"
+    choices: [ "yes", "no" ]
+    aliases: []
+  wait_timeout:
+    description:
+      - how long before wait gives up, in seconds
+    default: 600
+    aliases: []
```
bcoca: no need for empty aliases

cloud/profitbricks/profitbricks.py
```diff
@@ -0,0 +1,632 @@
...
+        if not module.params.get('name'):
+            module.fail_json(msg='name parameter is required for new instance')
+        if not module.params.get('image'):
+            module.fail_json(msg='image parameter is required for new instance')
+        if not module.params.get('subscription_user'):
+            module.fail_json(msg='subscription_user parameter is required for new instance')
+        if not module.params.get('subscription_password'):
+            module.fail_json(msg='subscription_password parameter is required for new instance')
+
+        (machine_dict_array) = create_virtual_machine(module, profitbricks)
```
bcoca: not all paths capture exceptions and turn them into fail_json messages, for example create_virtual_machine here.

cloud/profitbricks/profitbricks.py
baldwinSPC: Great, thanks @bcoca I will motor through these and update and re-push. Tnx!

bcoca:
minor issues but otherwise the module looks good

baldwinSPC:
@bcoca Let me know if it looks better. Thanks!

abadger:
What's the minimum python version of the profitbricks library?  If it's python 2.4 compatible then we should modify this module to run with python2.4.  If it's python2.6+ then we should list python >= 2.6 in the requirements section.

cloud/profitbricks/profitbricks.py
```diff
@@ -0,0 +1,639 @@
...
+      - 'web003.stackpointcloud.com'
+    wait_timeout: 500
+    state: stopped
+
+'''
+
+import re
+import uuid
+import time
+import json
```
abadger: I don't think we use json within this modules code so this can be removed.

cloud/profitbricks/profitbricks.py
```diff
@@ -0,0 +1,639 @@
...
+import sys
```
abadger: Looks like import sys can be removed as well.

cloud/profitbricks/profitbricks.py
```diff
@@ -0,0 +1,639 @@
...
+        )
+    )
+
+    if not HAS_PB_SDK:
+        module.fail_json(msg='profitbricks required for this module')
+
+    subscription_user = module.params.get('subscription_user')
+    subscription_password = module.params.get('subscription_password')
+    wait = module.params.get('wait')
+    wait_timeout = int(module.params.get('wait_timeout'))
```
abadger: Instead of doing this, set the type of wait_timeout to int.

abadger:
Some minor comments.  Otherwise looks good.

(Note: the wait_timeout change applies in multiple locations, I just noted it in one of them).

gregdek:
Oh wait... @abadger made requests for fixes, so I'll put this one in needs_revision.

baldwinSPC:
@gregdek Hi Greg, I've corrected all issues raised by both @abadger and @bcoca Let me know if there are any further changes that need to be done or if this can be merged now. If so, I can move onto the other PRs that build on this one.

baldwinSPC:
w00t. Thanks @bcoca

PR #211, opened 2015-01-20
An Ansible module to manage github keys

leocomelli:
# Ansible Module - Github Keys

An Ansible module to manage github public keys. The Github lets you add many public ssh keys, this is way that Github uses to identify trusted computers, without involving passwords.
## Features:
- Lists keys of non-authenticated user;
- Lists keys of authenticated user;
- Adds a new public ssh key;
- Gets a public ssh key by GH id;
- Removes an existing public ssh key.
## Sample

```
# file: gh_keys.yml

- hosts: devel
  user: root

  tasks:
    - name: Generate a new public key
      user:
        name: root
        generate_ssh_key: yes
        ssh_key_bits: 2048

    - name: Add a new GH public key
      gh_keys:
        action: add_key
        user: leocomelli
        passwd: secret
        title: devel
        key: /root/.ssh/id_rsa.pub
      register: keys

    - debug:
        var: keys.stdout
```
### Lists keys of a non-authenticated user

```
...
- gh_keys:
    action: list_keys
    user: leocomelli

# output
# [{"id":999999999,"key":"ssh-rsa AAA..."}]
```
### Lists keys of an authenticated user

```
...
- gh_keys:
    action: list_keys
    user: leocomelli
    passwd: secret

# output
# [{"id":999999999,"key":"ssh-rsa AAA...","url":"https://api.github.com/user/keys/999999999","title":"test-ghkeys","verified":true}]
```
### Adds a new public ssh key (authentication required)

```
...
- gh_keys:
  action: add_key
  user: leocomelli
  passwd: secret
  title: my-new-pkey
  key: /home/leocomelli/.ssh/id_rsa.pub

# output
# {"id":9999999991,"key":"ssh-rsa AAA2...","url":"https://api.github.com/user/keys/9999999991","title":"my-new-pkey","verified":true}
```
### Gets a public ssh key by GH id (authentication required)

```
...
- gh_keys:
  action: get_key
  user: leocomelli
  passwd: secret
  key_id: 9999999991

# output
# {"id":9999999991,"key":"ssh-rsa AAA2...","url":"https://api.github.com/user/keys/9999999991","title":"my-new-pkey","verified":true}
```
### Removes an existing public ssh key (authentication required)

```
...
- gh_keys:
  action: remove_key
  user: leocomelli
  passwd: secret
  key_id: 9999999991

# output
# N/A
```
## Authentication

[...]

PR #692, opened 2015-07-08
New module - github_key

rwe:


source_control/github_key.py
```diff
@@ -0,0 +1,213 @@
...
+      - SSH key name
+    required: true
+  pubkey:
+    description:
+      - SSH public key value. Required when state=present
+    required: false
+  state:
+    description:
+      - Whether to remove a key, ensure that it exists, or update its value.
+    choices: ['present', 'absent', 'updated']
```
bcoca: present should take care of updated

source_control/github_key.py
```diff
@@ -0,0 +1,213 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+module: github_key
+short_description: Manage GitHub access keys.
+description:
```
bcoca, line 20: missing version_added

source_control/github_key.py
```diff
@@ -0,0 +1,213 @@
...
+  name:
+    description:
+      - SSH key name
+    required: true
+  pubkey:
+    description:
+      - SSH public key value. Required when state=present
+    required: false
+  state:
+    description:
```
bcoca, line 38: missing required: false

source_control/github_key.py
```diff
@@ -0,0 +1,213 @@
...
+      - GitHub Access Token with permission to list and create public keys.
+    required: true
+  name:
+    description:
+      - SSH key name
+    required: true
+  pubkey:
+    description:
+      - SSH public key value. Required when state=present
+    required: false
```
bcoca, line 35: when required is false, you need to show default

bcoca:
some general issues with not handling any data for check mode nor trying to handle changed status and no error checking on the return, which should then be a fail_json call and not exit_json.

source_control/github_key.py
rwe, line 35: Is that true even if it's a conditionally required parameter? There is no meaningful default value.

source_control/github_key.py
```diff
@@ -0,0 +1,213 @@
...
+    description:
+      - GitHub Access Token with permission to list and create public keys.
+    required: true
+  name:
+    description:
+      - SSH key name
+    required: true
+  pubkey:
+    description:
+      - SSH public key value. Required when state=present
```
rwe: This should be `state=present or state=updated`

source_control/github_key.py
rwe, line 38: Isn't that implicit when specifying a default? When could a parameter have a default but still be required?

I just went through other PRs in `ansible-modules-core` where changes of adding redundant `required` to something with `default` were removed. I'm getting mixed signals on what the policy is there.

source_control/github_key.py
rwe: There are scenarios you want to ensure a key is present by name, but don't want to update it to a new value if it already exists. Is the expectation that this should overwrite that key every time?

rwe:
> not handling any data for check mode

In `check_mode`, this module does everything except for send the mutating `POST` or `DELETE` changes. It performs all the searches, calculating of what modifications need to happen, etc. What behavior seems to be missing?

> nor trying to handle changed status and no error checking on the return, which should then be a fail_json call and not exit_json.

Good catch, I've added HTTP error handling  in `GitHubSession.request`. `fetch_url` handles the other error types (connection, etc.) and calls `fail_json` appropriately.

source_control/github_key.py
rwe, line 20: Got it. What version should that be? `VERSION` on the root of this repo is currently marked as `1.8.2`, but not sure if that was the last version number or next release number.

bcoca, line 20: 2.0

rwe, line 20: Done

source_control/github_key.py
bcoca: yes, in the cases that you don't you can add a force=yes|no parameter (look at copy module)

bcoca:
for check_mode you need to return changed=True or False, possibly detailing what changes are required.

source_control/github_key.py
rwe: Got it! That makes sense, I'll make the change. Thanks

source_control/github_key.py
```diff
@@ -0,0 +1,217 @@
...
+      - GitHub Access Token with permission to list and create public keys.
+    required: true
+  name:
+    description:
+      - SSH key name
+    required: true
+  pubkey:
+    description:
+      - SSH public key value. Required when state=present or state=updated.
+    required: false
```
bcoca, line 35: if required: false a default should be present, when required is true it is not (would make no sense)

rwe:
`check_mode` is returning `changed`, see https://github.com/ansible/ansible-modules-extras/pull/692/files#diff-bb556342af3f47c7143fa9278a6e1909R157 and https://github.com/ansible/ansible-modules-extras/pull/692/files#diff-bb556342af3f47c7143fa9278a6e1909R174

rwe:
This should be ready for another look.

source_control/github_key.py
```diff
@@ -0,0 +1,223 @@
...
+    if pubkey:
+        pubkey_parts = pubkey.split(' ')
+        # Keys consist of a protocol, the key data, and an optional comment.
+        if len(pubkey_parts) < 2:
+            module.fail_json(msg='"pubkey" parameter has an invalid format')
+
+        # Strip out comment so we can compare to the keys GitHub returns.
+        pubkey = ' '.join(pubkey_parts[:2])
+    elif state == 'present':
+        module.fail_json(msg='"pubkey" is requred when state=present')
```
mscherer: typo, a 'i' is missing in required

source_control/github_key.py
```diff
@@ -0,0 +1,223 @@
...
+      - GitHub Access Token with permission to list and create public keys.
+    required: true
+  name:
+    description:
+      - SSH key name
+    required: true
+  pubkey:
+    description:
+      - SSH public key value. Required when state=present.
+    required: false
```
resmo, line 35: add  `default: none` or `default: null` here

source_control/github_key.py
```diff
@@ -0,0 +1,223 @@
...
+    description:
+      - GitHub Access Token with permission to list and create public keys.
+    required: true
+  name:
+    description:
+      - SSH key name
+    required: true
+  pubkey:
+    description:
+      - SSH public key value. Required when state=present.
```
resmo: hint: use C(state=present) for better readability

source_control/github_key.py
```diff
@@ -0,0 +1,223 @@
...
+- name: Read SSH public key to authorize
+  shell: cat /home/foo/.ssh/id_rsa.pub
+  register: ssh_pub_key
+
+- name: Authorize key with GitHub
+  local_action:
+    module: github_key
+    name: 'Access Key for Some Machine'
+    token: '{{github_access_token}}'
+    pubkey: '{{ssh_pub_key.stdout}}'
```
resmo, line 65: your example is good for registering a public key of a _remote_ system. Maybe an example of a local key using lookup would be nice:

```
- name: Authorize local key with GitHub
  local_action:
    module: github_key
    name: 'Access Key of {{ ansible_hostname }}'
    token: '{{ github_access_token }}'
    pubkey: '{{ lookup('file', '~/.ssh/id_rsa.pub') }}'
```

gregdek:

# 

gregdek:
Looks like this PR needs revision. Please make the suggested revisions, and when you're done, please add a comment with the text "ready_for_review" and we will continue with the review process.  Thanks!

gregdek:
@erydo A friendly reminder: this pull request has been marked as needing your action. If you still believe that this PR applies, and you intend to address the issues with this PR, just let us know in the PR itself and we will keep it open pending your changes.

rwe:
Thanks @gregdek, I will be addressing @resmo's comments—this fell off my radar.

source_control/github_key.py
rwe, line 65: This module is not intended to register a local machine. Doing so would require saving a privileged github access token to that machine first, and there's little or no benefit to going through that process versus just adding your local key by hand.

The module is intended for registering remote systems: servers, build machines and so on, and I think that showing the local example might be confusing. A newcomer may see that example and fail to make the distinction, and perhaps even not realize that they should have separate keys on different machines.

rwe:
ready_for_review

gregdek:
Thanks @erydo for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

gundalow:
See also https://github.com/ansible/ansible-modules-extras/pull/211 which is similar

bcoca:
this overlaps with #211

rwe:
Could we please get some eyes on this to merge in? All previous comments were addressed and this should be good to go…

gundalow:
Note to self: gundalow_review_this

gundalow:
Thank you for this module

PEP8 and Python 3 checks are good.

It's most likely something wrong I'm doing, but I can't get this module to work.

```
    - debug: msg='{{ssh_pub_key.stdout}}'

    - name: Authorize key with GitHub
      local_action:
        module: github_key
        name: 'Access Key for Some Machine'
        token: 'REDACTED'
        pubkey: '{{ssh_pub_key.stdout}}'
        force: false
```

`fatal: [localhost -> localhost]: FAILED! => {"changed": false, "failed": true, "invocation": {"module_args": {"force": false, "name": "Access Key for Some Machine", "pubkey": "ssh-rsa REDACTED user@host", "state": "present", "token": "REDACTED"}, "module_name": "github_key"}, "msg": " failed to send request POST to https://api.github.com/user/keys: HTTP Error 422: Unprocessable Entity"}
`

rwe:
@gundalow Hmm, just to be clear, the token you provided was a personal access token via https://github.com/settings/tokens, not actually the string 'REDACTED' ? I just tried this successfully:

``` yml
- hosts: all
  tasks:
    - name : Get machine public SSH key
      shell: cat ./foo_rsa.pub
      register: ssh_pub_key

    - name: Printing Key
      debug:
        msg: '{{ssh_pub_key.stdout}}'

    - name: Authorize key with GitHub
      local_action:
        module: github_key
        name: 'Ansible Test'
        token: '{{github_provisioning_token}}'
        pubkey: '{{ssh_pub_key.stdout}}'
        force: false
```

gundalow:
I gave given the token just "write:public_key" permissions. If I change the token I get an authentication error.

rwe:
@gundalow The token needs both `write:public_key` as well as `read:public_key`. Otherwise it can't determine whether to perform no change or to delete/create the key. See `ensure_key_present`. Does that help?

gundalow:
Even with `admin:public_key` (which sets `write:public_key` and `read:public_key`) this fails for me.

If I use the command line

```
curl -vu gundalow:mytoken https://api.github.com/user/keys
```

I can see my existing SSH keys

If I remove the `read:public_key` permission and retry the curl I get a 404

gundalow:
Urgh, the none obvious error means this pubic part already exists under a key with a different name.

Module is good to merge

gundalow:
shipit

gregdek:
Thanks again to @erydo for this PR, and thanks @ansible/core for reviewing. Marking for inclusion.

alikins:
Shipping it.

dfjaimes:
seems that this is missing return documentation, leading to broken builds

PR #695, opened 2015-07-08
Adding win_nssm module

georgefrank:
This module provides an interface to the "Non-Sucking Service Manager" program available for windows. The module assumes "nssm" is available via command line (it can be installed via chocolatey).

### DIFF
--- a/windows/win_nssm.ps1
+++ b/windows/win_nssm.ps1
@@ -1,0 +1,547 @@
+#!powershell
+# This file is part of Ansible
+#
+# Copyright 2015, George Frank <george@georgefrank.net>
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+$ErrorActionPreference = "Stop"
+
+# WANT_JSON
+# POWERSHELL_COMMON
+
+$params = Parse-Args $args;
+$result = New-Object PSObject;
+Set-Attr $result "changed" $false;
+
+
+If ($params.name)
+{
+    $name = $params.name
+}
+Else
+{
+    Fail-Json $result "missing required argument: name"
+}
+
+If ($params.state)
+{
+    $state = $params.state.ToString().ToLower()
+    $validStates = "present", "absent", "started", "stopped", "restarted"
+    
+    # These don't really fit the declarative style of ansible
+    # If you need to do these things, you can just write a command for it
+    # "paused", "continued", "rotated"
+
+    If ($validStates -notcontains $state)
+    {
+        Fail-Json $result "state is $state; must be one of: $validStates"
+    }
+}
+else
+{
+    $state = "present"
+}
+
+If ($params.application)
+{
+    $application = $params.application
+}
+Else
+{
+    $application = $null
+}
+
+If ($params.app_parameters)
+{
+    $appParameters = $params.app_parameters
+}
+Else
+{
+    $appParameters = $null
+}
+
+If ($params.stdout_file)
+{
+    $stdoutFile = $params.stdout_file
+}
+Else
+{
+    $stdoutFile = $null
+}
+
+If ($params.stderr_file)
+{
+    $stderrFile = $params.stderr_file
+}
+Else
+{
+    $stderrFile = $null
+}
+
+Function Service-Exists
+{
+    [CmdletBinding()]
+    param(
+        [Parameter(Mandatory=$true)]
+        [string]$name
+    )
+
+    return ,[bool](Get-Service "$name" -ErrorAction SilentlyContinue)
+}
+
+Function Nssm-Remove
+{
+    [CmdletBinding()]
+    param(
+        [Parameter(Mandatory=$true)]
+        [string]$name
+    )
+
+    if (Service-Exists -name $name)
+    {
+        $cmd = "nssm stop ""$name"""
+        $results = invoke-expression $cmd
+
+        $cmd = "nssm remove ""$name"" confirm"
+        $results = invoke-expression $cmd
+
+        if ($LastExitCode -ne 0)
+        {
+            Set-Attr $result "nssm_error_cmd" $cmd
+            Set-Attr $result "nssm_error_log" "$results"
+            Throw "Error removing service ""$name"""
+        }
+
+        $result.changed = $true
+     }
+}
+
+Function Nssm-Install
+{
+    [CmdletBinding()]
+    param(
+        [Parameter(Mandatory=$true)]
+        [string]$name,
+        [Parameter(Mandatory=$true)]
+        [string]$application
+    )
+
+    #note: the application name must look like the following, if the directory includes spaces:
+    # nssm install service "c:\Program Files\app.exe" """C:\Path with spaces"""
+    #see https://git.nssm.cc/?p=nssm/nssm.git;a=commit;h=0b386fc1984ab74ee59b7bed14b7e8f57212c22b for more info
+
+
+    if (!$application)
+    {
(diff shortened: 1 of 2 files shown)


jhawkesworth:
Thanks for this, this is going to be really useful - have had good experiences with nssm already so will be great to have a module for it.  

First thought on the PR is to mention that you need to install nssm actually in the documentation .py file.  I think there might even be a 'requirements' section in the documentation options which would probably be the place to mention it - have a look at some core modules.

Hope to test soon although not sure just when at the moment.

gregdek:

cc: @trondhindenes @petemounce @elventear @smadam813 @jhawkesworth @angstwad @sivel @chrishoffman @cchurch

chrishoffman:
+1

smadam813:
Looks good!  +1

windows/win_nssm.py
```diff
@@ -0,0 +1,138 @@
...
+  app_parameters:
+    description:
+      - Parameters to be passed to the application when it starts
+    required: false
+    default: null
+    aliases: []
+author: "Adam Keech (@smadam813), George Frank (@georgefrank)"
+'''
+
+# TODO:
```
petemounce: Is this comment relevant to this module?

windows/win_nssm.ps1
```diff
@@ -0,0 +1,547 @@
...
+
+Function Service-Exists
+{
+    [CmdletBinding()]
+    param(
+        [Parameter(Mandatory=$true)]
+        [string]$name
+    )
+
+    return ,[bool](Get-Service "$name" -ErrorAction SilentlyContinue)
```
petemounce: Leading comma typo...?

windows/win_nssm.py
georgefrank: Correct - it's not relevant. I'll remove that.

windows/win_nssm.ps1
georgefrank: This one isn't needed, so I'll remove it (it was just habit!). The leading comma does, however, have a use when multiple items are returned (more info: http://stackoverflow.com/questions/25279112/returning-a-liststring-using-a-comma-vs-not-using-a-comma)

petemounce: Ah, thank you - every day's a school day!

windows/win_nssm.ps1
```diff
@@ -0,0 +1,547 @@
...
+    [CmdletBinding()]
+    param(
+        [Parameter(Mandatory=$true)]
+        [string]$name
+    )
+
+    if (Service-Exists -name $name)
+    {
+        $cmd = "nssm stop ""$name"""
+        $results = invoke-expression $cmd
```
petemounce, line 111: Can this fail/throw?

windows/win_nssm.ps1
```diff
@@ -0,0 +1,547 @@
...
+    param(
+        [Parameter(Mandatory=$true)]
+        [string]$name,
+        [Parameter(Mandatory=$true)]
+        [string]$application
+    )
+
+    #note: the application name must look like the following, if the directory includes spaces:
+    # nssm install service "c:\Program Files\app.exe" """C:\Path with spaces"""
+    #see https://git.nssm.cc/?p=nssm/nssm.git;a=commit;h=0b386fc1984ab74ee59b7bed14b7e8f57212c22b for more info
```
petemounce: This would be better in the `py` doc, so it appears in the documentation.

windows/win_nssm.ps1
```diff
@@ -0,0 +1,547 @@
...
+
+
+    if (!$application)
+    {
+        Throw "Error installing service ""$name"". No application was supplied."
+    }
+
+    if (!(Service-Exists -name $name))
```
petemounce, line 142: Currently, `Service-Exists` returns a collection with an element in it. Does the `!` operator do the intended thing there, if `@(true)` comes back? Unfamiliar with the operator unwrapping as I discovered from earlier reply.

windows/win_nssm.py
```diff
@@ -0,0 +1,138 @@
...
+    - nssm is a service helper which doesn't suck. See https://nssm.cc/ for more information.
+requirements:
+  - "nssm >= 2.24.0 # (install via win_chocolatey) win_chocolatey: name=nssm"
+options:
+  name:
+    description:
+      - Name of the service to operate on
+    required: true
+    default: null
+    aliases: []
```
petemounce: I think an empty alias list can be omitted.

windows/win_nssm.py
```diff
@@ -0,0 +1,138 @@
...
+description:
+    - nssm is a service helper which doesn't suck. See https://nssm.cc/ for more information.
+requirements:
+  - "nssm >= 2.24.0 # (install via win_chocolatey) win_chocolatey: name=nssm"
+options:
+  name:
+    description:
+      - Name of the service to operate on
+    required: true
+    default: null
```
petemounce: I think if required is true, default is unnecessary to specify.

windows/win_nssm.py
```diff
@@ -0,0 +1,138 @@
...
+      - started
+      - stopped
+      - restarted
+      - absent
+    default: started
+    aliases: []
+
+  application:
+    description:
+      - The application binary to run as a service
```
petemounce, line 54: Since required is false and default is null, it would be nice to expand the description to mention that is a parameter to use when the service is already installed (I assume)

windows/win_nssm.py
```diff
@@ -0,0 +1,138 @@
...
+    required: false
+    default: null
+    aliases: []
+  
+  stdout_file:
+    description:
+      - Path to receive output
```
petemounce, line 64: A note about what happens with relative/absolute path would be nice - are relative paths relative to the application binary, or the nssm binary, or...?

petemounce:
@georgefrank thanks very much for the contribution; I'm +1 for it becoming merged in terms of code review, but am not in a position to actually test it at the moment.

windows/win_nssm.ps1
georgefrank, line 142: `!@($true)` yields the same result as `!$true` (also, `@($true)` and `$true` are equivalent as well)

windows/win_nssm.ps1
georgefrank: Completely agree! I'll move it there.

georgefrank:
Thanks for all the feedback, @petemounce! I'm not used to writing formal documentation, so all of the feedback there is especially appreciated. I'll fix things up and re-submit soon(:tm:).

windows/win_nssm.py
georgefrank: There are a bunch of empty alias collections, so I'll remove all of them.

windows/win_nssm.py
georgefrank, line 54: It's needed any time the service needs to be installed, which may happen if you specify any state other than "absent". This clearly needs a better description!

windows/win_nssm.ps1
georgefrank, line 111: Can you elaborate on what you're asking here? I'm not sure I understand the intent of the question.

petemounce, line 111: In the statement below you check and handle an error code, but not here. I
don't know nssm's behaviour. Is it not necessary to check the exit code
here?

Sorry for not being clear.

##

windows/win_nssm.ps1
petemounce, line 142: Cool

petemounce:
@georgefrank lgtm pending an answer on the exit-code check line 115. Otherwise, +1 and thanks very much for contributing!

windows/win_nssm.ps1
georgefrank, line 111: The stop command can fail, yes; however, we're only interested in whether the service can be removed (since we're already sure it exists, per the `Service-Exists` check). Removing a service using nssm only works well if the service is stopped. Here are the scenarios that can play out:
1. The service is stopped. In this case, `nssm stop` will fail, but `nssm remove` will succeed, so we don't want to do anything about the `nssm stop` failure.
2. The service is hanging and won't respond. In this case, `nssm stop` will fail, and also `nssm remove` will fail, which is caught below.
3. The service is responding, but is not currently stopped. In this case, `nssm stop` will succeed, and so will `nssm remove`.

So this command is fine without an error check.

petemounce, line 111: Gotcha

petemounce:
Ship it!

robynbergeron:
@georgefrank thanks for the PR; @chrishoffman, @smadam813, @jhawkesworth, @petemounce, thanks for the reviews and helping work through this. :)

Marking for inclusion; cc: @ansible/core team.

PR #702, opened 2015-07-09
Added support to assign attached mac address interface id and port options.

markleehamilton:
Updated code to pass pep8 v1.6.2 pylint v0.25.0.

### DIFF
--- a/network/openvswitch_port.py
+++ b/network/openvswitch_port.py
@@ -5,6 +5,8 @@
 
 # (c) 2013, David Stygstra <david.stygstra@gmail.com>
 #
+# Portions copyright @ 2015 VMware, Inc.
+#
 # This file is part of Ansible
 #
 # This module is free software: you can redistribute it and/or modify
@@ -19,8 +21,6 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this software.  If not, see <http://www.gnu.org/licenses/>.
-#
-# Portions copyright @ 2015 VMware, Inc.  All rights reserved.
 
 import syslog
 import os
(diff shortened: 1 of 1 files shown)


network/openvswitch_port.py
```diff
@@ -47,44 +54,188 @@
...
+        description:
+            - Used when port is created to set the external_ids:iface-id.
+    attached-mac:
+        required: false
+        description:
+            - MAC address when port is created using external_ids:attached-mac.
+    set:
+        required: false
+        description:
+            - Additional set options to apply to the port
```
bcoca: need version_added for new options

network/openvswitch_port.py
```diff
@@ -47,44 +54,188 @@
         default: 5
         description:
             - How long to wait for ovs-vswitchd to respond
+    iface-id:
+        required: false
```
bcoca: when required: false, default must be specified

network/openvswitch_port.py
```diff
@@ -17,6 +19,11 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this software.  If not, see <http://www.gnu.org/licenses/>.
+#
+# Portions copyright @ 2015 VMware, Inc.  All rights reserved.
```
bcoca: keep copyright with existing above, also not sure 'All rights reserved' works with GPL code

markleehamilton: I asked this specific question to our lawyers. They say this is standard and does not conflict with GPL.
-Mark

network/openvswitch_port.py
markleehamilton: Brian,
What is the process here? How should I address these changes?
-Mark

bcoca: just update the code and push again, the PR will have an additional commit (or you can rebase and force push)

markleehamilton:
I believe I've address all of your comments, bcoca.

PR #703, opened 2015-07-10
Prevent Zabbix config from being overwritten by providing a create_on…

CaptainParalytic:
Prevent Zabbix config from getting overwritten by providing an create_only optional parameter

### DIFF
--- a/monitoring/zabbix_host.py
+++ b/monitoring/zabbix_host.py
@@ -90,7 +90,11 @@ options:
             - Please review the interface documentation for more information on the supported properties
             - 'https://www.zabbix.com/documentation/2.0/manual/appendix/api/hostinterface/definitions#host_interface'
         required: false
-        default: []
+    create_only:
+        description:
+            - Will only create a host in Zabbix if it does not already exist, but will not update an existing host.e'
+        required: false
+        default: false
 '''
 
 EXAMPLES = '''
@@ -370,6 +374,7 @@ def main():
             state=dict(default="present", choices=['present', 'absent']),
             timeout=dict(type='int', default=10),
             interfaces=dict(required=False),
+            create_only=dict(required=False, choices=BOOLEANS, default=False),
             proxy=dict(required=False)
         ),
         supports_check_mode=True
@@ -388,6 +393,7 @@ def main():
     state = module.params['state']
     timeout = module.params['timeout']
     interfaces = module.params['interfaces']
+    create_only = module.params['create_only']
     proxy = module.params['proxy']
 
     # convert enabled to 0; disabled to 1
@@ -402,6 +408,13 @@ def main():
         module.fail_json(msg="Failed to connect to Zabbix server: %s" % e)
 
     host = Host(module, zbx)
+    
+    # check if host exist
+    is_host_exist = host.is_host_exist(host_name)
+
+    if is_host_exist and create_only:
+        module.exit_json(changed=False,
+                         result="'create_only' specified so existing host not updated")  
 
     template_ids = []
     if link_templates:
@@ -423,10 +436,7 @@ def main():
     if proxy:
         proxy_id = host.get_proxyid_by_proxy_name(proxy)
 
-    # check if host exist
-    is_host_exist = host.is_host_exist(host_name)
-
-    if is_host_exist:
+    if is_host_exist:     
         # get host id by host name
         zabbix_host_obj = host.get_host_by_host_name(host_name)
         host_id = zabbix_host_obj['hostid']
--- a/monitoring/zabbix_host.py
+++ b/monitoring/zabbix_host.py
@@ -92,7 +92,7 @@ options:
         required: false
     create_only:
         description:
-            - Will only create a host in Zabbix if it does not already exist, but will not update an existing host.e'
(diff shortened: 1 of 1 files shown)


monitoring/zabbix_host.py
```diff
@@ -90,7 +90,11 @@
             - Please review the interface documentation for more information on the supported properties
             - 'https://www.zabbix.com/documentation/2.0/manual/appendix/api/hostinterface/definitions#host_interface'
         required: false
-        default: []
+    create_only:
+        description:
+            - Will only create a host in Zabbix if it does not already exist, but will not update an existing host.e'
```
resmo: typo at the end of the line

bcoca:
i would just change the paramter to force: default to yes/true and use the no/false to not update existing entries, this keeps it in sync with other modules that have the same option.

resmo:
+1 for force and, default it to true. @harrisongu, @cove what do you say?

CaptainParalytic:
My only concern is that the default for force tends to be false. At least
create_only says exactly what it does.

But when all is said and done, it is the facility that is more important
than what it is called.

cove:
defaulting force to true is odd IMO too, I'd lean towards state: create_once.

CaptainParalytic:
But even without the flag it would only create it once. Subsequently it
would update the host that would by then already exist, which is what the
create_only flag is supposed to achieve.

cove:
I don't think I should default to create once, if that's what you mean. That wouldn't fit in with ansible's idempotent model.

Hum I may not be clear on the use case, is this an optimization actually?

CaptainParalytic:
No that's not what I was saying.  If you look at what it does at present,
if the host doesn't exist on Zabbix ,  it will be created. If it does exist
if cannot be created because it had already been been created,.  To put it
another way, it  already can only be created once. Subsequently it cannot
be created again because it already exists. One doesn't need a flag to say
only create this once because it cannot be created a second time I once it
already exists.

The idea of the create_only flag is to ensure that the only action that the
module can take is to create the host of it does not already exist. It is
to prevent it from updating a host if subsequent configuration has been
done in  Zabbix so that that configuration does not get overridden.

So I chose the flag name of  create_only because that describes EXACTLY
what it does.  If the create_only flag is set to true,  then the module
will only create a host,  it will not update an existing host.

So create_only means that the module will only create.

create_once is what the module does now, because the second or subsequent
time it runs it will update and not create.

cove:
Got it thanks. Seems to me state: create_only would be a little more intuitive than having another flag.

CaptainParalytic:
Really?

A state describes what the situation should be after the module has been
invoked.  So "present"  is a state.  Started is a state,.  Stopped is a
state."Do not update if it already exists" (another way of saying
create_only) is not a state,  unless you want to change the rules of the
English language.

Force (defaulting to true,  which is totally nonstandard) doesn't tell you
anything about what effect having it set to false will have.

If you want to mess with the English language and pretend that an
instruction is a state (which it clearly isn't) then fine, but please don't
suggest that having words mean what they do not mean is intuitive, because
it isn't.

But I think I'm losing the will to live here. I needed a way to ensure that
the module would only create a Zabbix host if it did not already exist and
to ensure that if it did already exist that it did not update it, thus
destroying any configuration that had been performed directly in Zabbix.

I decided to use a flag that told the module that it was only allowed to
create a new host and was to do nothing else. If it is generally felt that
using words that do not mean what they actually say is somehow easier to
understand than using words that mean exactly what they say is what we
should have, then so be it.

My concern is to have a way to ensure that the module will only create a
new host and will not update an existing one.

If the community wants to call it something that is not clear then fine by
me.

Got it thanks. Seems to me state: create_only would be a little more
intuitive than having another flag.

cove:
I see your point about using an additional flag, make sense. In which case I think create_only: makes sense as the flag name too, and defaulting to false, since the default assumption should just be to keep the settings in sync. My 2c anyway. ;)

resmo:
so @cove you, as owner of the module, are you +1 ?

`create_only` as name make sense at it makes clear what it does. I voted for `force` because there are several modules which already uses `force` in a similar manner. But hey, `create_only` is really fine. No need for a war about that naming...

cove:
I looked but didn't see the other modules that use force. Can you give me an example?

Yes, don't want anyone to lose the will to live over this. :)

bcoca:
copy, get_url, rabbitmq_user, lvg ... and more, those just from recent memory

CaptainParalytic:
But most of them have the default for force as false.

cove:
The copy module does default to true though, which sets a precedent. From just the module perspective create_only makes sense, but overall I think @bcoca and @resmo are right, force would be the right thing. 

+1 for force as well.

gregdek:
OK, so looking back at this, it looks as though @cove wants force -- and since he's the owner, that's that. :)

@CaptainParalytic Please make the suggested change, and when you're done, please add a comment with the text "ready_for_review" and we will continue with the review process.  It should go very quickly from here in. Thanks!

robynbergeron:
@CaptainParalytic -- A friendly reminder: this pull request has been marked as needing your action, and it hasn't been updated in the last month. 

gregdek:
Closing due to inactivity.

PR #527, opened 2015-05-22
[WIP] dnf rewrite

igor-raits:
It will use only API, no external tools and we will get 5x less code.

PR #704, opened 2015-07-10
DNF Module

shaded-enmity:
This is reimplementation of the DNF module using exclusively DNF's Python APIs.
Besides generic package(s) `install/upgrade/remove` functionality, I'd highlight:
- Supports group `install/upgrade/remove`
- Supports combinations of`{disable,enable}repo=`

I'd appreciate some input as to how to better report errors, as currently in some places I let the original exception bubble up into the logs.
I've tested the module with my toy playbooks, if someone could help with testing that'd be great :)

I'm currently working on documenting the functions.

### DIFF
--- a/packaging/os/dnf.py
+++ b/packaging/os/dnf.py
@@ -1,7 +1,8 @@
 #!/usr/bin/python -tt
-# -*- coding: utf-8 -*-
-
-# Written by Cristian van Ee <cristian at cvee.org>
+# vim: set fileencoding=utf-8
+#
+# Copyright 2015 Pavel Odvody <podvody@redhat.com>
+# Copyright 2015 Cristian van Ee <cristian at cvee.org>
 #
 # This file is part of Ansible
 #
@@ -19,17 +20,8 @@
 # along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
 #
 
-
-import traceback
 import os
 import dnf
-
-try:
-    from dnf import find_unfinished_transactions, find_ts_remaining
-    from rpmUtils.miscutils import splitFilename
-    transaction_helpers = True
-except:
-    transaction_helpers = False
 
 DOCUMENTATION = '''
 ---
@@ -94,8 +86,9 @@ notes: []
 # informational: requirements for nodes
 requirements:
   - dnf
-  - yum-utils (for repoquery)
-author: "Cristian van Ee (@DJMuggs)"
+author: 
+  - '"Pavel Odvody" <podvody@redhat.com>'
+  - '"Cristian van Ee (@DJMuggs)" <cristian at cvee.org>'
 '''
 
 EXAMPLES = '''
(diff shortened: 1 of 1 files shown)


shaded-enmity:
The CI build fails at the `with` keyword during compilation for Python 2.4

Not sure how relevant that is given that `DNF` itself doesn't work with that version.

bcoca:
most modules are checked for 2.4 compatibility unless we specifically exclude them, dnf is not on that list, yet.

bcoca:
this seems to overlap with #527

packaging/os/dnf.py
```diff
@@ -1,7 +1,8 @@
 #!/usr/bin/python -tt
-# -*- coding: utf-8 -*-
-
-# Written by Cristian van Ee <cristian at cvee.org>
+# vim: set fileencoding=utf-8
+#
```
bcoca, line 3: you removed -_\- coding: utf-8 -_-,  please add back as that should be the 2nd line for python's use to set file encoding

packaging/os/dnf.py
```diff
@@ -19,18 +20,9 @@
 # along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
 #
 
-
-import traceback
 import os
 import dnf
```
bcoca: dnf library import should catch import exception as there was in the previous version

bcoca:
requirements should specify the higher version of python

packaging/os/dnf.py
```diff
@@ -122,709 +115,245 @@
...
+                'changed':  changed,
+                'results':  results
+        }
+        module.exit_json(**o)
 
-def install_dnf_utils(module):
+def get_names(name):
+        if ',' not in name:
+                return [name]
+        return name.split(',')
```
bcoca: you should not need this function, just define names as a list in the module spec

packaging/os/dnf.py
shaded-enmity, line 3: That's emacs notation, I'm using vim's notation (`vim: set fileencoding=utf8`). Both notation works, although if it's standard to use the emacs notation I'm happy to change it.

packaging/os/dnf.py
shaded-enmity: The previous version was catching the import exception around non-existent code, that had never been part of the DNF API. There's no direct checking whether the `DNF` package can be actually imported, but it sounds like nice functionality. I'll add it. Thanks for the suggestion

packaging/os/dnf.py
shaded-enmity: I'll take a look, thanks

shaded-enmity:
I'm currently finishing up docstrings, once I have that done, I'll incorporate the changes as per the comments.

shaded-enmity:
Seems that the CI ignores the Python >= 2.6 requirement (or is my syntax incorrect?).

bcoca:
the CI does not pick up the requirements from the docs, its hardcoded in the travis file which modules are checked for what version

shaded-enmity:
Ping @DJMuggs

abadger:
I've added the python-2.6 requirement for dnf to .travis.yml.  I had talked with @ignatenkobrain earlier -- dnf python bindings need python2.6+ so that's a valid requirement for the module above.

@shaded-enmity, @ignatenkobrain can you two collaborate on your two pull requests?

One thing I noticed still missing in both pull requests is there doesn't appear to be any handling of check-mode.

gregdek:
This one now fails Travis. 

@shaded-enmity @ignatenkobrain would be great to figure out which of this or #527 should be our focus...

shaded-enmity:
I'll look into this tomorrow, although at a first glance it seems that Travis tries to compile the module with python 2.4, which isn't supported.

```
$ python2.4 -m compileall -fq -x 'cloud/|monitoring/zabbix.*\.py|/layman\.py|/maven_artifact\.py|clustering/consul.*\.py|notification/pushbullet\.py' .
Compiling ./packaging/os/dnf.py ...
  File "./packaging/os/dnf.py", line 432
    with init_dnf(
                ^
SyntaxError: invalid syntax
```

shaded-enmity:
Please refer to: https://github.com/ansible/ansible-modules-extras/pull/527

PR #712, opened 2015-07-13
Update small typo error on documentation of gluster_volume module, param is brick and not bricks

nadley:
Hi,

Here is a small pull request on documentation typo error about gluster_volume module. In fact "bricks" is not a param when using is ansible will complain with : "msg: unsupported parameter for module: bricks".

And ansible-doc gluster_volume give this documentation : 

> GLUSTER_VOLUME

  Create, remove, start, stop and tune GlusterFS volumes

Options (= is mandatory):
- brick
      Brick path on servers [Default: None]

Hope it helps regards.

### DIFF
--- a/system/gluster_volume.py
+++ b/system/gluster_volume.py
@@ -62,7 +62,7 @@ options:
     default: 'tcp'
     description:
       - Transport type for volume
-  bricks:
+  brick:
     required: false
     default: null
     description:
@@ -108,7 +108,7 @@ author: "Taneli Leppä (@rosmo)"
 
 EXAMPLES = """
 - name: create gluster volume
-  gluster_volume: state=present name=test1 bricks=/bricks/brick1/g1 rebalance=yes cluster="192.168.1.10,192.168.1.11"
+  gluster_volume: state=present name=test1 brick=/bricks/brick1/g1 rebalance=yes cluster="192.168.1.10,192.168.1.11"
   run_once: true
 
 - name: tune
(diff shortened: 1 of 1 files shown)


bcoca:
the code actually expects 'bricks', 'brick' is also allowed as an alias. I tested and cannot reproduce the error.

nadley:
Ok I understand the trouble, in the devel branch the code support the alias and 'bricks' and 'brick' works but as I'm using ansible 1.9.2 in the branch stable-1.9  the code only supports 'brick' there is no alias.

https://github.com/ansible/ansible-modules-extras/blob/stable-1.9/system/gluster_volume.py#L314

This new feature has been introduce by this commit : https://github.com/ansible/ansible-modules-extras/commit/6d66beb7936fc13b4b873d88504fdda067bab87b

bcoca:
so ansible-doc of 1.9 does show the correct version, I can backport the fix to 1.9 in case a 1.9.3 gets released

nadley:
yes ansible-doc give the correct version, This is the online documentation which is not correct regarding the last stable, but if the online doc is following the devel branch then it's ok but a little bit confusing for new users.

bcoca:
we have plans on making it follow stable and have a /devel for development, but have not had time to implement yet, so closing this ticket as the doc update would not be needed for devel

prupert:
I am using Ansible 1.9.4 and still seeing the error "msg: unsupported parameter for module: bricks"

`- name: Ensure GlusterFS volumes are created.
  gluster_volume:
    state=present
    name={{ item.name }}
    bricks={{ item.bricks | default(glusterfs_server_default_brick + "/" + item.name) }}
    force=yes
    cluster={{ glusterfs_server_cluster_nodes | join(",") }}
    replicas={{ item.replicas | default(glusterfs_server_volume_default_replicas) }}
  with_items: glusterfs_server_volumes
  run_once: true`

It does work when using "brick=" in stead of "bricks=".

Icedroid:
have the same error with ansible 1.9.4

Issue #664, opened 2015-06-30
known_hosts module not working ?


mmallejac:
##### Issue Type:

Bug Report
##### Component Name:

known_hosts
##### Ansible Version:

ansible --version
ansible 1.9.2
##### Ansible Configuration:

N/A
##### Environment:

apt-add-repository ppa:ansible/ansible
aptitude update
aptitude install ansible python-pip 

on Ubuntu 14.04.2 LTS
##### Summary:

This is a copy/confirmation of https://github.com/ansible/ansible/issues/11222
Trying for instance to add his own public key to itself, so ssh root@localhost would work, maybe not using the module correctly ?
##### Steps To Reproduce:

```
  - name: add itself to known_hosts list
    known_hosts: host={{ inventory_hostname }} key="{{ lookup('file', dataDir + '/' + inventory_hostname + '/id_rsa.pub') }}"
```
##### Expected Results:

on target host, ssh localhost should work without asking add fingerprint
##### Actual Results:

```
failed: [myHost] => {"cmd": "/usr/bin/ssh-keygen -F myHost -f /tmp/tmpE7kC4N", "failed": true, "rc": 1}
stderr: line 1 invalid key: ssh-rsa...
/tmp/tmpE7kC4N is not a valid known_hosts file.

msg: line 1 invalid key: ssh-rsa...
/tmp/tmpE7kC4N is not a valid known_hosts file.
```

sivel:
I think this may possibly be a misunderstanding.

The error shown is telling you that the key is not a proper SSH host key.

Looking at your lookup, I am assuming that the key you are giving it is a ssh public key, that would normally be used with `authorized_keys` as opposed to `known_hosts`.

The `known_hosts` module, is meant to manage `~/.ssh/known_hosts`, whereas the `authorized_keys` module would manage `~/.ssh/authorized_keys`.

ansible/ansible#11222 is a different issue, in that the key string was missing the key type.

costela:
This might be partly a misunderstanding, but the issue seems real enough. Host keys and user pubkeys have TTBOMK the same format, which means this should work, regardless of semantics.

My first guess, without having looked at the code yet, is that the module checks the syntax of the generated known_hosts file before moving it to its final location. This check fails if the added line uses the pubkey format, because the hostname is missing as first token in the line. 
[...]

PR #716, opened 2015-07-14
Allow multiple keys per host + minor improvements

### LINKED ISSUE #691: Add abilit for known_hosts module to manage multiple keys per host
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.9.2
##### Ansible Configuration:

Default
##### Environment:

N/A
##### Summary:

The known_hosts module seems to replace the file instead of editing it.
##### Steps To Reproduce:

```
- name: tell the host about our servers it might want to ssh to
  known_hosts:
    host='example.com'
    path='/etc/ssh/ssh_known_hosts'
    key="{{ item }}"
  with_items:
    - "example.com ssh-rsa AAAAAxxxxxxxx"
    - "example.com ssh-ed25519 AAAAAyyyyyyyy"
```
##### Expected Results:

The known_host file should have both keys added in.
##### Actual Results:

The known_hosts file only has the last key added in, the first one has been overwritten.

costela:
The biggest change in this PR (4d18347) introduces the allow_multiple parameter in an attempt to fix #691. 

Both other commits are largely cosmetic refactoring (6bc9ee3) and documentation improvement (1e19723) to address the misunderstanding in #664.


bcoca:
I don't even think we need allow_multiple as this should be how it works normally.

costela:
@bcoca: don't think so. If you do something like this:

```
known_hosts:
    host: somehost
    key: "{{ item }}"
    state: present
    allow_multiple: yes
with_items:
    - "somehost ssh-rsa ..."
    - "somehost ssh-dss ..."
```

You get changes on every playbook run, because while adding the second key, the module removes the first.

Or did you mean something else?

bcoca:
i mean we don't need the option, just the functionality

costela:
@bcoca, oh, that makes more sense :)
However, if we don't include an option we might leave users that rely on the current semantics vulnerable if we change the default behaviour. Imagine the following scenario:
- you have a central server with an ssh-dss key, accessed by many clients set up via the ansible.
- since there's only one central server, you decide to hard-code the single key to the ansible playbook.
- the server gets compromised and you set it up from scratch with a new ssh-ed25519 key, updating your hard-coded copy of the ansible playbook with the new key.
- with the allow_multiple semantics as default, the clients are now vulnerable to a MITM because the attacker has the original ssh-dss key, and it's still in the known_hosts files for every client.

fdutheil:
I agree that changing any default behaviour can be touchy, but in that case, the module's default behaviour is not respecting the behaviour of the tools it acts on. known_hosts supports multiple keys, and people using ssh_keyscan will import all the keys, not only the first one of the list...
I really thought it was something bad on my side, then a bug in the module, before I read this issue report.
@costela : the use case you report is valid, for sure, but people using  authorized_keys module face the same issue and will deal with it by removing the faulty key (which is anyway way saner that counting on any tool default behaviour when security is involved....).

So, may I suggest to get something similar to the authorized_keys module? It's `exclusive` parameter cover what `allow_multiple` would. Better use same keywords IMHO.

costela:
@fdutheil sounds reasonable. We'd still be inverting the default behavior, but at least providing a way to fall back to the previous one.
@bcoca does that sound ok to you?

bcoca:
sounds like we might not be talking about the same thing, authorized_keys has always allowed multiple keys per user, this is a problem with the current implementation not allowing more than 1 key per host. It should be 1 key per host and per key type.

gregdek:
It looks like we agree some revisions need to be made here, but we're not entirely clear on what. :)

So putting in needs_revision for bookkeeping purposes, and we can continue to discuss what those revisions should be.

Once those revisions are made, please add a comment with the text "ready_for_review" and we will continue with the review process.  Thanks!

gregdek:
Let's ask the actual owner of this module for his opinion. :)

@mcv21 Please review this PR to make sure it adheres to the following guidelines:

gregdek:
@mcv21 This change is still pending your review; do you have time to take a look and comment? Please comment with text 'shipit' or 'needs_revision' as appropriate.

costela:
@bcoca the underlying limitation here is the module's reliance on the ssh-keygen CLI tool. It's currently pretty limited when it comes to removing keys from the known_hosts file and does an "all or nothing" based on hostname. This means we can either say "remove everything for this host and just add this one key", but can't simply say "replace **only** the RSA key for this host". That's why I took the route of giving the user two options: either don't attempt removing the host (allow duplicates) or remove the host before adding (not possible to add two key formats for the same host).

Besides the current (admittedly suboptimal) solution, I see two alternative options:
- using the output from "ssh-keygen -F" to search for the host, loop over the found keys and remove only the one for the same type being added. This is easier to implement but has the problem of being brittle: if ssh-keygen changes its output format for that subcommand, the module stops working (it seems to have remained stable for a while and the output isn't localized, so it's hard to say how big of a risk this would be)
- incorporating ssh-keygen's functionality into the module, including its non-standard hashing mechanism. This would allow more fine-grained control over the known_host file, but is harder to implement (I've been trying to mimic the hashing behavior with python for the last hour, to no avail) and is also somewhat brittle: if the hashing changes, things break.

Does anyone have any preference between those options? Or did I maybe miss some other solution?

bcoca:
since both seem to depend on the implementation in ssh, i would go with the first option as minor output changes will be easier to fix than encoding changes.

costela:
Ok, I'll try to cook something up.
Just for future reference, in case ssh-keygen does change its output: it has been stable since at least OpenSSH 4.0, so I guess it's relatively safe to depend on it.

mcv21:
My feeling is that we should try and use ssh-keygen as much as possible (despite its failings); that was the point of this module!

I'm not sure the question of multiple keys vs replacing keys is quite correct yet. While it makes some sense to allow multiple keytypes per host (though any given ssh client / server combination will only use 1, so I'm not sure how desirable this actually is), it doesn't make sense (and may be actively harmful) to allow multiple keys of the same type per host. I don't think this PR addresses that as yet.

Accordingly, I think: needs_revision. Sorry for the delay in getting back to you.

gregdek:
Thanks @costela for this PR. A maintainer of this module has asked for revisions to this PR. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

robynbergeron:
@costela A friendly reminder: this pull request has been marked as needing your action. If you still believe that this PR applies, and you intend to address the issues with this PR, just let us know in the PR itself and we will keep it open pending your changes.

gregdek:
@costela Another friendly reminder: this pull request has been marked as needing your action. If you still believe that this PR applies, and you intend to address the issues with this PR, just let us know in the PR itself and we will keep it open. If we don't hear from you within another 14 days, we will close this pull request.

costela:
ready_for_review

costela:
@mcv21 yes, absolutely. The latest version of this PR works like this:
- with state=present:
  - if the host isn't found, add the provided key.
  - if the host is found, but with a different key type, add the provided key.
  - if the host is found with the same key type, but different keys, replace the old key with the provided.
  - if the host is found with the same key type and content, noop.
- with state=absent:
  - without "key" argument:
    - if the host is found, remove all keys (via "ssh-keygen -R")
    - if the host isn't found, noop
  - with "key" argument:
    - if the host isn't found, noop.
    - if the host is found, but a different key, noop.
    - if the host and the exact same key are found, remove the provided key.

This means one of my original concerns still applies: if the users updates a key and at the same time changes its type, the old key will remain in the known_hosts file, possibly allowing a MITM. So maybe @fdutheil 's idea of an `exclusive` argument isn't bad.

As a sidenote, my original motivation for writing this PR was enabling automatic known_hosts propagation across some subset of my nodes, based on the generated host keys. Having that, I can let ssh decide which of the available keys has higher priority. This way I can decouple the mutual authentication of nodes from the key details and ssh is free to use the best possible key supported by both client and server.

gregdek:
Thanks @costela. To the current maintainers, @mcv21 please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

system/known_hosts.py
```diff
@@ -185,54 +189,76 @@ def sanity_check(module,host,key,sshkeygen):
...
-        if l=='': 
-            continue
-        if l[0]=='#': #comment
+        if l=='':
             continue
-        if k==l: #found a match
-            return True, False #current, not-replace
-    #No match found, return current and replace
-    return True, True
+        elif l[0]=='#': #comment
```
mcv21: This is really picky, but the comment '#comment' here is now wrong - in the old version, we just skipped the comment line, but now we're parsing it for the line-number match.

mcv21:
Hi,

I'm happy with the content of this. I have one really picky quibble, which I've attached a comment to the diff about (hopefully you can see this) - we now parse the first line of ssh-keygen -F output for the line number, so leaving the note '#comment' there (which made sense when we just skipped the line) is confusing. It should be replaced with something like '# line from ssh-keygen telling us which line number matched'

Thanks,

Matthew

system/known_hosts.py
costela: Fair enough. I've expanded the comment to be a bit more explicit.

costela:
@mcv21 just incorporated your suggestion. If that's your only problem with the PR, care to give your 'shipit'? ;)

mcv21:
Thanks!

shipit.

gregdek:
Thanks again to @costela for this PR, and thanks @mcv21 for reviewing. Marking for inclusion.

resmo:
Could you please rebase this PR to the current HEAD of branch devel, e.g.

```
git remote add upstream git@github.com:ansible/ansible-modules-extras.git
git checkout devel #since your PR is applied from the devel branch
git pull --rebase upstream devel
git push origin devel -f
```

costela:
@resmo done

PR #717, opened 2015-07-15
Composer - Removed No Progress

### LINKED ISSUE #323: composer: command=dump-autoload fails because it doesn't recognize the --no-progress option
## repro

If the following two tasks are executed after each other

```
 - composer: command=update working_dir={{ your_php_app_dir }} 
# assuming {{ your_php_app_dir }} has a valid composer.json
 - composer: command=dump-autoloads working_dir={{ your_php_app_dir }}
```

The second one fails with the following error:

```
msg: [RuntimeException] The "--no-progress" option does not exist. dump-autoload [-o|--optimize] [--no-dev]
```
## guess at the root cause

always appended default options. As per http://docs.ansible.com/composer_module.html

> Default options that are always appended in each execution are –no-ansi, –no-progress, and –no-interaction
## Versions, environment
- Composer version 1.0-dev (829199c0530ea131262c804924f921447c71d4e8) 2015-03-16 13:11:02
- ansible 1.8.4
- ubuntu/trusty64

TomKita:
Removed no progress as default resolves #323

### DIFF
--- a/packaging/language/composer.py
+++ b/packaging/language/composer.py
@@ -141,7 +141,6 @@ def main():
 
     # Default options
     options.append('--no-ansi')
-    options.append('--no-progress')
     options.append('--no-interaction')
 
     options.extend(['--working-dir', os.path.abspath(module.params['working_dir'])])


bcoca:
i'm ok with removing it from defaults, but won't this create an issue with commands that DO support it? it will probably tie up stdout and mess with the feedback

TomKita:
That is a good point could be changed to default only for install/update?

bcoca:
that would be great

resmo:
don't forget to correct the `notes:` in the doc.

resmo:
@TomKita May I help you with the revision?

TomKita:
@resmo Of course go for it I'm a bit tied up with other projects at the moment. Unfortunately this is on the lower end of my list.

resmo:
@TomKita ok, I'll take it over.

gregdek:
@resmo want to just open a new pr? If so, feel free to close this one and create a new one.

robynbergeron:
Closing this one, since it looks like @resmo is now handling this in ansible/ansible-modules-extras/#965. Thanks @resmo, and thanks to @TomKita for the initial work :)

robynbergeron:
And someday I'll learn the ropes of proper github referencing. Yay me!

PR #725, opened 2015-07-18
Adding apk module

kbrebanov:
Adding apk module to manage Alpine Linux packages.

Currently supports:
- Installing a package
- Removing a package
- Ensuring a package is at the latest version
- Upgrading all installed packages
- Updating the repository indexes

### DIFF
--- a/packaging/os/apk.py
+++ b/packaging/os/apk.py
@@ -1,0 +1,196 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2015, Kevin Brebanov <https://github.com/kbrebanov>
+# Based on pacman (Afterburn <http://github.com/afterburn>, Aaron Bull Schaefer <aaron@elasticdog.com>)
+# and apt (Matthew Williams <matthew@flowroute.com>>) modules.
+#
+# This module is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This software is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this software.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: apk
+short_description: Manages apk packages
+description:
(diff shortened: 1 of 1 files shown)


resmo:
LGTM! 

One thing I noticed is that you can not pass a list of packages to be installed at once, similar to this `apk add openssh openntp vim` which would reduce the time used for install a bunch of packages.

kbrebanov:
@resmo Added support for installing and removing multiple packages at once.

packaging/os/apk.py
```diff
@@ -0,0 +1,219 @@
...
+    module.exit_json(changed=True, msg="removed %s package(s)" % (names))
+        
+# ==========================================
+# Main control flow.
+
+def main():
+    module = AnsibleModule(
+        argument_spec = dict(
+            state = dict(default='present', choices=['present', 'installed', 'absent', 'removed', 'latest']),
+            name = dict(type='str'),
```
resmo: use `type='list'` here, so users can pass a list like this:

```
- apk: name={{ item }}
  with_items:
  - foo
  - bar
```

or even comma separated as well (out of the box ansible magic)!

```
- apk: name=foo,bar
```

In your code, you will get python list to use, so your code must be adjusted accordantly.

kbrebanov:
@resmo Added support for 'with_items'

resmo:
+1 from me, handover to core review @gregdek

packaging/os/apk.py
```diff
@@ -0,0 +1,215 @@
...
+- apk: upgrade=yes
+
+# Update repositories as a separate step
+- apk: update_cache=yes
+'''
+
+import os
+import re
+
+APK_PATH="/sbin/apk"
```
bcoca: use the module's function to find binaries instead of hardcoding this

packaging/os/apk.py
```diff
@@ -0,0 +1,215 @@
...
+def update_package_db(module):
+    cmd = "apk update"
+    rc, stdout, stderr = module.run_command(cmd, check_rc=False)
+    if rc == 0:
+        return True
+    else:
+        module.fail_json(msg="could not update package db")
+
+def query_package(module, name):
+    cmd = "apk -v info --installed %s" % (name)
```
bcoca: use the commong apk you found already

packaging/os/apk.py
```diff
@@ -0,0 +1,215 @@
...
+    rc, stdout, stderr = module.run_command(cmd, check_rc=False)
+    if rc == 0:
+        return True
+    else:
+        return False
+
+def query_latest(module, name):
+    cmd = "apk version %s" % (name)
```
bcoca: same as above

kbrebanov:
@bcoca Used the module's get_bin_path function

PR #729, opened 2015-07-20
clc_group : the ansible module to manage groups on a centurylink cloud

popurisiva:
New module for group has been added for centurylink cloud.
This module adhere to the module development standards and best practices.
With this module, users can create  or delete groups in centurylink clould


bcoca:
fixing the version_added to "2.0" on both modules

bcoca:
also had remove some :  in notes, changed them for ,

Issue #252, opened 2015-02-12
Brew install when package name contains "/"


kdisneur:
I tried to install `cask` with `ansible` and I have the following errors: `msg: Invalid package: caskroom/cask/brew-cask.`

It's because there is some slashes in the name and, in the ansible task, the `/` seems to not be allowed. I tried to add the slash in the `VALID_PACKAGE_CHARS` constant, it works like a charm.

I've not opened a pull request yet, because I would like to have your feedback before: there is a reason the slash was not allowed in the regex or is it just an oversight?

Cheers,

tobywf:
I've also hit this issue. `homebrew_tap` does provide a workaround in some cases, but it's not ideal. This sounds similar to ansible/ansible#7991 and ansible/ansible#8413 where "+" and "." where left out of the filter.

Edit: `VALID_PACKAGE_CHARS` isn't the only change. The module calls `brew outdated` and then checks the complete output, but this only outputs short names (`caskroom/cask/brew-cask` -> `brew-cask`). `brew outdated <package>` accepts full names and returns a non-zero return code if `<package>` needs updating, which is probably simpler in the long run. So the methods `_current_package_is_outdated`  and `_outdated_packages` need changing, I'll prepare a PR this week.

kdisneur:
@tobywf Nope. I am sorry but I've no longer a PullRequest ready for this one and I think I will not have time to work on it this week. 
If you don't have time this week, I can take a look at this issue next week I guess. But feel free to create one if you have time before. :)

tresni:
To add to this, I have a custom Forumla that I use hosted in a gist.
`brew install --HEAD https://gist.githubusercontent.com/tresni/0abf5c8017b082d4717a/raw/superspark.rb` is a completely acceptable command and works.  But I can't do:

``` yaml
homebrew:
        name: https://gist.githubusercontent.com/tresni/0abf5c8017b082d4717a/raw/superspark.rb
        state: head
```

Fails with the "Invalid package".  Adding `:` & `/` allows it to work.

tobywf:
@tresni: Interesting, I'll have a look at that in the morning, try and get it incorporated into the PR I've prepared. I'm thinking there isn't much point in "validating" the package name at all, since brew will do that for us anyway? Interestingly enough, the URI works with `brew outdated`, so keeping the action idempotent should be no problem:

``` bash
[...]

PR #732, opened 2015-07-20
Homebrew: Allow colons, direct check for outdated formula

tobywf:
See issue #252. Colons should be in the validation filter so that formulas from e.g. gists can be used.

Previously, to determine if a formula is out of date a list of all retrieved formulas is retrieved, and the formula name is checked against that list. A formula name can be passed directly to `brew outdated`, this will return 1 if the formula is out of date or not installed which has the same result as checking if the name is in the list. This works well and is more robust for formulas specified by a URL.

### DIFF
--- a/packaging/os/homebrew.py
+++ b/packaging/os/homebrew.py
@@ -119,6 +119,7 @@ class Homebrew(object):
         /                   # slash (for taps)
         \+                  # plusses
         -                   # dashes
+        :                   # colons (for URLs)
     '''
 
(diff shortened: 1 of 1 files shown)


resmo:
LGTM. +1

gregdek:
@danieljaouen @andrew-d Please review this PR to make sure it adheres to the following guidelines:

indrajitr:
shipit

gregdek:
Marking for inclusion. Thanks @tobywf for the PR, and thanks @indrajitr for the review!

PR #735, opened 2015-07-21
Port campfire to fetch_url

abadger:
notification/campfire was using raw urllib2 which does not verify the server's SSL certificate. Ported to ansible's fetch_url() which does certificate checking.

@fabulops - I don't use campfire so if you are able to review this and verify that it still works, that would be greatly appreciated. Since this change will prevent man-in-the-middle attacks I will merge this even if you don't get back to me but I would vastly prefer you (or someone else using campfire) take a look to make sure this works.

### DIFF
--- a/notification/campfire.py
+++ b/notification/campfire.py
@@ -42,7 +42,7 @@ options:
               "vuvuzela", "what", "whoomp", "yeah", "yodel"]
 
 # informational: requirements for nodes
-requirements: [ urllib2, cgi ]
+requirements: [ ]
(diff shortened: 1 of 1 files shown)


fabulops:
@abadger - Unfortunately, I'm at a different job and don't have access to campfire to test. I'd go ahead and merge it since it seems that it should work, though I'm not familiar with Ansible's fetch_url()

abadger:
Thanks for taking a look.

# Merged

Hi!

issue is resolved.

Thank you!

PR #739, opened 2015-07-21
Port maven_artifact to fetch_url

abadger:
notification/maven_artifact.py was using raw urllib2 which does not verify the server's SSL certificate. Ported to ansible's fetch_url() which does certificate checking.

@chrisisbeef - I don't use maven so if you are able to review this and verify that it still works, that would be greatly appreciated. Since this change will prevent man-in-the-middle attacks I will merge this even if you don't get back to me but I would vastly prefer you (or someone else using maven_artifact) take a look to make sure this works.

### DIFF
--- a/packaging/language/maven_artifact.py
+++ b/packaging/language/maven_artifact.py
@@ -22,11 +22,9 @@
 __author__ = 'cschmidt'
 
 from lxml import etree
-from urllib2 import Request, urlopen, URLError, HTTPError
 import os
 import hashlib
 import sys
-import base64
 
 DOCUMENTATION = '''
 ---
@@ -69,7 +67,7 @@ options:
         required: false
         default: null
     password:
-        description: The passwor to authenticate with to the Maven Repository
+        description: The password to authenticate with to the Maven Repository
         required: false
         default: null
     dest:
@@ -81,6 +79,12 @@ options:
         required: true
         default: present
         choices: [present,absent]
+    validate_certs:
+        description: If C(no), SSL certificates will not be validated. This should only be set to C(no) when no other option exists.
+        required: false
+        default: 'yes'
+        choices: ['yes', 'no']
+        version_added: "1.9.3"
 '''
 
 EXAMPLES = '''
@@ -165,13 +169,12 @@ class Artifact(object):
 
 
 class MavenDownloader:
-    def __init__(self, base="http://repo1.maven.org/maven2", username=None, password=None):
+    def __init__(self, module, base="http://repo1.maven.org/maven2"):
+        self.module = module
         if base.endswith("/"):
             base = base.rstrip("/")
         self.base = base
         self.user_agent = "Maven Artifact Downloader/1.0"
-        self.username = username
-        self.password = password
 
     def _find_latest_version_available(self, artifact):
         path = "/%s/maven-metadata.xml" % (artifact.path(False))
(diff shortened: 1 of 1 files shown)


chrisisbeef:
I will take a look at this today.

On Tue, Jul 21, 2015 at 12:56 AM, Toshio Kuratomi notifications@github.com
wrote:

> notification/maven_artifact.py was using raw urllib2 which does not verify
> the server's SSL certificate. Ported to ansible's fetch_url() which does
> certificate checking.
> 
> @chrisisbeef https://github.com/chrisisbeef - I don't use maven so if
> you are able to review this and verify that it still works, that would be
> greatly appreciated. Since this change will prevent man-in-the-middle
> attacks I will merge this even if you don't get back to me but I would
> vastly prefer you (or someone else using maven_artifact) take a look to
> 
> ## make sure this works.
> 
> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/ansible/ansible-modules-extras/pull/739
> Commit Summary
> - Port maven_artifact to fetch_url
> 
> File Changes
> - _M_ packaging/language/maven_artifact.py
>   https://github.com/ansible/ansible-modules-extras/pull/739/files#diff-0
>   (20)
> 
> Patch Links:
> - https://github.com/ansible/ansible-modules-extras/pull/739.patch
> - https://github.com/ansible/ansible-modules-extras/pull/739.diff
> 
> —
> https://github.com/ansible/ansible-modules-extras/pull/739.

## 

Chris Schmidt

OWASP ESAPI Developer
http://www.owasp.org/index.php/Category:OWASP_Enterprise_Security_API

Check out OWASP ESAPI for Java
http://code.google.com/p/owasp-esapi-java/

OWASP ESAPI for JavaScript
http://code.google.com/p/owasp-esapi-js/

Yet Another Developers Blog
http://yet-another-dev.blogspot.com

Bio and Resume
http://www.digital-ritual.net/resume.html

packaging/language/maven_artifact.py
```diff
@@ -81,6 +80,12 @@
         required: true
         default: present
         choices: [present,absent]
+    validate_certs:
+        description: If C(no), SSL certificates will not be validated. This should only be used on personally controlled sites using self-signed certificates.
```
chrisisbeef: To be accurate - this can (and should) also be used with Maven Central. Can we update the documentation to reflect this information

http://blog.sonatype.com/2012/10/now-available-ssl-connectivity-to-central/#.Va-59RNVhBd

abadger: Hmm... I think the documentation is unclear.  I meant that you should only change it from the default of validate-certs=yes if using a personally controlled repository.  I'l change it to be more clear.

abadger: Changed to: If C(no), SSL certificates will not be validated. This should only be set to no when no other   option exists.

packaging/language/maven_artifact.py
```diff
@@ -208,13 +214,10 @@ def _request(self, url, failmsg, f):
...
             }
-        req = Request(url, None, headers)
-        try:
-            response = urlopen(req)
-        except HTTPError, e:
-            raise ValueError(failmsg + " because of " + str(e) + "for URL " + url)
-        except URLError, e:
-            raise ValueError(failmsg + " because of " + str(e) + "for URL " + url)
+
+        response, info = fetch_url(self.module, url, headers=headers)
```
chrisisbeef: I looked briefly at the fetch_url function - it's not clear to me where a user would specify where the trusted ca keystore is if they are using an internal CA.

Additionally, since we're switching to fetch_url should we not also let it handle the username/password functionality?

abadger: there's not currently a way to specify a specific file for an additional CA.  You can drop files into the directories specified in https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/urls.py#L427  /etc/ansible is probably the right directory to use if you only wanted it to affect tasks run by ansible.

A pull request could be made to add a parameter to allow a custom CA directory if you think it's needed.

Regarding basic auth, yeah -- I missed that when I made the port.

abadger:
Okay, basic auth responsibility transferred to fetch_url()

chrisisbeef:
@abadger - all looks good to me, there is still the open question about how to allow a user to specify a custom trusted certificate truststore for working with internal CAs; but I think that is a concern for fetch_url not for this module - when fetch_url supports it we can update the plugin to pass along the parameters

abadger:
@chrisisbeef Works for me.  Thanks for the review and helpful suggestions!

# Merged

Hi!

issue is resolved.

Thank you!

PR #743, opened 2015-07-21
ISO-8859-15 locale normalization in locale_gen module, fixes #741

### LINKED ISSUE #741: locale_gen is_present check fails with some locales
##### Issue Type:

Bug Report
##### Ansible Version:

```
ansible 1.9.2
```
##### Ansible Configuration:

No custom config
##### Environment:

Managing OS: OS X 10.10.4
Target OS: Debian Jessie
##### Summary:

`locale_gen` uses the ouput of `locale -a` to test if a given locale is already enabled ([See here](https://github.com/ansible/ansible-modules-extras/blob/devel/system/locale_gen.py#L67)). If the name of the locale differs from the name in `/etc/locale.gen` the test will fail and the locale will be enabled every time. Example:

```
# egrep -i '^en_us.iso.+15$' /etc/locale.gen
en_US.ISO-8859-15 ISO-8859-15
# locale -a | egrep -i '^en_us.iso.+15$'
en_US.iso885915
```
##### Steps To Reproduce:

Run the following playbook multiple times:

```

---
- hosts: all
  vars:
    locales:
      - "en_US"
      - "en_US.ISO-8859-15"
  tasks:
  - locale_gen: name="{{ item }}" state=present
    with_items: "{{ locales }}"
```
##### Expected Results:

Both tasks should report "changed" only after the first run, if at all. The following run should report "changed=0" because both locales are already enabled.
##### Actual Results:

While "en_US" only gets enabled once, "en_US.ISO-8859-15" will be enabled every time.

madddi:


abadger:
This looks good and I'm going to merge it.  However, the code that does these checks looks a little fragile.  I'm not sure of all the rules and conventions around locale naming, but maybe it should be rewritten to something like this?

``` python
def normalize_locale_name(name):
    """locale -a might return the encoding in either lower or upper case or with dashes
    Passing through this function makes them uniform for comparisons."""
    new_name = name.lower()
    for s, r in LOCALE_NORMALIZATION.items():
        if new_name.endswith(s):
            new_name = '%s%s' % (new_name[:-len(s)], r)
    return new_name
```

abadger:
# Merged

Hi!

issue is resolved.

Thank you!

PR #278, opened 2015-02-24
Update win_updates.py

adematte:
just added logPath parameter to the documentation.

PR #745, opened 2015-07-22
re-submit @adematte's PR #278 against HEAD

petemounce:
@adematte & @gregdek Re-submitting as suggested, replaces #278.


cchurch:
Looks good to me. Thanks!

PR #746, opened 2015-07-23
Module for creating vapp/vm in vcloud or vcd

bennojoy:
This  modules helps in creating vapp with a vm in vcloud Air/ subscription/ vcd environments

### DIFF
--- a/cloud/vmware/vca_vapp.py
+++ b/cloud/vmware/vca_vapp.py
@@ -1,0 +1,747 @@
+#!/usr/bin/python
+
+# Copyright (c) 2015 VMware, Inc. All Rights Reserved.
+#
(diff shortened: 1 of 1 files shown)


cloud/vmware/vca_vapp.py
```diff
@@ -0,0 +1,747 @@
...
+DOCUMENTATION = '''
+---
+module: vca_vapp
+short_description: create, terminate, start or stop a vm in vca
+description:
+  - Creates or terminates vca vms.
+version_added: "2.0"
+options:
+    username:
+      version_added: "2.0"
```
bcoca, line 28: no need for version_added on options for a new module, they are assumed to have been added when modules was added

PR #748, opened 2015-07-23
Module for modifying nat rules  in vcloud air or vcd

bennojoy:
This  modules helps with creating/modifying/deleting NAT rules in  vcloud Air/ subscription/ vcd environments


bcoca:
same issues as the other module

bennojoy:
@bcoa updated the license .. thanks

PR #749, opened 2015-07-23
zypper: Add local/remote rpm installation

alxgu:
Extend the zypper module to install remote/local rpm files like yum is able.

Example:
# Install local rpm file
- zypper: name=/tmp/fancy-software.rpm state=present

# Install the nginx rpm from a remote repo
- zypper: name=http://nginx.org/packages/sles/12/x86_64/RPMS/nginx-1.8.0-1.sles12.ngx.x86_64.rpm state=present

### DIFF
--- a/packaging/os/zypper.py
+++ b/packaging/os/zypper.py
@@ -31,7 +31,9 @@ import re
 DOCUMENTATION = '''
 ---
 module: zypper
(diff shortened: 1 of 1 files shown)


jpmens:
@bcoca I've looked at this and it seems a) very sane and b) extremely useful to me. +1

PR #756, opened 2015-07-24
libvirt: virt_pool module

drybjed:
This module manages storage pool configuration in `libvirt`.

`virt_pool` design is based on `virt` module. You can use it to:
- define or undefine storage pools using XML
- initialize a build of a storage pool, where supported (create a required directory in the filesystem, create new LVM volume, format a block device)
- start, stop, set autostart parameter of a storage pool
- delete storage pool contents

### DIFF
--- a/cloud/misc/virt_pool.py
+++ b/cloud/misc/virt_pool.py
@@ -1,0 +1,705 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2015, Maciej Delmanowski <drybjed@gmail.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


gregdek:

resmo:
Nice module overall. I am a bit concerned about the mixing of command and state. I would like to know why you decided to use both, wouldn't it be enough to have _only_ state? 

It also covers a lot of functionality, which maybe would make sense to separate e.g. facts or add it to a common virt fact module.

drybjed:
Configuration in `libvirt` cannot easily be changed without either editing XML files by hand or having more code that handles XML transformation internally. In many cases, like for example changing network interface configuration beyond simply adding or removing DHCP host entries you need to remove existing configuration and re-define it again.

Commands are more or less mapped to various `libvirt` actions, like definition of pools or networks, for example. They most certainly aren't idempotent - `libvirt` will return an error if you try to define the same network a second time.

States try to check existing state using commands and change to desired state in a controlled manner, for example if a network should be undefined, it will be destroyed (stopped) first. This should let `libvirt` send an error if any step cannot be finished correctly.

You most likely will want to use states in the playbook, but if you want more control, direct access to commands is there to for example execute actions ad-hoc from the command line.

I thought about moving the fact gathering to a separate module, but with only pools and networks it seemed not very much to warrant that. I imagine with more modules it could be easily doable.

resmo:
LGTM. 
+1

PR #757, opened 2015-07-24
libvirt: virt_net module

drybjed:
This module manages network configuration in `libvirt`.

`virt_net` is based on `virt` module. With it you can:
- define or undefine networks using XML
- start, stop, set an autostart status a given network

### DIFF
--- a/cloud/misc/virt_net.py
+++ b/cloud/misc/virt_net.py
@@ -501,6 +501,7 @@ def core(module):
 
     module.fail_json(msg="expected state or command parameter to be specified")
 
+
 def main():
 
     module = AnsibleModule (
--- a/cloud/misc/virt_net.py
+++ b/cloud/misc/virt_net.py
@@ -177,14 +177,12 @@ class LibvirtConnection(object):
         results = []
 
         # Get active entries
-        entries = self.conn.listNetworks()
-        for name in entries:
+        for name in self.conn.listNetworks():
             entry = self.conn.networkLookupByName(name)
             results.append(entry)
 
         # Get inactive entries
-        entries = self.conn.listDefinedNetworks()
-        for name in entries:
+        for name in self.conn.listDefinedNetworks():
             entry = self.conn.networkLookupByName(name)
             results.append(entry)
 
@@ -334,9 +332,8 @@ class VirtNetwork(object):
         return results
 
     def state(self):
-        entries = self.list_nets()
         results = []
-        for entry in entries:
+        for entry in self.list_nets():
             state_blurb = self.conn.get_status(entry)
             results.append("%s %s" % (entry,state_blurb))
         return results
@@ -378,9 +375,8 @@ class VirtNetwork(object):
         return self.facts(facts_mode)
 
     def facts(self, facts_mode='facts'):
-        entries = self.list_nets()
         results = dict()
-        for entry in entries:
+        for entry in self.list_nets():
             results[entry] = dict()
             results[entry]["autostart"] = self.conn.get_autostart(entry)
             results[entry]["persistent"] = self.conn.get_persistent(entry)
--- a/cloud/misc/virt_net.py
+++ b/cloud/misc/virt_net.py
@@ -317,18 +317,13 @@ class VirtNetwork(object):
         return self.conn.find_entry(entryid)
 
     def list_nets(self, state=None):
-        entries = self.conn.find_entry(-1)
         results = []
-        for x in entries:
-            try:
-                if state:
-                    entrystate = self.conn.get_status2(x)
-                    if entrystate == state:
-                        results.append(x.name())
-                else:
-                    results.append(x.name())
-            except:
-                pass
+        for entry in self.conn.find_entry(-1):
+            if state:
+                if state == self.conn.get_status2(entry):
+                    results.append(entry.name())
+            else:
+                results.append(entry.name())
         return results
 
     def state(self):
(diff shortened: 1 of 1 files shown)


cloud/misc/virt_net.py
```diff
@@ -0,0 +1,557 @@
...
+
+
+class LibvirtConnection(object):
+
+    def __init__(self, uri, module):
+
+        self.module = module
+
+        cmd = "uname -r"
+        rc, stdout, stderr = self.module.run_command(cmd)
```
mscherer: Why not use os.uname() ? ( the less fork, the better, IMHO )

cloud/misc/virt_net.py
```diff
@@ -0,0 +1,557 @@
...
+
+        if "xen" in stdout:
```
mscherer: So, if I have a VM on xen, I cannot use uri to connect to another libvirt, this doesn't sound good. I am also a bit surprised on the detection used, what if there is not 'xen' in the release ( like that seems be the case for xenserver ) and I use a xen guest, or what if there is xen in the release as a tag ( for xenodon, xena, or mixen as release tag ? ).

cloud/misc/virt_net.py
```diff
@@ -0,0 +1,557 @@
...
+
+    def find_entry(self, entryid):
+        # entryid = -1 returns a list of everything
+
+        results = []
+
+        # Get active entries
+        entries = self.conn.listNetworks()
+        for name in entries:
+            entry = self.conn.networkLookupByName(name)
```
mscherer, line 181: I suggest that it would be cleaner and more pythonic to do for name in self.conn.listNetworks(): and this kind of code. ( just nitpicking )

cloud/misc/virt_net.py
```diff
@@ -0,0 +1,557 @@
...
+                return self.module.exit_json(changed=True)
+
+    def undefine(self, entryid):
+        if not self.module.check_mode:
+            return self.find_entry(entryid).undefine()
+        else:
+            if not self.find_entry(entryid):
+                return self.module.exit_json(changed=True)
+
+    def get_status2(self, entry):
```
mscherer, line 223: I think a better name could be found than "get_status2", but I do not have much ideas.

cloud/misc/virt_net.py
```diff
@@ -0,0 +1,557 @@
...
+    def status(self, entryid):
+        return self.conn.get_status(entryid)
+
+    def get_xml(self, entryid):
+        return self.conn.get_xml(entryid)
+
+    def define(self, entryid, xml):
+        return self.conn.define_from_xml(entryid, xml)
+
+    def info(self, facts_mode='info'):
```
mscherer: Since there is no way to set facts_mode to anything but info, I think this argument should be removed.

cloud/misc/virt_net.py
```diff
@@ -0,0 +1,557 @@
...
+        results = []
+        for x in entries:
+            try:
+                if state:
+                    entrystate = self.conn.get_status2(x)
+                    if entrystate == state:
+                        results.append(x.name())
+                else:
+                    results.append(x.name())
+            except:
```
mscherer: A wide try/except would likely catch too much ( like out of memory, network failure, etc ), could it be restricted to the error that would appear ?

cloud/misc/virt_net.py
drybjed: Actually, I took that code from `virt` module, which is kinda old, and I don't use Xen so hadn't had a chance to try it. Since Xen [has a valid URI](https://libvirt.org/drvxen.html) I could just drop above code and go stright to connecting to a specified URI without checking for Xen in the first place, not sure why that was/is there. I will do modifications with other changes. Thanks for the heads up.

cloud/misc/virt_net.py
drybjed, line 181: Not a Python programmer by trads, so thanks. :-) Will do.

cloud/misc/virt_net.py
drybjed: Good idea, will do.

gregdek:
Looks like you've already got some action here, so glad to see that. Boilerplate text follows.

---

cloud/misc/virt_net.py
```diff
@@ -0,0 +1,543 @@
...
+            facts["ansible_facts"] = dict()
+            facts["ansible_facts"]["ansible_libvirt_networks"] = results
+        elif facts_mode == 'info':
+            facts['networks'] = results
+        return facts
+
+
+def core(module):
+
+    state     = module.params.get('state', None)
```
resmo, line 408: May I give a little note here, I know explicit is better than implicit, but get() does set `None` as default, no need to set it explicitly.

cloud/misc/virt_net.py
```diff
@@ -0,0 +1,543 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: virt_net
+author: "Maciej Delmanowski"
```
resmo: there is a new format for the author also including the github handle, would you mind change it to 

```
author: "Maciej Delmanowski (@drybjed)"
```

resmo:
LGTM
+1

PR #760, opened 2015-07-24
ec2_remote_facts module

michaeljs1990:
This module lets you get information about any number of ec2 instances
in your environment. It also has the option of creating hostnames based
on the ip of your server.

I recently created this so I could easily cluster rabbitmq nodes together and thought that this may be helpful for many other things since it's brings back a large amount of information about any instance you ask.

### DIFF
--- a/cloud/amazon/ec2_search.py
+++ b/cloud/amazon/ec2_search.py
@@ -135,7 +135,7 @@ def main():
                 for instance in [i for r in reservations for i in r.instances]:
                     if instance.private_ip_address != None:
                         instance.hostname = 'ip-' + instance.private_ip_address.replace('.', '-')
-                    if instance._state.name not in module.params.get('ignore_state') and :
+                    if instance._state.name not in module.params.get('ignore_state'):
                         server_info.append(todict(instance))
         except:
             print module.jsonify('error getting instances from: ' + region.name)
--- a/cloud/amazon/ec2_search.py
+++ b/cloud/amazon/ec2_search.py
@@ -146,8 +146,7 @@ def main():
         except:
             print module.jsonify('error getting instances from: ' + region.name)
 
-    ansible_facts = {'info': server_info}
-    ec2_facts_result = dict(changed=True, ec2=ansible_facts)
+    ec2_facts_result = dict(changed=True, ec2=server_info)
 
     module.exit_json(**ec2_facts_result)
 
--- a/cloud/amazon/ec2_search.py
+++ b/cloud/amazon/ec2_search.py
@@ -1,0 +1,152 @@
+#!/usr/bin/python
+#
+# This is a free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This Ansible library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this library.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: ec2_search
+short_description: ask EC2 for information about other instances.
+description:
+    - Only supports seatch for hostname by tags currently. Looking to add more later.
+version_added: "1.9"
+options:
+  key:
+    description:
+      - instance tag key in EC2
+    required: false
+    default: Name
+    aliases: []
+  value:
+    description:
+      - instance tag value in EC2
+    required: false
+    default: null
+    aliases: []
+  lookup:
+    description:
+      - What type of lookup to use when searching EC2 instance info.
+    required: false
+    default: tags
+    aliases: []
+  region:
+    description:
+      - EC2 region that it should look for tags in
+    required: false
+    default: All Regions
+    aliases: []
+author:
+    - "Michael Schuett (@michaeljs1990)"
+extends_documentation_fragment: aws
+'''
+
+EXAMPLES = '''
+# Note: These examples do not set authentication details, see the AWS Guide for details.
+
+# Basic provisioning example
+- ec2_search:
+    key: mykey
+    value: myvalue
+
+'''
+try:
+    import boto
+    import boto.ec2
+    HAS_BOTO = True
+except ImportError:
+    HAS_BOTO = False
+
+def todict(obj, classkey=None):
+    if isinstance(obj, dict):
+        data = {}
+        for (k, v) in obj.items():
+            data[k] = todict(v, classkey)
+        return data
+    elif hasattr(obj, "_ast"):
+        return todict(obj._ast())
+    elif hasattr(obj, "__iter__"):
+        return [todict(v, classkey) for v in obj]
+    elif hasattr(obj, "__dict__"):
+        # This Class causes a recursive loop and at this time is not worth
+        # debugging. If it's useful later I'll look into it.
+        if not isinstance(obj, boto.ec2.blockdevicemapping.BlockDeviceType):
+            data = dict([(key, todict(value, classkey))
+                for key, value in obj.__dict__.iteritems()
+                if not callable(value) and not key.startswith('_')])
+            if classkey is not None and hasattr(obj, "__class__"):
+                data[classkey] = obj.__class__.__name__
+            return data
+    else:
+        return obj
+
+def get_all_ec2_regions(module):
+    try:
+        regions = boto.ec2.regions()
+    except Exception, e:
+        module.fail_json('Boto authentication issue: %s' % e)
+
+    return regions
+
+# Connect to ec2 region
+def connect_to_region(region, module):
+    try:
+        conn = boto.ec2.connect_to_region(region.name)
+    except Exception, e:
+        print module.jsonify('error connecting to region: ' + region.name)
+        conn = None
+    # connect_to_region will fail "silently" by returning
+    # None if the region name is wrong or not supported
+    return conn
+
+def main():
+    module = AnsibleModule(
+        argument_spec = dict(
+            key = dict(default='Name'),
+            value = dict(),
+            lookup = dict(default='tags'),
+            ignore_state = dict(default='terminated'),
+            region = dict(),
+        )
+    )
+
+    if not HAS_BOTO:
+        module.fail_json(msg='boto required for this module')
+
+    server_info = list()
+
+    for region in get_all_ec2_regions(module):
+        conn = connect_to_region(region, module)
+        try:
+            # Run when looking up by tag names, only returning hostname currently
+            if module.params.get('lookup') == 'tags':
+                ec2_key = 'tag:' + module.params.get('key')
+                ec2_value = module.params.get('value')
+                reservations = conn.get_all_instances(filters={ec2_key : ec2_value})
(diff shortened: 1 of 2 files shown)


wimnat:
Isn't most of this functionality available by using the ec2.py dynamic inventory script?

michaeljs1990:
@wimnat as far as I know ec2.py dynamic inventory would only get you tag / ip and possibly public DNS. This brings back almost all the information about a given host plus is much more extensible and I think makes it easier to understand what is going on in the playbook as it brings the logic to a role level instead of keeping it at the playbook level.

michaeljs1990:
The other thing is that if I run the playbook with the dynamic inventory and get all the hosts i would have to run this on a machine that has access to all of them since in clustering scenarios the machines will all be using the same tag. This makes it into a manual process that can't be used when creating launch configurations or auto scale groups.

https://github.com/nowait-tools/ansible-rabbitmq

The readme does a good job of showing the problem I believe this solves well. I don't believe a dynamic inventory can be used as a solution but if it can I would be glad to give it a try.

wimnat:
If you run ec2.py --list you should see the complete set of info returned.

As an example...

"172.23.1.174": {
        "ec2__in_monitoring_element": false, 
        "ec2_ami_launch_index": "0", 
        "ec2_architecture": "x86_64", 
        "ec2_client_token": "8fa9sdf14-2787-480e-8f64-94d7ac875402", 
        "ec2_dns_name": "", 
        "ec2_ebs_optimized": false, 
        "ec2_eventsSet": "", 
        "ec2_group_name": "", 
        "ec2_hypervisor": "xen", 
        "ec2_id": "i-aaaddffrree", 
        "ec2_image_id": "ami-11223344", 
        "ec2_instance_type": "m3.large", 
        "ec2_ip_address": "", 
        "ec2_item": "", 
        "ec2_kernel": "aki-31990e0b", 
        "ec2_key_name": "mykey", 
        "ec2_launch_time": "2015-07-16T23:32:03.000Z", 
        "ec2_monitored": true, 
        "ec2_monitoring": "", 
        "ec2_monitoring_state": "enabled", 
        "ec2_persistent": false, 
        "ec2_placement": "ap-southeast-2b", 
        "ec2_platform": "", 
        "ec2_previous_state": "", 
        "ec2_previous_state_code": 0, 
        "ec2_private_dns_name": "ip-172-23-1-174.private.dns", 
        "ec2_private_ip_address": "172.23.1.174", 
        "ec2_public_dns_name": "", 
        "ec2_ramdisk": "", 
        "ec2_reason": "", 
        "ec2_region": "ap-southeast-2", 
        "ec2_requester_id": "", 
        "ec2_root_device_name": "/dev/sda1", 
        "ec2_root_device_type": "ebs", 
        "ec2_security_group_ids": "sg-aaddffee", 
        "ec2_security_group_names": "sec-group", 
        "ec2_sourceDestCheck": "true", 
        "ec2_spot_instance_request_id": "", 
        "ec2_state": "running", 
        "ec2_state_code": 16, 
        "ec2_state_reason": "", 
        "ec2_subnet_id": "subnet-aabbccdd", 
        "ec2_tag_Env": "prod", 
        "ec2_tag_Index": "3", 
        "ec2_tag_Name": "search01", 
        "ec2_tag_Status": "ready", 
        "ec2_tag_Type": "search", 
        "ec2_tag_Version": "1.2.3", 
        "ec2_virtualization_type": "paravirtual", 
        "ec2_vpc_id": "vpc-85jf85d"
      }

Looking at your example, you look up your hosts using two variables that represent a tag name and tag key.  Using dynamic inventory that would be grouped as 'tag_Name_Key' so if you wanted to apply the rabbitmq role to all of your rabbitmq hosts you would just use:

```
name: Playbook to apply rabbitmq role to rabbitmq hosts
hosts: tag_Name_Key // you can use variables here if you want
roles:
  - ansible-rabbitmq
```

Your role would need some tweaking (i.e. remove the ec2_search stuff) but it should do what you want.

michaeljs1990:
@wimnat Say I have 10 servers in with the same tag wouldn't the above cause the playbook to run on all of them every single time? Since it's in an auto scale group if one comes up it would build the current server but also try and run it on the other 10. I would also have to put my private key that I use for access on all of my servers.

michaeljs1990:
@wimnat spent an hour messing around with this tonight just to make sure that this was not something easily done with the dynamic inventory. The issue I run into is that i start an ASG and they all start coming up 2 or even 3 different playbooks are being run against different servers causing all kinds of trouble. I don't see any way to run this with dynamic inventory and then say get all hosts but only run on this current host. This module seems to be the optimal solution in my mind when you need to run the playbooks locally.

Once again i could be wrong but I don't think that dynamic inventory is suited for LCs.

wimnat:
Thanks for taking the time to try out dynamic inventory.  Maybe I'm not quite understanding your setup.  Where are you running your playbook from?  It sounds like Ansible is on each server and you're running the playbook on the server itself?  If that's the case, can't you use the ec2_facts module to get the info you need?

michaeljs1990:
@wimnat I had tried that a while ago but couldn't remember what the issue was so i tested it out again. The information brought back from that is only the current host. I looked through the module a bit but didn't see any way to alter this to bring back all hosts. since the server that is running this needs to know about the X amount of other servers with the same tags it doesn't work for clustering.

wimnat:
Why does the server need to know about the X amount of other servers?  It only needs to know one server in the cluster to join to right?  I'm just going of the rabbitmq clustering doc that says "the node chosen to cluster to does not matter".

Also, as a side-note, is there a particular reason you are configuring rabbitmq from scratch for an instance in an ASG?  Have you considered creating an AMI with most of this rabbitmq configuring leg work done and then you only need to think about the clustering part when the server starts?

I think this module might be useful though.  The current ec2_facts module uses metadata so maybe you could look at a PR for that module to make it optionally get remote data or perhaps rename this module to ec2_remote_facts.  Checkout http://docs.ansible.com/ansible/developing_modules.html checklist because you are missing a few things in this module, for example, don't use print statements.

michaeljs1990:
@wimnat You are correct it does only need to know about one other server so as long as you can get one other instance it doesn't matter. However connecting to multiple makes no difference.

I could create an AMI and use that but as far as the clustering issue it wouldn't solve any of that for me. I also just like having all of the logic in one place otherwise I would likely have a playbook just for making a standard rabbitmq server and then another that handles configuration. Not the worst thing in the world just not my preference.

I do like the idea of the name change that does make much more sense. I looked over the link you listed and it looks like print is ok as long as you are outputting valid json. I couldn't find anything that explicitly said don't use print. I want to let the user know that connecting to something failed such as trying to connect to the gov region in the off chance that they did have access to this.

@bcoca what do you think about using this module as opposed to making modifications to the existing ec2_facts module?

@wimnat I really appreciate you taking the time you have too look over this!

cloud/amazon/ec2_search.py
```diff
@@ -0,0 +1,157 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this library.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: ec2_search
+short_description: ask EC2 for information about other instances.
+description:
+    - Only supports seatch for hostname by tags currently. Looking to add more later.
+version_added: "1.9"
```
wimnat: Version 2.0

cloud/amazon/ec2_search.py
```diff
@@ -0,0 +1,157 @@
...
+options:
+  key:
+    description:
+      - instance tag key in EC2
+    required: false
+    default: Name
+    aliases: []
```
wimnat: No need for empty aliases

cloud/amazon/ec2_search.py
```diff
@@ -0,0 +1,157 @@
...
+        )
+    )
+
+    if not HAS_BOTO:
+        module.fail_json(msg='boto required for this module')
+
+    server_info = list()
+
+    for region in get_all_ec2_regions(module):
+        conn = connect_to_region(region, module)
```
wimnat: I would connect using the inbuilt ansible methods. Something like...

```
region, ec2_url, aws_connect_params = get_aws_connection_info(module)

if region:
    try:
        conn = connect_to_aws(boto.ec2, region, **aws_connect_params)
    except (boto.exception.NoAuthHandlerFound, StandardError), e:
        module.fail_json(msg=str(e))
else:
```

cloud/amazon/ec2_search.py
```diff
@@ -0,0 +1,157 @@
...
+                ec2_key = 'tag:' + module.params.get('key')
+                ec2_value = module.params.get('value')
+                reservations = conn.get_all_instances(filters={ec2_key : ec2_value})
+                for instance in [i for r in reservations for i in r.instances]:
+                    if instance.private_ip_address != None:
+                        instance.hostname = 'ip-' + instance.private_ip_address.replace('.', '-')
+                    if instance._state.name not in module.params.get('ignore_state'):
+                        server_info.append(todict(instance))
+        except:
+            print module.jsonify('error getting instances from: ' + region.name)
```
wimnat: Use module.fail_json()

cloud/amazon/ec2_search.py
```diff
@@ -0,0 +1,157 @@
...
+
+    ec2_facts_result = dict(changed=True, ec2=server_info)
```
wimnat: What's changed?

cloud/amazon/ec2_search.py
```diff
@@ -0,0 +1,157 @@
...
+            if module.params.get('lookup') == 'tags':
+                ec2_key = 'tag:' + module.params.get('key')
+                ec2_value = module.params.get('value')
+                reservations = conn.get_all_instances(filters={ec2_key : ec2_value})
+                for instance in [i for r in reservations for i in r.instances]:
+                    if instance.private_ip_address != None:
+                        instance.hostname = 'ip-' + instance.private_ip_address.replace('.', '-')
+                    if instance._state.name not in module.params.get('ignore_state'):
+                        server_info.append(todict(instance))
+        except:
```
wimnat: Be specific with what exception you're catching.  Maybe just boto exceptions?

cloud/amazon/ec2_search.py
michaeljs1990: I found another module that did this and they had changed set to true. I am not sure what the normal requirement for changed is. It does gather new data but it doesn't change anything so should it be false always?

cloud/amazon/ec2_search.py
michaeljs1990: sorry that was me just being lazy when i first wrote this. I'll look up what the exact exception is being thrown here is.

cloud/amazon/ec2_search.py
michaeljs1990: wont fail json cause the script to exit? it will always fail on the us-gov region. would it be better to add a param to filter out this region since very few people will use it but it is still possible that government contractors use ansible.

cloud/amazon/ec2_search.py
wimnat: As this module never changes any state, I think changed=False is correct

cloud/amazon/ec2_search.py
wimnat: Yes, fail_json causes an exit. Which is what you want.  You can't proceed without a region specified.  There's no need for any filter.

michaeljs1990: so the issue is this is looking in all regions so say i us-west comes back first and we get all the data. Then us-gov is up next and it fails then the entire script would fail when we really don't care if us-gov fails. Maybe I should force region to be a parameter for this role instead of automatically getting info about tags in all regions?

PR #761, opened 2015-07-25
Initial creation of Kubernetes Cluster Manager

kenjones-cisco:
Provides the ability to manage pods, replication controllers, and
services within a kubernetes cluster.


cloud/misc/kube.py
```diff
@@ -0,0 +1,297 @@
...
+    choices: ['present', 'absent', 'latest', 'reloaded', 'stopped']
+    default: present
+    description:
+      - present handles checking existence or creating if definition file provided,
+        absent handles deleting resource(s) based on other options,
+        latest handles creating ore updating based on existence,
+        reloaded handles updating resource(s) definition using definition file,
+        stopped handles stopping resource(s) based on other options.
+notes:
+  - "Requires kubectl for Kubernetes on servers"
```
bcoca: this should be in the requirements doc field

cloud/misc/kube.py
```diff
@@ -0,0 +1,297 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+DOCUMENTATION = """
+---
+module: kube
+short_description: Manage Kubernetes Cluster
+description:
+  - Create, replace, remove, and stop resources within a Kubernetes Cluster
+version_added: "1.9"
```
bcoca: at this point it should be version 2.0

cloud/misc/kube.py
```diff
@@ -0,0 +1,297 @@
...
+        else:
+            result = manager.create(check=False)
+
+    else:
+        module.fail_json(msg='Unrecognized state %s.' % state)
+
+    if result:
+        changed = True
+    module.exit_json(changed=changed,
+                     msg='success: {}'.format(' '.join(result))
```
bcoca: this forces python version 2.6 and above, normally we keep modules 2.4 compatible unless some library dependency forces us otherwise.

bcoca:
module looks good, just some minor stuff easy to correct.

one question, is kubectl the best way to interface with the cluster? is there no web api?

kenjones-cisco:
kubernetes does have REST APIs (http://kubernetes.io/third_party/swagger-ui/)

kubectl is the command line utility provided by kubernetes for managing the cluster.

mscherer:
The module will return "changed" almost each time, from a cursory look ( and if I am not wrong ), that's not very idempotent.

jimi-c:
In general, we try to avoid merging modules which simply shell out to a management program rather than access an API provided. I would much rather see this using the available API than kubectl.

kenjones-cisco:
Fair enough. I'll close for now and create a new one when I have spare cycles for that. This was created for a project I'm working on that wraps shortly. Thanks for all the feedback!

jimi-c:
@kenjones-cisco awesome, looking forward to it!

PR #773, opened 2015-07-27
clc_server: Ansible module for creating or deleting servers in Centurylink Cloud

popurisiva:
New module for servers in Centurylink Cloud.
This module adhere to the module development standards and best practices.
With this module, users can create or delete servers in Centurylink Cloud.

### DIFF
--- a/cloud/centurylink/clc_server.py
+++ b/cloud/centurylink/clc_server.py
@@ -1,0 +1,1373 @@
+#!/usr/bin/python
+
+#
+# Copyright (c) 2015 CenturyLink
+#
+# This file is part of Ansible.
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


cloud/centurylink/clc_server.py
```diff
@@ -0,0 +1,1373 @@
...
+        argument_spec = dict(
+            name=dict(),
+            template=dict(),
+            group=dict(default='Default Group'),
+            network_id=dict(),
+            location=dict(default=None),
+            cpu=dict(default=1),
+            memory=dict(default=1),
+            alias=dict(default=None),
+            password=dict(default=None),
```
resmo: Prevent passwords to be logged by adding `no_log=True`

popurisiva: Thank you @resmo for the comment. I have added the no_log attribute to password argument.

PR #778, opened 2015-07-28
New module - ec2_vpc_route_table_facts

wimnat:
Complements new ec2_vpc_route_table module


juliedavila:
+1
@wimnat in the future, for new modules can you start using boto3?

PR #783, opened 2015-07-29
New module - ec2_vpc_subnet_facts

wimnat:
Complements ec2_vpc_subnet module.  Similar to recently committed ec2_vpc_route_table_facts.

### DIFF
--- a/cloud/amazon/ec2_vpc_subnet_facts.py
+++ b/cloud/amazon/ec2_vpc_subnet_facts.py
@@ -1,0 +1,130 @@
+#!/usr/bin/python
+#
+# This is a free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This Ansible library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this library.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: ec2_vpc_subnet_facts
+short_description: Gather facts about ec2 VPC subnets in AWS
+description:
+    - Gather facts about ec2 VPC subnets in AWS
+version_added: "2.0"
+author: "Rob White (@wimnat)"
+options:
+  filters:
+    description:
+      - A dict of filters to apply. Each dict item consists of a filter key and a filter value. See U(http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeSubnets.html) for possible filters.
+    required: false
+    default: null
+  region:
+    description:
+      - The AWS region to use. If not specified then the value of the EC2_REGION environment variable, if any, is used. See U(http://docs.aws.amazon.com/general/latest/gr/rande.html#ec2_region)
+    required: false
+    default: null
+    aliases: [ 'aws_region', 'ec2_region' ]
+
+extends_documentation_fragment: aws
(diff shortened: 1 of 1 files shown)


wimnat:
@defionscode @rickmendes can you take a look at this and +1 if it's ok.  It is pretty much the same as the already merged route_table_facts module so i'd like to get this in too.

cloud/amazon/ec2_vpc_subnet_facts.py
```diff
@@ -0,0 +1,130 @@
...
+
+# Gather facts about any VPC subnet with a tag key Name and value Example
+- ec2_vpc_subnet_facts:
+    filters:
+      - "tag:Name": Example
+
+# Gather facts about any VPC subnet within VPC with ID vpc-abcdef00
+- ec2_vpc_subnet_facts:
+    filters:
+      - vpc-id: vpc-abcdef00
```
ghost: While testing I hit this error:

``` python
msg: argument filters is of invalid type: <type 'list'>, required: dict
```

I got around it by using:

``` python
filters:
  vpc_id: vpc-xxxxx
```

ghost:
@wimnat I tested this with my VPCs and it works well. See comment above about documentation issue.

Also, one of the things my team has been doing is updating ansible modules so they take names or ids. This is because we have name driven inventory in our shop. We find it more convenient if the module looks up the id from a name than our playbooks having to do it.

Any chance you would do that here for the filter objects like subnets and vpcs? I could see that being an additional PR if you don't want to do it here. Let me know. I can always put it on our TODO list.

Thanks.

wimnat:
@rickmendes thanks for reviewing.  I'm not quite clear on your requirement.  When you say you want to look up the id from a name, where is this name?  I assume this is different to a tag of 'Name'?

ghost:
@wimnat What I had in mind is similar to the feature I added to the ec2_elb_lb module. Call either get_all_vpcs or get_all_subnets and loop on the results looking for the name passed to the module. If found, grab the ID of vpc or subnet found.

This worked in my case because get_all_security_groups returns the name of each group in its results. I don't have firsthand knowledge of the vpc and subnet calls. I would hope if a name was set using the Name tag it would come back in those results along with the ID.

If they do not return those, I apologize for sending you on a wild goose chase.

wimnat:
Ahh i see. Yeh ELBs are a bit special as they have an actual "name" rather than route tables or subnets which don't have actual names but of course they can have a tag of "Name": "Something".

So you can just use the existing tag filter in this module if you want to do that.

E.g.
- ec2_vpc_subnet_facts:
  filters:
  - "tag:Name": Example

ghost:
@wimnat Nice. Thanks for the tip. I am good with that.

ghost:
@wimnat Sorry I forgot about coming back to this. It works well for me, so it gets a +1.

cloud/amazon/ec2_vpc_subnet_facts.py
wimnat: What version of boto?  vpc-id and vpc_id both work for me. I've got 2.38.0

ghost: @wimnat The workspace I tested this in still had boto 2.35.2. I upgraded it to 2.38.0 and got the same error message when using a list.

wimnat: Ah yeh i see, sorry i was looking at the filter itself rather than it being a list.  I've updated the examples.

wimnat:
@gregdek Greg, it says some checks are pending but it looks like the build passed and i think there's an +1 and my implicit +1 here for merge.  Anything i need to do?

gregdek:
Reran the travis checks and everything looks ok. Thanks @wimnat for the PR and thanks @rickmendes for the +1. Marking for inclusion.

robynbergeron:
Taking out of pending_action and applying the actual label for inclusion, as I think @gregdek meant to do above :)

PR #787, opened 2015-07-29
postgresql_ext: bugfix - change check mode call db_exists() -> ext_exists()

bmoar:
When in check mode, we want to check if the module exists
right here, not the database. If we are at this line,
we are connected to the database already so a db_exists
is not only redundant, but an error, because the db_exists()
function is undefined. Changing db_exists() to ext_exists()
is the only logical choice I could see to make this run
correctly in check mode.

As of now, check mode for this module doesn't work because db_exists() is undefined.

### DIFF
--- a/database/postgresql/postgresql_ext.py
+++ b/database/postgresql/postgresql_ext.py
@@ -137,7 +137,7 @@ def main():
     state = module.params["state"]
     changed = False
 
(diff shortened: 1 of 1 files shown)


lieutdan13:
+1 @bmoar I was searching for a fix and found this PR from you! Nice work.

PR #788, opened 2015-07-30
Port vsphere_copy from httplib to open_url for TLS cert validation

abadger:
vsphere_copy was using raw httplib which does not verify the server's SSL certificate. Ported to ansible's open_url() which does certificate checking.

@dagwieers - I don't use vsphere so if you are able to review this and verify that it still works, that would be greatly appreciated. Since this change will prevent man-in-the-middle attacks I will merge this even if you don't get back to me but I would vastly prefer you (or someone else using vsphere_copy) take a look to make sure this works.

### DIFF
--- a/cloud/vmware/vsphere_copy.py
+++ b/cloud/vmware/vsphere_copy.py
@@ -141,17 +141,17 @@ def main():
     except socket.error, e:
         if isinstance(e.args, tuple) and e[0] == errno.ECONNRESET:
             # VSphere resets connection if the file is in use and cannot be replaced
-            module.fail_json(msg='Failed to upload, image probably in use', status=e[0], reason=str(e), url=url)
+            module.fail_json(msg='Failed to upload, image probably in use', status=None, errno=e[0], reason=str(e), url=url)
         else:
-            module.fail_json(msg=str(e), status=e[0], reason=str(e), url=url)
+            module.fail_json(msg=str(e), status=None, errno=e[0], reason=str(e), url=url)
     except Exception, e:
-        status = -1
+        error_code = -1
         try:
             if isinstance(e[0], int):
-                status = e[0]
+                error_code = e[0]
         except KeyError:
             pass
-        module.fail_json(msg=str(e), status=status, reason=str(e), url=url)
+        module.fail_json(msg=str(e), status=None, errno=error_code, reason=str(e), url=url)
 
     status = r.getcode()
     if 200 <= status < 300:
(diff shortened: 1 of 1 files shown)


dagwieers:
@abadger I support this change (provided it still works ;-)

This module was written for a client. As I don't work there anymore I will ask them to test your change.

Do you know what cert bundle ansible/python uses in this case ?

PS While reading your code I notice a typo (satus instead of status).
PS2 I am also not in favor of using the same variable for two different things in the same block (status is used for the HTTP response, but also for the Exception handling errno). It confused me when going over the code.
PS3 Maybe it is easier to simply do: `if 200 <= status < 300:`. I prefer reading such "interval comparisons" over a double test.

abadger:
Fixed the status conditional.  About the naming of the status variable... I was just following the existing code there... The existing code uses a field called status in the return value.  If there's a socket error, the status is the erno otherwise it's the http status code.  I can rename the variable but it seems like the return value is the bigger problem as that's visible from user's playbooks.  Since this hasn't gone out in a released ansible yet, perhaps we want to change the name of that as well?

dagwieers:
@abadger I agree with your assessment theoretically wrt. using status for the JSON output.

However considering the implications this has on playbooks it is important to reuse the same JSON output where possible IMO. Even when the content may have a different meaning.
So my suggestion to change the variable name status within the block has no impact whatsoever (except readability) whereas changing the JSON status variable name changes the interface to the user (up to the point that users have to test for the existence of a property before using it, which breaks readability and possibly compatibility). At least if the code clearly uses a different variable name for different origins it indicates the JSON status is being used for two different purposes (depending on success/failure type).

I don't mind a second opinion regarding this from another developer. (And yes, we're splitting hairs here ;-)).

bcoca:
I would make both status distinct, but also always return them, since this is new to 2.0 I don't expect many playbooks to break by the alteration.

abadger:
Split out separate errno and status fields as @bcoca requested.  If one is not applicable I return None in that field.  I only changed the fail_json() cases to return both.  I think we should treat the fail_json() and exit_json() pathways separately (fail_json() is more like an exception route and exit_json() more like the normal return from a function).  Sending extraneous fields does have an impact on speed (because more data has to be transferred) so I think splitting the fail and success cases is a logical place to differentiate what is returned.

We could rename the status variable to http_status and keep the current status as it existed before for backwards compatibility if we wanted to, however, the module was only added to the repository at the end of June and hasn't gone out into a released ansible yet.  So it seems like we should still be free to change the interface at this point.  @dagwieers sounds like your client may have been using this for a while before it was added here.  Do you know if they were making use of the status field?  It's not a big deal to keep backwards compat if it solves a real problem for them... just disappointing that we can't keep the code clean of deprecations when it's so newly added.

Oh and for your cert bundle question:  There's a number of locations that are merged and checked, take a look here: https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/urls.py#L436

abadger:
@dagwieers any information here?

gregdek:
Is the hair-splitting worth keeping this change in limbo? :)

dagwieers:
No, please proceed without me.
If it has been tested by others and it works, I have no objections !

PR #794, opened 2015-07-31
Modules to support AWS ECS and some helper fact-getters

Java1Guy:
(This PR depends on #11818 in ansible/ansible and will be tested by a PR to come.)
These modules implement AWS ECS functionality.  For reference, see http://docs.aws.amazon.com/AmazonECS/latest/developerguide/ECS_GetStarted.html.  The idea is that enough of the Boto API is exposed to enable successful creation and operation of ECS clusters and tasks.

I have also added some helpers which gather facts about the AWS environment (IAM roles, EC2 VPCs, etc.)  The main one, ec2_remote_facts, was based off the ansible/contrib/inventory/ec2.py module.  As such, it requires the ec2.ini file.  Ideally, this would be abstracted away.  These are mostly required in order to execute a successful test run of creating a cluster, launching an ec2 instance in that cluster to run a task against it.


cloud/amazon/ec2.ini
```diff
@@ -0,0 +1,136 @@
+# Ansible EC2 external inventory script settings
```
jimi-c: Why is this file here? No modules should be using an ini file for data. If there are common areas for the facts to check, they should be moved to `module_common/ec2.py` as a special param list for ECS. All settings for modules must go through parameters.

Java1Guy:
@jimi-c I commited code which removes the ini file and puts the parameter processing in the module.

gregdek:
@Java1Guy...

Hmmm... this is kind of a huge PR, and it now appears that there are merge conflicts, although not immediately evident to me what those conflicts are. 

I guess I would ask if there's a way to break this up into a bunch of smaller PRs with one or two modules per PR. We've got a pretty big backlog of new modules as it is, and it seems like this will be one of the last ones to be reviewed if it has to be swallowed in one big chunk like this.

gregdek:
Marking as needs_rebase, which it apparently does, but I really think we should break this one up.

bennojoy:
@Java1Guy  i agree with @gregdek can you please break it up into one pr per module as it makes the reviewing easier. Also i think there  are a lots of places in documentation were version mentioned is 1.9 also content seems like copied from other modules retaining the previous authors etc.. can you please clean it up and also try doing an ansible-doc -m <module_name> to check if the documentation is rendered right.

Also i may not have understood fully why we need facts modules like vpc and remote_facts etc.. so when you split up the PR could you also mentioned how this facts module are used/required for ecs.

Java1Guy:
That is fine - I will break this up... the down side of that is that all the modules are needed to really do anything productive in AWS ECS, but not a problem.   FYI - the conflict now comes about because somebody checked in a file of the same name I was adding.  I will review to figure out the merge or other proper strategy.
@bennojoy The facts modules are required for the tests to know how/where to create instances and revised facts modules were required because the existing modules login to the ec2s, etc. to determine the facts -- all that is really needed is to use the AWS API to characterize the VPCs etc.

Java1Guy:
All files have been merged into other pull requests, so I am closing this one.

Issue #795, opened 2015-08-01
zypper_repository can't add repositories if current repository list is emty


toabctl:
##### Issue Type:

Bug Report
##### Ansible Version:

$ ansible --version
ansible 1.9.1
  configured module search path = None
##### Ansible Configuration:

Nothing changed
##### Environment:

openSUSE Tumbleweed
##### Summary:

When adding new repositories with "zypper_repository" and the list of current repositories is empty, "zypper_reppository" fais. The reason is that "zypper_repository" checks the returncode of "zypper -x -lr" which return 6 (see "man zypper") instead of 0. 6 means that there are no repositories defined.
##### Steps To Reproduce:
- delete all the repositories
- run on that machine:

```

---
tasks:
- zypper_repository: repo="/tmp/" name="foobar"
```
##### Expected Results:

Repository should be added
##### Actual Results:

```
failed: [192.168.122.102] => {"cmd": "/usr/bin/zypper -x lr", "failed": true, "rc": 6}
stdout: <?xml version='1.0'?>
<stream>
<repo-list>
</repo-list>
</stream>


FATAL: all hosts have already failed -- aborting
```

PR #796, opened 2015-08-01
zypper_repository: Fix repo parsing for empty list

toabctl:
When no repositories are defined in zypper, the return code
of "zypper repos" is 6. Handle that case and don't fail
if zypper_repository has to deal with an empty repo list.

Fixes https://github.com/ansible/ansible-modules-extras/issues/795

### DIFF
--- a/packaging/os/zypper_repository.py
+++ b/packaging/os/zypper_repository.py
@@ -95,21 +95,30 @@ def zypper_version(module):
 def _parse_repos(module):
     """parses the output of zypper -x lr and returns a parse repo dictionary"""
     cmd = ['/usr/bin/zypper', '-x', 'lr']
-    repos = []
-
     from xml.dom.minidom import parseString as parseXML
-    rc, stdout, stderr = module.run_command(cmd, check_rc=True)
-    dom = parseXML(stdout)
-    repo_list = dom.getElementsByTagName('repo')
-    for repo in repo_list:
-        opts = {}
-        for o in REPO_OPTS:
-            opts[o] = repo.getAttribute(o)
-        opts['url'] = repo.getElementsByTagName('url')[0].firstChild.data
-        # A repo can be uniquely identified by an alias + url
-        repos.append(opts)
-
-    return repos
+    rc, stdout, stderr = module.run_command(cmd, check_rc=False)
+    if rc == 0:
+        repos = []
+        dom = parseXML(stdout)
+        repo_list = dom.getElementsByTagName('repo')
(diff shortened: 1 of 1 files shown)


packaging/os/zypper_repository.py
```diff
@@ -95,21 +95,29 @@ def zypper_version(module):
...
+        return repos
+    # exit code 6 is ZYPPER_EXIT_NO_REPOS (no repositories defined)
+    elif rc == 6:
+        return []
+    else:
+        #module.fail_json(msg=stderr if stderr else stdout)
+        if stderr:
+            module.fail_json(msg=stderr)
+        else:
+            module.fail_json(msg=stdout)
```
abadger: The old code didn't fail if it couldn't parse the repo files.  I don't use zypper but I think is reasonable to fail in this case.  Probably want to include the rc code as a field of the return value.

I also don't know what the stdout and stderr might contain.  If they aren't really detaialed might be good to include some information about where this is failing.  Something like:

``` python

msg='Failed to parse existing zypper repository files: %s' % stderr
```

abadger:
Just one comment about including more information if we fail in this step.  Once that's fixed, this is good to merge.

toabctl:
> Just one comment about including more information if we fail in this step. Once that's fixed, this is good to merge.

@abadger Done in https://github.com/ansible/ansible-modules-extras/commit/7d9988c96119ce03525d792f7d4d3e8fd9ecfe22 . But I wonder if I should just say in the `msg` the exact command and add `stderr` and `stdout`  to the json as extra keys. Something like:

```
d = { 'zypper_exit_code': rc }
if stderr:
    d['stderr'] = stderr
if stdout:
    d['stdout'] = stdout                                                                                                                                            
module.fail_json(msg='Failed to execute "%s"' % " ".join(cmd), **d)
```

But `stdout` and `stderr` are anyway printed in an error case, right? So that's just useless overhead I guess.
No idea what the common pattern in ansible is. Or is this good enough as it is now?

abadger:
Your idea sounds good.  stdout and stderr I believe are only printed if the
module exits with a non-zero exit code.  Returning via fail_json() doesn't
print stdout and stderr (this lets us handle errors more gracefully as we
can control precisely what goes into the fail_json() data.

toabctl:
@abadger Done now. Fine to merge it or still something todo?

abadger:
Looks good.  Thanks!

# 

Hi!

issue is resolved.

Thank you!

PR #797, opened 2015-08-02
win_iis_website: Workaround bug in Get-Website cmdlet.

cchurch:
Workaround for issue described here: http://stackoverflow.com/questions/4170530/powershell-get-website-name-parameter-is-ignored


rriggs:
:+1:  Fixes the problem for me.

PR #798, opened 2015-08-03
strange results with services=all

vaupelt:
I issued a command with action=disable_alerts host=webserver services=all set and get this results:

```
"nagios_commands": [
    "[1438593631] DISABLE_SVC_NOTIFICATIONS;webserver;a", 
    "[1438593631] DISABLE_SVC_NOTIFICATIONS;webserver;l", 
    "[1438593631] DISABLE_SVC_NOTIFICATIONS;webserver;l"
]
```

This is not a big deal because i have just overlooked the action=silence command. Nevertheless a more predictable result would be a nice thing to have.

### DIFF
--- a/monitoring/nagios.py
+++ b/monitoring/nagios.py
@@ -913,6 +913,8 @@ class Nagios(object):
         elif self.action == 'enable_alerts':
             if self.services == 'host':
                 self.enable_host_notifications(self.host)
+            elif self.services == 'all':
+                self.enable_host_svc_notifications(self.host)
             else:
                 self.enable_svc_notifications(self.host,
                                               services=self.services)
@@ -920,6 +922,8 @@ class Nagios(object):
         elif self.action == 'disable_alerts':
             if self.services == 'host':
                 self.disable_host_notifications(self.host)
+            elif self.services == 'all':
+                self.disable_host_svc_notifications(self.host)
             else:
                 self.disable_svc_notifications(self.host,
                                                services=self.services)


gregdek:
@tbielawa Please review this PR to make sure it adheres to the following guidelines:

timlnx:
@gregdek @vaupelt  I'll gladly check this over.

I've got some Dr appointments this afternoon. I'll check back probably closer to 5pm EST.

robynbergeron:
@tbielawa -- any progress on this? :) Just checking in.

timlnx:
Oh I see what it did there. Without that conditional @vaupelt added it was splitting 'all' into a character list. Whoops. Hahaha, I have no idea how this didn't get noticed like.... 3 years ago :confused: Great catch.

With respect to what the patch is intended to fix, I would say it is shippable. However:

@gregdek You pointed out those module guidelines, I can see some which are in violation:
- `The version is not a float number and value the current development version` all the 'version' related strings in the doc header are using floats, i.e., `version_added: "0.7"` and `version_added: "2.0"` Not sure if that's a blocker. To be honest, this guideline in particular is not written that clearly to me.
- `default is not needed for required: true` There are at least 2 (maybe more) instances in the doc string where a parameter is marked as `required: true` and the `default` parameter is present (usually set to `null`)

The PR by itself is A+ mergable, it addresses the root cause of the OP's issue. I don't think it's their responsibility to fix the docs. As pointed out above, there are some lingering issues with the docstring.

Sorry for dragging my feet on this. Thanks for the follow-up poke.

PR #806, opened 2015-08-03
monit: Add retry for pending/initializing services

gaqzi:
This is the same as the pull request for #572. :)
ping @bcoca @gregdek

> If there are already ongoing actions for a process managed by monit, the
> module would exit unsuccessfully. It could also give off false positives
> because it did not determine whether the service was started/stopped
> when it was in a pending state. Which might be turning the service off,
> but the action was to start it.
> 
> For example "Running - pending stop" would be regarded as the service
> running and "state=enabled" would do nothing.
> 
> This will make Ansible wait for the state to finalize, or a timeout decided
> by the new `max_retries` option, before it decides what to do.
> 
> This fixes issue #244.

### DIFF
--- a/monitoring/monit.py
+++ b/monitoring/monit.py
@@ -18,6 +18,7 @@
 # You should have received a copy of the GNU General Public License
 # along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
 #
+from time import sleep
 
 DOCUMENTATION = '''
 ---
@@ -38,6 +39,12 @@ options:
     required: true
     default: null
     choices: [ "present", "started", "stopped", "restarted", "monitored", "unmonitored", "reloaded" ]
+  max_retries:
+    description:
+      - If there are pending actions for the service monitoried by monit Ansible will retry this
+        many times to perform the requested action. Between each retry Ansible will sleep for 1 second.
+    required: false
+    default: 10
 requirements: [ ]
 author: "Darryl Stoflet (@dstoflet)" 
 '''
@@ -50,6 +57,7 @@ EXAMPLES = '''
 def main():
     arg_spec = dict(
         name=dict(required=True),
+        max_retries=dict(default=10),
         state=dict(required=True, choices=['present', 'started', 'restarted', 'stopped', 'monitored', 'unmonitored', 'reloaded'])
     )
 
@@ -57,6 +65,7 @@ def main():
 
     name = module.params['name']
     state = module.params['state']
+    max_retries = module.params['max_retries']
 
     MONIT = module.get_bin_path('monit', True)
 
@@ -103,7 +112,21 @@ def main():
                 module.exit_json(changed=True, name=name, state=state)
         module.exit_json(changed=False, name=name, state=state)
 
-    running = 'running' in process_status 
+    running_status = status()
+    retries = 0
+    while 'pending' in running_status or 'initializing' in running_status:
+        if retries >= max_retries:
+            module.fail_json(
+                msg='too many retries waiting for "pending" or "initiating" to go away (%s)' % running_status,
+                retries=retries,
+                state=state
+            )
+
+        sleep(1)
+        retries += 1
+        running_status = status()
+
+    running = 'running' in status()
 
     if running and state in ['started', 'monitored']:
         module.exit_json(changed=False, name=name, state=state)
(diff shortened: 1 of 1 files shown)


monitoring/monit.py
```diff
@@ -38,6 +39,12 @@
     required: true
     default: null
     choices: [ "present", "started", "stopped", "restarted", "monitored", "unmonitored", "reloaded" ]
+  max_retries:
+    description:
+      - If there are pending actions for the service monitoried by monit Ansible will retry this
+        many times to perform the requested action. Between each retry Ansible will sleep for 1 second.
+    required: false
+    default: 10
```
resmo: `version_added: "2.0"` is missing.

monitoring/monit.py
```diff
@@ -50,6 +57,7 @@
 def main():
     arg_spec = dict(
         name=dict(required=True),
+        max_retries=dict(default=10),
```
resmo: I would add a `type='int'` as an integer is expected for `max_retries`

gaqzi:
@resmo: Thanks, I've added the `version_added` as well as the type declaration for `max_retries`! :)

gregdek:
@dstoflet Please review this PR to make sure it adheres to the following guidelines:

gregdek:
Hi @dstoflet -- are you available to review this PR? Please let us know either way.

gregdek:
Looks like @dstoflet is unavailable to review.

@brian-brazil any chance you might be able to help with maintainership of this module? I ask because you have second-most lines of code committed in this module.

monitoring/monit.py
```diff
@@ -86,8 +88,35 @@ def run_command(command):
         module.run_command('%s %s %s' % (MONIT, command, name), check_rc=True)
         return status()
 
-    process_status = status()
-    present = process_status != ''
+    def wait_for_monit_to_stop_pending(sleep_time=1):
```
brian-brazil: I think you need more than 10s, some services can take minutes to startup or shutdown. I'd suggest a timeout of say 5m rather than a retry limit, and have say 10s between attempts.

brian-brazil:
@gregdek I'm actually no longer using monit, though in my previous company I did something similar to this change that never made it upstream.

monitoring/monit.py
gaqzi: sure, I can absolutely up that. I'm a bit wary of 10s between checks though. The services I'm running usually go into start/stop state in about 2-3 seconds and waiting that long would seem excessive. 

Maybe look at a gradual back-off up to 10s?

brian-brazil: I'd say keep it simple and go with say a straight 5s.

We need to balance wasting CPU and slowing things down by polling too much, versus increased latency.

gaqzi: :+1:

gaqzi: @brian-brazil I've changed it to do a 300s timeout instead. With a 5 second wait between each check. :)

monitoring/monit.py
```diff
@@ -38,6 +39,14 @@
     required: true
     default: null
     choices: [ "present", "started", "stopped", "restarted", "monitored", "unmonitored", "reloaded" ]
+  timeout:
+    description:
+      - If there are pending actions for the service monitored by monit, then Ansible will check
+        for up to this long to verify the the requested action has been performed.
```
brian-brazil: for up to this many seconds

This makes the unit explicit

monitoring/monit.py
```diff
@@ -86,8 +89,35 @@ def run_command(command):
         module.run_command('%s %s %s' % (MONIT, command, name), check_rc=True)
         return status()
 
-    process_status = status()
-    present = process_status != ''
+    def wait_for_monit_to_stop_pending():
+        """Fails this run if there is no status or it's pending/initalizing for timeout"""
+        start_time = time.time()
+        timeout_time = start_time + timeout
```
brian-brazil: You can do `timeout_time = time.time() + timeout`

These are simple expressions, so fitting it into one line makes it a tad more readable.

brian-brazil:
:+1:

monitoring/monit.py
gaqzi: fixed

monitoring/monit.py
gaqzi: fixed. note to self: when refactoring check for variables used only once :P

gregdek:
Thanks @brian-brazil -- I'll add you to the maintainers list for this one. Even though you're not using it, you're probably still the best option. :) If it becomes too onerous, let us know.

Marking this one for inclusion. Thanks @gaqzi for the PR.

PR #815, opened 2015-08-07
Update lldp.py

wenottingham:
Add that it requires lldpctl.


bcoca:
requirements should be documented in the requirements entry

gregdek:
@wenottingham pls to put requirement in requirements section :)

abadger:
Thanks!

PR #820, opened 2015-08-10
Add cs_ip_address module

DazWorrall:
Hopefully @resmo can give this a look over. To my knowledge there's no way to make the association idempotent, so I've made notes in the documentation to that effect.

### DIFF
--- a/cloud/cloudstack/cs_ip_address.py
+++ b/cloud/cloudstack/cs_ip_address.py
@@ -1,0 +1,459 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+#
+# (c) 2015, Darren Worrall <darren@iweb.co.uk>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible. If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: cs_ip_address
+short_description: Manages Public/Secondary IP address associations
+description:
+    - Acquires and associates a public IP to an account. Due to API limitations,
+    - this is not an idempotent call, so be sure to only conditionally call this
+    - when C(state=present)
+version_added: '2.0'
+author: "Darren Worrall @dazworrall"
+options:
+  ip_address:
+    description:
+      - Public IP address.
+      - Required if C(state=absent)
+    required: true
+  domain:
+    description:
+      - Domain the IP address is related to.
+    required: false
+    default: null
+  network:
+    description:
+      - Network the IP address is related to.
+    required: false
+    default: null
+  account:
+    description:
+      - Account the IP address is related to.
+    required: false
+    default: null
+  project:
+    description:
+      - Name of the project the IP address is related to.
+    required: false
+    default: null
+  zone:
+    description:
+      - Name of the zone in which the virtual machine is in.
+      - If not set, default zone is used.
+    required: false
+    default: null
+  poll_async:
+    description:
+      - Poll async jobs until job has finished.
+    required: false
+    default: true
+extends_documentation_fragment: cloudstack
+'''
+
+EXAMPLES = '''
+# Associate an IP address
+- local_action:
+    module: cs_ip_address
+    account: My Account
+  register: ip_address
+  when: create_instance|changed
+
+# Disassociate an IP address
+- local_action:
+    module: cs_ip_address
+    ip_address: 1.2.3.4
+    state: absent
+'''
+
+RETURN = '''
+---
+ip_address:
+  description: Public IP address.
+  returned: success
+  type: string
+  sample: 1.2.3.4
+zone:
+  description: Name of zone the IP address is related to.
+  returned: success
+  type: string
+  sample: ch-gva-2
+project:
+  description: Name of project the IP address is related to.
+  returned: success
+  type: string
+  sample: Production
+account:
+  description: Account the IP address is related to.
+  returned: success
+  type: string
+  sample: example account
+domain:
+  description: Domain the IP address is related to.
+  returned: success
+  type: string
+  sample: example domain
+'''
+
+
+try:
+    from cs import CloudStack, CloudStackException, read_config
+    has_lib_cs = True
+except ImportError:
+    has_lib_cs = False
+
+# import cloudstack common
+class AnsibleCloudStack:
+
+    def __init__(self, module):
+        if not has_lib_cs:
+            module.fail_json(msg="python library cs required: pip install cs")
+
+        self.result = {
+            'changed': False,
+        }
+
+        self.module = module
+        self._connect()
+
+        self.domain = None
+        self.account = None
+        self.project = None
+        self.ip_address = None
+        self.zone = None
+        self.vm = None
+        self.os_type = None
+        self.hypervisor = None
+        self.capabilities = None
+
+
+    def _connect(self):
+        api_key = self.module.params.get('api_key')
+        api_secret = self.module.params.get('secret_key')
+        api_url = self.module.params.get('api_url')
+        api_http_method = self.module.params.get('api_http_method')
+        api_timeout = self.module.params.get('api_timeout')
+
+        if api_key and api_secret and api_url:
+            self.cs = CloudStack(
+                endpoint=api_url,
+                key=api_key,
+                secret=api_secret,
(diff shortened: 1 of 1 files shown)


resmo:
Thanks @DazWorrall, will look at it.

cloud/cloudstack/cs_ip_address.py
```diff
@@ -0,0 +1,459 @@
...
+
+
+try:
+    from cs import CloudStack, CloudStackException, read_config
+    has_lib_cs = True
+except ImportError:
+    has_lib_cs = False
+
+# import cloudstack common
+class AnsibleCloudStack:
```
resmo: This class should be replaced with 

```
from ansible.module_utils.cloudstack import *
```

in official ansible extras modules. 

I know it is not that easy to test it and I use it this way in inofficial https://github.com/resmo/ansible-cloudstack.

DazWorrall: Ah, thats where I based the module on. I actually tweak the behavior of AnsibleCloudStack.get_ip_address so it doesn't fail if the ip address cant be found (to handle the idempotent absent case). I'll take a look and make some changes.

resmo: I see, you could override it in `AnsibleCloudStackIPAddress`. That would be fine.

cloud/cloudstack/cs_ip_address.py
```diff
@@ -0,0 +1,459 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible. If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: cs_ip_address
+short_description: Manages Public/Secondary IP address associations
+description:
+    - Acquires and associates a public IP to an account. Due to API limitations,
```
resmo: I would remove `to an account` or add `or project` since it also works for projects. Btw you don't need to make a yaml list for this description, a single list item braked into lines will be fine.

cloud/cloudstack/cs_ip_address.py
```diff
@@ -0,0 +1,459 @@
...
+                self.result['ip_address'] = ip_address['ipaddress']
+            if 'id' in ip_address:
+                self.result['id'] = ip_address['id']
+        return self.result
+
+
+def main():
+    module = AnsibleModule(
+        argument_spec = dict(
+            ip_address = dict(required=False),
```
resmo: as per doc definition, this should be `required=True`

DazWorrall, line 229: It's not required for state=present, only state=absent. Can I configure that on the module?

DazWorrall:
`get_networks` is now duplicated in a couple of places, I'm happy to help put that on the parent class and test separately in another PR. I need to change the behavior of `get_ip_address` so I've overridden it for now, but again am happy to help make changes to the parent class.

I'm using this play for testing by the way, if it's any use to you: https://gist.github.com/DazWorrall/0d203266989418abfa1a

resmo:
I would not change the behaviour of `get_ip_address()` in the parent class. Overriding is really fine in this case, for now at least.

I thought about how we could make this idempotent. It looks to me we can use `listPublicIpAddresses` as you already did but return `None` if ip has not been found. This way we could check if the IP is already acquired and do not need to always return `changed=True`.

What do you think about? Or why did you write about an "API limitation"?

cloud/cloudstack/cs_ip_address.py
resmo: ok I see.

I was already one step further in my thinking about impotence. See the https://github.com/ansible/ansible-modules-extras/pull/820#issuecomment-129501956.

DazWorrall:
You cant ask [`associateIpAddress`](https://cloudstack.apache.org/api/apidocs-4.5/domain_admin/associateIpAddress.html) for a particular address, and there's no way to know ahead of time what address you're going to get allocated. My use case is assigning any available public IP from the pool to a newly created instance, without knowing or caring what it is.

I suppose we could create a higher level module which wrapped up some more API calls with some opinionated logic - `cs_instance_public_ip` maybe. If the instance already had a static nat; do nothing, if not; associate a public IP if there's none spare and then add the static nat.

resmo:
Yep, that was it. You never now what IP you will get.

I was actually already about making this module few months ago, and remember struggling about this too, I didn't find a satisfying answer.  I am fine with the way as you implemented it then!

cloud/cloudstack/cs_ip_address.py
```diff
@@ -0,0 +1,269 @@
...
+    - Acquires and associates a public IP to an account or project. Due to API
+      limitations this is not an idempotent call, so be sure to only
+      conditionally call this when C(state=present)
+version_added: '2.0'
+author: "Darren Worrall @dazworrall"
+options:
+  ip_address:
+    description:
+      - Public IP address. Required if C(state=absent)
+    required: true
```
resmo: in this case this should be
~~~
required: true
default: null
~~~

cloud/cloudstack/cs_ip_address.py
```diff
@@ -0,0 +1,269 @@
...
+extends_documentation_fragment: cloudstack
+'''
+
+EXAMPLES = '''
+# Associate an IP address
+- local_action:
+    module: cs_ip_address
+    network: My Network
+  register: ip_address
+  when: create_instance|changed
```
resmo: or 

```
when: instance_result.public_ip is not defined
```

?

cloud/cloudstack/cs_ip_address.py
resmo, line 229: No, just in the doc as you did.

cloud/cloudstack/cs_ip_address.py
resmo: change it to 

```
required: false
default: null
```

as discussed.

resmo:
these 2 final tweaks and this is good to go! Appreciate your work.

cloud/cloudstack/cs_ip_address.py
DazWorrall: That is better yes, good idea!

DazWorrall:
That's done then, thanks for your help on this and for the cs_\* modules.

gregdek:
Looks like @resmo approves! Marked for inclusion.

cloud/cloudstack/cs_ip_address.py
```diff
@@ -0,0 +1,270 @@
...
+            self.ip_address = ip_addresses['publicipaddress'][0]
+        return self._get_by_key(key, self.ip_address)
+
+    def associate_ip_address(self):
+        self.result['changed'] = True
+        args = {}
+        args['account'] = self.get_account(key='id')
+        args['networkid'] = self.get_network(key='id')
+        args['zoneid'] = self.get_zone(key='id')
+        ip_address = {}
```
resmo: could you also pass domain and project

resmo:
@gregdek found some more tweaks to do. Not ready yet.

cloud/cloudstack/cs_ip_address.py
```diff
@@ -0,0 +1,270 @@
...
+        ip_addresses = self.cs.listPublicIpAddresses(**args)
+
+        if ip_addresses:
+            self.ip_address = ip_addresses['publicipaddress'][0]
+        return self._get_by_key(key, self.ip_address)
+
+    def associate_ip_address(self):
+        self.result['changed'] = True
+        args = {}
+        args['account'] = self.get_account(key='id')
```
resmo: this must be `self.get_account(key='name')` and also pass domainid, projectid, and zoneid e.g.

```
args['account'] = self.get_account(key='name')
args['domainid'] = self.get_domain(key='id')
args['projectid'] = self.get_project(key='id')
args['zoneid'] = self.get_zone('id')
```

cloud/cloudstack/cs_ip_address.py
```diff
@@ -0,0 +1,270 @@
...
+    required: false
+    default: null
+  project:
+    description:
+      - Name of the project the IP address is related to.
+    required: false
+    default: null
+  zone:
+    description:
+      - Name of the zone in which the virtual machine is in.
```
resmo: virtual machine? should be `network` or `ip_address` or generic `resource`

DazWorrall:
Parameter changes and doc fixes done

cloud/cloudstack/cs_ip_address.py
```diff
@@ -0,0 +1,272 @@
...
+        ip_address = self.module.params.get('ip_address')
+        if not ip_address:
+            self.module.fail_json(msg="IP address param 'ip_address' is required")
+
+        args = {}
+        args['ipaddress'] = ip_address
+        args['account'] = self.get_account(key='name')
+        args['domainid'] = self.get_domain(key='id')
+        args['projectid'] = self.get_project(key='id')
+        args['listall'] = True
```
resmo: hmm I try to avoid `args['listall'] = True`. It returns more values as needed and tend to make problems when using as root admin. It seems not be an issue here, since public IP addresses are unique (hopefully). Would you mind removing it? Or is there any good reason?

DazWorrall: Done, I understand where I went wrong now after that other PR :)

resmo:
@gregdek LGTM shipit :+1: :shipit:

Issue #821, opened 2015-08-10
puppet module: puppetmaster and show_diff options are broken


bethanylong:
**Ansible Version/Configuration:**

```
$ ansible --version
ansible 2.0.0
  config file = 
  configured module search path = None
```

**Environment:** Running from FreeBSD 9.3; managing FreeBSD 9.3 and 10.1, and Debian 7

**Summary:** The puppet module tries to use the (nonexistent) `--show-diff` Puppet option instead of (correct) `--show_diff`, and it misspells the `--server` Puppet option as `-- server`.

**Steps To Reproduce:** Try to use the puppet module with the `show_diff` option set, or with the `puppetmaster` option set and no way to resolve the default "puppet" host name on the Puppet agent.

**Expected results:** Any evidence of a successful Puppet agent run when these options are used

**Actual results:**

```
TASK [puppet puppetmaster=puppetmaster.fqdn] ***********************************
fatal: [agent.fqdn]: FAILED! => {"changed": false, "disabled": false, "error": true, "failed": true, "msg": "puppet did not run", "rc": 1, "stderr": "", "stdout": "\u001b[1;35merr: Could not retrieve catalog from remote server: getaddrinfo: Name or service not known\u001b[0m\n\u001b[0;33mwarning: Not using cache on failed catalog\u001b[0m\n\u001b[1;35merr: Could not retrieve catalog; skipping run\u001b[0m\n\u001b[1;35merr: Could not send report: getaddrinfo: Name or service not known\u001b[0m\n", "stdout_lines": ["\u001b[1;35merr: Could not retrieve catalog from remote server: getaddrinfo: Name or service not known\u001b[0m", "\u001b[0;33mwarning: Not using cache on failed catalog\u001b[0m", "\u001b[1;35merr: Could not retrieve catalog; skipping run\u001b[0m", "\u001b[1;35merr: Could not send report: getaddrinfo: Name or service not known\u001b[0m"]}

TASK [puppet show_diff=yes] ****************************************************
fatal: [agent.fqdn]: FAILED! => {"changed": false, "disabled": false, "error": true, "failed": true, "msg": "puppet did not run", "rc": 1, "stderr": "Could not parse options: invalid option: --show-diff\n", "stdout": "", "stdout_lines": []}
```

bcoca:
cc @emonty

bethanylong:
Oops, you beat me to it :)

PR #822, opened 2015-08-10
Fix puppet module's puppetmaster and show_diff options

bethanylong:
Issue #821


bethanylong:
@bcoca just fixed this.

PR #824, opened 2015-08-10
Add listall=True param to listFirewallRules

DazWorrall:
When creating firewall rules for a different account than you have authenticated with - for example as an administrator setting up an account for a third party - the `listall` param is [required](https://cloudstack.apache.org/api/apidocs-4.5/domain_admin/listFirewallRules.html) to prevent the module trying to create the same rule again and getting the 'New rule conflicts with existing rule' error from Cloudstack

/cc @resmo


resmo:
Did you also pass `account` and `domain`?

DazWorrall:
Was that comment meant for #820?

resmo:
No :) you used cs_firewall with root admin permissions and also used account=someuser, domain=somedomain? this is really important, since as root admin I do not like to return all values, just the once you want to manage.

DazWorrall:
Sorry, didnt understand the question initially. No, this was calling without `account` or `domain`, and adding them does fix it. I understand your reasoning so lets just close this :)

resmo:
no problem! Thanks for your time.

PR #826, opened 2015-08-11
Add iptonetwork parameter to cs_instance

DazWorrall:
This supports the `iptonetworklist` parameter to [deployVirtualMachine](https://cloudstack.apache.org/api/apidocs-4.5/root_admin/deployVirtualMachine.html).

/cc @resmo

### DIFF
--- a/cloud/cloudstack/cs_instance.py
+++ b/cloud/cloudstack/cs_instance.py
@@ -473,7 +473,7 @@ class AnsibleCloudStackInstance(AnsibleCloudStack):
             self.module.fail_json(msg="networks and ip_to_networks are mutually exclusive.")
 
         network_names = [n['network'] for n in network_mappings]
-        ids = self.get_network_ids(network_names).split(',')
+        ids = self.get_network_ids(network_names)
         res = []
         for i, data in enumerate(network_mappings):
             res.append({'networkid': ids[i], 'ip': data['ip']})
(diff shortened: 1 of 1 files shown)


cloud/cloudstack/cs_instance.py
```diff
@@ -97,6 +97,12 @@
       - IPv6 address for default instance's network.
     required: false
     default: null
+  iptonetwork:
```
resmo: would you mind to change the name to `ip_to_networks` (plural)? And add an alias `ip_to_network`

```
...
 ip_to_networks = dict(type='list', aliases=['ip_to_network'], default=None),
...
```

cloud/cloudstack/cs_instance.py
```diff
@@ -448,9 +464,25 @@ def get_instance(self):
...
+    def get_iptonetwork_mappings(self):
+        network_mappings = self.module.params.get('iptonetwork')
+        if network_mappings is None:
+            return
+
+        if network_mappings and self.module.params.get('networks'):
+            self.module.fail_json(msg="networks and iptonetwork are mutually exclusive.")
+
+        network_names = [n['network'] for n in network_mappings]
+        ids = self.get_network_ids(network_names).split(',')
```
resmo: we should change `get_network_ids()` to return a list and make the `','.join()` where we use `self.get_network_ids()`. This way we do not need to split here again.

DazWorrall:
Changes made @resmo

resmo:
@gregdek LGTM shipit :+1:

PR #831, opened 2015-08-13
Enable managing multiple user permissions from one shot

mrusu91:
One of inconvinence this address is the the fact that
you have to pass user's tags even if you want to just add a permission

Signed-off-by: Marian Rusu rusumarian91@gmail.com

### DIFF
--- a/messaging/rabbitmq_user.py
+++ b/messaging/rabbitmq_user.py
@@ -45,9 +45,19 @@ options:
       - User tags specified as comma delimited
     required: false
     default: null
+  permissions:
+    description:
+      - a list of dicts, each dict contains vhost, configure_priv, write_priv, and read_priv,
+        and represents a permission rule for that vhost.
+      - This option should be preferable when you care about all permissions of the user.
+      - You should use vhost, configure_priv, write_priv, and read_priv options instead
+        if you care about permissions for just some vhosts.
+    required: false
+    default: []
   vhost:
     description:
       - vhost to apply access privileges.
+      - This option will be ignored when permissions option is used.
     required: false
     default: /
   node:
@@ -61,6 +71,7 @@ options:
       - Regular expression to restrict configure actions on a resource
         for the specified vhost.
       - By default all actions are restricted.
+      - This option will be ignored when permissions option is used.
     required: false
     default: ^$
   write_priv:
@@ -68,6 +79,7 @@ options:
       - Regular expression to restrict configure actions on a resource
         for the specified vhost.
       - By default all actions are restricted.
+      - This option will be ignored when permissions option is used.
     required: false
     default: ^$
   read_priv:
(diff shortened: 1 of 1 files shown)


bcoca:
cc @chrishoffman

mrusu91:
The fact that I have to pass the user's tags when I add permissions made me write this commit. I can add backwards compatibility, but if this doesn't make sense to you I will cancel it.

Can you confirm that?

chrishoffman:
This thing I like about bulk permissions is that it sets the final state and adds and removes permissions as necessary.  I think this is a good idea but would like to see backwards compatibility.

mrusu91:
@chrishoffman happy to hear that, I've updated the commit to contain backwards compatibility

gregdek:
@chrishoffman Please review this PR to make sure it adheres to the following guidelines:

gregdek:
Ping @chrishoffman -- have you had a chance to take a look at this PR? Thanks!

gregdek:
@chrishoffman can you please review? If you're unavailable to review PRs for this module, let us know and we'll be happy to look for additional maintainers to help.

robynbergeron:
@chrishoffman This change is still pending your review; do you have time to take a look and comment? Please comment with text 'shipit' or 'needs_revision' as appropriate.

chrishoffman:
I still see use cases for the single method and don't think it should be listed as deprecated.  If you have a perfectly clean and new setup, the bulk method is probably the way to go.  But if you have an existing server that you need to add or remove permissions from, then the single method would be preferable. I would just make both methods available and documented on when to use which.

needs_revision

gregdek:
Thanks @viruxel for this PR. A maintainer of this module has asked for revisions to this PR. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

mrusu91:
ready_for_review

chrishoffman:
shipit

gregdek:
Thanks again to @viruxel for this PR, and thanks @chrishoffman for reviewing. Marking for inclusion.

messaging/rabbitmq_user.py
```diff
@@ -175,14 +198,21 @@ def set_tags(self):
...
     def set_permissions(self):
-        cmd = ['set_permissions']
-        cmd.append('-p')
-        cmd.append(self.permissions['vhost'])
-        cmd.append(self.username)
-        cmd.append(self.permissions['configure_priv'])
-        cmd.append(self.permissions['write_priv'])
-        cmd.append(self.permissions['read_priv'])
-        self._exec(cmd)
+        for permission in self._permissions:
```
chengrcli, line 201: When adding a new user, self._permissions is None.
Will meet error: 'NoneType is not iterable'
I suggest initialize self._permissions = [] instead of None

mrusu91, line 201: pushed a commit here:
https://github.com/ansible/ansible-modules-extras/pull/2049

PR #837, opened 2015-08-16
Add consul http checks to consul.py

jemmyw:
Consul module already supports ttl and script checks. This commit adds
http checks.

### DIFF
--- a/clustering/consul.py
+++ b/clustering/consul.py
@@ -123,6 +123,20 @@ options:
             is supplied, m will be used by default e.g. 1 will be 1m
         required: false
         default: None
+    http:
+        description:
+          - checks can be registered with an http endpoint. This means that consul
+            will check that the http endpoint returns a successful http status.
+            Interval must also be provided with this option.
+        required: false
+        default: None
+    timeout:
+        description:
+          - A custom HTTP check timeout. The consul default is 10 seconds.
+            Similar to the interval this is a number with a s or m suffix to
+            signify the units of seconds or minutes, e.g. 15s or 1m.
+        required: false
+        default: None
     token:
         description:
           - the token key indentifying an ACL rule set. May be required to register services.
@@ -142,6 +156,13 @@ EXAMPLES = '''
       service_port: 80
       script: "curl http://localhost"
       interval: 60s
+
+  - name: register nginx with an http check
+    consul:
+      name: nginx
+      service_port: 80
+      interval: 60s
+      http: /status
 
   - name: register nginx with some service tags
     consul:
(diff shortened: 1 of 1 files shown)


bcoca:
cc @sgargan

sgargan:
@jemmyw change looks good, travis is reporting a failure but its not related to your checkin. Looking at the documentation for these http checks, they apparently take a timeout too so it can break out from a service that is non responsive. The default is apparently 10 seconds. I'm not sure if the python-consul api for http check exposed this, but if it does it might be an idea to allow this timeout to be configured too.

There are a bunch of integration tests in the core project that verify all the consul behavior, take a look at these, would be good if you could add tests for your changes there too.

jemmyw:
@sgargan OK, I've added the timeout option and fixed a bug with the script thanks to the integration tests. I'm not sure how to make a referencing PR for the integration tests, but they are here: jemmyw/ansible@9f9f476b6e97da926d6482b8fd737346733667ae

gregdek:
Adding to core_review to advise on integration tests.

evanccnyc:
Be nice to have the ability to specify an additional url to the http check. So you could do http://blah/status etc.

abadger:
Code looks good to me too.  @evanccnyc, did you test that this works for you?  If you could +1 from that standpoint that would be good to know.

@jemmyw For the integration tests, just open a PR against the ansible/ansible repository and post the link here.  Then we can merge both this one and that one at the same time.

evanccnyc:
@jemmyw any chance you can add a way to specify the status url: ie it could check /status or /lb etc?

robynbergeron:
Hi @jemmyw -- 

Hopefully this gives you enough info re: integration tests to go forward. If you want to open a PR for the integration test (as mentioned by @abadger above) and reference it here that would be perfect -- then we can move forward! Thanks.

jemmyw:
@robynbergeron PR for the integration tests https://github.com/ansible/ansible/pull/12589

jemmyw:
@evanccnyc the status url can be specified by setting the http param. i.e. interval=10s http=/status

evanccnyc:
@jemmyw I see, any chance you can provide an example in the example of that or clarify the documentation?

robynbergeron:
@jemmyw -- can you possibly add the the example as requested by @evanccnyc? It looks like the integration tests have already been merged. Let us know, thanks!

jemmyw:
@robynbergeron @evanccnyc example added, and yes the integration tests have been merged.

robynbergeron:
@jemmyw -- thanks for the revision to include examples.  

@evanccnyc -- does this look good to you as far as examples go? Also double-checking again with @sgargan as the module maintainer.

evanccnyc:
Looks good to me shipit.

PR #838, opened 2015-08-16
Adding support for service ACLs in consul_acl module

chrishoffman:
This pull request adds support to the consul_acl module for ACLs on services.

Consul 0.5.0 added support for service discovery ACLs.  The current version of the module only supports ACLs on the kv part of Consul.

### DIFF
--- a/clustering/consul_acl.py
+++ b/clustering/consul_acl.py
@@ -174,7 +174,7 @@ def update_acl(module):
         module.fail_json(msg="Could not create/update acl %s" % e)
 
     module.exit_json(changed=changed,
-                     token=obfuscate_token(token),
+                     token=token,
                      rules=rules,
                      name=name,
                      type=token_type)
@@ -190,10 +190,7 @@ def remove_acl(module):
     if changed:
         token = consul.acl.destroy(token)
 
-    module.exit_json(changed=changed, token=obfuscate_token(token))
-
-def obfuscate_token(token):
-    return token[:4] + "*" * (len(token) - 5)
+    module.exit_json(changed=changed, token=token)
 
 def load_rules_for_token(module, consul_api, token):
     try:
--- a/clustering/consul_acl.py
+++ b/clustering/consul_acl.py
@@ -22,7 +22,7 @@ module: consul_acl
 short_description: "manipulate consul acl keys and rules"
 description:
  - allows the addition, modification and deletion of ACL keys and associated
-   rules in a consul cluster via the agent. For more details on using and 
+   rules in a consul cluster via the agent. For more details on using and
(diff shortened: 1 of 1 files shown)


bcoca:
cc @sgargan

sgargan:
Service changes look good thanks for adding them. I have a minor concern about the token obfuscation, the reasoning behind returning it is so that it could be registered in a variable and subsequently used in further operations e.g. stored via a template onto client machines that will use the token, obfuscating it will break this. What is your concern here? If you're worried about it being passed back you could add an optional flag to omit or obfuscate it. There are probably other ways it can be excluded from the logs too.

chrishoffman:
You are absolutely correct about registering variables.  My concern was the -v output but it completely makes sense and was a use case I wasn't considering.  I personally generate tokens and set them explicitly for new acl policies so I was more concerned about tokens leaking into the logs.  I will remove the token obfuscation.

bcoca:
add no_log to the token spec if it should be kept 'confidential'

gregdek:
+1 to no_log being needed.

@chrishoffman Please make the suggested revisions, and when you're done -- or if we're incorrect and no revision to no_log needs to be made -- please add a comment with the text "ready_for_review" and we will continue with the review process.  Thanks!

chrishoffman:
The revisions were made to no longer obfuscate the token.  I have updated the pull request text accordingly.

ready_for_review

gregdek:
Thanks @chrishoffman.

@sgargan can you give a quick re-review? If you think it's good, please add a comment with "shipit" in the text, and we will flag it for inclusion. If not, please add a comment with "needs_revision" in the text, along with the needed changes, and we will flag it for revision.

sgargan:
Obfuscation is gone, looks good. Ship it (or shipit as the case may be).

robynbergeron:
Thanks @chrishoffman for the PR, and to @sgargan for the reviews. 

Marking for inclusion, cc: @ansible/core team.

abadger:
If there is concern about logging the token parameters we need to update the argument spec to mark those as no_log.  It won't affect returning a value wth token in it but it will affect logging of the parameter that is given to the module.  So Line 322 and possibly Line 316 need to be updated to have no_log.  Since token and mgmt_token are already parameters of this module, I'll go ahead and merge this now but I'd appreciate either a pull request or information about whether both Line 322 and Line 316 should be marked no_log.

If you're taking the PR route, this is what the argument spec would look like:

``` patch
- token=dict(required=False),
+ token=dict(required=False, no_log=True),
```

abadger:
# Merged

Hi!

issue is resolved.

Thank you!

PR #843, opened 2015-08-17
seport requires 2.6

jeffb-stell:
The system/seport.py module requires Python 2.6, it is not Python 2.4 safe.


jeffb-stell:
Unneeded, @bcoca made it 2.4 compatible

Issue #813, opened 2015-08-07
maven_artifact not working


carcarlo:
Hello
I've installed ansible 2.0 from GitHub building this RPM ansible-2.0.0-0.git201508070432.48840ad.devel.noarch.rpm.

When I try to download an artifact whit the command:
ansible 1.1.2.195 -m maven_artifact -a "group_id=org.apache.commons artifact_id=commons-collections dest=/tmp/commons-collections-latest.jar"  -u root  -k

I get the following error:

atal: [1.1.2.195]: FAILED! => {"changed": false, "failed": true, "msg": "Traceback (most recent call last):\r\n  File \"/root/.ansible/tmp/ansible-tmp-1438940214.19-9647185694305/maven_artifact\", line 2825, in <module>\r\n    main()\r\n  File \"/root/.ansible/tmp/ansible-tmp-1438940214.19-9647185694305/maven_artifact\", line 313, in main\r\n    downloader = MavenDownloader(module, repository_url, repository_username, repository_password)\r\nTypeError: **init**() takes at most 3 arguments (5 given)\r\n", "parsed": false}

bcoca:
cc @chrisisbeef

sirkubax:
ansible -i ~/ansible/etc/inv/inventory-qa some-host -m maven_artifact -a "group_id=org.jmxtrans artifact_id=jmxtrans version=251 repository_url=http://central.maven.org/maven2 dest=/tmp" -vvv

msg": "Traceback (most recent call last):\r\n  File \"/home/ubuntu/.ansible/tmp/ansible-tmp-1439822083.39-247708723227590/maven_artifact\", line 2645, in <module>\r\n    main()\r\n  File \"/home/ubuntu/.ansible/tmp/ansible-tmp-1439822083.39-247708723227590/maven_artifact\", line 309, in main\r\n    downloader = MavenDownloader(module, repository_url, repository_username, repository_password)\r\nTypeError: **init**() takes at most 3 arguments (5 given)\r\nOpenSSH_6.7p1 Ubuntu-5ubuntu1.2, OpenSSL 1.0.1f 6 Jan 2014\r\ndebug1: Reading configuration data

sirkubax:
quick print in line 309:
print '%s, %s, %s' % (repository_url, repository_username, repository_password)
gives output:
http://central.maven.org/maven2, None, None

while the provided class requires:
class MavenDownloader:
[...]

PR #845, opened 2015-08-17
Fix for maven_artifact failing on function signature

sirkubax:
QuickFix of 
https://github.com/ansible/ansible-modules-extras/issues/813

Note:
repository_username, repository_password - currently not in use
extension - set to "required=True" since is expected in the code. This could be written nicer: 
    extension = dict(default='', required=True) 
or some other way of securing code

### DIFF
--- a/packaging/language/maven_artifact.py
+++ b/packaging/language/maven_artifact.py
@@ -285,7 +285,7 @@ def main():
             artifact_id = dict(default=None),
             version = dict(default=None),
             classifier = dict(default=None),
-            extension = dict(default=None),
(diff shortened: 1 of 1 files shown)


chrisisbeef:
This looks good - the parameters in question were left over from the refactor that removed username and password digest generation (along with the https stuff) when it was switch to use the .request function to initiate the download instead of trying to generate the URL itself with all the right stuff. 

+1 for Merge

chrisisbeef:
@sirkubax - Great job! Thanks for the patch!

PR #849, opened 2015-08-19
Add probe functionality to modprobe

dlundquist:
Often times you want to use modprobe to load a module if a hardware
device is present. Adding `state=probe` that will attempt to load a
module and report changed if the module was loaded, no change if the
hardware device was not present and error if another error occurred.
This is the same behavior as 'present' except no error is raised or
change is reported if the driver reports no such device (ENODEV).

Previously submitted to ansible: ansible/ansible#6959

### DIFF
--- a/system/modprobe.py
+++ b/system/modprobe.py
@@ -26,7 +26,7 @@ short_description: Add or remove kernel modules
 requirements: []
 version_added: 1.4
 author:
-    - "David Stygstra (@stygstra)" 
+    - "David Stygstra (@stygstra)"
     - "Julien Dauphant"
     - "Matt Jeffery"
 description:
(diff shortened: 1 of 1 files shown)


bcoca:
cc @stygstra

system/modprobe.py
```diff
@@ -104,6 +107,13 @@ def main():
             if rc != 0:
                 module.fail_json(msg=err, **args)
             args['changed'] = True
+    elif args['state'] == 'probe':
+        if not present:
+            rc, _, err = module.run_command([module.get_bin_path('modprobe', True), args['name'], args['params']])
+            if rc == 0:
```
bcoca, line 113: Instead of a new probe state, this seems like it would be better as a probe option (no/yes, no by default) that would work the same as the current present, which actually runs the same command.

gregdek:
@dlundquist After some review, it looks as though this PR needs some revisions. 

robynbergeron:
@dlundquist -- A friendly reminder: this pull request has been marked as needing your action, and it hasn't been updated in the last month. 

gregdek:
Closed due to inactivity.

PR #851, opened 2015-08-20
[Amazon] Add EC2 ELB Facts module to gather facts about ELBs

mjschultz:
The ability to simply get facts about Elastic Load Balancers doesn't seem to be present in the existing [ec2_elb](http://docs.ansible.com/ansible/ec2_elb_module.html) or [ec2_elb_lb](http://docs.ansible.com/ansible/ec2_elb_lb_module.html) modules (unless I'm missing something).

We've found it useful, so perhaps others will as well.

### DIFF
--- a/cloud/amazon/ec2_elb_facts.py
+++ b/cloud/amazon/ec2_elb_facts.py
@@ -1,0 +1,198 @@
+#!/usr/bin/python
+#
+# This is a free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This Ansible library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this library.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: ec2_elb_facts
+short_description: Gather facts about EC2 Elastic Load Balancers in AWS
+description:
+    - Gather facts about EC2 Elastic Load Balancers in AWS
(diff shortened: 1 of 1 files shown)


bcoca:
cc @jarv @defionscode @rickmendes

cloud/amazon/ec2_elb_facts.py
```diff
@@ -0,0 +1,198 @@
...
+DOCUMENTATION = '''
+---
+module: ec2_elb_facts
+short_description: Gather facts about EC2 Elastic Load Balancers in AWS
+description:
+    - Gather facts about EC2 Elastic Load Balancers in AWS
+version_added: "2.0"
+author: "Michael Schultz (github.com/mjschultz)"
+options:
+  name:
```
ghost: If name represents a list, why not call it names?

mjschultz: It was a last minute decision to allow lists, renamed in https://github.com/ansible/ansible-modules-extras/commit/be59c0063b87dcb056fefaf2a1fc1c1f9086d608.

ghost:
@bcoca @mjschultz Looks good to me. +1

cloud/amazon/ec2_elb_facts.py
```diff
@@ -0,0 +1,198 @@
...
+    else:
+        module.fail_json(msg="region must be specified")
+
+    list_elb(connection, module)
+
+from ansible.module_utils.basic import *
+from ansible.module_utils.ec2 import *
+
+# this is magic, see lib/ansible/module_common.py
+#<<INCLUDE_ANSIBLE_MODULE_COMMON>>
```
bcoca: this is not needed as it is the same as:
`from ansible.module_utils.basic import *`

mjschultz: Ah, I thought that was the case now-a-days I just picked a module that hasn't been updated since the change I suppose. Fixed in https://github.com/mjschultz/ansible-modules-extras/commit/913266c04f62b775f30d82ffb8c163d0b77c24fa. (ec2_eni_facts are the modules I used as the examples, by the way.)

PR #855, opened 2015-08-21
Openvswitch db

markleehamilton:
Module for changing values in openvswitch tables.


network/openvswitch_db.py
```diff
@@ -0,0 +1,129 @@
...
+
+    changed = False
+
+    fmt = "ovs-vsctl -t %(timeout)s get %(table)s %(record)s %(col)s:%(key)s"
+
+    cmd = fmt % module.params
+
+    (_, output, _) = cmd_run(module, cmd, False)
+    if module.params['value'] not in output:
+        fmt = "ovs-vsctl -t %(timeout)s set %(table)s %(record)s " \
```
bcoca: use modules find_bin function instead of assuming this is in PATH

markleehamilton:
bcoca, I updated the module to use get_bin_path.

network/openvswitch_db.py
```diff
@@ -0,0 +1,135 @@
...
+        description:
+            - Expected value for the table, record, column and key.
+    timeout:
+        required: false
+        default: 5
+        description:
+            - How long to wait for ovs-vswitchd to respond
+"""
+
+DOCUMENTATION = __doc__
```
bcoca: not sure why a 2nd 'DOCUMENTATION is here, but this will break the docs parser

network/openvswitch_db.py
```diff
@@ -0,0 +1,135 @@
...
+
+# Disable in band copy
+- openvswitch_db: table=Bridge record=br-int col=other_config
+                  key=disable-in-band value=true
+'''
+
+
+def cmd_run(module, cmd, check_rc=True):
+    """ Log and run ovs-vsctl command. """
+    syslog.syslog(syslog.LOG_NOTICE, cmd)
```
bcoca: you should not invoke syslog, the module already does this with pertinent information

network/openvswitch_db.py
```diff
@@ -0,0 +1,135 @@
...
+    if module.params['value'] not in output:
+        fmt = "%(ovs-vsctl)s -t %(timeout)s set %(table)s %(record)s " \
+              "%(col)s:%(key)s=%(value)s"
+        cmd = fmt % module.params
+        ##
+        # Check if flow exists and is the same.
+        (rtc, _, err) = cmd_run(module, cmd)
+        if rtc != 0:
+            module.fail_json(msg=err)
+        changed = True
```
bcoca, line 104: this seems to always return 'changed' unless there is an error, is there no way to check if the values already exist as expected?

markleehamilton, line 104: Line 98 does the check to see if the value is already set. The module can return not changed.

markleehamilton:
I removed the extra syslog in cmd_run method. I find the way I posted to syslog easier to read but I get your concern. I also removed the second DOCUMENTATION variable.

markleehamilton:
bcoca,
 I addressed all of your concerns.

PR #856, opened 2015-08-21
Add ZFS backing store support to lxc_container

daniel5gh:
Hi!

Added the `zfs` option to the list of allowed backing stores. This allowed me the create an lxc container with its backing store on a ZFS dataset.

Thanks for the great work!


daniel5gh:
Thanks for merging.

I forgot to add the ZFS option to the backing store choices in the documentation string.

PR #860, opened 2015-08-24
Fixing empty tags check on rabbitmq_user

Draiken:
Right now even if you pass in an empty `tags` list to the module (either with  an empty string or null) it will erroneously think the tags list have changed and re-apply the tags on every run.

This happens because the default value of `None` is not used, and it tries to split the string by commas generating `['']` as a tags list, instead of an empty list.

### DIFF
--- a/messaging/rabbitmq_user.py
+++ b/messaging/rabbitmq_user.py
@@ -108,7 +108,7 @@ class RabbitMqUser(object):
         self.username = username
         self.password = password
         self.node = node
-        if tags is None:
+        if not tags:
             self.tags = list()
         else:
             self.tags = tags.split(',')


messaging/rabbitmq_user.py
```diff
@@ -108,7 +108,7 @@ def __init__(self, module, username, password, tags, vhost, configure_priv, writ
         self.username = username
         self.password = password
         self.node = node
-        if tags is None:
+        if tags is None or tags == '':
```
bcoca: why not the simpler?:
`if not tags:`

Draiken: I'm not a python programmer, so I just made it work locally with this or :)

Draiken:
@bcoca done!

Sorry I have no idea what's truthy or falsey on python

resmo:
LGTM

PR #883, opened 2015-08-26
Add new module: pagerduty_alert

ApsOps:
- trigger, acknowledge or resolve pagerduty incidents

### DIFF
--- a/monitoring/pagerduty_alert.py
+++ b/monitoring/pagerduty_alert.py
@@ -1,0 +1,157 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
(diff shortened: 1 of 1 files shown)


bcoca:
I would move to monitoring/ dir as that is where we have modules that deal with monitoring services

monitoring/pagerduty_alert.py
```diff
@@ -0,0 +1,157 @@
...
+    - "Amanpreet Singh (@aps-sids)"
+requirements:
+    - PagerDuty API access
+options:
+    service_key:
+        description:
+            - The GUID of one of your "Generic API" services.
+            - This is the "service key" listed on a Generic API's service detail page.
+        required: true
+    event_type:
```
bcoca: A more imperative language here might be more inline with anisble modules:

state: triggered, acknowledged, resolved

also check if the incident is already in that state to report on changed correctly

aps-sids:
Thanks @bcoca . Did that.

monitoring/pagerduty_alert.py
ApsOps: Okay, makes sense.

ApsOps:
@bcoca Did the changes as per your suggestions. Please take a look.

PR #886, opened 2015-08-27
Add support for custom service offerings

DazWorrall:
This adds 3 new params: cpu, cpu_speed, and memory, which are required together /cc @resmo

### DIFF
--- a/cloud/cloudstack/cs_instance.py
+++ b/cloud/cloudstack/cs_instance.py
@@ -65,7 +65,7 @@ options:
     default: null
   memory:
     description:
-      - The clock speed/shares allocated to the instance, used with custom service offerings
+      - The memory allocated to the instance, used with custom service offerings
     required: false
     default: null
   template:
--- a/cloud/cloudstack/cs_instance.py
+++ b/cloud/cloudstack/cs_instance.py
@@ -53,7 +53,7 @@ options:
       - If not set, first found service offering is used.
     required: false
     default: null
-  cpu_count:
+  cpu:
     description:
       - The number of CPUs to allocate to the instance, used with custom service offerings
     required: false
--- a/cloud/cloudstack/cs_instance.py
+++ b/cloud/cloudstack/cs_instance.py
@@ -53,6 +53,21 @@ options:
       - If not set, first found service offering is used.
     required: false
     default: null
+  cpu_count:
+    description:
+      - The number of CPUs to allocate to the instance, used with custom service offerings
+    required: false
+    default: null
+  cpu_speed:
+    description:
+      - The clock speed/shares allocated to the instance, used with custom service offerings
+    required: false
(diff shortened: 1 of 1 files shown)


cloud/cloudstack/cs_instance.py
```diff
@@ -547,6 +562,20 @@ def get_user_data(self):
...
+        cpu = self.module.params.get('cpu')
+        cpu_speed = self.module.params.get('cpu_speed')
+        memory = self.module.params.get('memory')
+        if all([cpu, cpu_speed, memory]):
+            res = [{
+                'cpuNumber': cpu,
+                'cpuSpeed': cpu_speed,
+                'memory': memory,
+            }]
+        elif any([cpu, cpu_speed, memory]):
```
resmo: there is a ansible module feature, where you can specify which arguments must used together, see https://github.com/DazWorrall/ansible-modules-extras/blob/feature/cs_instance_custom/cloud/cloudstack/cs_instance.py#L858. 

```
required_together = (
     ['api_key', 'api_secret', 'api_url'],
     ['cpu', 'cpu_speed', 'memory'],
),
```

resmo:
Hi @DazWorrall thanks for this PR. 

Question, would it make sense to say: you can either pass `service_offering` or ('cpu','cpu_speed','memory') right? In this case I would also add a

```
mutually_exclusive = (
   ['cpu', 'service_offering'],
),
```

DazWorrall:
You still need to specify a service offering that supports custom resource settings I believe - the default offering may not always do that, so I think in most cases you would need to specify both.

resmo:
right, I see the offering has to be flagged as custom, so you can pass the custom settings. Makes sense.

LGTM. shipit @gregdek

cloud/cloudstack/cs_instance.py
```diff
@@ -53,6 +53,21 @@
       - If not set, first found service offering is used.
     required: false
     default: null
+  cpu_count:
```
resmo: this should be `cpu` as you used in code

DazWorrall:
Thanks for your time @resmo :)

cloud/cloudstack/cs_instance.py
```diff
@@ -53,6 +53,21 @@
...
+    required: false
+    default: null
+  cpu_speed:
+    description:
+      - The clock speed/shares allocated to the instance, used with custom service offerings
+    required: false
+    default: null
+  memory:
+    description:
+      - The clock speed/shares allocated to the instance, used with custom service offerings
```
resmo: this seems also not quite right :)

DazWorrall: Yeah that's just embarassing... whoops!

gregdek:
cc: @ansible/core for merging

PR #890, opened 2015-08-28
Add some more locales to LOCALE_NORMALIZATION

kmalov:
I found some Russian locales missed in system/locale_gen.py dict LOCALE_NORMALIZATION. I added these locales to LOCALE_NORMALIZATION and some more.

### DIFF
--- a/system/locale_gen.py
+++ b/system/locale_gen.py
@@ -52,6 +52,13 @@ LOCALE_NORMALIZATION = {
     ".utf8": ".UTF-8",
     ".eucjp": ".EUC-JP",
     ".iso885915": ".ISO-8859-15",
+    ".cp1251": ".CP1251",
+    ".koi8r": ".KOI8-R",
+    ".armscii8": ".ARMSCII-8",
+    ".euckr": ".EUC-KR",
+    ".gbk": ".GBK",
+    ".gb18030": ".GB18030",
+    ".euctw": ".EUC-TW",
 }
 
 # ===========================================


gregdek:
@AugustusKling Please review this PR to make sure it adheres to the following guidelines:

AugustusKling:
Looks good to me, so "shipit".

I'm wondering though if the already existing mapping is a good idea at all. Perhaps the problem of the differenct representations could be solved by lowercasing and stripping `-`, too. That would get rid of an ever expanding mapping. I'm just not sure if we can generalize all variants like that.
Such an improvement does not need to hinder the proposed commit, tough.

gregdek:
Thanks @AugustusKling for the review and @kmalov for the patch. Marking for inclusion.

kmalov:
I agree mapping is bad idea but it works. I thought about lowering and
stripping and compared it with number of locales need to be added to
LOCALE_NORMALIZATION. The number of "broken" locales seemed to me quite
small so adding all of them to LOCALE_NORMALIZATION looks better for me
then rewriting fix_case().

2015-08-29 13:26 GMT+03:00 Augustus Kling notifications@github.com:

> Looks good to me, so "shipit".
> 
> I'm wondering though if the already existing mapping is a good idea at
> all. Perhaps the problem of the differenct representations could be solved
> by lowercasing and stripping -, too. That would get rid of an ever
> expanding mapping. I'm just not sure if we can generalize all variants like
> that.
> Such an improvement does not need to hinder the proposed commit, tough.
> 
> —
> https://github.com/ansible/ansible-modules-extras/pull/890#issuecomment-135968988
> .

## 

Малов Константин
e-mail: constantine.malov@gmail.com
моб.: +7 919 966 222 0

gregdek:
cc: @ansible/core for merge

PR #891, opened 2015-08-28
fix creation containers on btrfs subvolumes

sapun:


gregdek:
@cloudnull Please review this PR to make sure it adheres to the following guidelines:

cloud/lxc/lxc_container.py
```diff
@@ -448,7 +448,7 @@
         'zfs_root'
     ],
     'btrfs': [
-        'lv_name', 'vg_name', 'thinpool', 'zfs_root'
+        'lv_name', 'vg_name', 'thinpool', 'zfs_root', 'fs_type', 'fs_size'
```
cloudnull, line 451: Quick question here but does it really need to NOT have storage type and size when using a btrfs container?  I've not seen that though admittedly I've not tested btrfs in some time. If so, the change lgtm but I'd like to know why this change is required first.

sapun, line 451: Because without this fix i catch error like this
"err": "filesystem type and size are only valid with block devices\n"

cloudnull, line 451: fair enough.

cloudnull:
@gregdek  shipit - LGTM!

gregdek:
Thanks @sapun for the patch, and @cloudnull for the review. Marked for inclusion.

PR #905, opened 2015-09-01
Add new option "noprompt" for win_chocolatey

mrknmc:
The `noprompt` option allows you to install chocolatey packages with the `-y` flag.

I am not convinced `noprompt` is the best name for this flag but chocolatey documentations refers to `-y` as:

> Add -y for previous behavior with no prompt.

This flag is useful when installing packages that have dependencies, for example the rabbitmq package has a depedency on the erlang package. With `-y` you automatically confirm the prompt asking you to install the erlang package.

Following is an example prompt when trying to uninstall the erlang package:

```
C:\Users\vagrant>C:\ProgramData\chocolatey\choco.exe uninstall erlang
Chocolatey v0.9.9.8
Uninstalling the following packages:
erlang

erlang v18.0.20150709.2
The package erlang wants to run 'chocolateyUninstall.ps1'.
Note: If you don't run this script, the installation will fail.
Note: To confirm automatically next time, use '-y' or consider setting
 'allowGlobalConfirmation'. Run 'choco feature -h' for more details.
Do you want to run the script?
 1) yes
 2) no
 3) print
```


mrknmc:
Following doc refers to the option as

```
-y, --yes, --confirm
  Confirm all prompts - Chooses default answer instead of prompting.
  Implies --accept-license
```

Might rename `noprompt` to `confirm` with `yes` as alias.

https://github.com/chocolatey/choco/wiki/CommandsReference#how-to-pass-options--switches

bcoca:
this should not really be an option, the module should always avoid prompts, I would consider it an 'option' if when no it would run with '--no' which would work as the opposite of --yes and still avoid prompts (I don't know if chocolatey supports this).

mrknmc:
If I understand what you're saying then we should change the module to automatically include the -y flag, correct?

bcoca:
yep, modules should avoid interactive prompts by default, one more question, is -y available in all versions of chocolatey?

mrknmc:
Just tried `-y` with chocolatey v0.9.8.33 which doesn't require `-y` and everything worked OK:

```
PS C:\Windows\system32> choco install -y rabbitmq
Chocolatey (v0.9.8.33) is installing 'rabbitmq' and dependencies. By installing you accept the license for 'rabbitmq
d each dependency you are installing.

erlang v15.3.1
Downloading erlang 64 bit
  from 'http://www.erlang.org/download/otp_win64_R15B03-1.exe'
Installing erlang...
erlang has been installed.
erlang has finished successfully! The chocolatey gods have answered your request!
Erlang has finished successfully! The chocolatey gods have answered your request!

rabbitmq v3.5.3
Downloading rabbitmq 32 bit
  from 'http://www.rabbitmq.com/releases/rabbitmq-server/v3.5.3/rabbitmq-server-3.5.3.exe'
Installing rabbitmq...
rabbitmq has been installed.
rabbitmq has finished successfully! The chocolatey gods have answered your request!

RabbitMQ Management Plugin enabled by default at http://localhost:15672

Finished installing 'rabbitmq' and dependencies - if errors not shown in console, none detected. Check log for error
 unsure.
```

bcoca:
wouldn't people using older versions still benefit from this feature?

mrknmc:
Hey Brian, so I was going to implement -y by default in a new pull request but then realised it's already done, see https://github.com/ansible/ansible-modules-extras/blob/devel/windows/win_chocolatey.ps1#L231. Dunno how I missed that the first time round! :see_no_evil:

PR #907, opened 2015-09-01
Retention

greg-hellings:
Adds a sub-module to the File section that allows you to define a retention/removal policy for files or folders under a given root.

### DIFF
--- a/files/retention.py
+++ b/files/retention.py
@@ -1,6 +1,60 @@
 #!/usr/bin/env python
 
+DOCUMENTATION='''
+module: retention
+author: "Greg Hellings (@greghellings)"
+version_added: 2.0
+description:
+    - Define a retention policy for files or directories matching a particular shell pattern. This module will only
+      indicate that something has changed if it deletes files or folders.
+short_description: Remove files/folders outside of retention policy
+options:
+  src: 
+    description:
+      - The folder where files/folders live that should be culled. If this folder does not exist, then this module
+        will fail to execute.
+    required: true
+    aliases: [ "directory" ]
+  count:
+    description:
+      - The number of files matching the pattern to retain. If you set this to 0, then all files matching the
+        pattern will be deleted.
+    required: true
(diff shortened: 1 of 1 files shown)


bcoca:
doesn't this overlap with find + file modules?

greg-hellings:
You can replicate the results with find + file, but I wouldn't say it overlaps with them. It removes the need to customize sorting and array slicing in the playbooks and gives single-line access to the behavior.

bcoca:
it seems better suited to a role (I need to publish my 'tidy' role), modules that just aggregate what other modules can already do are not really that useful.

greg-hellings:
Why is there a template module, which treads on the ability to do copy + lineinfile? Then why are there any modules outside of shell? The decomposition could become drastic if having a module wrap other functionality is considered a bad idea. :-/

bcoca:
He, I would be happy to delete lineinfile (i always recommend template use above this module). But those are not the same comparisons. 

Shell is there as a fallback for when a module does not exist that can handle an issue, it is not recommended for general use.

If we follow your logic to the other extreme to have a module for every time you can avoid using 2 modules or a module + filter would require us to have millions of modules. Lets keep the discussion from absurd extremes, it is not productive.

gregdek:
Apologies @greg-hellings, but a strong -1 from a core contributor is sufficient grounds to close this PR.

greg-hellings:
I'm still befuddled by @bcoca's opposition. This is not a trivial thing to accomplish with a role. The steps that must be gone through to replicate this module are:

Find appropriate entries within a directory
Store the results
Sort them appropriately
Slice the array
Delete the remainder of the slice

For something as fundamental as retention, this seems worth inserting into a module to me. Is it possible to achieve without a crafted module? Yes. But then again, anything is attainable with only a few basic modules.

Asking people to craft roles to handle logic this complicated seems unnecessary. As someone once told me: Ansible is not a programming language.

PR #926, opened 2015-09-07
Fix hang at creation of detached ENIs

bretmartin:
`ec2_eni` currently waits forever for an ENI's `attachment.status` to be `attached` even if the ENI was specified as being created without being attached to an instance. This PR corrects the hang by only waiting for that status if `instance_id` is set.

@wimnat @evanccnyc @rickmendes


wimnat:
+1

evanccnyc:
Looks good to me. +1

Issue #932, opened 2015-09-07
pkg5 failed if package is installed


Steffeng5:
-Solaris 11

When I try to install for example SUNWfont-xorg-core via pkg5 then the first time it is working (because it is not installed)

If it is already installed the command pkg list -- does not know this package name, therefore it tries to install it again. The install command then returns RC 4 and "No updates necessary for this image.", but this is missinterpreted as failed.

I'm not so formular with this strange package managers in solaris, but if there should be another check like "pkginfo SUNWfont-xorg-core" then everythings should works fine.

robynbergeron:
[module: packaging/os/pkg5.py]

@Steffeng5 Thanks for submitting this bug report.

@mavit -- thanks in advance for taking a look at this bug report and for your ongoing work in maintaining this module.

It looks like you were both conversing in #933 after it got closed -- though not sure if there was a resolution from the conversation. Just giving a friendly reminder -- if this can be closed, let us know! :)

mavit:
The outcome of the discussion in #933 was that we should treat rc == 4 as not chaged (having made sure that this does the right thing when state=installed rather than state=latest).  This is still outstanding.

ansibot:
@mavit, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@mavit, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@mavit, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@mavit, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@mavit, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@mavit, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
[...]

PR #933, opened 2015-09-07
pkg5 failed if package is installed #932

Steffeng5:
Working solution for issue #932

### DIFF
--- a/packaging/os/pkg5.py
+++ b/packaging/os/pkg5.py
@@ -156,6 +156,8 @@ def ensure(module, state, packages, params):
 
 def is_installed(module, package):
     rc, out, err = module.run_command(['pkg', 'list', '--', package])
+	 if rc != 0:
+      rc, out, err = module.run_command(['pkginfo', package])
     return not bool(int(rc))
 
 
--- a/packaging/os/pkg5.py
+++ b/packaging/os/pkg5.py
@@ -156,8 +156,8 @@ def ensure(module, state, packages, params):
 
 def is_installed(module, package):
     rc, out, err = module.run_command(['pkg', 'list', '--', package])
-	 if rc != 0:
-      rc, out, err = module.run_command(['pkginfo', package])
+    if rc != 0:
+        rc, out, err = module.run_command(['pkginfo', package])
     return not bool(int(rc))
 
 


bcoca:
why not just remove the previous failing check and use pkginfo for all?

bcoca:
cc @mavit

Steffeng5:
Is pkginfo working for all types of packages or is it as buggy as pkg list?

bcoca:
good question, as someone that has avoided Solaris for the past 10 yrs, I'm not the best to answer this

Steffeng5:
Okay. I've tried sth. to get sure the code is with the double check: 

root@solaris:~# pkginfo motif
ERROR: information for "motif" was not found
root@solaris:~# pkg list -- motif
NAME (PUBLISHER)                                  VERSION                    IFO
library/motif                                     0.5.11-0.175.0.0.0.0.0     i--

Therefore we should retain both checks :sweat:

bcoca:
So this made no sense to me and I looked at it deeper.

It seems we are mixing package managers here, pkginfo is part of SVR4 packaging and pkg is IPS packaging (which has it's own `pkg info` command).

The reason that it fails with one and succeeds with the other is because they are different package management systems with different package tracking, so a package appears in the one you installed it with. As such this PR is incorrect, pkg5 should not be tracking v4 packaging and viceversa.

Steffeng5:
hm okay, mixing the packagemanagers is not so good. 

Another idea: 

pkg install returns rc = 4 if the package is installed but no update is required. 

root@solaris:~# pkg install SUNWfont-xorg-core
No updates necessary for this image.
root@solaris:~# echo $?
4

Maybe we should change the code to accept also rc 4 as "ok"?

bcoca:
since we already check before trying to install, it is not needed (unless someone installs in the milliseconds between check and attempt to install.

Steffeng5:
As you can see above or in my bugreport the check is not working for all packages starting with "SUNW" ;)

And interpret rc 4 No updates necessary as FAILED is in my opinion wrong.

bcoca:
is that package installed using the SRV4 system?

Steffeng5:
No, using pkg5 -> pkg install SUNWfont-xorg-core

bcoca:
so this looks like a pkg bug, but i would accept a special case for rc == 4 that sets a  custom message and changed=False

mavit:
The reason that `pkg list` doesn't show `SUNWfont-xorg-core` is because it technically isn't installed.  Because it has its `renamed` attribute set to `true`, `pkg install` will chase down and install its dependencies but skip the package itself.  See the manifest at http://pkg.oracle.com/solaris/release/manifest/0/SUNWfont-xorg-core%400.5.11%2C5.11-0.133%3A20101027T183825Z

So, I don't think running `pkginfo` to check for installed packages would hurt, but it wouldn't help in the case of package renames unrelated to SVR4 packages.

Agreed, treating rc 4 as not changed seems like the way to go.  @Steffeng5: will you prepare a pull request?  I think we need to do some testing to make sure this does the right thing when `state=installed` rather than `state=latest`, though.

That said, my advice is that if one wants the functionality provided by `SUNWfont-xorg-core`, it would be probably be best to run `pkg info -r SUNWfont-xorg-core`, observe that that functionality is now provided by `system/font/xorg/xorg-core`, and use `system/font/xorg/xorg-core` in your playbook instead.

bcoca:
would passing `-r` to `pkg list` also work out this renaming? should we switch the test to `pkg info -r`? it seems that fixing the detection error is better than handling the outcome of it's failure.

mavit:
I'm afraid not.  Here `-r` means, "display information about packages in the repositories", which doesn't help when checking for installed packages.

Presumably it would be possible to resolve these chains by calling the pkg Python API, but it looks pretty fiddly and low-level and I don't think it's supposed to be user-facing.

bcoca:
_grumble_solaris_grumble_, ok, so it seems rc == 4 detection it is, at this point i would even take out the existing check as it is ineffective.

PR #934, opened 2015-09-08
fix #894 by actually updating with the modified settings

### LINKED ISSUE #894: Firewalld - adding IP to zone doesn't work
Pretty simple line like this just isn't working:

`- firewalld: source=1.2.3.4 state=enabled zone=trusted`

Adding the rule manually with firewall-cmd results in ansible showing OK instead of changed, so it's checking the existing state properly but isn't adding new rules.

This is with a centos 7 machine (up to date) running firewalld 0.3.9

debugloop:
The latest version of `firewalld.py` does not apply the settings modified by the `get_source` and `remove_source` methods. This PR fixes this by writing the settings as done by other methods of this class and was tested with v2.0.0 against CentOS 7.

### DIFF
--- a/system/firewalld.py
+++ b/system/firewalld.py
@@ -76,7 +76,7 @@ options:
 notes:
   - Not tested on any Debian based system.
 requirements: [ 'firewalld >= 0.2.11' ]
-author: "Adam Miller (@maxamillion)" 
+author: "Adam Miller (@maxamillion)"
 '''
 
 EXAMPLES = '''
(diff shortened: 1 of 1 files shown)


bcoca:
cc @maxamillion

maxamillion:
:+1:

debugloop:
@maxamillion not sure if this warrants it's own bug report, but I think the firewalld module is behaving unexpected in some cases. For example, using `immediate=true` does not apply any rule for me. Are you working on this module and have this on your todo list?

maxamillion:
It's not currently on my TODO list, please open an issue for it and we can get it sorted from there. Thanks again for this fix! :)

PR #936, opened 2015-09-08
added missing serialize_instances function

privateip:
same issue as https://github.com/ansible/ansible-modules-extras/pull/935 

This PR adds the function.  Module is not currently usable until this PR is merged.


bcoca:
hmm, now i'm thinking this function should be in shared code in module_utils/vca.py

bcoca:
cc @bennojoy

privateip:
makes sense, all of the vca authentication should move there as well at the
same time.  should also pick up vca_vapp as well as there will most likely
be additional vca modules added in the future.

gregdek:
No author: field in this module? Can I presume it's @privateip, or @bennojoy, or...?

Marking needs_revision to make sure somebody adds an author: field :)

PR #938, opened 2015-09-09
NetworkManager reimplementation - Initial release

alesz:
- Allows free-form configuration of all parameters
- Interfaces with NetworkManager through nmcli tool only
- Tested ethernet, vlan, bridge and wifi interfaces

### DIFF
--- a/network/nmcli2.py
+++ b/network/nmcli2.py
@@ -1,0 +1,340 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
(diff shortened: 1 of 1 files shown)


sivel:
What is wrong with the nmcli module that was _just_ added to Ansible a few months ago?

https://github.com/ansible/ansible-modules-extras/blob/devel/network/nmcli.py

http://docs.ansible.com/ansible/nmcli_module.html

bcoca:
if there is a problem with the existing module we should fix it, not create competing implementations.

PR #954, opened 2015-09-10
fixing hang after creation in ec2_eni

evanccnyc:
It hangs after creating the interface if not attaching.


jimi-c:
Merged, thanks!

PR #971, opened 2015-09-13
Ensure tag values get updated in ec2_vpc_subnet

manicminer:
This minor fix ensures that tag values are updated for a VPC subnet, when tags already exist but the values differ. Previously, if you updated a tag value, the module would report a change but the value would not be updated.

For example, if you were to run the following task, then update either the `Name` or `env` tags, and run it again, would report that the subnet was updated but the tag values would stay the same.

``` yaml
- ec2_vpc_subnet:
    region: us-west-2
    az: us-west-2a
    cidr: 10.0.1.0/24
    state: present
    tags:
      Name: "my subnet"
      env: "production"
    vpc_id: vpc-12345678
```

Mentioning @erydo as the module author

### DIFF
--- a/cloud/amazon/ec2_vpc_subnet.py
+++ b/cloud/amazon/ec2_vpc_subnet.py
@@ -163,7 +163,7 @@ def ensure_tags(vpc_conn, resource_id, tags, add_only, check_mode):
         if to_delete and not add_only:
             vpc_conn.delete_tags(resource_id, to_delete, dry_run=check_mode)
 
-        to_add = dict((k, tags[k]) for k in tags if k not in cur_tags)
+        to_add = dict((k, tags[k]) for k in tags if k not in cur_tags or cur_tags[k] != tags[k])
         if to_add:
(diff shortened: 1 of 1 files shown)


gregdek:
@erydo Please review this PR to make sure it adheres to the following guidelines:

robynbergeron:
@erydo -- still looking for a review on this PR.

Also pinging other AWS modules folks for their possible review as well -- @jsmartin @scicoin-project @tombamford @garethr @lorin @jarv @jsdalton @silviud @adq @zbal @zeekin @willthames @lwade @carsongee @defionscode @tastychutney @bpennypacker @loia

rwe:
Sorry about that—my head has been out of the ansible stack for several months. That said, this patch looks entirely correct.

As for the module checklist @robynbergeron @gregdek, this PR doesn't introduce a new module nor does it modify any of the major pieces referred to in the guidelines and so I'm not sure if the procedure here is to do a full fresh check. The changes at least do not violate the checklist.

So, from my end: :shipit:

juliedavila:
shipit

PR #974, opened 2015-09-14
Pushbullet.com no longer uses camel case, so we shouldn't either.

kovacsbalu:


bcoca:
we might want to make this conditional as people that use the older lib will have the camel case

kovacsbalu:
In fact, pushbullet.py support old camel case style, so skip that fix.

robynbergeron:
@kovacsbalu -- I think what I'm reading here is that this PR is not actually needed because pushbullet.py does support old camel case, so I'm going to close this PR. You are welcome to ping here and reopen it if I am mistaken in my understanding :)

Thanks!

PR #982, opened 2015-09-15
Add cloudstack loadbalancer modules

DazWorrall:
This adds `cs_loadbalancer_rule` and `cs_loadbalancer_rule_member`, /cc @resmo for review

### DIFF
--- a/cloud/cloudstack/cs_loadbalancer_rule.py
+++ b/cloud/cloudstack/cs_loadbalancer_rule.py
@@ -1,0 +1,280 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+#
+# (c) 2015, Darren Worrall <darren@iweb.co.uk>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible. If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: cs_loadbalancer_rule
+short_description: Manages load balancer rules
+description:
+    - Add or remove load balancer rules
+version_added: '2.0'
+author: "Darren Worrall @dazworrall"
+options:
+  name:
+    description:
+      - The name of the load balancer rule
+    required: true
+  algorithm:
+    description:
+      - Load balancer algorithm
+      - Required when using C(state=present).
+    required: false
+    choices: [ 'source', 'roundrobin', 'leastconn' ]
+    default: null
+  private_port:
+    description:
+      - The private port of the private ip address/virtual machine where the
+        network traffic will be load balanced to
+      - Required when using C(state=present).
+    required: false
+    default: null
+  public_port:
+    description:
+      - The public port from where the network traffic will be load balanced from
+      - Required when using C(state=present).
+    required: true
+    default: null
+  public_ip:
+    description:
+      - Public ip address from where the network traffic will be load balanced from
+    required: false
+    default: null
+  open_firewall:
+    description:
+      - Whether the firewall rule for public port should be created, while creating the new rule.
+      - Use M(cs_firewall) for managing firewall rules.
+    required: false
+    default: false
+  state:
+    description:
+      - State of the instance.
+    required: true
+    default: 'present'
+    choices: [ 'present', 'absent' ]
+  domain:
+    description:
+      - Domain the rule is related to.
+    required: false
+    default: null
+  account:
+    description:
+      - Account the rule is related to.
+    required: false
+    default: null
+  zone:
+    description:
+      - Name of the zone in which the rule shoud be created.
+      - If not set, default zone is used.
+    required: false
+    default: null
+extends_documentation_fragment: cloudstack
+'''
+
+EXAMPLES = '''
+# Create a load balancer rule
+- local_action:
+    module: cs_loadbalancer_rule
+    name: balance_http
+    public_ip: 1.2.3.4
+    algorithm: leastconn
+    public_port: 80
+    private_port: 8080
+
+# Delete a load balancer rule
+- local_action:
+    module: cs_loadbalancer_rule
+    name: balance_http
+    state: absent
+'''
+
+RETURN = '''
+---
+id:
+  description: UUID of the rule.
+  returned: success
+  type: string
+  sample: a6f7a5fc-43f8-11e5-a151-feff819cdc9f
+zone:
+  description: Name of zone the rule is related to.
+  returned: success
+  type: string
+  sample: ch-gva-2
+project:
+  description: Name of project the rule is related to.
+  returned: success
+  type: string
+  sample: Production
+account:
+  description: Account the rule is related to.
+  returned: success
+  type: string
+  sample: example account
+domain:
+  description: Domain the rule is related to.
+  returned: success
+  type: string
+  sample: example domain
+'''
+
+
+try:
+    from cs import CloudStack, CloudStackException, read_config
+    has_lib_cs = True
+except ImportError:
+    has_lib_cs = False
+
+# import cloudstack common
+from ansible.module_utils.cloudstack import *
+
+
+class AnsibleCloudStackLBRule(AnsibleCloudStack):
+
+    def __init__(self, module):
+        super(AnsibleCloudStackLBRule, self).__init__(module)
+        self.returns = {
+            'publicip': 'public_ip',
+            'publicport': 'public_port',
+            'privateport': 'private_port',
+            'algorithm': 'algorithm',
+            'cidrlist': 'cidrlist',
+        }
+
+    def get_ip_address(self, ip_address=None, key=None):
+        if ip_address is None:
+            ip_address = self.module.params.get('ip_address')
+        if not ip_address:
+            self.module.fail_json(msg="IP address param 'ip_address' is required")
+
+        args = {}
+        args['ipaddress'] = ip_address
+        args['account'] = self.get_account(key='name')
+        args['domainid'] = self.get_domain(key='id')
+        args['projectid'] = self.get_project(key='id')
+        ip_addresses = self.cs.listPublicIpAddresses(**args)
+
+        if ip_addresses:
+            self.ip_address = ip_addresses['publicipaddress'][0]
+        return self._get_by_key(key, self.ip_address)
+
+    def get_rule(self, **kwargs):
+        rules = self.cs.listLoadBalancerRules(**kwargs)
+        if rules:
+            return rules['loadbalancerrule'][0]
+
+    def create_lb_rule(self):
+        args = {
+            'account': self.get_account(key='name'),
+            'domainid': self.get_domain(key='id'),
+            'projectid': self.get_project(key='id'),
+            'zoneid': self.get_zone(key='id'),
+            'publicipid': self.get_ip_address(self.module.params.get('public_ip'), key='id'),
+            'name': self.module.params.get('name'),
+        }
+        rule = self.get_rule(**args)
+        if not rule and not self.module.check_mode:
+            args.update({
+                'algorithm': self.module.params.get('algorithm'),
+                'privateport': self.module.params.get('private_port'),
+                'publicport': self.module.params.get('public_port'),
+            })
+            res = self.cs.createLoadBalancerRule(**args)
+            if 'errortext' in res:
+                self.module.fail_json(msg="Failed: '%s'" % res['errortext'])
+
+            poll_async = self.module.params.get('poll_async')
+            if poll_async:
+                res = self.poll_job(res, 'loadbalancer')
+            rule = res
+            self.result['changed'] = True
+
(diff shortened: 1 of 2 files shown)


cloud/cloudstack/cs_loadbalancer_rule_member.py
```diff
@@ -0,0 +1,253 @@
...
+        return self._change_members('remove')
+
+
+def main():
+    module = AnsibleModule(
+        argument_spec = dict(
+            name = dict(),
+            vms = dict(required=True),
+            state = dict(choices=['present', 'absent'], default='present'),
+            zone = dict(default=None),
```
resmo, line 222: project_id is missing here as well in the docs, but is used on L155.

DazWorrall, line 222: It shouldn't be in _member at all, I'll fix that, and add it correctly in the main module - you can specify a project when searching for a load balancer by IP address.

cloud/cloudstack/cs_loadbalancer_rule_member.py
```diff
@@ -0,0 +1,253 @@
...
+        return {
+            'account': self.get_account(key='name'),
+            'domainid': self.get_domain(key='id'),
+            'projectid': self.get_project(key='id'),
+            'zoneid': self.get_zone(key='id'),
+        }
+
+    def _change_members(self, operation):
+        if operation not in ['add', 'remove']:
+            raise RuntimeError('Bad operation: %s' % operation)
```
resmo: `self.module.fail_json(msg="Bad operation: %s" % operation)` is preferred, because any traceback to the user would is defined to be a bug.

cloud/cloudstack/cs_loadbalancer_rule_member.py
resmo, line 222: hmm but here https://github.com/ansible/ansible-modules-extras/pull/982/files#diff-3b35a6b36e38ac06c3856c928c352854R147 you wouldn't find the rule for a project right? IMHO project_id is needed also for _member

DazWorrall, line 222: You're right, I'll add it in

cloud/cloudstack/cs_loadbalancer_rule_member.py
```diff
@@ -0,0 +1,252 @@
...
+        }
+
+    def _change_members(self, operation):
+        if operation not in ['add', 'remove']:
+            self.module.fail_json(msg="Bad operation: %s" % operation)
+        args = self._get_common_args()
+        rule = self.get_rule(name=self.module.params.get('name'), **args)
+        if not rule:
+            self.module.fail_json(msg="Unknown rule: %s" % self.module.params.get('name'))
+        if not self.module.check_mode:
```
resmo: imho this condition is not on the right place. check mode should be right before the execution of removeFromLoadBalancerRule() or assignToLoadBalancerRule(). otherwise --check would always return `changed: false`

cloud/cloudstack/cs_loadbalancer_rule.py
```diff
@@ -0,0 +1,283 @@
...
+class AnsibleCloudStackLBRule(AnsibleCloudStack):
+
+    def __init__(self, module):
+        super(AnsibleCloudStackLBRule, self).__init__(module)
+        self.returns = {
+            'publicip': 'public_ip',
+            'publicport': 'public_port',
+            'privateport': 'private_port',
+            'algorithm': 'algorithm',
+            'cidrlist': 'cidrlist',
```
resmo: in other module I use just `cidr` for `clidrlist` e.g. in `cs_firewall` https://github.com/ansible/ansible-modules-extras/blob/devel/cloud/cloudstack/cs_firewall.py#L222

cloud/cloudstack/cs_loadbalancer_rule.py
```diff
@@ -0,0 +1,283 @@
...
+def main():
+    module = AnsibleModule(
+        argument_spec = dict(
+            name = dict(),
+            algorithm = dict(choices=['source', 'roundrobin', 'leastconn'], required=False),
+            private_port = dict(type='int', default=10, required=False),
+            public_port = dict(type='int', default=10, required=False),
+            state = dict(choices=['present', 'absent'], default='present'),
+            public_ip = dict(required=False),
+            cidr_list = dict(required=False),
```
resmo: also use `cidr` here for convention to other modules.

cloud/cloudstack/cs_loadbalancer_rule_member.py
```diff
@@ -0,0 +1,259 @@
...
+
+    def remove_members(self):
+        return self._change_members('remove')
+
+
+def main():
+    module = AnsibleModule(
+        argument_spec = dict(
+            name = dict(),
+            vms = dict(required=True),
```
resmo: set `type=list` for param vms so you can pass a yaml list or a single value,  as well as a comma separated list, also add an `alias=vm` so user can use `vm=myvm.example.com` as well.

cloud/cloudstack/cs_loadbalancer_rule.py
```diff
@@ -0,0 +1,283 @@
...
+        return rule
+
+
+def main():
+    module = AnsibleModule(
+        argument_spec = dict(
+            name = dict(),
+            algorithm = dict(choices=['source', 'roundrobin', 'leastconn'], required=False),
+            private_port = dict(type='int', default=10, required=False),
+            public_port = dict(type='int', default=10, required=False),
```
resmo: default=10 makes no sense :)

cloud/cloudstack/cs_loadbalancer_rule.py
```diff
@@ -0,0 +1,283 @@
...
+            self.result['changed'] = True
+        return rule
+
+
+def main():
+    module = AnsibleModule(
+        argument_spec = dict(
+            name = dict(),
+            algorithm = dict(choices=['source', 'roundrobin', 'leastconn'], required=False),
+            private_port = dict(type='int', default=10, required=False),
```
resmo: default=10 makes no sense :)

cloud/cloudstack/cs_loadbalancer_rule.py
```diff
@@ -0,0 +1,283 @@
...
+            'publicipid': self.get_ip_address(self.module.params.get('public_ip'), key='id'),
+            'name': self.module.params.get('name'),
+        }
+
+    def create_lb_rule(self):
+        args = self._get_common_args()
+        rule = self.get_rule(**args)
+        if not rule and not self.module.check_mode:
+            args.update({
+                'algorithm': self.module.params.get('algorithm'),
```
resmo: This would result in a cloudstack exception if algorithm is None (you specified this is required on state=present which is the way it should be). however for integration testing, it would be easier to tests against a fail_json msg, e.g. https://github.com/ansible/ansible-modules-extras/pull/941/files#diff-7b26e22953353b1b6a5d9bb0ac7ba754R312

cloud/cloudstack/cs_loadbalancer_rule.py
```diff
@@ -0,0 +1,288 @@
...
+            if poll_async:
+                res = self._poll_job(res, 'loadbalancer')
+            self.result['changed'] = True
+        return rule
+
+
+def main():
+    module = AnsibleModule(
+        argument_spec = dict(
+            name = dict(),
```
resmo: should be required=True

cloud/cloudstack/cs_loadbalancer_rule.py
```diff
@@ -0,0 +1,288 @@
...
+            })
+            res = self.cs.createLoadBalancerRule(**args)
+            if 'errortext' in res:
+                self.module.fail_json(msg="Failed: '%s'" % res['errortext'])
+
+            poll_async = self.module.params.get('poll_async')
+            if poll_async:
+                res = self.poll_job(res, 'loadbalancer')
+            rule = res
+            self.result['changed'] = True
```
resmo: this must be outside the check_mode condition

cloud/cloudstack/cs_loadbalancer_rule.py
```diff
@@ -0,0 +1,288 @@
...
+        args = self._get_common_args()
+        rule = self.get_rule(**args)
+        if rule and not self.module.check_mode:
+            res = self.cs.deleteLoadBalancerRule(id=rule['id'])
+            if 'errortext' in res:
+                self.module.fail_json(msg="Failed: '%s'" % res['errortext'])
+            poll_async = self.module.params.get('poll_async')
+            if poll_async:
+                res = self._poll_job(res, 'loadbalancer')
+            self.result['changed'] = True
```
resmo: same here, outside check_mode condition

gregdek:
Looks like @resmo has been doing a good job of reviewing so far, so thanks for that.

Assuming that this is now ready for review, so putting back into community_review.

@resmo, when you're happy with these, just give us the "shipit" sign.

cloud/cloudstack/cs_loadbalancer_rule_member.py
```diff
@@ -0,0 +1,262 @@
...
+        return self._change_members('add')
+
+    def remove_members(self):
+        return self._change_members('remove')
+
+
+def main():
+    module = AnsibleModule(
+        argument_spec = dict(
+            name = dict(),
```
resmo: name is required

resmo:
I thought about `public_ip` and IMHO it should be required for `cs_loadbalancer_rule` even if the API does not force it (i guess it looks if there is just one IP so it would take this one?) and it should also `publicipid` passed to `createLoadBalancerRule()`. This way we have a very clear defined state.

What is your opinion?

cloud/cloudstack/cs_loadbalancer_rule.py
```diff
@@ -0,0 +1,290 @@
...
+            if poll_async:
+                res = self._poll_job(res, 'loadbalancer')
+        return rule
+
+
+def main():
+    module = AnsibleModule(
+        argument_spec = dict(
+            name = dict(required=True),
+            algorithm = dict(choices=['source', 'roundrobin', 'leastconn'], required=False),
```
resmo: I would add here a `default='source'`, user can still change it if they want but source is the one I use most. What do you think?

DazWorrall:
It passes `publicipid` [already](https://github.com/DazWorrall/ansible-modules-extras/blob/22b16931f2af1923da265cd4e2edf885d00cc810/cloud/cloudstack/cs_loadbalancer_rule.py#L192), as for making it mandatory - I was a bit wary about deviating from the api in case there was some other use case I couldn't think of and it hurts someone down the line. It doesn't affect any of my use cases yet though, so if you'd prefer it I'm happy to change it

cloud/cloudstack/cs_loadbalancer_rule.py
DazWorrall: I was just following what the api does, which doesnt have a default; I personally prefer `leastconn` for example! But I'll go with whatever you prefer, it's easily selected in a play :)

resmo:
So in this case (you prefer to leave it optional, I prefer to make it mandatory) I would like to do some tests and report back the results later on.

BTW I will create some integration tests (living in https://github.com/ansible/ansible/blob/devel/test/integration/cloudstack.yml) anyway and these will be automatically tested on my build server http://build.ngine.io:8080/job/ansible.cloudstack.base/. currently only basic network has been tested but working for an advanced networking setup.

I already found another edge case if you have more than one isolated network, you also must pass `networkid`, but this is also not yet handled in other modules as well. And not see such an setup in real world.

cloud/cloudstack/cs_loadbalancer_rule.py
resmo: yes sure, I mean we should have a default no matter what it is, this saves some user input.
If there is no default, user have to define it anyway... so make a default, take whatever you like. :)

DazWorrall:
Yeah, we certainly dont do that here at the moment. Let me know what you decide with `public_ip` and I'll sort it, but if we run out of time I'll make it mandatory so we can get it merged for v2.

resmo:
I am not afraid about the timing. No worry. There will be a beta and RCs before a release of v2. These two modules are certainly the most requested ones and they should be as awesome as possible :)

resmo:
I looked at it more detailed and have a few improvements, too much to discuss those here. I'll send a PR to your repo for discussion.

robynbergeron:
@DazWorrall @resmo -- I see there was a new commit after @resmo's comment above -- is this ready for @resmo to review again, or are you still working collaborating on this elsewhere?

resmo:
@DazWorrall I was not able to make a PR to your repo. I assume this is a github bug, because it seems related that you forked form emonty/ansible-modules-extras not form ansible/ansible-module-extras. However you find the changes here https://github.com/resmo/ansible-modules-extras/tree/feature/cs_loadbalancer_rule, maybe you can try to delete your "fork" and clone it again from ansible/ansible-module-extras and push your branches again.

A few notes to my changes. I make the public_ip address required, so it makes sure we are identify the right rule and i renamed it to ip_address but made an alias so the existing functionality in utils can be used.

The module now updates a rule, e.g. algorithm. Added all returns in the correct format (some int casts for bad returns from the API), and added tags handling.

IMHO we should close this PR and make pull requests to each branch `feature/cs_loadbalancer_rule` unless we are done with the modules. I also started with making integration tests for it. 

I didn't looked add _rule_member yet and the next 14 days I am on vacation.

DazWorrall:
I've deleted and re-forked, which may well have hosed this pull request but if you want to send that PR now @resmo we should be fine, or just make a new PR directly to this repo from your own fork, I dont mind. I've done a few tests with the playbooks I was using here and they still work fine :+1:

resmo:
@DazWorrall ok, PR made to your repo. coming back to you in 2 weeks.

PR #994, opened 2015-09-18
Module for managing OS X launchd (system/launchd)

multimac:
Created this module for managing launchd on OS X via launchctl. This removes the need for fumbling around with the command/shell module to manage jobs in launchd. It also correctly reports whether the system was changed (jobs loaded/unloaded/started/stopped/etc.).

It's currently built for the OS X 10.10 version of launchd, as major changes were made between this and the 10.9 version.


multimac:
Closing due to messed up commit history

PR #995, opened 2015-09-18
Module for managing launchd on OS X (system/launchd)

multimac:
Created this module for managing launchd on OS X via launchctl. This removes the need for fumbling around with the command/shell module to manage jobs in launchd. It also correctly reports whether the system was changed (jobs loaded/unloaded/started/stopped/etc.).

It's currently built for the OS X 10.10 version of launchd, as major changes were made between this and the 10.9 version.

### DIFF
--- a/system/launchd.py
+++ b/system/launchd.py
@@ -1,4 +1,22 @@
 #!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2015, David Symons <Mult1m4c@gmail.com>
+#
+# This file is part of Ansible
+#
+# This module is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This software is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


robynbergeron:

robynbergeron:

# 

mmulich:
This works well. Some of the newer `launchctl` verbiage and domain space stuff is confusing. I'd suggest an example for using the `gui/$UID` domain space.

:bug: This needs to `os.path.expanduser` on `path`.

system/launchd.py
```diff
@@ -0,0 +1,405 @@
...
+
+
+def main():
+    state_choices = ['loaded', 'unloaded', 'enabled', 'disabled', 'started', 'stopped', 'restarted', 'status']
+    module = AnsibleModule(
+        argument_spec = dict(
+            state=dict(default='loaded', choices=state_choices, type='str'),
+            domain=dict(required=True, type='str'),
+            label=dict(default=None, alias=['name'], type='str'),
+            path=dict(default=None, type='str'),
```
bcoca, line 354: make this a 'path' type and both ~ and $HOME as well as relative paths will be automatically resolved.

kpaleniu:
It would be nice to have a feature to reload the service again. If the plist for the service is also managed using ansible and the plist changes, you need to unload & load the service for the changes to  effect. Unless of course there is a command that already does this that I am not aware of.

juju4:
:+1:

PR #999, opened 2015-09-19
Added module win_zip

davidobrien1985:
This module supports compressing local files and folders into a zip.

Requirements: At least .net Framework 4.5 installed.


robynbergeron:

windows/win_zip.ps1
```diff
@@ -0,0 +1,161 @@
...
+
+$params = Parse-Args $args;
+
+$result = New-Object psobject @{
+  win_zip = New-Object psobject
+  changed = $false
+}
+
+# Check if all arguments were provided and set defaults for optional params
+If ($params.src) {
```
trondhindenes, line 67: Now that we're moving to strict mode, line 67 isnt gonna work. Please use `Get-AnsibleParam` to get specific attributes from the param object.

trondhindenes:
There are now two modules in PR that do pretty much the same: #999 and #177. Maybe figure out which one to go forward with so we don't waste anyone's coding cycles? ping @schwartzmx @davidobrien1985

windows/win_zip.ps1
```diff
@@ -0,0 +1,161 @@
...
+  }
+  catch {
+    $ErrorMessage = $_.Exception.Message
+    Fail-Json $result $ErrorMessage
+  }
+}
+
+# compress the folder
+try {
+  [Reflection.Assembly]::LoadWithPartialName( "System.IO.Compression.FileSystem" )
```
schwartzmx, line 145: Is this available in PowerShell 3.0?  I don't think it is iirc

ah you are checking [here](https://github.com/ansible/ansible-modules-extras/pull/999/files#diff-459bce6ea74e00108d7c4b2f2f6cacd9R88) for .NET

windows/win_zip.py
```diff
@@ -0,0 +1,62 @@
...
+             will be renamed with a bak extension.
+        required: false
+        default: null
+        aliases: []
+author: "David O'Brien (obrien.david@outlook.com) @david_obrien"
+'''
+
+EXAMPLES = '''
+# Compress a single file, do not overwrite existing foo.zip. Will result in existing foo.zip to be renamed to foo.bak
+- win_copy: src=C:\\folder\\foo.bar dest=c:\\TEMP\\foo.zip
```
schwartzmx: Using win_copy here but the module is `win_zip`?  Maybe lingering from a previous change?

windows/win_zip.py
```diff
@@ -0,0 +1,62 @@
...
+# Compress a folder, overwrite an existing foo.zip.
+- win_copy: src=C:\\folder dest=c:\\TEMP\\foo.zip force=true
```
schwartzmx: Using win_copy here but the module is `win_zip`?  Maybe lingering from a previous change?

trondhindenes:
David, I have a unzip function which falls back to using COM if .Net 4.5 is not available. Maybe look at it and incorporate it in your code? https://gist.github.com/trondhindenes/2ce428c329771ce04214

davidobrien1985:
@trondhindenes my experience using COM to zip files / folders was never great. It works, but it was never great. Does it really need to support that?

windows/win_zip.ps1
davidobrien1985, line 67: I don't have a 2.0 dev environment at the moment. So I have to postpone this for now.

@robynbergeron which version are contributors expected to develop on? If I go and use the new Get-AnsibleParam function, then this module will only work on 2.0 and up and it can't be back ported to 1.9.3 for example.

PR #1005, opened 2015-09-21
zabbix_host : add the ability to set inventory_mode

ritzk:
Allows user to set inventory_mode


monitoring/zabbix_host.py
```diff
@@ -356,6 +362,28 @@ def link_or_clear_template(self, host_id, template_id_list):
...
+            inventory_mode = int(-1)
+
+
+        # watch for - https://support.zabbix.com/browse/ZBX-6033
+        request_str = {'hostid': host_id, 'inventory_mode': inventory_mode}
+        try:
+            if self._module.check_mode:
+                self._module.exit_json(changed=True)
+            self._zapi.host.update(request_str)
+        except Exception, e:
```
resmo, line 385: catch all exception should be avoided.

resmo:
would you mind squash the commits into one?

ritzk:
Hi @resmo 
 Not sure, which all Exception I should be catching. Any guidelines would be helpful.

resmo:
@ritzk I see there are a few catch all exceptions in this module. So one more doesn't hurt. Can be addressed later.

gregdek:
@cove Please review this PR to make sure it adheres to the following guidelines:

PR #1007, opened 2015-09-21
Contribute new module win_invoke_from_scheduled_job

breathe:
This adds a new windows module win_invoke_from_scheduled_job

The module expects an argument that is a powershell command string and immediately executes that command from within a windows scheduled task.  This is needed to escape the winrm jail that prevents usage of some kinds of windows rpc mechanisms ...

The module is implemented using the powershell 3 api's -- register-scheduledjob, wait-job, and receive-job.  register-scheduledjob automatically creates a scheduled task in which to run the given powershell snippet.  wait-job blocks until a job in progress has completed.  receive-job receives back the results of the scheduled-job execution.

More info on winrm execution environment: http://www.hurryupandwait.io/blog/safely-running-windows-automation-operations-that-typically-fail-over-winrm-or-powershell-remoting

Approach adapted from here: https://github.com/mwrock/boxstarter/blob/master/BoxStarter.Common/Invoke-FromTask.ps1

ScheduledJob api intro: http://blogs.technet.com/b/heyscriptingguy/archive/2014/05/12/introduction-to-powershell-scheduled-jobs.aspx

My first powershell script -- I'm by no means a powershell expert ...  Feedback welcome


robynbergeron:
Hi @breathe --

robynbergeron:
Also: pinging our windows module friends in case they are able to take a look at this. 

@trondhindenes @petemounce @elventear @smadam813 @jhawkesworth @angstwad @sivel @chrishoffman @cchurch

trondhindenes:
Nice, looking forward to testing it. @breathe, have you considered adding credentials support in order to support scenarios where the task needs to be run with a different credential than the calling user?

breathe:
I had considered credential support but ended up not needing it for my immediate use case -- and wasn't sure if anyone else would want this module.  I think it should be straightforward to add -- register-scheduledjob accepts -Credential <credential> and -Authentication <AuthenticationTypeEnum> arguments -- I'm not sure what the right value would be for the authentication type though ...  It seems like -Authentication Basic is most straightforward and sensible since the module's arguments have to be transferred over the network anyway ...?

It might also be good to support a 'RunElevated' option and perhaps some other switches which can be passed to the -ScheduledJobOption argument of register-scheduledjob:

```
             @{StartIfOnBattery=$False; StopIfGoingOnBattery=$True; WakeToRun=$False; StartIfNotIdle=$False; 
        IdleDuration="00:10:00"; IdleTimeout="01:00:00"; StopIfGoingOffIdle=$True; RestartOnIdleResume=$False; 
        ShowInTaskScheduler=$True; RunElevated=$False; RunWithoutNetwork=$False; DoNotAllowDemandStart=$False; 
        MultipleInstancePolicy=IgnoreNew}
```

windows/win_invoke_from_scheduled_job.ps1
```diff
@@ -0,0 +1,87 @@
...
+
+# ensure a job of the same name is not already registered
+Unregister-ScheduledJob -name $jobName -Force -errorAction SilentlyContinue
+
+# register the command to run as a scheduled task and tell it to execute immediately
+$job = Register-ScheduledJob -RunNow -ScriptBlock $wrapperCommandClosure -Name $jobName
+
+# poll for the job to have started ... then wait for it to finish 
+# wait-job is called repeatedly in a loop because wait-job returns $null if the job does not exist yet ...
+while ($true)
```
elventear: This is minor, but I think it would improve the code.

```
while((wait-job -Name $jobname) -eq $null) {
    start-sleep -s 1
}
```

breathe:
@elventear - thanks for the suggestion, I changed the while loop.

I also added support for a RunElevated option and changed the defaults so the scheduled job is more likely to run reliably when run on a battery powered device.

breathe:
works for me ...  I only have a windows 10 testing environment though at the moment, so I can't validate whether it works on powershell 3/4 ...

nitzmahone:
Unfortunately, getting this working robustly requires more hassle. The PS ScheduledJob stuff is pretty cantankerous (as I discovered when writing the new win_updates module), and there are undocumented failure cases that can occur pretty repeatably at each call in here. The most egregious one is the delayed appearance of the actual job output in the return value from Receive-Job (ie, Wait/Receive-Job eventually returns a non-null object whose Output property must then be polled). I had several other complications for win_updates that may or may not be relevant here (eg, ensuring that only one update job would run at a time in the case of dropped connections/retries).

breathe:
Thanks for the feedback @nitzmahone -- I haven't run into this issue of receive-job returning anything other than the job results yet ...  Do you know how to reproduce?  I've only tested on windows 10 -- possible the issues you ran into are fixed in powershell5?

Looking at win_update.ps1 -- I see you are using the -ArgumentList parameter to register-scheduledjob -- I had tried to use that but was apparently mis-using it as I couldn't get it to work -- I must've been using it incorrectly and should look at that again.

I see you opted not to use receive-job -- I guess this is due to experimentally determined issues you ran into?  Had you tried the pattern that's in this module during that process ... -- call wait_job repeatedly until non-null and then at that point calling receive-job with -wait parameter ...?

Another look at the docs also makes me think I should add the -Force parameter to the Wait-Job and Receive-Job calls -- although the need to do this _might_ be sidestepped by the StopIfGoingOnBatteries=$false scheduledjoboption (assuming that StopIfGoingOnBatteries=$false is sufficient to prevent the scheduled job from ever idling ...  This I do not really know as its not clear to me what all the causes of a scheduled job going into idle state might be.  The -Force parameter on Wait-Job and Receive-Job claims to handle any such cases though so probably is a good idea to add) ...

breathe:
As for preventing multiple instances from running at the same time ...  This module uses a hard-coded job name -- and enforces a job of that name is not already running via unregister-scheduledjob prior to running ...  This seems relatively simple and straightforward -- is there a better way?

nitzmahone:
Hey @breathe: Unfortunately, there isn't much in the way of published best practices to follow for Windows modules yet. It looks like you've got a good start here, but there are a lot of issues that would keep me from +1'ing this right now. I can tell you some of the things I personally look for to +1 for inclusion:
- We generally support back to Powershell 3.0 / Windows Server 2008R2 / Windows 7. It's OK if your module doesn't, but the requirements need to be clearly documented, and ideally, your module will validate its assumptions and fail gracefully on unsupported platforms.
- Reasonably robust error-handling/recovery. I think if you test this module much on older PS/Windows versions, you'll find that it's still got a lot of issues, and that some of them will require manual intervention to fix (eg, won't be re-runnable without hackage) the way things are implemented now.
- A description of the testing you've done (which platforms, PS versions, types/sizes of scripts, etc). Sans that, I'll assume, "works on my machine" is as far as it went. ;)

I've spent a lot of time solving most of these issues already in the win_updates scheduled task stuff. Early on, there was some talk about moving my scheduled task wrapper into powershell.ps1 (which I'd personally oppose for various reasons), or maybe into another opt-in module replacer file (which I would support). If we got some traction on that, having this module leverage it would be a much easier sell, and prevent us from needing to maintain duplicate task scheduler wrappers. I'd be happy to engage in that discussion, or kick it off if you like.

breathe:
I would support you kicking off that discussion.  I'd love to remove this module in favor of native ansible support.  I think ansible should include a supported way to invoke processes outside the winrm jail -- I think this is probably needed within a variety of automation contexts ...

My main thought in offering up this module was -- the winrm jail is pretty obscure -- not well documented, and can cause non-obvious failures of scripts which run fine outside of it.  I think anyone using ansible to automated windows systems is likely to run into the need to escape this jail at some point (maybe my needs are just different and that's not true?).  Here was the first (hacky) way I came across to escape the jail and thought I'd offer it up since there didn't seem to a way anywhere else (at the time I had no idea there was an implementation of this embedded in win_updates.ps1 -- and that implementation is not currently useable outside of that module ...)

nitzmahone:
Done: https://groups.google.com/forum/#!topic/ansible-devel/ueEyAfFwW14

nitzmahone:
... and @breathe, you're definitely not alone. The various things that don't "just work" under WinRM are a constant source of frustration to many (myself included). I had enough issues with the PS scheduled job support though, that I've seriously considered a completely different approach to busting out of WinRM (ephemeral services a la PSExec). The PS Scheduled Job stuff is _slow_\- depends on platform, but in the worst cases, I've seen it add 20s+ of overhead to a no-op.

gregdek:
Thanks again to @breathe for this PR, and thanks @ansible/core for reviewing. Marking for inclusion.

nitzmahone:
Sorry, bot auto-tagged this with the word-that-shall-not-be-mentioned-until-it's-ready-to-be-included...

gregdek:
Thanks again to @breathe for this PR, and thanks @ansible/core for reviewing. Marking for inclusion.

nitzmahone:
Since we can't seem to keep the bot from re-tagging this: _DON'T SHIP IT_

gregdek:
Thanks @breathe for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

breathe:
I don't use ansible on Windows anymore.  And cannot test any of this code.  I'm not sure what's going on with the bot but I'll close this pull request.

PR #1008, opened 2015-09-22
cleaned up vca_vapp module to focus on managing vapps

privateip:
cleaned up vca_vapp module and simplified functionality to focus on vapp management


robynbergeron:
Marking for inclusion because owner_pr. Shipit, shipit on the shippy ship! 

cc @ansible/core :)

PR #1015, opened 2015-09-22
module: system/iptables

LinusU:
This is a new module for manipulating iptable entries. It supports both adding and removing based on state (`absent`/`present`) and has full check mode support.

### DIFF
--- a/system/iptables.py
+++ b/system/iptables.py
@@ -21,6 +21,11 @@
 # import module snippets
 from ansible.module_utils.basic import *
 
+BINS = dict(
+    ipv4='iptables',
+    ipv6='ip6tables',
+)
+
 DOCUMENTATION = '''
 ---
 module: iptables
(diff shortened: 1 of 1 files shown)


system/iptables.py
```diff
@@ -0,0 +1,130 @@
...
+
+# import module snippets
+from ansible.module_utils.basic import *
+
+DOCUMENTATION = '''
+---
+module: iptables
+short_description: Modify the systems iptables
+requirements: []
+version_added: 1.9
```
resmo: this should be `version_added: "2.0"`

system/iptables.py
```diff
@@ -0,0 +1,130 @@
...
+author:
+  - "Linus Unnebäck"
```
resmo: please add your github handle as well: `"Linus Unnebäck (@LinusU)"` and btw because you are the only author, it must not be a list...

system/iptables.py
```diff
@@ -0,0 +1,130 @@
...
+description:
+  - Modify the systems iptables.
```
resmo: this should be more clear: e.g. it does not change the startup iptables config only the running config, etc.

system/iptables.py
```diff
@@ -0,0 +1,130 @@
...
+# along with this software.  If not, see <http://www.gnu.org/licenses/>.
+
+# import module snippets
+from ansible.module_utils.basic import *
+
+DOCUMENTATION = '''
+---
+module: iptables
+short_description: Modify the systems iptables
+requirements: []
```
resmo, line 33: obviously it depends on `iptables`

system/iptables.py
```diff
@@ -0,0 +1,130 @@
...
+  table:
+    description: Packet matching table to operate on.
+    required: false
+    default: filter
+    choices: [ "filter", "nat", "mangle", "raw", "security" ]
+  chain:
+    description: Chain to operate on.
+    required: true
+    choices: [ "INPUT", "FORWARD", "OUTPUT", "PREROUTING", "POSTROUTING", "SECMARK", "CONNSECMARK" ]
+  rule:
```
bcoca: this seems just to be a 'catchall' for iptables command line options, i would prefer to break down interface (in, out), ip (source, destination), protocol, target, etc. this just removes 3 things from a raw shell command as is.

system/iptables.py
bcoca, line 33: no need to state that, i think that is obvious from the module name

system/iptables.py
LinusU: That is one way of doing it, I started looking at it but there is **very many** options that we need to support in this case.

First of all we need all the basic standard options, that's not too bad. It would be ~11 options, but that wouldn't even cover my basic example of port forwarding which I think is fairly common. To make this work we would need to support the extended set of options which seems to be **very many**. Here is the list: http://ipset.netfilter.org/iptables-extensions.man.html

All rules becomes a bit longer but it looks kinda nice.

``` diff
+ iptables: table=nat chain=PREROUTING in_interface=eth0 protocol=tcp match=tcp destination_port=80 jump=REDIRECT to_ports=8600'
- iptables: table=nat chain=PREROUTING rule='-i eth0 -p tcp -m tcp --dport 80 -j REDIRECT --to-ports 8600'
```

I don't know, what do you think?

LinusU: @bcoca I added a commit with the style you suggested, I actually think it's the way to go. It's easy to add new arguments if needed.

system/iptables.py
```diff
@@ -0,0 +1,216 @@
...
+    cmd = push_arguments(iptables_path, '-D', params)
+    module.run_command(cmd, check_rc=True)
+
+
+def main():
+    module = AnsibleModule(
+        supports_check_mode=True,
+        argument_spec=dict(
+            table=dict(required=False, default='filter', choices=['filter', 'nat', 'mangle', 'raw', 'security']),
+            chain=dict(required=True, default=None, choices=['INPUT', 'FORWARD', 'OUTPUT', 'PREROUTING', 'POSTROUTING', 'SECMARK', 'CONNSECMARK']),
```
bcoca: we might not want to restrict chains as you can have user defined chains, make it a string and we can document the standard choices

bcoca:
looks much better now. We'll probably need to make a note that this is only for the 'in memory' rules and that it does not save/restore them, those can be managed with the iptables init script/startup configs

LinusU:
@bcoca Looks good?

``` diff
diff --git a/system/iptables.py b/system/iptables.py
index 9acdeca..2b79f1d 100644
--- a/system/iptables.py
+++ b/system/iptables.py
@@ -33,19 +33,15 @@ short_description: Modify the systems iptables
 requirements: []
 version_added: "2.0"
 author: Linus Unnebäck (@LinusU) <linus@folkdatorn.se>
-description: Iptables is used to set up, maintain, and inspect the tables of IPv4 packet filter rules in the Linux kernel.
+description: Iptables is used to set up, maintain, and inspect the tables of IP packet filter rules in the Linux kernel. This module does not handle the saving and/or loading of rules, but rather only manipulates the current rules that are present in memory. This is the same as the behaviour of the "iptables" and "ip6tables" command which this module uses internally.
 options:
   table:
-    description: Packet matching table to operate on.
+    description: This option specifies the packet matching table which the command should operate on. If the kernel is configured with automatic module loading, an attempt will be made to load the appropriate module for that table if it is not already there.
     required: false
     default: filter
     choices: [ "filter", "nat", "mangle", "raw", "security" ]
-  chain:
-    description: Chain to operate on.
-    required: true
-    choices: [ "INPUT", "FORWARD", "OUTPUT", "PREROUTING", "POSTROUTING", "SECMARK", "CONNSECMARK" ]
   state:
-    description: Wheter the rule should be absent or present
+    description: Wheter the rule should be absent or present.
     required: false
     default: present
     choices: [ "present", "absent" ]
@@ -54,6 +50,9 @@ options:
     required: false
     default: ipv4
     choices: [ "ipv4", "ipv6" ]
+  chain:
+    description: Chain to operate on. This option can either be the name of a user defined chain or any of the builtin chains: "INPUT", "FORWARD", "OUTPUT", "PREROUTING", "POSTROUTING", "SECMARK", "CONNSECMARK"
+    required: true
   protocol:
     description: The protocol of the rule or of the packet to check. The specified protocol can be one of tcp, udp, udplite, icmp, esp, ah, sctp or the special keyword "all", or it can be a numeric value, representing one of these protocols or a different one. A protocol name from /etc/protocols is also allowed. A "!" argument before the protocol inverts the test. The number zero is equivalent to all. "all" will match with all protocols and is taken as default when this option is omitted.
     required: false
@@ -162,9 +161,9 @@ def main():
         supports_check_mode=True,
         argument_spec=dict(
             table=dict(required=False, default='filter', choices=['filter', 'nat', 'mangle', 'raw', 'security']),
-            chain=dict(required=True, default=None, choices=['INPUT', 'FORWARD', 'OUTPUT', 'PREROUTING', 'POSTROUTING', 'SECMARK', 'CONNSECMARK']),
             state=dict(required=False, default='present', choices=['present', 'absent']),
             ip_version=dict(required=False, default='ipv4', choices=['ipv4', 'ipv6']),
+            chain=dict(required=True, default=None, type='str'),
             protocol=dict(required=False, default=None, type='str'),
             source=dict(required=False, default=None, type='str'),
             destination=dict(required=False, default=None, type='str'),
```

bcoca:
yes, it looks good

wenottingham:
Looking at this - this module does not at all try to handle ordering, correct? As rules are a chain, order can matter, especially when jump/goto is involved.

LinusU:
That is true, I haven't personally ran into that problem thought. I don't know exactly how it would work actually, like what the syntax would be like.

Do you have any ideas?

wenottingham:
The iptables commandline only supports specifying rules by numerical place in the chain, I'm not sure if there's a good way to wrap that in the module without parsing all the rules and attempting to maintain that state in the module - it's probably just worth documenting.

bcoca:
just making a list of rules with the 'correct' order and using with_items would work in this case.

i would also argue that if you want to make a comprehensive set of rules that require these kinds of dependencies you should template the iptables restore file. This module just deals with individual rules.

robynbergeron:
Hi, everyone:

This PR somehow escaped the templated info for reviewing new modules -- putting that good stuff in now, so that @LinusU's continued work can result in good things :)

---

robynbergeron:
Also: I'm going to stick this in needs_revision -- just to address the ordering comments from @wenottingham and @bcoca (thanks, guys!). 

@LinusU -- when you've made the revisions, please leave a ready_for_review comment and we can move this along. 

In the meantime, of course, if anyone would like to review this new awesome module for adherence to guidelines (more info on that in the previous comment) or test it as is for functionality, you are more than welcome to do so!

LinusU:
Added this note:

> This module just deals with individual rules. If you need advanced chaining of rules the recommended way is to template the iptables restore file.

ready_for_review

LinusU:
Super \o/

Thanks for the merge, when will this be released into an Ansible version?

bcoca:
when 2.0 is released, which should be soon

LinusU:
Nice, thank you

PR #1016, opened 2015-09-22
Adding yumrepo module

jtyr:
This patch is adding a new module which allows to add and remove YUM repository definitions. The module implements all repository options as described in the `yum.conf` manual page.

### DIFF
--- a/packaging/os/yumrepo.py
+++ b/packaging/os/yumrepo.py
@@ -1,0 +1,560 @@
+#!/usr/bin/python
+# encoding: utf-8
+
+# (c) 2015, Jiri Tyr <jiri.tyr@gmail.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+
+import ConfigParser
+import os
+
+
+DOCUMENTATION = '''
+---
+module: yumrepo
+author: Jiri Tyr (@jtyr)
+version_added: '2.0'
+short_description: Add and remove YUM repositories
+description:
+  - Add or remove YUM repositories in RPM-based Linux distributions.
+
+options:
+  bandwidth:
+    required: false
+    default: 0
+    description:
+      - Maximum available network bandwidth in bytes/second. Used with the
+        I(throttle) option.
+      - If I(throttle) is a percentage and bandwidth is C(0) then bandwidth
+        throttling will be disabled. If I(throttle) is expressed as a data rate
+        (bytes/sec) then this option is ignored. Default is C(0) (no bandwidth
+        throttling).
+  baseurl:
+    required: false
+    default: None
+    description:
+      - URL to the directory where the yum repository's 'repodata' directory
+        lives.
+      - This or the I(mirrorlist) parameter is required.
+  cost:
+    required: false
+    default: 1000
+    description:
+      - Relative cost of accessing this repository. Useful for weighing one
+        repo's packages as greater/less than any other.
+  description:
+    required: false
+    default: None
+    description:
+      - A human readable string describing the repository.
+  enabled:
+    required: false
+    choices: ['yes', 'no']
+    default: 'yes'
+    description:
+      - This tells yum whether or not use this repository.
+  enablegroups:
+    required: false
+    choices: ['yes', 'no']
+    default: 'yes'
+    description:
+      - Determines whether yum will allow the use of package groups for this
+        repository.
+  exclude:
+    required: false
+    default: None
+    description:
+      - List of packages to exclude from updates or installs. This should be a
+        space separated list. Shell globs using wildcards (eg. C(*) and C(?))
+        are allowed.
+      - The list can also be a regular YAML array.
+  failovermethod:
+    required: false
+    choices: [roundrobin, priority]
+    default: roundrobin
+    description:
+      - C(roundrobin) randomly selects a URL out of the list of URLs to start
+        with and proceeds through each of them as it encounters a failure
+        contacting the host.
+      - C(priority) starts from the first baseurl listed and reads through them
+        sequentially.
+  file:
+    required: false
+    default: None
+    description:
+      - File to use to save the repo in. Defaults to the value of I(name).
+  gpgcakey:
+    required: false
+    default: None
+    description:
+      - A URL pointing to the ASCII-armored CA key file for the repository.
+  gpgcheck:
+    required: false
+    choices: ['yes', 'no']
+    default: 'no'
+    description:
+      - Tells yum whether or not it should perform a GPG signature check on
+        packages.
+  gpgkey:
+    required: false
+    default: None
+    description:
+      - A URL pointing to the ASCII-armored GPG key file for the repository.
+  http_caching:
+    required: false
+    choices: [all, packages, none]
+    default: all
+    description:
+      - Determines how upstream HTTP caches are instructed to handle any HTTP
+        downloads that Yum does.
+      - C(all) means that all HTTP downloads should be cached.
+      - C(packages) means that only RPM package downloads should be cached (but
+         not repository metadata downloads).
+      - C(none) means that no HTTP downloads should be cached.
+  includepkgs:
+    required: false
+    default: None
+    description:
+      - List of packages you want to only use from a repository. This should be
+        a space separated list. Shell globs using wildcards (eg. C(*) and C(?))
+        are allowed. Substitution variables (e.g. C($releasever)) are honored
+        here.
+      - The list can also be a regular YAML array.
+  keepalive:
+    required: false
+    choices: ['yes', 'no']
+    default: 'no'
+    description:
+      - This tells yum whether or not HTTP/1.1 keepalive should be used with
+        this repository. This can improve transfer speeds by using one
+        connection when downloading multiple files from a repository.
+  metadata_expire:
+    required: false
+    default: 21600
+    description:
+      - Time (in seconds) after which the metadata will expire.
+      - Default value is 6 hours.
+  metalink:
+    required: false
+    default: None
+    description:
+      - Specifies a URL to a metalink file for the repomd.xml, a list of
+        mirrors for the entire repository are generated by converting the
+        mirrors for the repomd.xml file to a baseurl.
+  mirrorlist:
+    required: false
+    default: None
+    description:
+      - Specifies a URL to a file containing a list of baseurls.
+      - This or the I(baseurl) parameter is required.
+  mirrorlist_expire:
+    required: false
+    default: 21600
+    description:
+      - Time (in seconds) after which the mirrorlist locally cached will
+        expire.
+      - Default value is 6 hours.
+  name:
+    required: true
+    description:
+      - Unique repository ID.
+  password:
+    required: false
+    default: None
+    description:
+      - Password to use with the username for basic authentication.
+  protect:
+    required: false
+    choices: ['yes', 'no']
+    default: 'no'
+    description:
+      - Protect packages from updates from other repositories.
+  proxy:
+    required: false
+    default: None
+    description:
+      - URL to the proxy server that yum should use.
+  proxy_password:
+    required: false
+    default: None
+    description:
+      - Username to use for proxy.
+  proxy_username:
+    required: false
+    default: None
+    description:
+      - Password for this proxy.
+  repo_gpgcheck:
+    required: false
+    choices: ['yes', 'no']
+    default: 'no'
+    description:
+      - This tells yum whether or not it should perform a GPG signature check
+        on the repodata from this repository.
+  reposdir:
+    required: false
+    default: /etc/yum.repos.d
+    description:
+      - Directory where the C(.repo) files will be stored.
+  retries:
+    required: false
+    default: 10
+    description:
+      - Set the number of times any attempt to retrieve a file should retry
+        before returning an error. Setting this to C(0) makes yum try forever.
+  skip_if_unavailable:
+    required: false
+    choices: ['yes', 'no']
+    default: 'no'
+    description:
+      - If set to C(yes) yum will continue running if this repository cannot be
+        contacted for any reason. This should be set carefully as all repos are
+        consulted for any given command.
+  sslcacert:
+    required: false
+    default: None
+    description:
+      - Path to the directory containing the databases of the certificate
+        authorities yum should use to verify SSL certificates.
+  ssl_check_cert_permissions:
+    required: false
+    choices: ['yes', 'no']
+    default: 'no'
+    description:
+      - Whether yum should check the permissions on the paths for the
+        certificates on the repository (both remote and local).
+      - If we can't read any of the files then yum will force
+        I(skip_if_unavailable) to be true. This is most useful for non-root
+        processes which use yum on repos that have client cert files which are
+        readable only by root.
+  sslclientcert:
+    required: false
+    default: None
+    description:
+      - Path to the SSL client certificate yum should use to connect to
+        repos/remote sites.
+  sslclientkey:
+    required: false
+    default: None
+    description:
+      - Path to the SSL client key yum should use to connect to repos/remote
+        sites.
+  sslverify:
+    required: false
+    choices: ['yes', 'no']
+    default: 'yes'
+    description:
+      - Defines whether yum should verify SSL certificates/hosts at all.
+  state:
+    required: false
+    choices: [absent, present]
+    default: present
+    description:
+      - A source string state.
+  throttle:
+    required: false
+    default: None
+    description:
+      - Enable bandwidth throttling for downloads.
+      - This option can be expressed as a absolute data rate in bytes/sec. An
+        SI prefix (k, M or G) may be appended to the bandwidth value.
+  timeout:
+    required: false
+    default: 30
+    description:
+      - Number of seconds to wait for a connection before timing out.
+  username:
+    required: false
+    default: None
+    description:
+      - Username to use for basic authentication to a repo or really any url.
+
+extends_documentation_fragment: files
+
+notes:
+  - All comments will be removed if modifying an existing repo file.
+  - Section order is preserved in an existing repo file.
+  - Parameters in a section are ordered alphabetically in an existing repo
+    file.
+  - The repo file will be automatically deleted if it contains no repository.
+'''
+
+EXAMPLES = '''
+- name: Add repository
+  yumrepo:
+    name: epel
+    description: EPEL YUM repo
+    baseurl: http://download.fedoraproject.org/pub/epel/$releasever/$basearch/
+
+- name: Add multiple repositories into the same file (1/2)
+  yumrepo:
+    name: epel
+    description: EPEL YUM repo
+    file: external_repos
+    baseurl: http://download.fedoraproject.org/pub/epel/$releasever/$basearch/
+    gpgcheck: no
+- name: Add multiple repositories into the same file (2/2)
+  yumrepo:
+    name: rpmforge
+    description: RPMforge YUM repo
+    file: external_repos
+    baseurl: http://apt.sw.be/redhat/el7/en/$basearch/rpmforge
+    mirrorlist: http://mirrorlist.repoforge.org/el7/mirrors-rpmforge
+    enabled: no
+
+- name: Remove repository
+  yumrepo:
+    name: epel
+    state: absent
+
+- name: Remove repository from a specific repo file
+  yumrepo:
+    name: epel
+    file: external_repos
+    state: absent
+'''
+
+RETURN = '''
+repo:
+    description: repository name
+    returned: success
+    type: string
+    sample: "epel"
+state:
+    description: state of the target, after execution
+    returned: success
+    type: string
+    sample: "present"
+'''
+
+
+class YumRepo(object):
+    # Class global variables
+    module = None
+    params = None
+    section = None
+    repofile = ConfigParser.RawConfigParser()
+
+    # List of parameters which will be allowed in the repo file output
+    allowed_params = [
+        'bandwidth', 'baseurl', 'cost', 'enabled', 'enablegroups', 'exclude',
+        'failovermethod', 'gpgcakey', 'gpgcheck', 'gpgkey', 'http_caching',
+        'includepkgs', 'keepalive', 'metadata_expire', 'metalink',
+        'mirrorlist', 'mirrorlist_expire', 'name', 'password', 'protect',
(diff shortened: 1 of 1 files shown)


packaging/os/yumrepo.py
```diff
@@ -0,0 +1,501 @@
...
+from ansible.module_utils.basic import *
+from ansible.module_utils.urls import *
+import ConfigParser
+import os
+
+
+DOCUMENTATION = '''
+---
+module: yumrepo
+version_added: 2.0
```
resmo: please add quotes to the 2.0: `version_added: "2.0"`

packaging/os/yumrepo.py
```diff
@@ -0,0 +1,501 @@
...
+
+notes:
+  - Default directory for the repositories is C(/etc/yum.repos.d). This can be
+    changed in the C(ansible.conf) file by declaring new section C([yumrepo])
+    and setting the parameter C(dir = /path/to/repos.d). The value in the
+    C(ansible.cfg) file can be overwritten by an environment variable
+    C(YUMREPO_DIR).
+  - The repo file will be automatically deleted if it contains no repository.
+
+author: Jiri Tyr
```
resmo: add you github handle as well. e.g. `author: "Jiri Tyr (@jtyr)"`

packaging/os/yumrepo.py
jtyr: Fixed

packaging/os/yumrepo.py
jtyr: Fixed

robynbergeron:
Hi @jtyr -- Thanks for submitting this new module to Ansible Extras! This module is now in community review, a process that is open to all Ansible users. In order for this module to be approved, it must gain the following votes:

robynbergeron:
(Also: thanks @resmo for comments thus far!)

jtyr:
@robynbergeron Thanks for the information. Please could you advise me what's the best way how to get somebody to review and tag the module with the “works_for_me” and “passes_guidelines” tags? Should I contact somebody from the [Subject Matter Experts](https://github.com/ansible/ansible-modules-extras/blob/devel/REVIEWERS.md#subject-matter-experts) or should I just wait till somebody is willing to review it?

robynbergeron:
Hi @jtyr -- 

You can read a bit more about the above referenced process here -- it is freshly introduced. ;) http://robyn.io/2015/09/28/ansible-extras-and-you/

We don't have a specific "best way" just yet -- and most of the subject matter expert folks that we have are more specific to areas other than this.  My suggestion would be to give it a little shout-out on twitter, facebook, or some other social media type place -- or maybe review someone else's outstanding new module PR and hope that the karma is returned :) We're hoping that widening the pool of potential reviewers will help out, but it may take a little bit of time for people to know that they really can help out :) 

I will give it a little bump on twitter and a shoutout to my Fedora friends and see if maybe anyone is willing to help out.

jtyr:
@robynbergeron Thanks for the explanation. I would definitely like to help to review some of the pull requests but first I would like to know if the one I have written satisfies the standard in order I can use the same measure for the others. I really appreciate your help and I hope I will be able to pay it back soon.

packaging/os/yumrepo.py
```diff
@@ -0,0 +1,501 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+
+from ansible.constants import get_config
+from ansible.constants import load_config_file
```
mgedmin: I don't think you can use those in modules.

The way Ansible modules work: your module's source code is pre-processed, treating `from ansible.module_utils.SOMETHING import *` as an include directive, then some placeholders in those includes get replaced with module arguments etc., then the preprocessed source is SCPed over to the remote and executed as a single file.

There's no requirement for Ansible to be installed on remote hosts, so you can't rely on ansible imports succeeding, with the exception of those special include directives that masquerade as Python imports.

packaging/os/yumrepo.py
```diff
@@ -0,0 +1,501 @@
...
+from ansible.module_utils.basic import *
+from ansible.module_utils.urls import *
```
mgedmin: These should appear at the bottom.  One reason for that is to maintain useful line number information in tracebacks (see the above comment for a description of the textual preprocessing taking place; if the contents of those module utils get included here, all the line numbers below will shift).

This is mentioned in the [Ansible module checklist](http://docs.ansible.com/ansible/developing_modules.html#module-checklist) BTW.

packaging/os/yumrepo.py
```diff
@@ -0,0 +1,501 @@
...
+    required: false
+    default: null
+    description:
+      - List of packages to exclude from updates or installs. This should be a
+        space separated list. Shell globs using wildcards (eg. C(*) and C(?))
+        are allowed.
+  failovermethod:
+    required: false
+    choices: [roundrobin, priority]
+    default: roundrobin
```
mgedmin, line 88: Stylstic issue: choices/default are shown as quoted strings elsewhere, but not here.

packaging/os/yumrepo.py
```diff
@@ -0,0 +1,501 @@
...
+        downloads that Yum does.
+      - C(all) means that all HTTP downloads should be cached.
+      - C(packages) means that only RPM package downloads should be cached (but
+         not repository metadata downloads).
+      - C(none) means that no HTTP downloads should be cached.
+  includepkgs:
+    required: false
+    default: null
+    description:
+      - .
```
mgedmin: This seems incomplete.

packaging/os/yumrepo.py
```diff
@@ -0,0 +1,501 @@
...
+            self.module.fail_json(
+                msg='Paramater "baseurl" or "mirrorlist" is required for '
+                'adding a new repo.')
+
+        # Set options
+        for key in sorted(self.params.keys()):
+            value = self.params[key]
+
+            # Convert boolean value to integer
+            if type(value).__name__ == 'bool':
```
mgedmin: Please use `if isinstance(value, bool):`

packaging/os/yumrepo.py
```diff
@@ -0,0 +1,501 @@
...
+        # Compose the repo file
+        for section in sorted(self.repofile.sections()):
+            repo_string += "[%s]\n" % section
+
+            for key, value in self.repofile.items(section):
+                repo_string += "%s = %s\n" % (key, value)
+
+            repo_string += "\n"
+
+        return repo_string
```
mgedmin, line 466: It would be simpler to 

``` python
from StringIO import StringIO
...
    def dump(self):
        buf = StringIO()
        self.repofile.write(buf)
        return buf.getvalue()
```

mgedmin, line 466: On second thought I'm not sure that would sort the sections the right way, especially if you remove and add back the same section.  Hm.

On the other hand, you're not sorting the keys.

packaging/os/yumrepo.py
jtyr: Thanks for the explanation. I have removed this kind of configuration and introduced a new module parameter instead.

packaging/os/yumrepo.py
jtyr: I have moved these two imports right above the `main()` call.

packaging/os/yumrepo.py
jtyr, line 88: I have fixed the quotation to be more consistent. I'm quoting only the values which might be converted to boolean (e.g. yes, no).

packaging/os/yumrepo.py
jtyr: I have added the description.

packaging/os/yumrepo.py
jtyr: Fixed.

packaging/os/yumrepo.py
jtyr, line 466: I have added the key sorting.

larsks:
This playbook is incorrect w/r/t to what the module requires:

```
- hosts: localhost
  tasks:
    - yumrepo:
        name: testrepo
        baseurl: http://localhost/packages/
        exclude:
          - broken-package
          - anotherpackage
```

The above should either (a) work, by translating the YAML list into
the correct syntax for a yum repo configuration, or (b) it should
raise an error.

As it stands, it results in:

```
[testrepo]
baseurl = http://localhost/packages/
exclude = ['broken-package', 'anotherpackage']
```

I would vote for (a), so that it works naturally with YAML lists.

larsks:
Should this module also accept `file`-like parameters for setting group/owner/etc?

packaging/os/yumrepo.py
```diff
@@ -0,0 +1,502 @@
...
+      - Specifies a URL to a file containing a list of baseurls.
+  mirrorlist_expire:
+    required: false
+    default: 21600
+    description:
+      - Time (in seconds) after which the mirrorlist locally cached will
+        expire.
+      - Default value is 6 hours.
+  name:
+    required: true
```
larsks, line 172: A minor request, but maybe place any `required` options at the top of this list so they're easier to spot?

jtyr, line 172: According to the documentation of other (core) modules (e.g. group, template) the parameters are sorted alphabetically.

larsks, line 172: That's fine.  Like I said, minor, and I figured it was probably canonical.

jtyr:
@larsks Thanks for the review. The `exclude` list should be "space separated list" as written in the documentation. I'm not sure if there is any other Ansile module which would allow to use data structure as the value of a parameter.

larsks:
So you should go with choice (b) then, which was "raise an error", rather than string-ifying the list.

jtyr:
@larsks Good point with the file-like parameters!

larsks:
For what it's worth, there _are_ ansible modules that accept data structures as parameters (e.g., cloud/openstack/os_server.py).

jtyr:
@larsks I have added the support for all file params. I have also added the support for the YAML array in  the case of the `exclude` and the `includepkgs` parameter. Please review and comment with “works_for_me” and/or “passes_guidelines” if you think it works well.

larsks:
works_for_me

With the most recent changes, the following playbook does "what I expect":

```
- hosts: localhost
  tasks:
    - yumrepo:
        name: testrepo
        baseurl: http://localhost/packages/
        exclude:
          - broken-package
          - anotherpackage
        includepkgs: thisone tahtone
        gpgcheck: false
        owner: root
        group: wheel
```

robynbergeron:
Hi @jtyr -- thanks for your updates to this PR and your work with @larsks and @mgedmin on this. (Also, HI @larsks! Nice to see you here, used to seeing you in Fedora / RDO land :D)

It looks like we now have a works_for_me vote -- which means we now just need a passes_guidelines vote. Perhaps @larsks or @mgedmin are willing to do that -- or anyone else can as well.

Friendly reminder for anyone who may want to review this for guidelines adherence:

robynbergeron:
Also: More out of curiosity -- I am wondering if / how easily someone could use this as a model to do a similar module for dnf, to go along with dnf.py in extras. Thoughts? @abadger -- maybe you have some thoughts on this as well? It would be nice just from a parity standpoint for some those folks who make the move from yum to dnf on their systems at some point in the future. (I can totally file a feature request / idea issue for this, but figured I'd just ask the question to a captive audience -- answers to this question either way definitely won't block the inclusion of this module once it has the required votes :D)

larsks:
@robynbergeron This module should work fine with dnf as is.  DNF currently uses the same repository configuration (in the same location) as yum.

larsks:
needs_revision: Pretty much everything looks fine, except that I don't see documentation for:
- Return: document the return structure of the module

jtyr:
@larsks This module returns only the standard structure (`changed`, `name`, `state`) which I think is not worth documenting.

larsks:
That's fine.  This is my first time reviewing a module, and I don't have a sense how strict the community is w/r/t to those requirements.  I'll wait a bit, but if nobody else has comments I'll just declare it good by fiat later today :)

packaging/os/yumrepo.py
bcoca, line 172: parameters get sorted on display, so in the file do whatever is more readable for the maintainers

packaging/os/yumrepo.py
```diff
@@ -0,0 +1,529 @@
...
+        # the repo name. This is why we check if the baseurl/mirrorlist is
+        # defined.
+        if (self.params['baseurl'], self.params['mirrorlist']) == (None, None):
+            self.module.fail_json(
+                msg='Paramater "baseurl" or "mirrorlist" is required for '
+                'adding a new repo.')
+
+        # Set options
+        for key in sorted(self.params.keys()):
+            value = self.params[key]
```
mgedmin: Minor: you could simplify these two lines to

``` python
for key, value in sorted(self.params.items()):
```

packaging/os/yumrepo.py
```diff
@@ -0,0 +1,529 @@
...
+
+    # Get repo status after change
+    yumrepo_after = yumrepo.dump()
+
+    # Compare repo states
+    changed = yumrepo_before != yumrepo_after
+
+    # Save the file only if not in check mode and if there was a change
+    if not module.check_mode and changed:
+        yumrepo.save()
```
mgedmin, line 544: This could raise (because of file permissions/missing directories/out of disk space/whatever).  AFAIU the [module guidelines](http://docs.ansible.com/ansible/developing_modules.html#module-checklist) ask that you catch the IOError here and replace it with a `module.fail_json(...)` call.

mgedmin:
LGTM, except for the missing error handling.

(Grain of salt: I am not an experienced Ansible module reviewer either.)

packaging/os/yumrepo.py
```diff
@@ -0,0 +1,529 @@
...
+            self.module.fail_json(
+                msg='Repo directory "%s" does not exist.' % repos_dir)
+
+        # Get the given or the default repo file name
+        repo_file = self.params['repoid']
+        if self.params['file'] is not None:
+            repo_file = self.params['file']
+
+        # Set dest; also used to set dest parameter for the FS attributes
+        self.params['dest'] = "%s/%s.repo" % (repos_dir, repo_file)
```
mgedmin: Why not `os.path.join(repos_dir, repo_file)`?  Would avoid double slashes if the user specifies `repos_dir=/with/a/trailing/slash/`.

larsks:
Yay, another reviewer! I like @mgedmin's suggestions, particularly w/r/t catching ioerror exceptions raised by file operations.

packaging/os/yumrepo.py
jtyr: Fixed.

packaging/os/yumrepo.py
jtyr, line 544: I have added the exception handling into the `save()` method.

packaging/os/yumrepo.py
jtyr: Fixed.

jtyr:
@mgedmin Thank you very much for your comments. I have fixed all the issues in the latest update (see the diff bellow). Please review the changes and and let me know if it looks good to you.

@larsks @mgedmin I have replaced the `hidden_params` variable with the `allowed_params` variable which helps to make sure that only params from this list will be in the final repo file (see the diff bellow). Please test the module and comment with "works_for_me" and/or "passes_guidelines" if there are no further problems.

``` diff
diff --git a/packaging/os/yumrepo.py b/packaging/os/yumrepo.py
index cb14a90..13dbaad 100644
--- a/packaging/os/yumrepo.py
+++ b/packaging/os/yumrepo.py
@@ -336,16 +336,22 @@ class YumRepo(object):
     section = None
     repofile = ConfigParser.RawConfigParser()

+    # List of parameters which will be allowed in the repo file output
+    allowed_params = [
+        'bandwidth', 'baseurl', 'cost', 'enabled', 'enablegroups', 'exclude',
+        'failovermethod', 'gpgcakey', 'gpgcheck', 'gpgkey', 'http_caching',
+        'includepkgs', 'keepalive', 'metadata_expire', 'metalink',
+        'mirrorlist', 'mirrorlist_expire', 'name', 'password', 'protect',
+        'proxy', 'proxy_password', 'proxy_username', 'repo_gpgcheck',
+        'retries', 'skip_if_unavailable', 'sslcacert',
+        'ssl_check_cert_permissions', 'sslclientcert', 'sslclientkey',
+        'sslverify', 'throttle', 'timeout', 'username']
+
     # List of parameters which can be a list
     list_params = ['exclude', 'includepkgs']

-    # List of keys which will be excluded from the final output
-    hidden_params = [
-        'file', 'repoid', 'reposdir', 'state', 'dest', 'checkmode',
-        'mode', 'owner', 'group', 'seuser', 'serole', 'setype', 'selevel']
-
     def __init__(self, module):
-        # To be albe to use fail_json
+        # To be able to use fail_json
         self.module = module
         # Shortcut for the params
         self.params = self.module.params
@@ -364,7 +370,7 @@ class YumRepo(object):
             repo_file = self.params['file']

         # Set dest; also used to set dest parameter for the FS attributes
-        self.params['dest'] = "%s/%s.repo" % (repos_dir, repo_file)
+        self.params['dest'] = os.path.join(repos_dir, "%s.repo" % repo_file)

         # Read the repo file if it exists
         if os.path.isfile(self.params['dest']):
@@ -387,9 +393,7 @@ class YumRepo(object):
                 'adding a new repo.')

         # Set options
-        for key in sorted(self.params.keys()):
-            value = self.params[key]
-
+        for key, value in sorted(self.params.items()):
             if key in self.list_params and isinstance(value, list):
                 # Join items into one string for specific parameters
                 value = ' '.join(value)
@@ -398,18 +402,36 @@ class YumRepo(object):
                 value = int(value)

             # Set the value only if it was defined (default is None)
-            if value is not None and key not in self.hidden_params:
+            if value is not None and key in self.allowed_params:
                 self.repofile.set(self.section, key, value)

     def save(self):
         if len(self.repofile.sections()):
             # Write data into the file
-            fd = open(self.params['dest'], 'wb')
-            self.repofile.write(fd)
-            fd.close()
+            try:
+                fd = open(self.params['dest'], 'wb')
+            except IOError:
+                self.module.fail_json(
+                    msg='Can not open repo file %s.' %
+                    self.params['dest'])
+
+            try:
+                try:
+                    self.repofile.write(fd)
+                except Error:
+                    self.module.fail_json(
+                        msg='Can not write repo file %s.' %
+                        self.params['dest'])
+            finally:
+                fd.close()
         else:
             # Remove the file if there are not repos
-            os.remove(self.params['dest'])
+            try:
+                os.remove(self.params['dest'])
+            except OSError:
+                self.module.fail_json(
+                    msg='Can not remove empty repo file %s.' %
+                    self.params['dest'])

     def remove(self):
         # Remove section if exists
@@ -424,11 +443,7 @@ class YumRepo(object):
             repo_string += "[%s]\n" % section

             for key, value in sorted(self.repofile.items(section)):
-                # This is here to prevent the checkmod param to appear in the
-                # dump output (Dunno how it appears here as it's filtered out
-                # in the add() method. Please help!).
-                if key not in self.hidden_params:
-                    repo_string += "%s = %s\n" % (key, value)
+                repo_string += "%s = %s\n" % (key, value)

             repo_string += "\n"

```

packaging/os/yumrepo.py
```diff
@@ -0,0 +1,547 @@
...
+                self.repofile.set(self.section, key, value)
+
+    def save(self):
+        if len(self.repofile.sections()):
+            # Write data into the file
+            try:
+                fd = open(self.params['dest'], 'wb')
+            except IOError:
+                self.module.fail_json(
+                    msg='Can not open repo file %s.' %
```
mgedmin: Typo: should be "Cannot" (or "Can't").

packaging/os/yumrepo.py
```diff
@@ -0,0 +1,547 @@
...
+                    self.params['dest'])
+
+            try:
+                try:
+                    self.repofile.write(fd)
+                except Error:
+                    self.module.fail_json(
+                        msg='Can not write repo file %s.' %
```
mgedmin: Same here.

packaging/os/yumrepo.py
```diff
@@ -0,0 +1,547 @@
...
+                        self.params['dest'])
+            finally:
+                fd.close()
+        else:
+            # Remove the file if there are not repos
+            try:
+                os.remove(self.params['dest'])
+            except OSError:
+                self.module.fail_json(
+                    msg='Can not remove empty repo file %s.' %
```
mgedmin: Same here.

packaging/os/yumrepo.py
jtyr: Fixed.

packaging/os/yumrepo.py
jtyr: Fixed.

packaging/os/yumrepo.py
jtyr: Fixed.

robynbergeron:
Hi @jtyr -- thanks for your quick revisions! :)

@mgedmin & @larsks  -- if you can take a look at this and let us know how these revisions work, that would be awesome.

I'd also like to take a moment to add that even though both of you are not "experienced" module reviewers, you're doing great and doing exactly what we believed community members would do when we moved to the new module review guidelines for extras: make reviews for quality, in good faith. 

You're both clearly paying attention to the guidelines, and making sure it works and is a good module, and that kind of care is largely what ensures that things that go in will be good :) 

Just to be clear about where we are at this point: we need a passes_guidelines vote, and a works_for_me vote; @larsks, you gave a works_for_me vote previously, but I suspect the module has changed enough since that point that giving it another test for functionality would be a good idea.

Thanks, everyone!

mgedmin:
passes_guidelines

I haven't actually gone through all of them _this time_ -- but last time I did, and the missing error handling was the only thing that stood out.  Well, error handling and the lack of description of the module return value, [mentioned earlier](https://github.com/ansible/ansible-modules-extras/pull/1016#issuecomment-147401562).

larsks:
passes_guidelines
works_for_me

The module continues to work as expected, and now provides a better error message if file permissions prevent an operation from succeeding.

jtyr:
@robynbergeron What about to celebrate [the great news](http://www.ansible.com/blog/red-hat) by merging this great module? ;o)

jtyr:
@robynbergeron @gregdek Any chance this PR gets some attention any time soon?

gregdek:
Thanks @jtyr -- marking for inclusion.

jtyr:
@robynbergeron @gregdek It has been more than 3 weeks since this PR was marked as "shipit" and it still wasn't merged. Is there any other step which needs to be done before it gets merged?

bcoca:
Sorry for the delay, but we had stopped merging new modules until we cut stable-2.0 branch and were concentrating on bugs to launch 2.0.

for 2.1 and above we are requiring return docs to be documented, which this module does not have.

jtyr:
@bcoca Thanks for the information. It's pity that the module won't make it into the v2.0 :o( I have added the missing `RETURN` documentation. Please review it.

bcoca:
sorry about the exclusion, but we had to cut off somewhere and missed this by a hair. On the bright side 2.1 will not take as long as 2.0, it is just an incremental release and not a core rewrite, I don't think anyone wants to do another any time soon.

jtyr:
@bcoca I understand. Would it be possible to cherry-picked this commit and shipped with v2.0.1?

PR #1017, opened 2015-09-22
Adding the --sudo flag for CPANM use since the sudo: yes on the task …

milad-soufastai:
Add the ability to use --sudo flag with the CPANM module.

### DIFF
--- a/packaging/language/cpanm.py
+++ b/packaging/language/cpanm.py
@@ -64,6 +64,12 @@ options:
     required: false
     default: false
     version_added: "2.0"
+  use_sudo:
+    description:
+     - Use sudo flag for cpanm
+    required: false
+    default: false
+    version_added: "2.0"
 notes:
    - Please note that U(http://search.cpan.org/dist/App-cpanminus/bin/cpanm, cpanm) must be installed on the remote host.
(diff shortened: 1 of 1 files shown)


bcoca:
'sudo: yes on the task does not work' <= exactly how is that not working?

sudo on the task should run the whole command under sudo, not add the --sudo flag to the underlying command. The result should be the same but the intermediate build steps would also be run by the sudo user, not just the install steps.

milad-soufastai:
Hi Brian,
Per CPANM: -S, --sudo
Switch to the root user with sudo when installing modules. Use this if you want to install modules to the system perl include path.

Currently using the ansible cpanm module with or without sudo: yes doesn't actually install the modules in system perl include path but under the user path, usually ~/perl5

bcoca:
we should add a explanation as use_sudo is really 'install in system path' in this case, the sudo at the task level does run it as root. Also that use_sudo will not benefit from ansible handling of sudo password prompts.

milad-soufastai:
@bcoca I've updated the description to match closer what the flag does. I'm not sure about the language around the ansible handling of sudo password.

bcoca:
i changed the option name but kept use_sudo as an alias and expanded the description a bit

PR #767, opened 2015-07-26
Additional upgrade mode fixes for pacman module with check_mode safety

indrajitr:
- ensure upgrade option honors and actually supports `check_mode`
- enabling just `upgrade` and `update_cache` should perform upgrade
- example added for the equivalent for `pacman -Syu`

NB: This is a followup of #762

PR #1018, opened 2015-09-23
Additional upgrade mode fixes for pacman module with check_mode safety

indrajitr:
- ensure upgrade option honors and actually supports `check_mode`
- enabling just `upgrade` and `update_cache` should perform upgrade
- example added for the equivalent for `pacman -Syu`

NB: This is a followup of #762, and replacement of #767


robynbergeron:
@indrajitr -- thanks for the PR. 

Marking as owner_pr. Also marking as shipit for inclusion. cc @ansible/core team to make that magic happen. :)

PR #1020, opened 2015-09-23
Added module for SoftLayer

mcltn:
This module allows the functionality of building and canceling SoftLayer virtual instances. This module requires the SoftLayer library to work. The parameters allow you to build a single instance as well as specify a quantity of instances to build in addition to the various options for the virtual instances themselves.

### DIFF
--- a/cloud/softlayer/sl.py
+++ b/cloud/softlayer/sl.py
@@ -1,0 +1,348 @@
+#!/usr/bin/python
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: SoftLayer
+short_description: create or cancel a virtual instance in SoftLayer
+description:
+  - Creates or cancels SoftLayer instances. When created, optionally waits for it to be 'running'.
+version_added: "2.1"
+options:
+  instance_id:
+    description:
+      - Instance Id of the virtual instance to perform action option
+    required: false
+    default: null
+  hostname:
+    description:
+      - Hostname to be provided to a virtual instance
+    required: false
+    default: null
+  domain:
+    description:
+      - Domain name to be provided to a virtual instance
+    required: false
+    default: null
+  datacenter:
+    description:
+      - Datacenter for the virtual instance to be deployed
+    required: false
+    default: null
+  tags:
+    description:
+      - Tag or list of tags to be provided to a virtual instance
+    required: false
+    default: null
+  hourly:
+    description:
+      - Flag to determine if the instance should be hourly billed
+    required: false
+    default: true
+  private:
+    description:
+      - Flag to determine if the instance should be private only
+    required: false
+    default: false
+  dedicated:
+    description:
+      - Falg to determine if the instance should be deployed in dedicated space
+    required: false
+    default: false
+  local_disk:
+    description:
+      - Flag to determine if local disk should be used for the new instance
+    required: false
+    default: true
+  cpus:
+    description:
+      - Count of cpus to be assigned to new virtual instance
+    required: true
+    default: null
+  memory:
+    description:
+      - Amount of memory to be assigned to new virtual instance
+    required: true
+    default: null
+  disks:
+    description:
+      - List of disk sizes to be assigned to new virtual instance
+    required: true
+    default: [25]
+  os_code:
+    description:
+      - OS Code to be used for new virtual instance
+    required: false
+    default: null
+  image_id:
+    description:
+      - Image Template to be used for new virtual instance
+    required: false
+    default: null
+  nic_speed:
+    description:
+      - NIC Speed to be assigned to new virtual instance
+    required: false
+    default: 10
+  public_vlan:
+    description:
+      - VLAN by its Id to be assigned to the public NIC
+    required: false
+    default: null
+  private_vlan:
+    description:
+      - VLAN by its Id to be assigned to the private NIC
+    required: false
+    default: null
+  ssh_keys:
+    description:
+      - List of ssh keys by their Id to be assigned to a virtual instance
+    required: false
+    default: null
+  post_uri:
+    description:
+      - URL of a post provisioning script ot be loaded and exectued on virtual instance
+    required: false
+    default: null
+  state:
+    description:
+      - Create, or cancel a virtual instance. Specify "present" for create, "absent" to cancel.
+    required: false
+    default: 'present'
+  wait:
+    description:
+      - Flag used to wait for active status before returning
+    required: false
+    default: true
+  wait_timeout:
+    description:
+      - time in seconds before wait returns
+    required: false
+    default: 600
+
+requirements:
+    - "python >= 2.6"
+    - "softlayer >= 4.1.1"
+author: "Matt Colton (@mcltn)"
+'''
+
+EXAMPLES = '''
+- name: Build instance
+  hosts: localhost
+  gather_facts: False
+  tasks:
+  - name: Build instance request
+    local_action:
+      module: sl
+      hostname: instance-1
+      domain: anydomain.com
+      datacenter: dal09
+      tags: ansible-module-test
+      hourly: True
+      private: False
+      dedicated: False
+      local_disk: True
+      cpus: 1
+      memory: 1024
+      disks: [25]
+      os_code: UBUNTU_LATEST
+      wait: False
+
+- name: Build additional instances
+  hosts: localhost
+  gather_facts: False
+  tasks:
+  - name: Build instances request
+    local_action:
+      module: sl
+      hostname: "{{ item.hostname }}"
+      domain: "{{ item.domain }}"
+      datacenter: "{{ item.datacenter }}"
+      tags: "{{ item.tags }}"
+      hourly: "{{ item.hourly }}"
+      private: "{{ item.private }}"
+      dedicated: "{{ item.dedicated }}"
+      local_disk: "{{ item.local_disk }}"
+      cpus: "{{ item.cpus }}"
+      memory: "{{ item.memory }}"
+      disks: "{{ item.disks }}"
+      os_code: "{{ item.os_code }}"
+      ssh_keys: "{{ item.ssh_keys }}"
+      wait: "{{ item.wait }}"
+    with_items:
+      - { hostname: 'instance-2', domain: 'anydomain.com', datacenter: 'dal09', tags: ['ansible-module-test', 'ansible-module-test-slaves'], hourly: True, private: False, dedicated: False, local_disk: True, cpus: 1, memory: 1024, disks: [25,100], os_code: 'UBUNTU_LATEST', ssh_keys: [], wait: True }
+      - { hostname: 'instance-3', domain: 'anydomain.com', datacenter: 'dal09', tags: ['ansible-module-test', 'ansible-module-test-slaves'], hourly: True, private: False, dedicated: False, local_disk: True, cpus: 1, memory: 1024, disks: [25,100], os_code: 'UBUNTU_LATEST', ssh_keys: [], wait: True }
+
+
+- name: Cancel instances
+  hosts: localhost
+  gather_facts: False
+  tasks:
+  - name: Cancel by tag
+    local_action:
+      module: sl
+      state: absent
+      tags: ansible-module-test
+'''
+
+# TODO: Disabled RETURN as it is breaking the build for docs. Needs to be fixed.
+RETURN = '''# '''
+
+import time
+
+STATES = ['present', 'absent']
+DATACENTERS = ['ams01','ams03','dal01','dal05','dal06','dal09','fra02','hkg02','hou02','lon2','mel01','mex01','mil01','mon01','par01','sjc01','sjc03','sao01','sea01','sng01','syd01','tok02','tor01','wdc01','wdc04']
+CPU_SIZES = [1,2,4,8,16]
+MEMORY_SIZES = [1024,2048,4096,6144,8192,12288,16384,32768,49152,65536]
+INITIALDISK_SIZES = [25,100]
+LOCALDISK_SIZES = [25,100,150,200,300]
+SANDISK_SIZES = [10,20,25,30,40,50,75,100,125,150,175,200,250,300,350,400,500,750,1000,1500,2000]
+NIC_SPEEDS = [10,100,1000]
+
+try:
+  import SoftLayer
+  from SoftLayer import VSManager
+
+  HAS_SL = True
+  vsManager = VSManager(SoftLayer.create_client_from_env())
+except ImportError:
+  HAS_SL = False
+
+
+def create_virtual_instance(module):
+
+  # Check if OS or Image Template is provided (Can't be both, defaults to OS)
+  if (module.params.get('os_code') != None and module.params.get('os_code') != ''):
+    module.params['image_id'] = ''
+  elif (module.params.get('image_id') != None and module.params.get('image_id') != ''):
+    module.params['os_code'] = ''
+    module.params['disks'] = [] # Blank out disks since it will use the template
+  else:
+    return False, None
+
+  tags = module.params.get('tags')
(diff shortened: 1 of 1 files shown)


robynbergeron:
Hi @mcltn --

sp4ceb4r:
"passes_guidelines"

Side Note: While PEP8 is not required the use of spaces instead of tabs would make any potential Pull Requests easier as spaces seem to be the more default option.

bcoca:
while we don't force PEP8, tabs as indentation do create issues, so please convert those to spaces.

mcltn:
All of the tabs contained in the code have been converted to spaces.

coreylane:
works_for_me

cloud/softlayer/softlayer.py
```diff
@@ -0,0 +1,428 @@
...
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: SoftLayer
+short_description: create or cancel a virtual instance in SoftLayer
+description:
+  - Creates or cancels SoftLayer instances. When created, optionally waits for it to be 'running'.
+options:
+  hostname:
```
bcoca: the docs need to be valid YAML, the indentation is off in many places

mcltn:
Issues with spacing within documentation yaml has been corrected.

cloud/softlayer/softlayer.py
```diff
@@ -0,0 +1,428 @@
...
+  tasks:
+  - name: Cancel by tag
+    local_action:
+      module: softlayer
+      state: absent
+      tag: ansible-module-test
+'''
+
+import time
+from ansible.module_utils.facts import *
```
bcoca: not sure why you have facts imported here

cloud/softlayer/softlayer.py
```diff
@@ -0,0 +1,428 @@
...
+      (changed, instance) = create_virtual_instance(module)
+      instances.append(instance)
+      counter += 1
+
+  return True, instances
+
+
+def create_virtual_instance(module):
+
+  hostname = module.params.get('hostname')
```
bcoca: most of these are just then reassigned down below, wouldn't it just be simpler to directly use the module.params?

bcoca:
also we would appreciate it if you could flatten this into a single commit

mcltn:
Flattened to a single commit.

kbroughton:
Looks good.  My one nit would be that softlayer.py is the name of the dynamic inventory script as well.
Perhaps following the lead of the ec2 modules we could call the sl creation
sl.py, thus allowing sl_tag, sl_volume etc to come in the future.  Names will become long if all modules have to be pre-pended by softlayer in full.  ec2 modules for reference: http://docs.ansible.com/ansible/list_of_cloud_modules.html

mcltn:
yep, been thinking about that as well. Going to rename to your suggestion.

kbroughton:
Testing now.  I'm seeing something odd.  Passing in an int for cpus works fine, but passing a variable that resolves to an int fails.  Could you try passing in a var for cpus and see if you can reproduce?

failed: [localhost -> 127.0.0.1] => (item={'count': 3, 'datacenter': 'dal06', 'hostname': u'worker', 'disks': [25, 100], 'tags': ['mesos_slave', 'hdfs_datanode', 'consul_client'], 'disk_info': [{'device': 'xvda', 'size': 25}, {'device': 'xvdc', 'size': 100}, {'device': 'xvdd', 'size': 200}], 'cpus': 1, 'local_disk': False}) => {"failed": true, }}
msg: value of cpus must be one of: 1,2,4,8,16, got: 1

as you can see, item.cpus is resolving to an int (no quotes around the 1 in 'cpus': 1)
  cpus: "{{item.cpus }}"        - fails when item.cpus = 1
  cpus: 1                              - works fine

mcltn:
This throws the error due to the "choices" option in the parameter. Only certain cpu sizes are allowed and the array is of ints.

kbroughton:
Docs don't specify if ssh_keys: var takes a keyname or id.  Name doesn't work, I have to use the key id found by
slcli sshkey list
:........:.............................:.................................................:.......:
:   id   :            label            :                   fingerprint                   : notes :
:........:.............................:.................................................:.......:
: 10623  :         dummy        : a3*****:b3 :   -   :

It would be nice to do id lookup of ssh_key names but i wouldn't hold up this pr over it.  Just clarify in the docs.

mcltn:
Specified in the docs that VLAN and SSH Keys are by Id.

kbroughton:
@mcltn 
Still trying to figure out why i can't pass variables to cpus: and memory:

Here is my debug using a jinja2 test to make sure it is a number
- debug: msg="{{item.cpus }} {{item.cpus is number}}"
  with_items: sl_requested_hosts

where 
sl_requested_hosts:
  cpus: 1

returns  
}
    "msg": "1 True"
}

So it seems I am passing in an int and it is in the list of allowed values.
Other modules take variables no problem.  Am i missing something?

mcltn:
looks like adding type='int' to the params fixes the issue. Adding to both cpus and memory.

kbroughton:
for consistency with ec2 module, could you move the wait logic to its own top-level arg rather than putting it into state?  I wasn't clear if state = wait would create, then wait, or just wait assuming it was already created.

ec2 docs:
parameter           required            default             choices              comments
wait                    no              no                    yes, no              wait for the instance to be 'running'  
                                                                                                     before returning. Does not wait for 
                                                                                                     SSH, see 'wait_for' example for details.
wait_timeout
                             no                 300                                         how long before wait gives up, in 
                                                                                                    seconds

mcltn:
I'm hesitant to do it this way since some users, as well as myself, would want to potentially run provisioning in parallel, but I guess it could place an order in the first play, say for web heads, then when running a second play, say for a db, this step could have a wait flag. It would still require a tag or something to use to know what instances to wait for. Other thoughts?

kbroughton:
@mcltn I'll have to check but I think the way "wait" works with ec2 is that all jobs are launched in parallel, but all jobs must complete before moving on to the next task.  I see no use case to waiting for each to complete serially.  You would use the "serial" keyword for that i guess.

kbroughton:
when I add a list of tags rather than a single tag I get "array" as the name of the tag rather than a list of the tags.

tag: [ tag1, tag2, tag3 ]

Output: 

```
"status": {
      "keyName": "ACTIVE", 
      "name": "Active"
    }, 
    "tagReferences": [
      {
        "empRecordId": "", 
        "id": 9756287, 
        "resourceTableId": 13964399, 
        "tag": {
          "name": "array"
        }, 
        "tagId": 333695, 
```

I even tried changing to
      tag=dict(type='list'),
from tag=dict()
in
def main():
  module = AnsibleModule(

mcltn:
The SoftLayer python library takes tags in a string with tags being separated by commas within the string. This was setup to work the same. Probably just need to update the docs to reflect the proper use. Also looking into doing some refactoring for the wait suggestion.

kbroughton:
cool.  I discovered this last night.  I can submit a PR that would allow
the module to ingest either comma delimited string or an ansible list.
 ec2.py accepts an ansible list, so again this is for consistency.  And
again for consistency with ec2, maybe you could make tags an alias for tag.

tag=dict(aliases=['tags'])

On Fri, Nov 20, 2015 at 6:30 AM, matt colton notifications@github.com
wrote:

> The SoftLayer python library takes tags in a string with tags being
> separated by commas within the string. This was setup to work the same.
> Probably just need to update the docs to reflect the proper use. Also
> looking into doing some refactoring for the wait suggestion.
> 
> —
> https://github.com/ansible/ansible-modules-extras/pull/1020#issuecomment-158385188
> .

## 

kesten broughton
devops @ cognitivescale

mcltn:
Made modifications to use wait logic within the creation of virtual machines. Optional parameters wait and wait_timeout, timeout defaults to 600 and wait is True by default. I also modified the tag param to tags since this was what the Softlayer python library is using also.

mcltn:
Just noticed an issue with wait and multiple tags. Should it just default to the first tag in a list for using for the wait? Thoughts?

kbroughton:
Regarding tags, I think exact match is most logical.  You could have an "any_tag" option which would switch between the two.  
But i would be very careful about cancelling with tags instead of instance ids.  Right now when I call list_instances(tags="tag1,tag2") i get returned instances that have neither tag1 nor tag2.  I'm running more tests, but it seems like searching multiply tagged instances with multiple search tags is broken.
Until we do thorough testing, i think it is unsafe to allow cancel by tag.

mcltn:
So it appears that the vsManager.list_instances() when using tags takes a list. I'm converting it to use it as a list, which is actually how it previously was when wait was a separate state, but was encapsulating the string into [] to make an array. Cancel was done the same way.

mcltn:
Also, I noticed the needs revision label had been removed earlier on some fixes/changes, but can you go ahead and put one back until we agree it is ready?

mcltn:
Latest has updates for proper use of tags per SoftLayer python library and VS Manager. Updated example section to reflect use of tags as well, between string and list use.

code-noggin:
You have an option to specify the private_vlan but do not have one for public_vlan. Would you mind including a public_vlan option same as private_vlan?

robynbergeron:
@mcltn -- sorry, just noticed the request to add the needs revision label. Looks like you've been doing some work since that point though -- is it now ready for review?

mcltn:
No worries Robyn, I do want to make one more edit to the code around the use of the wait parameters. It needs to only wait for the instances in which it was creating vs using the tag specified. This should be updated and tested soon.

mcltn:
Updates completed to "wait" logic to only look at instances that have been created in current process. Using "wait" will provide IP details. Should possibly also add ability to only wait for IP, since IP is not given on return of instances initially and when not waiting.

mcltn:
All changes have been completed and tested.

gregdek:
@mcltn thanks for all your work here.

@kbroughton thanks for your extensive feedback. Since we've had +1s from other folks, I'd like to wait for your feedback. If you think it's ready to go, give us a "shipit".

gregdek:
Actually, it's time to make a judgement call here. This has a lot of +1s, and the maintainer is clearly responsive, even if there may still be minor issues. Therefore, we're going to mark this as "shipit". Thanks @mcltn for all the hard work, and thanks in advance for maintaining this awesome module. :)

farioas:
Status?

mcltn:
No changes are being made currently or anticipated just yet. Just waiting to get pulled into the main repo. Thanks @gregdek for helping get this pushed along. Looking forward to the add, and next features and enhancements to be added to this module.

cloud/softlayer/sl.py
```diff
@@ -0,0 +1,422 @@
...
+    description:
+      - Hostname of the virtual instanceId
+    required: true
+    default: null
+  instance_id:
+    description:
+      - Instance Id of the virtual instance to perform action option
+    required: false
+    default: null
+  hostname:
```
bcoca, line 30: hostname is duplicated

cloud/softlayer/sl.py
```diff
@@ -0,0 +1,422 @@
...
+      - List of ssh keys by their Id to be assigned to a virtual instance
+    required: false
+    default: null
+  post_uri:
+    description:
+      - URL of a post provisioning script ot be loaded and exectued on virtual instance
+    required: false
+    default: null
+  state:
+    description:
```
bcoca, line 121: this should document the choices 'present' and 'absent', also they should be enforced in the module spec in the code.

mcltn:
Updated documentation and added the version_added requirement. Also added choices for the present and absent to the state param. CI build started failing on this commit though due to no return being specified, which was new, so added an empty return as well.

mcltn:
Thanks for pointing that out @farioas . Now updated and tested with the newer way to setup the SoftLayer client.

gregdek:
Putting back into shipit. Thanks @mcltn!

cloud/softlayer/sl.py
```diff
@@ -0,0 +1,430 @@
...
+      - Create, or cancel a virtual instance. Specify "present" for create, "absent" to cancel.
+    required: false
+    default: 'present'
+  quantity:
+    description:
+      - Quantity of virtual instances to provision
+    required: false
+    default: 0
+  increment:
```
bcoca: this is a bad pattern that crept into other modules, ansible already allows the use of sequences and loops to take care of this

cloud/softlayer/sl.py
```diff
@@ -0,0 +1,430 @@
...
+
+STATES = ['present', 'absent']
+DATACENTERS = ['ams01','ams03','dal01','dal05','dal06','dal09','fra02','hkg02','hou02','lon2','mel01','mex01','mil01','mon01','par01','sjc01','sjc03','sao01','sea01','sng01','syd01','tok02','tor01','wdc01','wdc04']
+CPU_SIZES = [1,2,4,8,16]
+MEMORY_SIZES = [1024,2048,4096,6144,8192,12288,16384,32768,49152,65536]
+INITIALDISK_SIZES = [25,100]
+LOCALDISK_SIZES = [25,100,150,200,300]
+SANDISK_SIZES = [10,20,25,30,40,50,75,100,125,150,175,200,250,300,350,400,500,750,1000,1500,2000]
+NIC_SPEEDS = [10,100,1000]
+
```
bcoca, line 214: is there no way to query the softlayer api for this info? it seems that it will need to be updated anytime the options change

cloud/softlayer/sl.py
```diff
@@ -0,0 +1,430 @@
...
+    required: false
+    default: true
+  wait_timeout:
+    description:
+      - time in seconds before wait returns
+    required: false
+    default: 600
+
+requirements:
+    - "python >= 2.6"
```
bcoca, line 137: wouldn't the softlayer python lib also be required?

gregdek:
@mcltn A friendly reminder: this pull request has been marked as needing your action. If you still believe that this PR applies, and you intend to address the issues with this PR, just let us know in the PR itself and we will keep it open pending your changes.

mcltn:
Planning on making some adjustments, may need some time. Should have updates coming soon. Thanks!

gregdek:
@mcltn A friendly reminder: this pull request has been marked as needing your action. If you still believe that this PR applies, and you intend to address the issues with this PR, just let us know in the PR itself and we will keep it open pending your changes.

kbroughton:
if @mcltn doesn't respond by end of week, i can try responding to the
action items on the ticket rather than closing it.

On Thu, Mar 3, 2016 at 9:14 AM, Greg DeKoenigsberg <notifications@github.com

> wrote:
> 
> @mcltn https://github.com/mcltn A friendly reminder: this pull request
> has been marked as needing your action. If you still believe that this PR
> applies, and you intend to address the issues with this PR, just let us
> know in the PR itself and we will keep it open pending your changes.
> 
> —
> https://github.com/ansible/ansible-modules-extras/pull/1020#issuecomment-191808185
> .

## 

kesten broughton
devops @ cognitivescale

cloud/softlayer/sl.py
mcltn: Making edits to remove this functionality within the module. Makes better sense to utilize it outside.

cloud/softlayer/sl.py
mcltn, line 214: These could possibly be pulled via Api calls, but are also pretty static. Only one that may change in near future would be data centers.

cloud/softlayer/sl.py
mcltn, line 137: Will add this on next commit.

gregdek:
@mcltn A friendly reminder: this pull request has been marked as needing your action. If you still believe that this PR applies, and you intend to address the issues with this PR, just let us know in the PR itself and we will keep it open pending your changes.

mcltn:
Removed the ability to specify quantity. The example documentation shows how to perform multiple provisions using the loop feature with_items. I'd still like to see a better way to handle the wait flag, since now this would wait for each instance in serial as they build for ones that are flagged. This could cause quite a delay when building a large volume of servers.

gregdek:
Thanks @mcltn. Presuming that you still intend to handle diffs from @bcoca above? Leaving in needs_revision for now, but if you feel that it's ready, reply with ready_for_review and we'll put it back into community review.

mcltn:
ready_for_review

JulienChampseix:
any update ? @bcoca @gregdek

gregdek:
Thanks @mcltn for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

darKoram:
@mcltn A little late, but shipit!  Sorry, missed your initial request.

rickji:
Question: where do I put in the SoftLayer username and APIkey for authentication?

Blake-:
Try putting it in ~/.softlayer
Takes the following form, but with real username & key. Also I think the endpoint may differ a bit depending on what you do?

```
[softlayer]
username = SL99999
api_key = 80924t2h4ti2h4t2jh4tk2j4thj2kht4j2kht
endpoint_url = https://api.softlayer.com/xmlrpc/v3.1/
```

PR #1022, opened 2015-09-23
win_chocolatey bugfixes

trondhindenes:
Cleaned up the code a bit (using Get-Attr more extensively), removed $erroractionpreference.


trondhindenes:
The bugs I was getting probably relates to strictmode (https://github.com/ansible/ansible/issues/12497) but still I feel that erroractionpreference shouldnt be used without the necessary try/catch handlers. We need to ensure that ansible will always return a result object.

robynbergeron:
Marking for inclusion as owner_pr, via the magical shipit label. cc: @ansible/core team :)

Thanks, @trondhindenes for the PR.

PR #1026, opened 2015-09-25
First cut of gsettings module

ag-wood:
A module to manage the desktop settings using a python gsettings equivalent.


system/gsettings.py
```diff
@@ -0,0 +1,225 @@
...
+---
+module: gsettings
+short_description: Module to manage dconf settings.
+description:
+    - This module provides a python equivalent of the gsettings command-line tool
+    - for use in ansible.
+version_added: "1.9"
+notes:
+    - become and become_user must be used to specify the target configuration set.
+requirements: [ "PyGObjects"]
```
mgedmin: What's that?  There's a library called PyGObject (singular).

system/gsettings.py
```diff
@@ -0,0 +1,225 @@
...
+import time
+import signal
+import subprocess
+
+try:
+    from gi.repository import Gio, GLib
+except ImportError:
+    HAS_GIO=False
+else:
+    HAS_GIO=True
```
mgedmin: PEP-8 says you should put spaces around the `=`

system/gsettings.py
```diff
@@ -0,0 +1,225 @@
...
+
+UID  = 1
```
mgedmin: This doesn't seem to be used anywhere.

system/gsettings.py
```diff
@@ -0,0 +1,225 @@
...
+ENV_DBUS = 'DBUS_SESSION_BUS_ADDRESS'
+SESSION_MANAGERS = [
+    'gnome-session', 'mate-session', 'xfce4-session',
+    'cinnamon-session','icewm-session', 'openbox-session' ]
+    
+change_msgs = []
```
mgedmin: Why is this global?  It could very well be a local variable inside `main()`.

system/gsettings.py
```diff
@@ -0,0 +1,225 @@
...
+
+class Gsettings(object):
+    def __init__(self):
+        def proc_name(pid):
+            '''Return process name of process pid'''
```
mgedmin: Docstrings in Python are usually enclosed in tripple double-quotes: `"""..."""`.  Using single quotes feels weird.

system/gsettings.py
```diff
@@ -0,0 +1,225 @@
...
+            try:
+                for attr in open("/proc/{0}/status".format(pid),'r'):
+                    if 'Name' in attr:
```
mgedmin: Maybe better check `if attr.startswith('Name:'):`?

system/gsettings.py
```diff
@@ -0,0 +1,225 @@
...
+            '''Return username of UID of process pid'''
+            for ln in open('/proc/{0}/status'.format(pid),'r'):
+                if ln.startswith('Uid:'):
+                    uid = int(ln.split()[UID])
+                    return pwd.getpwuid(uid).pw_name
+
+        def proc_environ(pid, varname=""):
+            '''Return process environment variable of process pid'''
+            proc_env = open("/proc/{0}/environ".format(pid),'r')
+            env = proc_env.readline()
```
mgedmin: Better use `.read()` here, environment variables might contain embedded newlines.

system/gsettings.py
```diff
@@ -0,0 +1,225 @@
...
+            for env_var in env.split('\x00'):
+                if varname != "":
+                    prefix = "{0}=".format(varname)
+                    if env_var.startswith(prefix):
+                        return env_var.replace(prefix, '')
```
mgedmin: Use `env_var[len(prefix):]` instead of `.replace()`.

system/gsettings.py
```diff
@@ -0,0 +1,225 @@
...
+        def bus_address(pid):
+            '''Return the bus address for the specified process'''
+            return proc_environ(pid, ENV_DBUS)
+
+        def new_session():
+            '''Create a new dbus session'''
+            dbus_process = subprocess.Popen('dbus-launch', stdout=subprocess.PIPE)
+            for env_string in dbus_process.stdout.readlines():
+                env_var = env_string[0:env_string.index('=')]
+                env_value = env_string[env_string.index('=')+1:]
```
mgedmin: Simpler/prettier code:

``` python
env_var, env_value = env_string.split('=', 1)
```

system/gsettings.py
```diff
@@ -0,0 +1,225 @@
...
+        self.session_pid = -1
+        self._cleanup = False
+
+        for procfile in os.listdir('/proc'):
+            if procfile.isdigit():
+                if proc_name(procfile) in SESSION_MANAGERS:
+                    self.session_pid = procfile
+
+        osuser = pwd.getpwuid(os.getuid()).pw_name
+        if self.session_pid <> -1:
```
mgedmin: `<>` is archaic and deprecated (and altogether removed in Python 3), please use `!=`.

system/gsettings.py
```diff
@@ -0,0 +1,225 @@
...
+            sessionuser = proc_owner(self.session_pid)
+        else:
+            sessionuser = osuser
+
+        # Setup the environment and connect to the bus.
+        if self.session_pid <> -1 and osuser == sessionuser:
```
mgedmin: Why convert UIDs to usernames if all you do with them is compare with each other?  Simpler to keep and compare the UIDs.

system/gsettings.py
```diff
@@ -0,0 +1,225 @@
...
+            self.session_pid = new_session()
+            self._cleanup = True
+
+        self.bus = Gio.bus_get_sync(Gio.BusType.SESSION, None)
+
+    def get_value(self, schema, key):
+        ''' Read the value of the specified key '''
+        if ':' in schema:
+            path = schema.split(':')[1]
+            schema = schema.split(':')[0]
```
mgedmin: You could do this on one line:

``` python
schema, path = schema.split(':', 1)
```

system/gsettings.py
```diff
@@ -0,0 +1,225 @@
...
+            path = None
+        self.settings = Gio.Settings(schema=schema, path=path)
+        self.settings.set_value(key, value)
+
+    def __del__(self):
+        ''' Class cleanup '''
+        if self._cleanup:
+           # Stop the dbus daemon if one was started.
+           time.sleep(1)
+           os.kill(int(self.session_pid), signal.SIGTERM)
```
mgedmin: `__del__` should be avoided if possible. If the new session should always be terminated, better install an [atexit](https://docs.python.org/2/library/atexit.html) handler or something.

system/gsettings.py
```diff
@@ -0,0 +1,225 @@
...
+    if not HAS_GIO:
+        module.fail_json(rc=1, msg='The python PyGObject module is a required dependency.')
+
+    gs = Gsettings()
+    gsettings = module.params['settings']
+    for gsetting in gsettings:
+        schema = gsetting['schema']
+        key = gsetting['key']
+        value = gsetting['value']
+        old_setting = gs.get_value(schema, key)
```
mgedmin: Could this possibly raise an exception, if the schema/key are invalid?  If so, you must try/except and convert that to a `module.fail_json()` call.

system/gsettings.py
```diff
@@ -0,0 +1,225 @@
...
+        old_value = old_setting.unpack()
+        if old_value != value:
+            has_changed = True
+            if module.check_mode:
+                change_msgs.append('{0}/{1} -> current: {2} ; proposed new: {3}'.format(schema, key, old_value, value))
+            else:
+                new_value = GLib.Variant(setting_type, value)
+                gs.set_value(schema, key, new_value)
+                change_msgs.append('{0}/{1} -> old: {2} ; new: {3}'.format(schema, key, old_value, value))
+    gs = None
```
mgedmin: This is not a very obvious way to trigger the cleanup.  Can you define and call a `.cleanup()` method instead?

system/gsettings.py
```diff
@@ -0,0 +1,225 @@
...
+except ImportError:
+    HAS_GIO=False
+else:
+    HAS_GIO=True
+
+UID  = 1
+ENV_DBUS = 'DBUS_SESSION_BUS_ADDRESS'
+SESSION_MANAGERS = [
+    'gnome-session', 'mate-session', 'xfce4-session',
+    'cinnamon-session','icewm-session', 'openbox-session' ]
```
mgedmin: PEP-8: can you please find all the commas that are not followed by a space and add the missing space?

There are [tools](https://pypi.python.org/pypi/flake8) that can check Python style for you.

system/gsettings.py
```diff
@@ -0,0 +1,225 @@
...
+                    uid = int(ln.split()[UID])
+                    return pwd.getpwuid(uid).pw_name
+
+        def proc_environ(pid, varname=""):
+            '''Return process environment variable of process pid'''
+            proc_env = open("/proc/{0}/environ".format(pid),'r')
+            env = proc_env.readline()
+            for env_var in env.split('\x00'):
+                if varname != "":
+                    prefix = "{0}=".format(varname)
```
mgedmin: You can hoist the emptiness check and prefix computation outside of the loop.

system/gsettings.py
```diff
@@ -0,0 +1,225 @@
...
+        setting_type = old_setting.get_type_string()
+
+        old_value = old_setting.unpack()
+        if old_value != value:
+            has_changed = True
+            if module.check_mode:
+                change_msgs.append('{0}/{1} -> current: {2} ; proposed new: {3}'.format(schema, key, old_value, value))
+            else:
+                new_value = GLib.Variant(setting_type, value)
+                gs.set_value(schema, key, new_value)
```
mgedmin: Could this raise an exception if you supply an invalid value?

system/gsettings.py
```diff
@@ -0,0 +1,225 @@
...
+        else:
+            path = None
+        self.settings = Gio.Settings(schema=schema, path=path)
+        self.settings.set_value(key, value)
+
+    def __del__(self):
+        ''' Class cleanup '''
+        if self._cleanup:
+           # Stop the dbus daemon if one was started.
+           time.sleep(1)
```
mgedmin: What is the purpose of this sleep?

ag-wood: The sleep is required to prevent a race condition after setting the value and destroying the process.

system/gsettings.py
ag-wood: Yes - sorry, I was a little over-enthusiastic.  Have done the pep-8 check and updated.

system/gsettings.py
ag-wood: Sorry - I don't follow what you are saying here...

ag-wood:
@mgedmin, I think I have addressed all of the issues you raised.  Thanks for your assistance, I think it is a much better result than what I originally posted.

system/gsettings.py
mgedmin: Sorry for being unclear -- I meant this

``` python
        def proc_environ(pid, varname=""):
            '''Return process environment variable of process pid'''
            if not varname:    # <- the if varname != "" check moved here
                return None
            proc_env = open("/proc/{0}/environ".format(pid),'r')
            env = proc_env.read()   # <- changed as per previous suggestion
            prefix = "{0}=".format(varname)  # <-- moved here
            for env_var in env.split('\x00'):
                if env_var.startswith(prefix):
                    ...
```

system/gsettings.py
```diff
@@ -0,0 +1,248 @@
...
+                return None
+            except:
+                return None
+
+        def proc_owner(pid):
+            """Return username of UID of process pid"""
+            UID = 1
+            for ln in open('/proc/{0}/status'.format(pid), 'r'):
+                if ln.startswith('Uid:'):
+                    return int(ln.split()[UID])
```
mgedmin: Uh.  I don't see why you use a named constant here but no such thing in `proc_name()`.

I don't think named constants help, you're just taking the 2nd space-separated field, so `ln.split()[1]` is fine.

system/gsettings.py
```diff
@@ -0,0 +1,248 @@
...
+
+ENV_DBUS = 'DBUS_SESSION_BUS_ADDRESS'
+SESSION_MANAGERS = [
+    'gnome-session', 'mate-session', 'xfce4-session',
+    'cinnamon-session', 'icewm-session', 'openbox-session']
+
+
+class Gsettings(object):
+    def __init__(self):
+        def proc_name(pid):
```
mgedmin: (This nesting of functions is a bit weird, why not move them to the top level?)

system/gsettings.py
```diff
@@ -0,0 +1,248 @@
...
+        # careful checking of values is done before
+        # commiting to get_value.
+        if schema in Gio.Settings.list_schemas():
+            settings = Gio.Settings(schema=schema, path=path)
+            if key in Gio.Settings.list_keys(settings):
+                return settings.get_value(key)
+            else:
+                raise ValueError('{0} is not valid in the schema: {1}'.format(key, schema))
+        else:
+            raise ValueError('{0} is not a valid schema'.format(schema))
```
mgedmin: This would flow a bit more naturally without nesting the checks:

``` python
if schema not in ...list_schemas():
    raise ...
settings = ...
if key not in ...list_keys(...):
    raise ...
return settings.get_value(key)
```

system/gsettings.py
```diff
@@ -0,0 +1,248 @@
...
+
+    def set_value(self, schema, key, value):
+        """ Write a value to the specified key """
+        if ':' in schema:
+            path = schema.split(':')[1]
+            schema = schema.split(':')[0]
```
mgedmin: The same sort of refactoring I suggested in get_value() applies here too:

``` python
schema, path = schema.split(':', 1)
```

robynbergeron:
Hi @ag-wood -- 

robynbergeron:
(Also, thanks to @mgedmin for your comments so far, much appreciated!)

system/gsettings.py
ag-wood: It's a technique I've seen used to hide the implementation of functions that are only used inside the class function.  I'm very happy to move them if it's not the done thing. Still learning lots - thanks!

PR #1040, opened 2015-09-29
Source Mage GNU/Linux package manager support

vaygr:
I'm glad to present the 'sorcery' module for Ansible.

[Sorcery](http://sourcemage.org/Sorcery) is a package management system for some super-flexible source-based Linux distro called [Source Mage GNU/Linux](http://sourcemage.org). It is one of the [oldest](http://sourcemage.org/linux-timeline.png) distros available, and even while being not so popular, it still takes [its niche](http://sourcemage.org/About) quite well.

The module provides wide-range functionality for controlling SMGL (so we call them) systems the way you want them to be (pretty minimal and up-to-date), including dependencies management, full/partial system rebuild, updating of the components (package manager, collections of software, packages), so we could provide a universal alternative to our outdated [Cabal command](http://sourcemage.org/Sorcery/Commands/Cabal), used to be a multi-system control tool.

[Intro page](http://sourcemage.org/Intro) is there if you need some more information on how our system works.
You can also find us lurking on #sourcemage at Freenode IRC network and ask a question!

Thanks!


gregdek:
@vaygr Thanks for submitting this new module to Ansible Extras! This module is now in community review, a process that is open to all Ansible users. In order for this module to be approved, it must gain the following votes:

vaygr:
works_for_me, passes_guidelines

gregdek:
Thanks @vaygr for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

vaygr:
ready_for_review

gregdek:
Thanks @vaygr for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

vaygr:
bump :)

packaging/os/sorcery.py
```diff
@@ -0,0 +1,637 @@
...
+    'cast': None,
+    'dispel': None,
+    'gaze': None
+}
+
+SORCERY_LOG_DIR = "/var/log/sorcery"
+SORCERY_STATE_DIR = "/var/state/sorcery"
+
+
+def exec_command(command, module):
```
bcoca: you can just assign a dictionary with envvars/values to `module.run_command_environ_update`

packaging/os/sorcery.py
```diff
@@ -0,0 +1,637 @@
...
+
+    prompt_env = "env PROMPT_DELAY=0 VOYEUR=0"
+
+    return module.run_command("%s %s" % (prompt_env, command))
+
+
+def get_sorcery_ver(module):
+    """ Get Sorcery version. """
+
+    cmd_sorcery = "%s --version" % SORCERY['sorcery']
```
bcoca, line 164: ~~for this and other command line tools, you should rely on `module.get_bin_path` and not just that the user's path is correctly set.~~ nvmd, just realized you did do this in the global

packaging/os/sorcery.py
```diff
@@ -0,0 +1,637 @@
...
+    depends_list = [s.split('(')[0] for s in depends]
+
+    cmd_gaze = "%s -q version %s" % (SORCERY['gaze'], ' '.join(depends_list))
+
+    rc, stdout, stderr = exec_command(cmd_gaze, module)
+
+    if rc != 0:
+        module.fail_json(msg="wrong dependencies for spell '%s'" % spell)
+
+    fi = fileinput.input(sorcery_depends, inplace=True)
```
bcoca, line 356: if the open fails you'll get a fatal error, this should be within an exception handler

packaging/os/sorcery.py
```diff
@@ -0,0 +1,637 @@
...
+            module.fail_json(msg="I/O error on the depends file")
+    finally:
+        fi.close()
+
+    depends_new = [v for v in depends if depends[v]]
+
+    if depends_new:
+        try:
+            try:
+                fl = open(sorcery_depends, 'a')
```
bcoca, line 408: is there a common way to lock this file? do we really need to write to it directly? 

this seems like it can easily lead to data corruption and not very friendly to multiple users

packaging/os/sorcery.py
```diff
@@ -0,0 +1,637 @@
...
+            update_cache = dict(default=False, aliases=['update_codex'],
+                                choices=BOOLEANS, type='bool'),
+            cache_valid_time = dict(default=0, type='int')
+        ),
+        required_one_of = [['name', 'update', 'update_cache']],
+        supports_check_mode = True
+    )
+
+    if os.geteuid() != 0:
+        module.fail_json(msg="sudo/become is required for this operation")
```
bcoca: or root? probably 'root privileges' is the best expression

vaygr: totally agreed

packaging/os/sorcery.py
vaygr: this is extremely useful, thanks

packaging/os/sorcery.py
vaygr: missed that one, thanks!

vaygr: wait a sec, isn't it handled later in the next `try` block right after this line?

packaging/os/sorcery.py
vaygr: @bcoca unfortunately, no -- there's no built-in functionality for this in the package manager itself, otherwise I'd definitely use it. Managing dependencies is interactively handled by Sorcery and requires user input if they're about to change, this is why we need to write to that file. It looks like this, as you may have noticed:

```
--[ snip ]--
python:gettext:on:required::
python:gdbm:off:optional::
python:db:off:optional::
python:sqlite:on:optional::
python:tk:off:optional::
python:ncurses:on:optional::
python:readline:on:optional::
python:libressl(SSL):on:optional::
python:libffi:on:optional:--with-system-ffi:
python:expat:on:optional:--with-system-expat:
python:valgrind:on:optional:--with-valgrind:--without-valgrind
python:gcc:on:optional:--with-cxx-main=g++:--without-cxx-main
libxslt:libxml2:on:required::
libxslt:icu:on:required::
libxslt:python:on:optional:--with-python:--without-python
libxslt:libgcrypt:on:optional:--with-crypto:--without-crypto
--[ snip ]--
```

We do not need to write to that file line by line though, we could work on a copy and then `os.rename()` it to improve consistency in case of failure?

PR #1043, opened 2015-09-29
Allow relative path for bower executable

javiplx:
When bower is installed locally with the npm module, the binary is not globally accessible. This PR allows to specify a relative path to bower executable from the directory where bower packages will be installed.

### DIFF
--- a/packaging/language/bower.py
+++ b/packaging/language/bower.py
@@ -51,6 +51,7 @@ options:
   localexec:
     description:
       - Relative path to bower executable from install path
+    default: false
     required: false
   state:
     description:
@@ -176,7 +177,7 @@ def main():
         offline=dict(default='no', type='bool'),
         production=dict(default='no', type='bool'),
         path=dict(required=True),
-        localexec=dict(required=False),
+        localexec=dict(default=False, required=False),
         state=dict(default='present', choices=['present', 'absent', 'latest', ]),
         version=dict(default=None),
     )
--- a/packaging/language/bower.py
+++ b/packaging/language/bower.py
@@ -48,6 +48,10 @@ options:
     description:
       - The base path where to install the bower packages
     required: true
+  localexec:
+    description:
+      - Relative path to bower executable from install path
+    required: false
   state:
     description:
       - The state of the bower package
@@ -85,6 +89,7 @@ class Bower(object):
         self.offline = kwargs['offline']
         self.production = kwargs['production']
         self.path = kwargs['path']
+        self.localexec = kwargs['localexec']
         self.version = kwargs['version']
 
         if kwargs['version']:
@@ -94,7 +99,17 @@ class Bower(object):
 
     def _exec(self, args, run_in_check_mode=False, check_rc=True):
         if not self.module.check_mode or (self.module.check_mode and run_in_check_mode):
-            cmd = ["bower"] + args + ['--config.interactive=false', '--allow-root']
+            cmd = []
+
+            if self.localexec:
+                if not os.path.isdir(os.path.join(self.path, self.localexec)):
+                    self.module.fail_json(msg="relative path %s is not a directory" % self.localexec)
+                cmd.append(os.path.join(self.path, self.localexec, "bower"))
+            else:
+                cmd.append("bower")
+
+            cmd.extend(args)
+            cmd.extend(['--config.interactive=false', '--allow-root'])
 
             if self.name:
                 cmd.append(self.name_version)
@@ -161,6 +176,7 @@ def main():
         offline=dict(default='no', type='bool'),
         production=dict(default='no', type='bool'),
         path=dict(required=True),
+        localexec=dict(required=False),
         state=dict(default='present', choices=['present', 'absent', 'latest', ]),
         version=dict(default=None),
     )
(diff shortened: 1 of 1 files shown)


gregdek:
@mwarkentin Please review this PR to make sure it adheres to the following guidelines:

mwarkentin:
@javiplx Thanks for the PR! A couple of notes:
- Seems like something like `relativepath` might be clearer than `localexec` as an argument name? It's also how it's referenced in the error message: `msg="relative path %s is not a directory" % self.localexec`
- Can you add an example of using this to the docs?

Besides that, everything else looks good I think!

packaging/language/bower.py
```diff
@@ -48,6 +48,11 @@
     description:
       - The base path where to install the bower packages
     required: true
+  relative_execpath:
+    description:
+      - Relative path to bower executable from install path
```
mwarkentin, line 53: Maybe a bit clearer if it was something like: `Relative path to bower executable from install path (if bower is not installed globally)`

Might help people understand when they would need to use this?

mwarkentin:
One more comment, and if you could add the example of how to use this new option, and I think we're good to go.

robynbergeron:
@mwarkentin -- Is this ready to go, or still looking for @javiplx to provide example on how to use the new option? (I am not seeing that in here, but just double checking in case I'm blind.)

mwarkentin:
@robynbergeron Still waiting on the example.

robynbergeron:
@mwarkentin -- thanks, that's what I thought. :)

@javiplx -- we're super close! Just looking for an example on how to use the new option to be added to the docs, per @mwarkentin's comment above. And I think once we have that + @mwarkentin is happy with it, we'll get this all merged and happy. 

Once you've made the revisions, please leave a comment saying ready_for_review and we'll keep this moving! :)

Thanks to both of you for your work on this.

javiplx:
Apologies for the delay with the example. I've used there also a call to npm module to make clearer. Hope this mix is acceptable.
ready_for_review

robynbergeron:
Hi @javiplx -- thanks for the update!

@mwarkentin -- can you take a look and let us know if this is now ready, now that we have an example? Thanks! Just leave a comment saying "shipit" or "needs_revision" depending on if it's ready or not :)

mwarkentin:
@javiplx @robynbergeron This looks good to me now.. shipit!

gregdek:
Thanks @mwarkentin for the review, and thanks @javiplx for the PR. Marking for inclusion.

packaging/language/bower.py
```diff
@@ -48,6 +48,11 @@
...
+    default: false
+    required: false
   state:
```
abadger, line 57: Need version_added: "2.0"

packaging/language/bower.py
```diff
@@ -75,6 +80,10 @@
 
 description: Update packages based on bower.json to their latest version.
 - bower: path=/app/location state=latest
+
+description: install bower locally and run from there
+- npm: path=/app/location name=bower global=no
+- bower: path=/app/location localexec=node_modules/.bin
 '''
```
abadger, line 88: Looks like the param was renamed to relative_execpath so the example needs to be changed as well.

packaging/language/bower.py
```diff
@@ -48,6 +48,11 @@
     description:
       - The base path where to install the bower packages
     required: true
+  relative_execpath:
+    description:
+      - Relative path to bower executable from install path
+    default: false
```
abadger: null is probably a more standard default than False.

packaging/language/bower.py
```diff
@@ -172,13 +193,14 @@ def main():
     offline = module.params['offline']
     production = module.params['production']
     path = os.path.expanduser(module.params['path'])
+    relative_execpath = os.path.expanduser(module.params['relative_execpath'])
```
abadger: I believe this will fail when relative_execpath is not specified:

``` python
>>> os.path.expanduser(False)
  File "<stdin>", line 1, in <module>
  File "/usr/lib64/python2.7/posixpath.py", line 261, in expanduser
    if not path.startswith('~'):
AttributeError: 'bool' object has no attribute 'startswith'
```

packaging/language/bower.py
```diff
@@ -161,6 +181,7 @@ def main():
         offline=dict(default='no', type='bool'),
         production=dict(default='no', type='bool'),
         path=dict(required=True),
+        relative_execpath=dict(default=False, required=False),
```
abadger, line 185: In current 2.0 we've added a type='path' that does expanduser and expandvars automatically on the parameter.  might that be better than just using generic str type?

javiplx, line 185: I wasn't even aware that arguments have a type. I would agree with your suggestion, but it should be done also on path. On one side, I don't know how to do that, and on the other side I would like this to be patcheable against 1.9, which is the versión we currently use.

gregdek:
@javiplx A friendly reminder: this pull request has been marked as needing your action. If you still believe that this PR applies, and you intend to address the issues with this PR, just let us know in the PR itself and we will keep it open pending your changes.

javiplx:
I don't know which kind of revision it needs. If it's about the type=path comment, my answer is still valid. I have no idea about how to implement that, and in any case it makes the patch incompatible with 1.9.x, which we currently use.

abadger:
Sorry for not getting back to this sooner -- we've been working to get 2.0 out the door since November.  I've made the final changes suggested, squashed the commits, and merged it here: https://github.com/ansible/ansible-modules-extras/commit/e9898190cd73a9945e52e65e094f27c711023ee1

Note: reagarding the 1.9.x compatibility... we wouldn't be able to push this into an official 1.9.x tarball since it is a new feature for the module and we only push features in new major releases (1.9.0, 2.0.0, 2.1.0, etc).

This feature should be in the 2.1 release.

Thanks!

PR #1046, opened 2015-09-30
Check mode support for VPC route table creation

shawnsi:
The existing check_mode flag on `create_route_table` triggers and error when sent.  This instead sets `changed` and exits the module if the route table does not already exist.

### DIFF
--- a/cloud/amazon/ec2_vpc_route_table.py
+++ b/cloud/amazon/ec2_vpc_route_table.py
@@ -496,13 +496,13 @@ def ensure_route_table_present(connection, module):
         
     # If no route table returned then create new route table
     if route_table is None:
-        if module.check_mode:
-            module.exit_json(changed=True)
-
         try:
-            route_table = connection.create_route_table(vpc_id)
+            route_table = connection.create_route_table(vpc_id, module.check_mode)
             changed = True
-        except EC2ResponseError, e:
+        except EC2ResponseError as e:
+            if e.error_code == 'DryRunOperation':
+                module.exit_json(changed=True)
+
             module.fail_json(msg=e.message)
         
     if routes is not None:
--- a/cloud/amazon/ec2_vpc_route_table.py
+++ b/cloud/amazon/ec2_vpc_route_table.py
@@ -496,8 +496,11 @@ def ensure_route_table_present(connection, module):
         
     # If no route table returned then create new route table
     if route_table is None:
+        if module.check_mode:
+            module.exit_json(changed=True)
+
         try:
-            route_table = connection.create_route_table(vpc_id, check_mode)
+            route_table = connection.create_route_table(vpc_id)
             changed = True
         except EC2ResponseError, e:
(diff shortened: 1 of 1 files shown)


robynbergeron:
Hi @shawnsi -- 

Thanks for submitting this pull request. 

@erydo @wimnat -- Please review this PR to make sure it adheres to the following guidelines:

wimnat:
What error do you get back?

shawnsi:
@wimnat I should clarify my original comment.

As is, this module while create a non existent route table _even in check mode_.  The culprit appears to be the `check_mode` variable set via `module.params.get('check_mode')`.  I tried changing the `create_route_table` call to use `module.check_mode` like this:

``` python
    # If no route table returned then create new route table
    if route_table is None:
        try:
            route_table = connection.create_route_table(vpc_id, module.check_mode)
            changed = True
        except EC2ResponseError, e:
```

This then causes ansible to fail with this message:

``` bash
TASK [subnet : Create Internet Route Table] ************************************
fatal: [us-east-1]: FAILED! => {"changed": false, "failed": true, "msg": "Request would have succeeded, but DryRun flag is set."}
```

My change works around that problem by simply skipping the `create_route_table` call and exiting the module.

wimnat:
Would it be possible to go with your original idea but just add an
exception handle for the boto error so that the module logic path is
preserved?

robynbergeron:
@shawnsi -- looks like @wimnat is looking for some revision to this PR. 

shawnsi:
@wimnat I've refactored per your request.  I still have to call `module.exit_json` in the exception handler though to avoid issues with the `route_table` object further down module evaluation.  This makes sense to me as we can't very well evaluate further changes against a non existent object.

ready_for_review

robynbergeron:
Hi @shawnsi -- thanks for the updates.

@wimnat @erydo -- sending back your way for re-review! Let us know if it's passing guidelines / is ready to merge with a shipit comment -- or if it needs more revision, leave a needs_revision comment.

Thanks, everyone!

wimnat:
Sorry to be picky, but could you compare based on the error_code rather than the msg as it's less likely to change.

E.g.
if e.error_code == 'DryRunOperation':
  module.exit_json(changed=True)

I think that's the right error code but you might want to check ;)

needs_revision

shawnsi:
Good call @wimnat.  I based that logic off boto unit tests.  Your comment made it clear that AWS API error codes are mapped to the python exception which is much more useful.  Thanks!

I amended my commit accordingly.

ready_for_review

wimnat:
shipit

robynbergeron:
Awesome. Thanks, @wimnat for your reviews, and to @shawnsi for all the work :)

Marking for inclusion -- cc @ansible/core team.

shawnsi:
@wimnat I made another update that covers the remaining check mode usage in the module.  Can you review?

ready_for_review

PR #1047, opened 2015-10-01
switched to use module loggigng

bcoca:


robynbergeron:
Hi @bcoca Thanks :)

Travis is failing -- it looks like on maybe 2.4 with znode.py, which @abadger just moved in/renamed last night. I could be wrong though. Still learning the ropes of Travis. :) So I think something needs revision? :)

robynbergeron:
https://github.com/ansible/ansible-modules-extras/commit/962afeed5434e944e39e37b835b514f9bb9ce9ab for reference

bcoca:
yep, that is not caused by this PR

gregdek:
Entrusting this one to the core team. :) 

Since (a) it was opened by a core team member and (b) it touches multiple files.

Issue #1033, opened 2015-09-28
pam_limits: value 'unlimited' isn't accepted


tomjakubowski:
If I try to use `'unlimited'` as the `value` argument to the `pam_limits` module, I get this error:

```
ValueError: invalid literal for int() with base 10: 'unlimited'
```

tomjakubowski:
I believe also that an existing `unlimited` value in the `limits.conf` file will cause an error here: https://github.com/ansible/ansible-modules-extras/blob/bc07c3b504eb5370b1a6e7db8864b6ac5b8bcaea/system/pam_limits.py#L184

robynbergeron:
[module: system/pam_limits.py]

robynbergeron:
Hi @tomjakubowski -- thanks for submitting this bug report.

@usawa -- pinging you to let you know about this bug report for pam_limits.py.

machacekondra:
Hi @tomjakubowski patch which fix this issue was merged, you can try if it works OK for you, now.

tomjakubowski:
:+1: thanks!

PR #1052, opened 2015-10-04
system: pam_limits: add support for unlimited/infinity/-1

machacekondra:
Early pam_limits module didn't support special values for items.
This patch is adding support to special values unlimited, infinity and -1.

Issue: https://github.com/ansible/ansible-modules-extras/issues/1033
Signed-off-by: Ondra Machacek machacek.ondra@gmail.com

### DIFF
--- a/system/pam_limits.py
+++ b/system/pam_limits.py
@@ -102,7 +102,7 @@ def main():
             domain            = dict(required=True, type='str'),
             limit_type        = dict(required=True, type='str', choices=pam_types),
             limit_item        = dict(required=True, type='str', choices=pam_items),
-            value             = dict(required=True, type='int'),
+            value             = dict(required=True, type='str'),
             use_max           = dict(default=False, type='bool'),
             use_min           = dict(default=False, type='bool'),
             backup            = dict(default=False, type='bool'),
@@ -131,6 +131,9 @@ def main():
 
     if use_max and use_min:
         module.fail_json(msg="Cannot use use_min and use_max at the same time." )
+
+    if not (value in ['unlimited', 'infinity', '-1'] or value.isdigit()):
+        module.fail_json(msg="Argument 'value' can be one of 'unlimited', 'infinity', '-1' or positive number. Refer to manual pages for more details.")
 
     # Backup
     if backup:
@@ -181,7 +184,10 @@ def main():
         line_domain     = line_fields[0]
         line_type       = line_fields[1]
         line_item       = line_fields[2]
-        actual_value    = int(line_fields[3])
+        actual_value    = line_fields[3]
+
+        if not (actual_value in ['unlimited', 'infinity', '-1'] or actual_value.isdigit()):
+            module.fail_json(msg="Invalid configuration of '%s'. Current value of %s is unsupported." % (limits_conf, line_item))
 
         # Found the line
         if line_domain == domain and line_type == limit_type and line_item == limit_item:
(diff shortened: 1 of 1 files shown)


system/pam_limits.py
```diff
@@ -191,11 +197,24 @@ def main():
                 nf.write(line)
                 continue
 
+            actual_value_unlimited = actual_value in ['unlimited', 'infinity', '-1']
+            value_unlimited = value in ['unlimited', 'infinity', '-1']
```
chrrrles, line 201: `value_unlimited` looks unused here

machacekondra, line 201: True, I'll re-submit patch

robynbergeron:
@machacekondra -- thanks for submitting this PR.

@usawa -- Please review this PR to make sure it adheres to the following guidelines:

Thanks!

system/pam_limits.py
```diff
@@ -181,7 +184,10 @@ def main():
         line_domain     = line_fields[0]
         line_type       = line_fields[1]
         line_item       = line_fields[2]
-        actual_value    = int(line_fields[3])
+        actual_value    = line_fields[3]
+
+        if not (actual_value in ['unlimited', 'infinity', '-1'] or actual_value.isdigit()):
+            module.fail_json(msg="Value of '%s' is unsupported, please fix it." % line_item)
```
bcoca: it could be clearer that the error resides in the existing limits file

machacekondra: Done

system/pam_limits.py
```diff
@@ -181,7 +184,10 @@ def main():
         line_domain     = line_fields[0]
         line_type       = line_fields[1]
         line_item       = line_fields[2]
-        actual_value    = int(line_fields[3])
+        actual_value    = line_fields[3]
+
+        if not (actual_value in ['unlimited', 'infinity', '-1'] or actual_value.isdigit()):
+            module.fail_json(msg="Invalid configuration of '%s'. Value '%s' is unsupported." % (limits_conf, line_item))
```
abadger, line 191: Should this be one of the following:

``` python
            module.fail_json(msg="Invalid configuration of '%s'. Value '%s' is unsupported." % (limits_conf, actual_value)) 
            module.fail_json(msg="Invalid configuration of '%s'. Current value of %s is unsupported." % (limits_conf, line_item)) 
```

system/pam_limits.py
```diff
@@ -132,6 +132,9 @@ def main():
     if use_max and use_min:
         module.fail_json(msg="Cannot use use_min and use_max at the same time." )
 
+    if not (value in ['unlimited', 'infinity', '-1'] or value.isdigit()):
+        module.fail_json(msg="Argument 'value' can be either 'unlimited', 'infinity', '-1' or positive number. Refer to manual pages for more details.")
+
     # Backup
```
abadger, line 138: minor nitpick:  either is used when there's two choices.  So this should be:

``` python
module.fail_json(msg="Argument 'value' can be one of 'unlimited', 'infinity', '-1' or a positive number. Refer to the manual pages for more details.") 
```

system/pam_limits.py
```diff
@@ -191,11 +197,23 @@ def main():
...
+                    new_value = value
 
             if use_min:
-                new_value = min(value,actual_value)
+                if value.isdigit() and actual_value.isdigit():
+                    new_value = min(value, actual_value)
+                elif actual_value_unlimited:
+                    new_value = value
+                else:
+                    new_value = actual_value
```
abadger, line 218: In order to keep idempotence in case both actual_value and value are set to different unlimited strings, I think the min check needs to check whether the value is an unlimited version and set to actual_value if so, otherwise fall back to new_value = value.  (that way we always prefer the current configuration if they are different strings but equivalent).

system/pam_limits.py
```diff
@@ -191,11 +197,23 @@ def main():
                 nf.write(line)
                 continue
 
+            actual_value_unlimited = actual_value in ['unlimited', 'infinity', '-1']
+
             if use_max:
-                new_value = max(value, actual_value)
+                if value.isdigit() and actual_value.isdigit():
+                    new_value = max(value, actual_value)
```
abadger: Not sure if value can be a two digit number but if so, this needs to convert to integers before comparing value and actual_value like this:

``` python
new_value = max(int(value), int(actual_value))
```

the reason is that something like `max('10', '9')` returns `'9'`.

The same goes for the call to `min(value, actual_value)` below.

abadger:
Noted some issues.  Once those are fixed, this is good to merge.

machacekondra:
All fixed

abadger:
@machacekondra Thanks for adding that feature!

# Merged

Hi!

issue is resolved.

Thank you!

PR #173, opened 2014-12-30
Homebrew: Allow repository_path to be defined

ashmckenzie:
This is useful when both Boxen and Ansible exist on the same node.  Also PR is to help facilitate https://github.com/ashmckenzie/ansible-modules-extras/pull/1.

PR #1039, opened 2015-09-29
homebrew: Improve 'argument_spec' handling

indrajitr:
- Remove choice list for boolean values in `argument_spec` and make it more consistent with core modules
- Add 'package' alias and support for list type for 'name' parameter

NB: Seems like this module is orphaned. Would be happy to take over in the interim to clear the backlogs :)

PR #1053, opened 2015-10-05
homebrew: Add explicit documentation for 'path' argument (with expected default)

indrajitr:
In Homebew, a formula is installed in a location relative to the actual
`brew` command. The documentation clarifies that.

Additionally, removed redundant 'path' reconstruction in multiple places.

See: [Homebrew documentation](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Installation.md#multiple-installations) for more.

/cc @ashmckenzie since this is related to PR #173. Would be great if you can quickly try this out.

### DIFF
--- a/packaging/os/homebrew.py
+++ b/packaging/os/homebrew.py
@@ -69,10 +69,22 @@ options:
 notes:  []
 '''
 EXAMPLES = '''
+# Install formula foo with 'brew' in default path (C(/usr/local/bin))
 - homebrew: name=foo state=present
+
+# Install formula foo with 'brew' in alternate path C(/my/other/location/bin)
+- homebrew: name=foo path=/my/other/location/bin state=present
+
+# Update homebrew first and install formula foo with 'brew' in default path
 - homebrew: name=foo state=present update_homebrew=yes
+
+# Update homebrew first and upgrade formula foo to latest available with 'brew' in default path
 - homebrew: name=foo state=latest update_homebrew=yes
+
+# Update homebrew and upgrade all packages
 - homebrew: update_homebrew=yes upgrade_all=yes
+
+# Miscellaneous other examples
 - homebrew: name=foo state=head
 - homebrew: name=foo state=linked
 - homebrew: name=foo state=absent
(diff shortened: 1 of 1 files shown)


robynbergeron:
Hi @indrajitr -- thanks for submitting this PR. Since you're the owner, I can mark this for inclusion -- did you want me to wait on @ashmckenzie to try it out, or just move ahead? :)

indrajitr:
We can move ahead because this extra cleanup and documentation is necessary regardless.
I pinged @ashmckenzie so that he can checkout and close #173 if necessary.

robynbergeron:
okay! Marking for inclusion .. cc @ansible/core team. 

Thanks! :)

robynbergeron:
Actually -- @indrajitr... i think it needs a rebase (I might be wrong on this, but I think i'm right... slowly getting the hang of more of this. You probably know better than I do :D)

indrajitr:
Right you are, 02c17c023a2c98333fb997429f2fbf6c308e5695 for #1039 needs to be accommodated. Rebase coming up.

indrajitr:
ready_for_review

robynbergeron:
Hi @indrajitr -- thanks for the rebase! (And thanks for validating that I was indeed correct! I'm getting smarter all the time :D)

Marking this owner_pr for inclusion -- pinging @ansible/core friends for that magic.

packaging/os/homebrew.py
```diff
@@ -37,6 +37,11 @@
             - name of package to install/remove
         required: false
         default: None
+    path:
+        description:
+            - ':' separated list of paths to search for 'brew' executable. Since A package (I(formula) in homebrew parlance) location is prefixed relative to the actual path of I(brew) command, providing an alternative I(brew) path enables managing different set of packages in an alternative location in the system.
+        required: false
+        default: '/usr/local/bin'
```
resmo, line 44: missing: `version_added: "2.0"`

indrajitr, line 44: I don't consider it so because the parameter `path` existed earlier. It just wasn't explicitly added in the `DOCUMENTATION` block and thus never showed up in the module documentation.

PR #1055, opened 2015-10-05
Gsettings module

ag-wood:
Accidentally closed previous pull request - re-submitting.

### DIFF
--- a/system/gsettings.py
+++ b/system/gsettings.py
@@ -175,9 +175,9 @@ class Gsettings(object):
             if key in Gio.Settings.list_keys(settings):
                 return settings.get_value(key)
             else:
-                module.fail_json(rc=1, msg='{0} is not valid in the schema: {1}'.format(key, schema))
+                raise ValueError('{0} is not valid in the schema: {1}'.format(key, schema))
         else:
-            module.fail_json(rc=1, msg='{0} is not a valid schema'.format(schema))
+            raise ValueError('{0} is not a valid schema'.format(schema))
 
     def set_value(self, schema, key, value):
         """ Write a value to the specified key """
@@ -220,7 +220,11 @@ def main():
         schema = gsetting['schema']
         key = gsetting['key']
         value = gsetting['value']
-        old_setting = gs.get_value(schema, key)
+        try:
+            old_setting = gs.get_value(schema, key)
+        except ValueError as err:
+            module.fail_json(rc=1, msg=err.message)
+
         setting_type = old_setting.get_type_string()
 
         old_value = old_setting.unpack()
--- a/system/gsettings.py
+++ b/system/gsettings.py
@@ -26,7 +26,7 @@ description:
 version_added: "1.9"
 notes:
     - become and become_user must be used to specify the target configuration set.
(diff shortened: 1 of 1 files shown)


chrrrles:
related:   https://github.com/ansible/ansible-modules-extras/pull/1026

robynbergeron:
Hi @ag-wood -- Thanks for submitting this new module to Ansible Extras! This module is now in community review, a process that is open to all Ansible users. In order for this module to be approved, it must gain the following votes:

system/gsettings.py
```diff
@@ -0,0 +1,258 @@
...
+class Gsettings(object):
+    def __init__(self):
+        # Determine the session pid.
+        self.session_pid = -1
+        self._cleanup = False
+
+        for procfile in os.listdir('/proc'):
+            if procfile.isdigit():
+                if proc_name(procfile) in SESSION_MANAGERS:
+                    self.session_pid = procfile
```
mscherer, line 144: So, what happen if there is more than one session manager running ?

system/gsettings.py
```diff
@@ -0,0 +1,258 @@
...
+
+        osuser = os.getuid()
+        if self.session_pid != -1:
+            sessionuser = proc_owner(self.session_pid)
```
mscherer, line 148: There is a race condition, if the PID is reused (or change between the time the PID is found and the time this file is opened again ).

system/gsettings.py
```diff
@@ -0,0 +1,258 @@
...
+        # already checked in the get_value function.
+        settings = Gio.Settings(schema=schema, path=path)
+        settings.set_value(key, value)
+
+    def cleanup(self):
+        """ Class cleanup """
+        if self._cleanup:
+            # Stop the dbus daemon if one was started.
+            time.sleep(1)
+            os.kill(int(self.session_pid), signal.SIGTERM)
```
mscherer, line 204: shouldn't we verify that the PID is still the one of the dbus daemon ? if it did crash, this would send a SIGTERM to a random process somewhere, which is likely a unattended side effect.

system/gsettings.py
```diff
@@ -0,0 +1,258 @@
...
+            has_changed = True
+            if module.check_mode:
+                change_msgs.append('{0}/{1} -> current: {2} ; proposed new: {3}'.format(schema, key, old_value, value))
+            else:
+                try:
+                    new_value = GLib.Variant(setting_type, value)
+                except Exception, err:
+                    module.fail_json(rc=1, msg='The value provided is invalid: {1}'.format(err))
+
+                gs.set_value(schema, key, new_value)
```
pilou-, line 250: The boolean value returned by `set_value` should be checked.

system/gsettings.py
```diff
@@ -0,0 +1,258 @@
...
+                change_msgs.append('{0}/{1} -> old: {2} ; new: {3}'.format(schema, key, old_value, value))
+    gs.cleanup()
+    module.exit_json(rc=0, changed=has_changed, msg=change_msgs)
```
pilou-, line 253: You could remove the `has_changed` variable and use instead `bool(change_msgs)`.

PR #1060, opened 2015-10-05
Fix open_iscsi EXAMPLES section

sivel:
This PR moves the examples for `open_iscsi` into an `EXAMPLES` section, and reformats to follow standards.

cc @srvg


srgvg:
:+1:

Issue #908, opened 2015-09-01
win_environment changes don't take into effect until after user logout


synapticarbors:
Issue Type:
- Bug Report

Ansible Version: 

```
ansible 1.9.2
  configured module search path = None
```

I attempted to test this off of the latest `devel` branch, but I was getting WinRM ssl errors that weren't resolved using the callback plugin that has been mentioned on the Google Group, and which works for `v1.9.2`.

Ansible Configuration:

```
[defaults]
callback_plugins=<path stuff>/test-ansible/callback_plugins
```

Environment:

Running Ansible off of OSX 10.9.6, targeting a Windows server 2012 R2 

Summary:

The `win_environment` module appears to successfully set an environment variable, but the env variable is not visible on the target machine until the user logs out and then logs back in. 

Steps To Reproduce:

I copied the `win_environment` module (both `.py` and `.ps1` files) into `./library` since they are only available in v2 (see above). I then use that module in a task to set an environment variable as:

```

---
- name: Test ansible
  hosts: windows
  remote_user: deploy
  tasks:
    - name: Set ABC environment variable
      win_environment:
          state: present
          name: 'ABC'
          value: "{{ abc }}"
          level: User
```

Expected Results:

I expect that given a fresh powershell that `$env:ABC` will be the value provided in the playbook. 

Actual Results:

If `ABC` is not set on the target machine, I get a result as:

```
ok: [win_machine] => {"before_value": "null", "changed": true, "level": "user", "name": "ABC", "value": "xxx"}
```

Switching over to an RDP on `win_machine`, `ABC` is not set. Re-running the ansible play, I get:

```
ok: [win_machine] => {"before_value": "xxx", "changed": false, "level": "user", "name": "ABC", "value": "xxx"}
```

so, the new value is being stored somewhere. 

If on the Windows machine I run either `setx ABC yyy` from a command prompt or `[Environment]::SetEnvironmentVariable('ABC', 'yyy', 'User')` from the powershell, and then re-run the ansible task on my OSX machine, I get:

```
ok: [win_machine] => {"before_value": "yyy", "changed": true, "level": "user", "name": "ABC", "value": "xxx"}
```

However, again, the change isn't reflected on the machine after starting up a new powershell terminal. The only way to get the value pushed by Ansible to be reflected on the user machine is to log the user off and then log them back on. This doesn't seem to be the correct behavior, or at least not what I'd anticipate given how running similar commands locally on the target machine in a script behaves. 

[...]

PR #1066, opened 2015-10-07
Document win_environment limitations 

jhawkesworth:
As result of https://github.com/ansible/ansible-modules-extras/issues/908


robynbergeron:
Hi @jhawkesworth -- thanks for submitting this PR.

Marking as owner_pr for inclusion -- cc @ansible/core team. Note that this is a documentation fix only to the module (which also resolves a bug report).

Issue #1073, opened 2015-10-08
win_package: The variable '$script:MsiTools' cannot be retrieved because it has not been set.


darrengruber:
##### Ansible Version:

```
ansible 2.0.0 (detached HEAD 5ae74f7d38) last updated 2015/10/08 14:32:55 (GMT -400)
  lib/ansible/modules/core: (detached HEAD d659c79db7) last updated 2015/10/08 14:31:12 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD 0ff1776a84) last updated 2015/10/08 14:31:24 (GMT -400)
  config file =
  configured module search path = None
```
##### Ansible Configuration:

N/A
##### Environment:

Managing from: `OSX 10.10.5`
Managing to: `Microsoft Windows NT 6.2.9200.0 (Microsoft Windows Server 2012 Standard)`
##### Summary:

Attempts to install an .msi from a URL using `win_package` will result in an error.
##### Steps To Reproduce:
- Create a playbook with this task defined on it:  

```
    - name: Install Microsoft ReportViewer 2012 Runtime
      win_package:
        path: http://download.microsoft.com/download/F/B/7/FB728406-A1EE-4AB5-9C56-74EB8BDDF2FF/ReportViewer.msi
        product_id: "{DFDB4411-54A2-41CE-AB89-CE54C176A69D}"
```
- Run this task against a Windows Server 2012 Standard host.
##### Expected Results:

When given a working URL for download along with its corresponding `product_id`, ansible will download and install this package in an idempotent manner.
##### Actual Results:

Task fails with the following error:

```
TASK [Install Microsoft® ReportViewer 2012 Runtime] ****************************
[...]

PR #1076, opened 2015-10-09
turned off strict msi code checking

trondhindenes:
We don't really need strict msi checking, as this introduces some unstabilities in the module.


robynbergeron:
Hi @trondhindenes -- thanks for this PR :) Labelling it as a bugfix and not a feature... since the word instability generally means bugs will or are happening :)

Marking as owner_pr and shipit for inclusion and all that jazz, cc @ansible/core friends.

robynbergeron:
(And since, yeah, #1073 ... definitely a bugfix PR!)

PR #1085, opened 2015-10-12
files/confparse - alternate ini parser

ritzk:
config parser for *nix configuration files, with minimal support for config files with sections. Allows one to read and write to config file.

old bz - https://github.com/ansible/ansible-modules-extras/pull/658


bcoca:
we don't normally allow for 'duplicate' modules that have the same purpose, we generally encourage adding any extra functionality to existing modules.

ritzk:
Done.

ritzk:
@bcoca  This works with unix configuration files, which lack section header.

ini_file is exclusively for ini files which have section header ( based on ConfigParser)

bcoca:
that is something that should be added as an option to the existing module

gregdek:
Thanks @ritzk for the PR, but @bcoca of the core team has made the call. Please make any changes in the existing module. Closing this PR.

PR #1089, opened 2015-10-13
Refactor ec2_remote_facts to use filters

wimnat:
Modified ec2_remote_facts to use the filters attribute similar to other aws facts modules.

Using filters the other attributes are unnecessary and since this module is yet to be released i think it's ok to remove them at this stage.

@michaeljs1990 i hope you can review and see where i'm going with this PR - i know it's a big one!

I removed your code that created a hostname fact as this seemed dangerous to me as it was assuming the hostname would be set to the AWS standard of ip-1-2-3-4 but this may not be the case.

An alternative method to do what you want would be to use this module to get your list of instances and then use the ec2_facts module to get the metadata which will include the actual hostname.

E.g.

ec2_remote_facts:
  filters:
    "tag:Name": get_these_instances
register: i

add_host:
  hostname: "{{ item.private_ip_address }}"
  groupname: hostname_group
- name: get hostnames
  hosts: hostname_group
  gather_facts: true
  ### After gathering facts hostname of each instance will be an ansible fact or you can go on to use ec2_facts to get the metadata

### DIFF
--- a/cloud/amazon/ec2_remote_facts.py
+++ b/cloud/amazon/ec2_remote_facts.py
@@ -16,137 +16,149 @@
 DOCUMENTATION = '''
 ---
 module: ec2_remote_facts
-short_description: ask EC2 for information about other instances.
+short_description: Gather facts about ec2 instances in AWS
 description:
-    - Only supports search for hostname by tags currently. Looking to add more later.
+    - Gather facts about ec2 instances in AWS
 version_added: "2.0"
 options:
-  key:
+  filters:
     description:
-      - instance tag key in EC2
-    required: false
-    default: Name
-  value:
-    description:
-      - instance tag value in EC2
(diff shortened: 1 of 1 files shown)


robynbergeron:
Hi @wimnat -- thanks for the PR.

@michaeljs1990 - Please review this PR to make sure it adheres to the following guidelines:

Thanks, everyone!

michaeljs1990:
just got a chance to look over this. This looks awesome to me and it's nice that it matches the AWS spec for filters. My only concern is breaking changes for users. Do we have a good way right now to let people know or is it just captured in the release notes?

Code looks great to me... can we just add a newline at the end of the file :+1: and then shipit

wimnat:
@michaeljs1990 as this is a 2.0 module and 2.0 isn't released yet, i think we're ok to make a breaking change at this point in time.

I added the newline at end.  If all looks good, just leave a shipit comment :)

michaeljs1990:
shipit

michaeljs1990:
nice work on this much better than my original version.

gregdek:
Thanks @wimnat for the PR, and thanks @michaeljs1990 for the review. Marking for inclusion.

PR #1090, opened 2015-10-13
Use filters attribute rather than eni id 

wimnat:
Matches the new ec2_vpc_route_table_facts and ec2_vpc_subnet_facts modules

shipit


robynbergeron:
Hi @wimnat -- thanks for the PR.

Marking as owner_pr and ready to ship :)

robynbergeron:
Apparently I am having a day where I can't differentiate between cloud and windows when labeling PRs. Yikes :\

bcoca:
this change breaks backwards compatiblity

wimnat:
The module hasn't been released yet so I think the change is ok

PR #1091, opened 2015-10-13
Add win_uri Module

blakfeld:
An attempt to port the "uri" module to Windows using "Invoke-WebRequest".

### DIFF
--- a/windows/win_uri.py
+++ b/windows/win_uri.py
@@ -24,7 +24,7 @@
 DOCUMENTATION = """
 ---
 module: win_uri
-version_added: ""
+version_added: "2.0"
 short_description: Interacts with webservices.
 description:
   - Interacts with HTTP and HTTPS services.
@@ -32,10 +32,12 @@ options:
   url:
     description:
       - HTTP or HTTPS URL in the form of (http|https)://host.domain:port/path
+    required: true
   method:
     description:
       - The HTTP Method of the request or response.
     default: GET
+    required: false
     choices:
       - GET
       - POST
@@ -53,13 +55,17 @@ options:
   body:
     description:
       - The body of the HTTP request/response to the web service.
+    required: false
+    default: None
   headers:
     description:
       - Key Value pairs for headers. Example "Host: www.somesite.com"
-author: Corwin Brown
+    required: false
+    default: None
+author: Corwin Brown (@blakfeld)
 """
 
-Examples= """
+Examples = """
 # Send a GET request and store the output:
 ---
 - name: Perform a GET and Store Output
@@ -90,4 +96,19 @@ Examples= """
     url: http://www.somesite.com
     method: POST
     body: "{ 'some': 'json' }"
+
+# Check if a file is available on a webserver
+---
+- name: Ensure Build is Available on Fileserver
+  when: ensure_build
+  win_uri:
+    url: "http://www.somesite.com"
+    method: HEAD
+    headers:
+      test: one
+      another: two
+  register: build_check_output
+  until: build_check_output.StatusCode == 200
+  retries: 30
+  delay: 10
 """
--- a/windows/win_uri.py
+++ b/windows/win_uri.py
@@ -1,7 +1,7 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
 
-# (c) 2015, Corwin Brown <corwin.brown@maxpoint.com>
+# (c) 2015, Corwin Brown <corwin@corwinbrown.com>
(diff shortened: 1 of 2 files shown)


robynbergeron:
Hi @blakfeld --

robynbergeron:
Pinging our windows crew also as this may be of interest to them. 
@trondhindenes @petemounce @elventear @smadam813 @jhawkesworth @angstwad @sivel @chrishoffman @cchurch

davidobrien1985:
I like it :) :+1:

windows/win_uri.py
```diff
@@ -0,0 +1,93 @@
...
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+# this is a windows documentation stub.  actual code lives in the .ps1
+# file of the same name
+
+DOCUMENTATION = """
+---
+module: win_uri
+version_added: ""
```
jhawkesworth: please put in version - current dev is "2.0" at the time of writing.

jhawkesworth:
This would be useful.  
Can you do (very minor) doc fixes please?
--jon

blakfeld:
Documentation fixed! I also poked through the official docs, and added "required" and better "Default" fields.

robynbergeron:
Hi @blakfeld -- thanks for the docs fixes!

@davidobrien1985 & @jhawkesworth -- a super friendly request that we need a +1 vote for each of the following things (as described above... but I'll paste it again for good measure):

Thanks for your help -- great to see new modules getting reviews! :dancers:

jhawkesworth:
ok, got back to this again.

Thanks for the doc fixes.

I know I'm being really picky here but obviously want to make it all as good as possible.

Using Get-AnsibleParam might reduce the lines of code a bit - I like small modules as they transfer to hosts more quickly.

Really tiny thing - adding your github id to the Author field in the docs would make you popular with the ansible maintainers too.

If you have any test playbooks can you share to give others a hand to get started testing too?

Jon

blakfeld:
@jhawkesworth -- Thanks for the nit-picks! Nit-picks are the only way to get quality code, and I'm a Linux guy by trade, so my Powershell likely needs nit-picking.

I went and switched over to use Get-AnsibleParams, which seems like a way better way to do what I was doing.

I added my GitHub name to the docs

I don't have a formal test playbook, but here's the task I'm using it for in production:

```
---
- name: Ensure Build is Available on Fileserver
  when: ensure_build
  win_uri:
    url: "http://www.somesite.com"
    method: HEAD
  register: build_check_output
  until: build_check_output.StatusCode == 200
  retries: 30
  delay: 10
```

blakfeld:
I've run into a bug with the updated params. I'm working on it now, I'll update when I get it sorted

EDIT:

Okay, sorted it all out! I just updated the new Params too quickly and didn't proof-read as much as I should've

gregdek:
Back to @jhawkesworth ? :)

jhawkesworth:
Thanks for turning this round so quickly @blakfeld 
Giving this a meets_guidelines +1 now.

blakfeld:
Anyone opposed to merging this?

blakfeld:
Found and fixed a bug. If you ran this with a service account that had never used Internet Explorer, it would fail. Under the covers I'm using Invoke-WebRequest, which uses the Internet Explorer Engine to parse the web page, and if you've never run Internet Explorer, some configuration isn't set, so it fails. The solution is to use the -UseBasicParsing flag. The only upside to using the IEEngine is that it allows the DOM to be used like a Powershell object, but since that isn't super useful to Ansible, I defaulted that to true.

jhawkesworth:
Good catch with the -UseBasicParsing and I like that you have explained it in the docs too.

Since you have revised, I checked again and this still

meets_guidelines

Just needs someone (other than me) to test now.

Jon

nitzmahone:
@blakfeld Looks pretty good- I played around with it a bit and it behaved as expected. One minor request, though- can you insert a little something to translate the return keys into ansible standard-looking return values (eg, as you're adding the keys to the result object, roll them through a dumb func to convert PasalCasedStringKey into pascal_cased_string_key), and adjust the docs accordingly? @bcoca or @cchurch might have a different opinion, but I think we want to abstract the result dict so it looks like a normal Ansible result dict.

blakfeld:
@nitzmahone seems reasonable. Hacked that in!

nitzmahone:
TIL "snake_case"... Sold!

PR #1094, opened 2015-10-13
lookupMib=False for pysnmp 4.3.0

ogenstad:
pysnmp 4.3.0 was released 2015-09-28. In that version the maintainers of pysnmp changed the default parameter to handle MIB lookups. In version 4.2.5 pysnmp just returned the numeric oids by default, but from 4.3.0 a lookup will be attempted which breaks this module.

This patch forces disables this lookup. Without the patch no data would be returned when using pysnmp 4.3.0.

I've tested the patch both with pysnmp 4.2.5 and 4.3.0.

### DIFF
--- a/network/snmp_facts.py
+++ b/network/snmp_facts.py
@@ -112,7 +112,7 @@ class DefineOid(object):
         self.sysContact  = dp + "1.3.6.1.2.1.1.4.0"
         self.sysName     = dp + "1.3.6.1.2.1.1.5.0"
         self.sysLocation = dp + "1.3.6.1.2.1.1.6.0"
-        
+
         # From IF-MIB
         self.ifIndex       = dp + "1.3.6.1.2.1.2.2.1.1"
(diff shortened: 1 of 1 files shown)


robynbergeron:
Hi @ogenstad -- thanks for the PR, and for your ongoing work in maintaining this module :)

Labeling as owner_pr and marking for inclusion -- cc @ansible/core team.

bcoca:
this seems to be exclusively non significant whitespace changes, we don't accept formatting requests as they break attribution

ogenstad:
I  think my editor got creative with the whitespaces.

There are actually two lines which are relevant, both of them look like this:
lookupMib=False

Should be line 252 and 290. Without those lines the module wouldn't return any data if the system is running pysnmp > 4.2.5.

bcoca:
since you are also author of the lines in question, i'm just merging this

PR #479, opened 2015-05-11
patch pacman's module for yaourt, a pacman-wrapper

RubenKelevra:
yaourt can access the aur-packages as well. the pacman module was patched to make aur-packages-updates as well.

else I changed a bit of internal behavoir and included a patch which added the system-upgrade function to pacman.

PR #1104, opened 2015-10-14
New Module: Yaourt

ghost:
Adding yaourt module for the ArchLinux pacman wrapper called yaourt

### DIFF
--- a/packaging/os/pacman.py
+++ b/packaging/os/pacman.py
@@ -124,13 +124,13 @@ def get_version(pacman_output):
     return None
 
 def query_package(module, pacman_path, name, state="present"):
-    """Query the package status in both the local system and the repository. Returns a boolean to indicate if the package is installed, and a second boolean to indicate if the package is up-to-date."""
+    """Query the package status in both the local system and the repository. Returns a boolean to indicate if the package is installed, a second boolean to indicate if the package is up-to-date and a third boolean to indicate whether online information were available"""
     if state == "present":
         lcmd = "%s -Qi %s" % (pacman_path, name)
         lrc, lstdout, lstderr = module.run_command(lcmd, check_rc=False)
         if lrc != 0:
             # package is not installed locally
-            return False, False
+            return False, False, False
 
         # get the version installed locally (if any)
         lversion = get_version(lstdout)
@@ -143,9 +143,10 @@ def query_package(module, pacman_path, name, state="present"):
         if rrc == 0:
             # Return True to indicate that the package is installed locally, and the result of the version number comparison
             # to determine if the package is up-to-date.
-            return True, (lversion == rversion)
-
-        return False, False
+            return True, (lversion == rversion), False
+        
+    # package is installed but cannot fetch remote Version. Last True stands for the error
+        return True, True, True
 
 
 def update_package_db(module, pacman_path):
@@ -165,7 +166,7 @@ def upgrade(module, pacman_path):
     if rc == 0:
         if module.check_mode:
             data = stdout.split('\n')
-            module.exit_json(changed=True, msg="%s package(s) would be upgraded" % len(data))
+            module.exit_json(changed=True, msg="%s package(s) would be upgraded" % (len(data) - 1))
         rc, stdout, stderr = module.run_command(cmdupgrade, check_rc=False)
         if rc == 0:
             module.exit_json(changed=True, msg='System upgraded')
@@ -190,7 +191,7 @@ def remove_packages(module, pacman_path, packages):
     # Using a for loop incase of error, we can report the package that failed
     for package in packages:
         # Query the package first, to see if we even need to remove
-        installed, updated = query_package(module, pacman_path, package)
+        installed, updated, unknown = query_package(module, pacman_path, package)
         if not installed:
             continue
 
@@ -211,10 +212,15 @@ def remove_packages(module, pacman_path, packages):
 
 def install_packages(module, pacman_path, state, packages, package_files):
     install_c = 0
+    package_err = []
+    message = ""
 
     for i, package in enumerate(packages):
         # if the package is installed and state == present or state == latest and is up-to-date then skip
-        installed, updated = query_package(module, pacman_path, package)
+        installed, updated, latestError = query_package(module, pacman_path, package)
+        if latestError and state == 'latest':
+            package_err.append(package)
+            
         if installed and (state == 'present' or (state == 'latest' and updated)):
             continue
 
@@ -230,17 +236,19 @@ def install_packages(module, pacman_path, state, packages, package_files):
             module.fail_json(msg="failed to install %s" % (package))
 
         install_c += 1
-
+    
+    if state == 'latest' and len(package_err) > 0:
+        message = "But could not ensure 'latest' state for %s package(s) as remote version could not be fetched." % (package_err)
+    
     if install_c > 0:
-        module.exit_json(changed=True, msg="installed %s package(s)" % (install_c))
-
-    module.exit_json(changed=False, msg="package(s) already installed")
-
-
+        module.exit_json(changed=True, msg="installed %s package(s). %s" % (install_c, message))
+        
+    module.exit_json(changed=False, msg="package(s) already installed. %s" % (message))
+    
 def check_packages(module, pacman_path, packages, state):
     would_be_changed = []
     for package in packages:
-        installed, updated = query_package(module, pacman_path, package)
+        installed, updated, unknown = query_package(module, pacman_path, package)
         if ((state in ["present", "latest"] and not installed) or
                 (state == "absent" and installed) or
                 (state == "latest" and not updated)):


robynbergeron:
Hi @TecSet -- Thanks for submitting this new module to Ansible Extras! This module is now in community review, a process that is open to all Ansible users. In order for this module to be approved, it must gain the following votes:

robynbergeron:
@TecSet -- it's worth noting that there is an outstanding PR for a yaourt module already. #479 

The person who submitted it hasn't responded to some requests for revisions recently, so it's somewhat stalled -- but it might be worth taking a look at for the feedback it has received, or to see if you can get some of the kind folks who have commented on that PR to take a look at this. 

Also pinging @indrajitr and @elasticdog for their awareness and possible eyeballs, since they are the kind and gracious maintainers of the pacman module and may have some thoughts here. :)

ghost:
The Yaourt Module is completely based on the current pacman module found in this repository. 
The only changes are from "pacman" to "yaourt" and a few extra arguments to be able to also install aur packages.

ghost:
works_for_me

ghost:
passes_guidelines

indrajitr:
Given the similarity between `yaourt` and `pacman` (and `yaourt` typically striving to stay compatible with `pacman`), how about having this functionality folded into `pacman`?

I have nothing against having a separate module `yaourt` in principle. However, we are potentially looking at maintaining two separate module with 90+% copy-paste overlap which probably isn't ideal.

Thoughts?

NB: Would be great to have Ansible core members weigh in with recommendation(s).

ghost:
Pacman cannot install AUR on its own. The only way to implement that functionality to pacman would be to add base-devel (makepkg and other) as dependency.

indrajitr:
I am talking about _externalizing_ pacman/yaourt via a new module variable (say, `pkgmgr`) which defaults to `pacman` (but can be optionally set to `yaourt`). It would be imperative that setting pkgmgr=yaourt would require yaourt being available.

robynbergeron:
@TecSet -- it is probably worth noting that the votes have to come from a community member other than the person who is submitting the pull request for the new module. Sorry if that was unclear :)

https://github.com/ansible/ansible-modules-extras/blob/devel/REVIEWERS.md describes this in more detail.

robynbergeron:
@ansible/core team -- your weigh-in here on @indrajitr's comments would be appreciated. Thanks! :)

ghost:
@robynbergeron but you said:
If you have tested [...], please add “works_for_me” in the comments.
If you have gone [....], please add “passes_guidelines” in the comments. 
:D but okay. The link does make it clear.

ghost:
I've modified the pacman module on a separate branch to demonstrated a way of transparently integrating yaourt (untested):
https://github.com/TecSet/ansible-modules-extras/blob/devel-pacman-yaourt-integration/packaging/os/pacman.py

Yaourt is only used, when the parameter aur is true (default false) but will be used for all commands if so.  Only problem left is whats happens at "get_bin_path" when yaourt is not installed

bhcopeland:
+1 

I gave your pacman.py a try, and it worked ok.
- I found running "sudo: yes" causes failure. Because building AUR packages are advised to run as a none sudo user, so you should add some error checking there.
- When aur's are building, packages can take a while to compile. Some output should come back, (no extra output came with -vvv either). It would be useful to know what it is doing. As I was able to make it hang for up to 10mins on larger packages.

ghost:
@bhcopeland Thanks for testing, 
- sudo: yes should definitely lead to an error inside the module.
- Printing something for projects which take a longer time will be slightly difficult I guess. As far as I know, ansible is not meant to output anything except using register and debug as additional step.

bhcopeland:
- Output. Agreed, I did try "debug=" but didn't show anything useful.
- Already installed packages. Granted, when you run yaourt from the command line, it will recompile the package, but in a ansible playbook this is messy. Yaourt, also alerts the user with "is up to date -- reinstalling". By default, it will recompile a package every time the playbook is ran (if the package is installed or not). ~~You should add a option for this, and assume by default if the package exists do not re-compile.~~

Thanks for your efforts on this, will be a very useful module :-)

ghost:
- About the Output, I will take a look at it to see whats possible
- About the reinstalling problem, I must admit that I'm not able to reproduce the problem you're facing. Using either the original pacman, my pacman or my yaourt module always results in a package not being installed if it already is. See the following output:

``` text
TASK: [Update Arch Packages (3/3)] ******************************************** 
ok: [localhost]

TASK: [debug var=io] ********************************************************** 
ok: [localhost] => {
    "var": {
        "io": {
            "changed": false, 
            "invocation": {
                "module_args": "update_cache=yes name=google-chrome", 
                "module_complex_args": {}, 
                "module_name": "yaourt"
            }, 
            "msg": "package(s) already installed"
        }
    }
}
```

If you want to specify what the program shall do, take a look at:

``` text
    state:
        description:
            - Desired state of the package.
        required: false
        default: "present"
        choices: ["present", "absent", "latest"]
```

bhcopeland:
Sorry TecSet for not being clear, it was when using `state: latest`.  With the pacman module it causes ansible to go "ok" and not to recompile/install. Same state should be given with using a aur.

ghost:
Sry for still being confused but is that a feature request or a bug you're facing.
To make it clear, I've tested state=latest and state=present and both do work with aur and non-aur being not installed if they are (except latest if package has an update)

As you can see in my last post, I had been using google-chrome which is an aur program. Installing it with present does not lead to a recompile/reinstalling.
Same for latest (Notice that I've used yaourt last time and pacman this time):

``` text
ok: [localhost] => {
    "var": {
        "io": {
            "changed": false, 
            "invocation": {
                "module_args": "update_cache=yes aur=yes name=google-chrome state=latest", 
                "module_complex_args": {}, 
                "module_name": "pacman"
            }, 
            "msg": "package(s) already installed"
        }
    }
}
```

bhcopeland:
So just tested it on my local env. If I run `state: present`, out of the 24 packages, 1 always re-installs. If I run `state: latest`, 7 packages re-install. I guess its just down the state of the PKGBUILD files.

ghost:
Ok, would you mind giving me some additional information?
What I need is a yaourt -Qi package_name and a yaourt -Si package_name for the 1 and the 7 packages.

bhcopeland:
Sure. Uploaded a [gist](https://gist.github.com/bhcopeland/7b8d3c54de769ae61019) for you.

bhcopeland:
Sure. Uploaded a [gist](https://gist.github.com/bhcopeland/7b8d3c54de769ae61019) for you.

ghost:
Thank you, that looks quite promising and interesting:
- About the 1 program which gets reinstalled by present: The Problem is, that there is no .git with that name. So yaourt cannot find its remote and returns an error. I will contact the yaourt creater and see what I can do to verbose a problem like that. 
- About the other 6 programs... well, seems like the versions you have installed are greater than the one's in the remote. My Module recognizes a Version differenz and leaves the rest to yaourt. Does yaourt update them by using yaourt -Sua? And what does yaourt say using yaourt -Sa package_name?

Dissonant-Tech:
needs_revision:

Using `upgrade` with  `--check` returns `(n+1) packages would be upgraded`, where `n` is the correct number of packages. Seems you forgot to remove the `\n` after `.split()` [here](https://github.com/ansible/ansible-modules-extras/blob/a7b1ef3f2db5800c7665207c5570f4eab93eb06b/packaging/os/yaourt.py#L167). 

**Command:** `yaourt: name=fasd state=present` **w/o** `sudo: yes`

Without `sudo: yes` it hangs. It hangs. Using htop on the server while the playbook was stalled I saw it was on the command `sudo -v` which is the final part of yaourt's install process and is run by `pacman -U` to install the package file.

Relevant journalctl:

```
ansible-yaourt[3093]: Invoked with upgrade=False force=False name=['fasd'] recurse=False state=present update_cache=False
admin : a password is required ; TTY=pts/2 ; PWD=/tmp/yaourt-tmp-admin/aur-fasd ; USER=root ;COMMAND=list /usr/bin/pacman --color never -U --noconfirm /tmp/yaourt-tmp-admin/PKGDEST.AKd/fasd-1.0.1-1-any.pkg.tar.xz
```

**Command:** Any **w/** `sudo: yes`

**Result:** Any command with `sudo: yes` also fails for me.

`yaourt: name=fasd state=present` returns: `FAILED! => {"changed": false, "failed": true, "msg": "failed to install fasd"}`.

`yaourt: update_cache=yes upgrade=yes` erroneously returns "System upgraded" without actually upgrading anything.

```
ansible-yaourt[26175]: Invoked with upgrade=True force=False name=None recurse=False state=present update_cache=True
```

I'm sorry this was so long, I was typing it as I was troubleshooting but a **tl;dr**:
- `sudo: no` hangs when `pacman -U` (used by yaourt to install packages) tries to authenticate with `sudo -v`
- `sudo: yes` also fails for me, though I can't exactly pinpoint why (likely because yaourt cannot be run as sudo)
- `sudo: yes` erroneously returns `changed` when using upgrade, and correctly returns `failed` when trying to install
- Your pacman.py module has the same errors for me

ghost:
@Dissonant-Tech  Thanks for your testing!
- upgrade with --check returns (n+1) packages would be upgraded
  Fixed
- sudo: no hangs
  Sry but that looks more like an User Error,  instead of a module Bug. Does yaourt install the package when you try it manualy? I'm only calling yaourt and not responsible for the Program to function itself.
- sudo: yes also fails
  Expected Result: It'll be made more verbose in time

Dissonant-Tech:
@TecSet Thanks for your work.

Manually installing packages with yaourt works fine, but yaourt does ask for the sudo password, which is where I see the process hang.

Example with `yaourt -S fasd --noconfirm`:

```
==> Continue installing fasd ? [Y/n]
==> [v]iew package contents [c]heck package with namcap
==> ---------------------------------------------------
==>
[sudo] password for USER:
```

The final `[sudo]` prompt is only skipped if you've recently `sudo`-ed and the session has not expired.

ghost:
@Dissonant-Tech Ah, now I got the problem. 

Unfortunately, I won't be able to fix that, 'cause modules are not able to handle passwords. 
The only thing that might work is using --ask-sudo-password and sudo:no. 
If that doesn't either, workaround would be to create a user which can sudo without password and using become.

vonpupp:
I would like to test this PR however I'm using Ansible installed via the system package manager (pacman). Is there a way to test only this module within my current playbook or system installed Ansible?

I think having a yaourt module it is very important for the Arch community, and it is an unsolved problem I'm currently going trough. So congrats on the PR!

Thanks.

bcoca:
@vonpupp download the file to a 'library' dir adjacent to the play you want to use it in (or inside role)

vonpupp:
Thank you @bcoca. I confirm that: works_for_me, so :+1:.

There are also other wrappers that accept pacman arguments like: yaourt, pacserve and powerpill, so perhaps some sort of inheritance from the pacman module would be a good idea. I personally _really would like an official_ yaourt solution within the extras modules. I have seen other solutions/hacks like [1](https://github.com/cdown/ansible-yaourt/) and [2](https://gist.github.com/wrecker/39ecb1eb1ab8ee1d0ce1).

@Dissonant-Tech, what you could do to solve that is to create a dedicated build user like "makepkg" and give him sudo powers passwordless. I have seen this already on github, you can check it on [Haevas.yaourt](https://github.com/Haevas/Haevas.yaourt), in which I basically used [Stouts.sudo](https://github.com/Stouts/Stouts.sudo) as follows:

```
sudo_users:
  - makepkg
sudo_passwordless:
  - makepkg
```

I hope this helps!

Regards.

coxley:
:+1:

bhcopeland:
Can someone explain why yaourt stuff was never merged this PR is confusing :+1:

PR #1109, opened 2015-10-15
New module - ec2_vol_facts

wimnat:
Module ec2_vol_facts.  Although ec2_vol module contains a "list" state, this state has been discouraged so if this module is merged then list state in ec2_vol could be deprecated.

This module goes beyond the list state by allowing vols returned to be filtered as per the amazon doc - http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeVolumes.html

### DIFF
--- a/cloud/amazon/ec2_vol_facts.py
+++ b/cloud/amazon/ec2_vol_facts.py
@@ -1,0 +1,135 @@
+#!/usr/bin/python
+#
+# This is a free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This Ansible library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this library.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: ec2_vol_facts
+short_description: Gather facts about ec2 volumes in AWS
+description:
+    - Gather facts about ec2 volumes in AWS
+version_added: "2.0"
+author: "Rob White (@wimnat)"
+options:
+  filters:
+    description:
+      - A dict of filters to apply. Each dict item consists of a filter key and a filter value. See U(http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeVolumes.html) for possible filters.
+    required: false
+    default: null
+extends_documentation_fragment:
+    - aws
+    - ec2
+'''
+
+EXAMPLES = '''
+# Note: These examples do not set authentication details, see the AWS Guide for details.
+
+# Gather facts about all volumes
+- ec2_vol_facts:
+
+# Gather facts about a particular volume using volume ID
+- ec2_vol_facts:
+    filters:
+      volume-id: vol-00112233
+
+# Gather facts about any volume with a tag key Name and value Example
+- ec2_vol_facts:
+    filters:
+      "tag:Name": Example
+
+# Gather facts about any volume that is attached
+- ec2_vol_facts:
+    filters:
+      attachment.status: attached
+
+'''
+
+try:
+    import boto.ec2
+    from boto.exception import BotoServerError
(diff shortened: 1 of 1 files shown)


robynbergeron:
Hi @wimnat --

robynbergeron:
Pinging our lovely AWS crew for feedback -- @jsmartin @scicoin-project @tombamford @garethr @lorin @jarv @jsdalton @silviud @adq @zbal @zeekin @willthames @carsongee @defionscode

And, pinging @lwade in particular for feedback, as maintainer of the ec2_vol module in core.

@wimnat -- is it a "could" or "should" be deprecated wrt list state in ec2_vol -- and if this is merged, are you planning to do a PR for that? (Apologies if you already have and it's not ringing a bell in my wee little memory :D)

robynbergeron:
I promise to all of you that I know the difference between windows and AWS. Except, apparently, when it comes to adding labels. Or when it's 4am. :person_with_pouting_face:

wimnat:
I've been told in the past that facts modules are preferred over a list
state so I would guess this module could lead to deprecation of ec2_vol
list state but I will leave that to @bcoca and the rest of the core team to
answer.

willthames:
@wimnat is that policy actually written down? 

I've tried to implement list state in modules before and been rejected, but it would be good if it were clarified somewhere. The filters selection is definitely a good feature, I just worry about a massive proliferation of modules and wonder if there isn't a better approach

wimnat:
@willthames It's just been mentioned to me either here or in the ansible-devel google group.  It would probably be a good idea to add it to the module guidelines, if it is indeed gospel.  (Another thing i've been told here is that state should never have a default parameter but i don't think that's in the guidelines either)

I personally really favor separate modules for facts.  It particularly works for Amazon infrastructure with the filters dicts making it easy to grab particular info like a vpc id or snapshots tagged with a certain something.

I would also favor more modules that are more specific over fewer, more complex modules.

tastychutney:
@wimnat That would be a change as well from what I've been told in the past. I had several facts modules rejected because "that's what inventory scripts are for". Never mind that they don't work for my use cases, which are AWS-based as well. :-)

gregdek:
Going to defer to the core team here for policy ruling.

cc @ansible/core

willthames:
@gregdek here is my opinion, for reference from an @ansible/aws point of view. 

I think @wimnat makes a bunch of good points - my main issue is that they should be written down with the guidelines. I'm happy enough to deprecate `state=list` on that basis, and replace it with this module (which has far better functionality due to the extra filtering that would complicate the ec2_vol module)

bcoca:
already there:
" Sometimes you want to split the module, specially if you are adding a list/info state, you want a _facts version"

willthames:
@bcoca - you're quite right, sorry. I think the statement could be stronger, but it is there. 

The only thing this is missing before passing guidelines is the RETURNS documentation structure.

bcoca:
Return is not mandatory, yet. Feel free to submit PR that you think clarifies the guidelines more.

willthames:
"Return: document the return structure of the module"

Feel free to update the documentation to clarify that that's not mandatory

willthames:
Raised ansible/ansible#12957 to clarify guidelines

wimnat:
@willthames, happy to ship this?

willthames:
@wimnat in general I think I'd prefer to have the RETURN documentation but it's not a blocker if @bcoca thinks it's not mandatory. And it can always be added in a future PR.

:shipit:

robynbergeron:
@wimnat @willthames -- as a friendly reminder, for new modules we need the +1 votes for guidelines and working (see https://github.com/ansible/ansible-modules-extras/pull/1109#issuecomment-149851917 above). Not because I <3 being picky, but just for consistency's sake :smile: 

If @willthames is willing to put up those votes I'm happy with shipping this, since as noted above the RETURN documentation isn't a blocker (though I would encourage filing a feature request against the module once it has shipped as a reminder for future things to do :D)

Let us know!

willthames:
passes_guidelines

willthames:
I don't have time right now to test if it works I'm afraid - I might be able to come back to it later.

manicminer:
Tested most of the documented filters.

works_for_me

wimnat:
@robynbergeron with the works_for_me and the passes_guidelines can this get shipped?

gregdek:
@wimnat I think the answer is yes. Thanks for your perseverance. :) 

Marking as "shipit" at long last.

Issue #879, opened 2015-08-26
ZFS volsize idempotence


mavit:
##### Issue Type:
- Bug Report
##### Ansible Version:

```
ansible 1.9.1
  configured module search path = None
```
##### Environment:

Solaris 10 and 11.
##### Steps To Reproduce:

The following task will return `changed` every time you run it, because it naively compares the volsize property to the string returned by the `zfs get` command:

```
- zfs: name=rpool/swap state=present volsize=2147483648
```

By default, the output of the `zfs get` command is formatted in human-readable style, meaning that the string that you set isn't necessarily the same string that you'll get.  For example:

```
> sudo zfs set volsize=6145M rpool/dump
> zfs get volsize rpool/dump
NAME        PROPERTY  VALUE  SOURCE
rpool/dump  volsize   6.00G  local
```

This confuses the Ansible zfs module.  

My particular use-case is that I wish to set the dump volume to be half the memory size, without having to do something messy with Jinja2 to format that into, say, a whole number of gigabytes.

I propose that the Ansible zfs module passes the `-p` option to `zfs get`, causing it to format sizes as integers.  This won't entirely solve the problem, but at least it'll be predictable what format I need to use to get idempotence.  If this sounds reasonable, I can create a pull request.

chrrrles:
This is tricky, as volsize could be an integer (100000), or an integer with a character at the end (10M).  If volsize is input in the latter form, will we be able to reliably calculate the true value for `volsize` that zfs will set / return?  If not, I lean towards restricting the volsize argument to integers-only.

chrrrles:
@mavit - Would you feel comfortable creating a pull request for this?

ansibot:
@johanwiren, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@johanwiren, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@johanwiren, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

johanwiren:
needs_contributor

ansibot:
This repository has been locked. All new issues and pull requests should be filed in https://github.com/ansible/ansible

[...]

PR #1121, opened 2015-10-19
Enable the use of different number formats e.g. 2G, 2048M or 2147483648

ag-wood:
Addresses issue #879 while preserving current functionality (i.e. no restriction on number format)


system/zfs.py
```diff
@@ -237,6 +237,36 @@
 
 import os
 
+def to_bytes(value):
```
bcoca: these functions would belong in a lib/ansible/module_utils/basic.py (see existing pretty_bytes)

ag-wood:
Thanks @bcoca for the suggestion re: module_utils.   Do I just submit a separate pull request to ansible/ansible?  Also, not being terribly experienced with how things get done, am curious about how the separate requests would get merged as this creates a dependency for the inclusion of functionality in the zfs module.  Sorry if these are dumb questions.

Yannig:
Hi @ag-wood,

I have already submited a pull request matching this needs at the following location: https://github.com/ansible/ansible/pull/12074

I'm using an ansible filter to convert size strings into bytes.

What do you think about @bcoca?

ag-wood:
Does that help in the context that I need it?  Does a filter get used like other functions within a module ?i.e. I wanted to use the function for testing a value passed into the zfs module.

ag-wood:
Based on @bcoca 's recommendation, I was planning to submit the following two functions:

```
def to_bytes(self, size):
    if isnumeric(size):
        return size
    suffixes = ['B', 'K', 'M', 'G', 'T', 'P', 'E', 'Z', 'Y']
    index = next(i for i, d in enumerate(size) if not d.isdigit())
    if index > 0:
        value = size[:index]
        suffix = (size[index:][0]).upper()
        if suffix in suffixes:
            power = suffixes.index(suffix)
            return int(float(value) * (1024 ** power))
    return None

def isnumeric(self, value):
    try:
        v = float(value)
        return True
    except:
        return False
```

bcoca:
i had forgotten about the filters, they can be used as input to the modules, not from the modules themselves. Having the fitlers I don't think the modules themselves need to handle these things.

also the is_numeric function is just check for floats, that can be handled by setting the type in the module spec to float.

ag-wood:
The is_numeric and to_bytes function are purely to manage the string values that can be passed (i.e. to check whether the string is numeric as opposed to any of the other allowable string values i.e. 2G, 2048M etc.)  I don't want to restrict the input to a numeric value as this hamstrings the module and requires the use of very long numbers that are easy to get mixed up.

Essentially, I was trying to add some flexibility to satisfy issue #879 without making it more difficult to use, which I think would be the outcome if we force numeric inputs for the size values.

I'm also having trouble seeing how the human_bytes filter can be used within the module for post processing inputs in the manner described above.

bcoca:
im saying that the 'general interface' will allow for this while not needing to add this complexity to modules which can take a stricter more specific input:

`size: "{{20G|to_bytes}}"`

having filters that do this removes the need for modules to handle it

ag-wood:
It still seems to be less straight forward than handling it within the module.  I get that an Ansible savvy operator might fall into the pattern you are suggesting, but I don't see that it fits with the 'Ansible is simple' mantra.

I'm willing to concede defeat and bow to the greater wisdom - so the summary then is that you prefer to change the interface (force numerics for quota, maybe volsize...) which may break existing playbooks.  I'm not prepared to make that sort of call as I'm not sure how backward compatibility should work.

bcoca:
or let me completely reverse my position, create a 'bytesize' module type that does this as part of the module spec

ag-wood:
So, do you mean double up on each parameter that can be represented as a string or numeric?

bcoca:
no, just have a type='bytes', which accepts strings and returns a normalized 'byte size'

Yannig:
@bcoca: It's a good idea.

I have start working on a little patch:

``` diff
diff --git a/lib/ansible/module_utils/basic.py b/lib/ansible/module_utils/basic.py
index 60fe323..cffad59 100644
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -470,6 +470,7 @@ class AnsibleModule(object):
                 'int': self._check_type_int,
                 'float': self._check_type_float,
                 'path': self._check_type_path,
+                'bytes': self._check_type_bytes,
             }
         if not bypass_checks:
             self._check_required_arguments()
@@ -1210,6 +1211,8 @@ class AnsibleModule(object):
         value = self._check_type_str(value)
         return os.path.expanduser(os.path.expandvars(value))

+    def _check_type_bytes(self, value):
+        return self.string_to_bytes(value)

     def _check_argument_types(self):
         ''' ensure all arguments have the requested type '''
@@ -1806,22 +1809,47 @@ class AnsibleModule(object):
         fh.write(str)
         fh.close()

+    size_ranges = (
+        (1<<80L, 'YB'),
+        (1<<70L, 'ZB'),
+        (1<<60L, 'EB'),
+        (1<<50L, 'PB'),
+        (1<<40L, 'TB'),
+        (1<<30L, 'GB'),
+        (1<<20L, 'MB'),
+        (1<<10L, 'KB'),
+        (1, 'Bytes')
+    )
     def pretty_bytes(self,size):
-        ranges = (
-                (1<<70, 'ZB'),
-                (1<<60, 'EB'),
-                (1<<50, 'PB'),
-                (1<<40, 'TB'),
-                (1<<30, 'GB'),
-                (1<<20, 'MB'),
-                (1<<10, 'KB'),
-                (1, 'Bytes')
-            )
-        for limit, suffix in ranges:
+        for limit, suffix in size_ranges:
             if size >= limit:
                 break
         return '%.2f %s' % (float(size)/ limit, suffix)

+    def string_to_bytes(size):
+        ''' Convert size in string format into bytes (ex: '2K' => 2048) '''
+        m = re.search('^\s*(\d*\.?\d*)\s*([A-Z])?', size, flags=re.IGNORECASE)
+        if m is None:
+            raise ValueError("string_to_bytes() can't interpret following string: %s" % str(size))
+        try:
+            num = float(m.group(1))
+        except:
+            raise ValueError("string_to_bytes() can't interpret following number: %s (original input string: %s)" % (m.group(1), size))
+
+        unit = m.group(2)
+        if unit is None:
+            ''' No unit given, returning raw number '''
+            return int(round(num))
+
+        unit = unit[0].upper() + 'B'
+
+        for limit, suffix in size_ranges:
+            if unit == suffix:
+                break
+        if unit != 'Bytes' and limit == 1:
+            raise ValueError("string_to_bytes() can't interpret following string: %s" % size)
+        return int(round(num * limit))
+
     #
     # Backwards compat
     #
```

I was starting to port lvol to this new type and find that size can be expressed in percent (50%).

@bcoca: what do you think about? Do we need to create a new type like bytes_or_percent?

robynbergeron:
Hi @ag-wood (thanks for the PR!), @Yannig (thanks for the feedback and the other PR), @bcoca -- thanks for looking at this thus far. 

Also pinging @johanwiren (zfs module maintainer) to let you know about this PR. Johan - it looks like this may be redone as a wider PR to ansible/ansible -- but just putting this on your radar just in case, and for feedback anyway :)

Thanks, everyone! :)

Yannig:
I have done some rework against my PR (https://github.com/ansible/ansible/pull/12074) in order to be able to call the same function from AnsibleModule and string_to_bytes from module_utils.

@bcoca, @ag-wood: what do you think about?

gregdek:
Thanks @ag-wood for this PR. Unfortunately, it is not mergeable in its current state due to merge conflicts. Please rebase your PR. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

gregdek:
@ag-wood A friendly reminder: this pull request has been marked as needing your action. If you still believe that this PR applies, and you intend to address the issues with this PR, just let us know in the PR itself and we will keep it open pending your changes.

ag-wood:
Closing this as I don't have the energy to convince anyone that my approach was worth implementing.

PR #1128, opened 2015-10-20
no-suggests is obsolete, switch to no-recommends

pmakowski:


robynbergeron:
Hi @pmakowski -- thanks for submitting this PR.

Marking as owner_pr and thus flagging for inclusion -- pinging @ansible/core team for the magic.

PR #1132, opened 2015-10-21
[Bug] exchange name contains chars that needs to be quoted

caduri:


robynbergeron:
Hi @caduri -- thanks for this PR.

@manuel-sousa -- Please review this PR to make sure it adheres to the following guidelines:

manuel-sousa:
shipit

@caduri what exchange were you trying to create for this to become an issue? 
I've made a quick test with allowed characters (https://www.rabbitmq.com/specification.html) and didn't give me any issue in my environment.

> The exchange name consists of a non-empty sequence of these characters: letters, digits, hyphen, underscore, period, or colon.
> Notes: The lexicon is not enforced by the server.

The submited patch shouldn't produce side effects on working configurations and might allow some names since the lexicon isn't enforced by RabbitMQ. As such i think we should allow it.

caduri:
Try exchange name with / like "ansible/input"

manuel-sousa:
According to specification / isn't in the allowed characters for the exchanges. However RabbitMQ doesn't enforce it and as such i think the plugin shouldn't either which is why I recomended to shipit.

Those names might cause issues elsewhere so use with caution.

caduri:
I weren't aware of that, thanks.

gregdek:
Thanks @caduri for the PR, and thanks @manuel-sousa for the detailed review. Marking for inclusion.

PR #1134, opened 2015-10-21
composer arguments

ramunasd:
This adds additional option 'arguments' for composer. Sometimes it's very useful with commands init, require, create-project and etc.

### DIFF
--- a/packaging/language/composer.py
+++ b/packaging/language/composer.py
@@ -133,8 +133,6 @@ def parse_out(string):
     return re.sub("\s+", " ", string).strip()
 
 def has_changed(string):
-    if string == "":
-        return False
     return "Nothing to install or update" not in string
 
 def get_available_options(module, command='install'):
--- a/packaging/language/composer.py
+++ b/packaging/language/composer.py
@@ -36,6 +36,11 @@ options:
             - Composer command like "install", "update" and so on
         required: false
         default: install
+    arguments:
+        description:
+            - Composer arguments like required package, version and so on
+        required: false
+        default: null
     working_dir:
         description:
             - Directory of your project ( see --working-dir )
@@ -102,6 +107,17 @@ notes:
 EXAMPLES = '''
 # Downloads and installs all the libs and dependencies outlined in the /path/to/project/composer.lock
 - composer: command=install working_dir=/path/to/project
+
+- composer:
+    command: "require my/package"
+    working_dir: "/path/to/project"
+
+# Clone project and install with all dependencies
+- composer:
+    command: "create-project"
+    arguments: "package/package /path/to/project ~1.0"
+    working_dir: "/path/to/project"
+    prefer_dist: "yes"
 '''
 
 import os
@@ -116,6 +132,8 @@ def parse_out(string):
     return re.sub("\s+", " ", string).strip()
 
 def has_changed(string):
+    if string == "":
+        return False
     return "Nothing to install or update" not in string
 
 def get_available_options(module, command='install'):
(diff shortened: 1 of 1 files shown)


ramunasd:
thanks @resmo, but how can i know to which version this will be added?

resmo:
set it to "2.0" for now

packaging/language/composer.py
```diff
@@ -116,6 +132,8 @@ def parse_out(string):
     return re.sub("\s+", " ", string).strip()
 
 def has_changed(string):
+    if string == "":
```
resmo: IMHO this is not needed, why did ou make it?

robynbergeron:
Hi @ramunasd -- thanks for this PR. And thanks, @resmo for your quick review, as always :)

@resmo is one of the module maintainers, but adding the below just for process / clarity. @ramunasd -- I'm putting this in needs_revision, per @resmo's questions. :)

@resmo @dmtrs -- Please review this PR to make sure it adheres to the following guidelines:

packaging/language/composer.py
ramunasd: Sometimes composer returns nothing, but ansible thinks that something changed, because empty string doesn't have string 'Nothing to install...'.

Maybe this condition should be improved somehow, but i'm not sure.

resmo: hmm I see. Do you have an example?

Generally we should return `changed=True` if we are not sure that something has changed, also for commands executed `changed=True` makes sense.

ramunasd: After deeper investigation i see only 'composer init' for existing project returns empty string. So this check mainly for command 'init'.

I can remove this check if You see something bad.

resmo: I would remove it. composer init would return changed=True which is what we want. So please remove it.

gregdek:
@ramunasd is this now ready for review, in your opinion? If so,  please add a comment with the text "ready_for_review" and we will continue with the review process.  Thanks!

ramunasd:
@gregdek Hi, it's ready_for_review

gregdek:
Thanks @ramunasd. 

@dmtrs @resmo Please review this PR to make sure it adheres to the following guidelines:

resmo:
Ok, I think this change needs even to go further, because it actually fixes the command arguments handling. 

We expect the command to be a composer command but it is actually a string (so you are allowed by the module to add composer commands with arguemnts. e.g. 

```
- composer:
    command: "require my/package"
    working_dir: "/path/to/project"
```

But to find out all options available we did not expect this to be possible, it can cause this to fail. 

So we should add a check that command is a string not having a whitespace in it (e.g. no arguments). So the check for the options would only fail if the command is unknown, which is what we want.

So the example should be:

```
- composer:
   command: "require"
   arguments: " my/package"
   working_dir: "/path/to/project"
```

So in short:
Please extend this with a check to make sure, command has no whitespace (no arguments)

```
import re
if re.search(r"\s", command):
    module.fail_json(msg="Use the 'arguments' param for passing arguments with the 'command'")
```

ramunasd:
Hi @resmo, check added.

packaging/language/composer.py
```diff
@@ -102,6 +108,17 @@
 EXAMPLES = '''
 # Downloads and installs all the libs and dependencies outlined in the /path/to/project/composer.lock
 - composer: command=install working_dir=/path/to/project
+
+- composer:
+    command: "require my/package"
```
resmo: Please fix this example with the new `arguments` param

resmo:
shipit

abadger:
Thanks @ramunasd and @resmo!

PR #1137, opened 2015-10-21
.exists() deprecated in zabbix 3.0 API calls have been removed

drrtuy:
There is a deprecated zabbix api function call in the zabbix-group.py code. Ansible module doesn't work with Zabbix 3.0 server. hostgroup.exists call must be droped according to [ ZBXNEXT-2724](https://support.zabbix.com/browse/ZBXNEXT-2724).

### DIFF
--- a/monitoring/zabbix_group.py
+++ b/monitoring/zabbix_group.py
@@ -130,11 +130,10 @@ def delete_group(zbx, group_id):
 
 def check_group(zbx, host_group):
     try:
-        result = zbx.hostgroup.exists(
-            {
-                'name': host_group
-            }
-        )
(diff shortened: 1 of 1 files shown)


robynbergeron:
Hi @drrtuy -- thanks for this PR.

@cove @harrisongu -- Please review this PR to make sure it adheres to the following guidelines:

drrtuy:
shipit

bcoca:
does this break against older versions of zabbix or was the alternate call always available?
if not a try/except using one call and falling back to the other might be a better fix.

bcoca:
@drrtuy you do not need to "shipit" on your own PR, that is implied by the PR already

drrtuy:
This function is available in zabbix API with the same syntax since 1.8.

PR #1025, opened 2015-09-24
remove ecs facts and fix docs for ecs_taskdefinition module

bennojoy:
remove ecs facts and fix docs for ecs_taskdefinition module

PR #1144, opened 2015-10-22
fix documentation contents and format

Java1Guy:
Apparently it can't auto-merge because the file was deleted in the devel branch...

### DIFF
--- a/cloud/amazon/ecs_task_facts.py
+++ b/cloud/amazon/ecs_task_facts.py
@@ -40,24 +40,35 @@ options:
             - Set this to a list of task identifiers.  If 'details' is false, this is required.
         required: false
     family:
+        description:
+            - the family of tasks to describe
         required: False
         type: str
-
     container_instance:
+        description:
+            - what container's tasks to describe
         required: False
         type: 'str'
     max_results:
+        description:
+            - how many results to return
         required: False
         type: 'int'
     started_by:
+        description:
+            - who started the task
         required: False
         type: 'str'
     service_name:
+        description:
+            - the service of the tasks
         required: False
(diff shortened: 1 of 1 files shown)


sivel:
You are correct.  That module is no longer part of Ansible.  It was removed in https://github.com/ansible/ansible-modules-extras/pull/1025

Java1Guy:
Indeed it was -- the $10,000 question is why?  My speculation was because the documentation was hosed...
fyi - @defionscode

sivel:
@bennojoy can probably explain more, but it seems specifically targeted at removing the facts plugins.  I know there were questions about their inclusion before merging, my guess it was related to that.

Java1Guy:
I guess we'll wait to hear from @bennojoy  - but, if it was due to docs issues, i believe they are resolved (at least i have run all the ecs modules through ansible-doc now).

gregdek:
This is probably a good question to ask in ansible-devel. Closing for now; if there's a reason to reopen down the line, we can do that. For now, though, the module no longer exists.

bennojoy:
@gregdek  @Java1Guy The main reason was i had less bandwidth to check the facts module and fix them and since both modules were in one commit i had to remove them. It is in my todo list to add them back and also we are also thinking if these should be seperate modules or just have one module ecs_facts. Also most of the facts are returned by the module itself to have a working ecs cluster setup but if you think the facts modules are absolutely mandatory i can give priority to add back those modules.

Java1Guy:
I was mostly going with what I perceived to be the convention already established.  The true test will come from anybody actually trying to use the modules, so we'll see.  Thanks for the reply.

PR #1145, opened 2015-10-22
support proxmox ve version 4

frogu:
Support for openvz was dropped in PVE 4.0. This change discovers the version running on the node and chooses the appropriate API call to make.


frogu:
sorry, left some dumb debugging in it.

PR #1147, opened 2015-10-22
pkgutil improvements

jaymell:
More consistent use of run_command() wrapper function -- all former
calls to module.run_command() now use this function, which now 
inserts the default path to the binary in the paths to search -- this is 
beneficial b/c the binary is not in the default $PATH directories, so the 
module would previously fail if you hadn't previously 
modified $PATH on your servers.

Tested on fresh installs of Solaris 10 and 11

### DIFF
--- a/packaging/os/pkgutil.py
+++ b/packaging/os/pkgutil.py
@@ -63,10 +63,10 @@ import os
 import pipes
 
 def package_installed(module, name):
-    cmd = [module.get_bin_path('pkginfo', True)]
+    cmd = ['pkginfo']
     cmd.append('-q')
     cmd.append(name)
-    rc, out, err = module.run_command(' '.join(cmd))
+    rc, out, err = run_command(module, cmd)
     if rc == 0:
         return True
     else:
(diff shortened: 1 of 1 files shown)


gregdek:
@dermute Please review this PR to make sure it adheres to the following guidelines:

jaymell:
shipit

bcoca:
@jaymell authoring the PR already lets us know you want to shipit, voting on it yourself does not count extra

jaymell:
@bcoca Derp. Apologies, still new at this -- thanks!

packaging/os/pkgutil.py
```diff
@@ -63,10 +63,10 @@
 import pipes
 
 def package_installed(module, name):
-    cmd = [module.get_bin_path('pkginfo', True)]
+    cmd = ['pkginfo']
```
bcoca, line 66: this is not a good change, get_bin_path makes sure it finds a pkginfo or returns a module fatal error, now you assume the pkginfo is in path and the run_command must deal with it not being there

jaymell, line 66: I definitely see your point. I was trying to make the module's current code more consistent -- notice that it only uses module.get_bin_path in the package_installed() function and in all the other 'package_' functions does what you're criticizing.

Your comments suggest that it may be better to remove the run_command function entirely, since all it effectively does is do a module.get_bin_path for calling functions. Do you think this is a better way to go? The only downside I see to this is adding several redundant lines of code where I do this in each function that runs a command:

cmd = [ module.get_bin_path(COMMAND, True, PATH) ]
cmd.extend(ADDITIONAL COMMAND ARGS)

Is it not better to do it in just one place?

-- Thanks

bcoca, line 66: ah, i stopped at this change, i see what you are doing in later in the local run_command, the original run_command does not make much sense to me either, but this seems like a smaller change over adding the extra search path to each one.

PR #1149, opened 2015-10-23
Enable stdout and stderr on sucessful runs, making show_diff useable

Jmainguy:
This adds stdout and stderr to the module, which makes the show_diff choice useful, as it will print it out now.

@emonty Let me know what you think

### DIFF
--- a/system/puppet.py
+++ b/system/puppet.py
@@ -171,8 +171,8 @@ def main():
 
     if not p['manifest']:
         cmd = ("%(base_cmd)s agent --onetime"
-               " --ignorecache --no-daemonize --no-usecacheonfailure"
-               " --no-splay --detailed-exitcodes --verbose") % dict(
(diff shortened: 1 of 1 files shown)


bcoca:
cc @emonty

Jmainguy:
@emonty @bcoca whats the deal, Why was https://github.com/ansible/ansible-modules-extras/pull/1260  merged but this wasn't. Would really like some feedback.

bcoca:
Merged cause it had shipit label, I wasn't aware of overlap either or would
have raised the question.

Jmainguy:
Cool, that makes sense. Looks like this still adds some additional stdout, stderr, and color=0 to the module. So I will keep this open. @emonty can you please review this PR?

gregdek:
Trivial. Adding shipit myself.

PR #1150, opened 2015-10-23
Change show_diff to default to yes, to match what puppet agent --test…

Jmainguy:
… is, since the rest of the options defaulted to on are grabbed from --test

@emonty What do you think. From the puppet-agent man page.

```
 --test Enable  the most common options used for testing. These are ´onetime´, ´verbose´, ´ignorecache´, ´no-daemonize´, ´no-usecacheonfailure´, ´detailed-exit-codes´, ´no-splay´,
              and ´show_diff´.
```

### DIFF
--- a/system/puppet.py
+++ b/system/puppet.py
@@ -45,9 +45,9 @@ options:
     default: None
   show_diff:
     description:
-      - Should puppet return diffs of changes applied. Defaults to off to avoid leaking secret changes by default.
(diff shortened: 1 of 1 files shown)


Jmainguy:
cc @emonty  just in case it doesn't email you from the opening comment =)

emonty:
shipit

PR #1180, opened 2015-10-30
Add insert support to iptables.

### LINKED ISSUE #2167: iptables module: add possibility to INSERT rules (right now is only APPEND)
##### ISSUE TYPE

- Feature Idea
##### COMPONENT NAME

iptables
##### ANSIBLE VERSION

```
ansible 2.0.1.0
```
##### OS / ENVIRONMENT

N/A

dvigueras:
This PR adds support to be able to insert new rules in iptables (iptables -I).

I think @LinusU should review this PR as he is the author of the iptables module.

Fixes #2167

### DIFF
--- a/system/iptables.py
+++ b/system/iptables.py
@@ -388,7 +388,7 @@ def main():
         argument_spec=dict(
             table=dict(required=False, default='filter', choices=['filter', 'nat', 'mangle', 'raw', 'security']),
             state=dict(required=False, default='present', choices=['present', 'absent']),
-            insert=dict(required=False, default=False, type='bool', choices=BOOLEANS),
+            action=dict(required=False, default='append', type='str', choices=['append', 'insert']),
             ip_version=dict(required=False, default='ipv4', choices=['ipv4', 'ipv6']),
             chain=dict(required=True, default=None, type='str'),
             protocol=dict(required=False, default=None, type='str'),
@@ -427,7 +427,7 @@ def main():
         rule=' '.join(construct_rule(module.params)),
         state=module.params['state'],
     )
-    insert = module.params['insert']
+    insert = (module.params['action'] == 'insert')
     ip_version = module.params['ip_version']
     iptables_path = module.get_bin_path(BINS[ip_version], True)
     rule_is_present = check_present(iptables_path, module, module.params)
--- a/system/iptables.py
+++ b/system/iptables.py
@@ -56,6 +56,14 @@ options:
     required: false
     default: present
     choices: [ "present", "absent" ]
+  insert:
+    description:
+      - Whether the rule should be inserted at the top or not. If the rule
+        already exists in the chain the rule is deleted and it is inserted
+        at the top of the chain.
+    required: false
+    default: no
+    choices: [ "yes", "no" ]
   ip_version:
     description:
       - Which version of the IP protocol this rule should apply to.
@@ -364,6 +372,11 @@ def append_rule(iptables_path, module, params):
     module.run_command(cmd, check_rc=True)
 
 
+def insert_rule(iptables_path, module, params):
+    cmd = push_arguments(iptables_path, '-I', params)
+    module.run_command(cmd, check_rc=True)
+
+
 def remove_rule(iptables_path, module, params):
     cmd = push_arguments(iptables_path, '-D', params)
     module.run_command(cmd, check_rc=True)
@@ -375,6 +388,7 @@ def main():
         argument_spec=dict(
             table=dict(required=False, default='filter', choices=['filter', 'nat', 'mangle', 'raw', 'security']),
             state=dict(required=False, default='present', choices=['present', 'absent']),
+            insert=dict(required=False, default=False, type='bool', choices=BOOLEANS),
             ip_version=dict(required=False, default='ipv4', choices=['ipv4', 'ipv6']),
             chain=dict(required=True, default=None, type='str'),
             protocol=dict(required=False, default=None, type='str'),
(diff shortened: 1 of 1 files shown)


bcoca:
what is the use case here? 
would the same behavior not be achieved with ensuring that the rule is absent and then present?

system/iptables.py
```diff
@@ -220,7 +234,6 @@ def append_param(rule, param, flag, is_list):
         if param is not None:
             rule.extend([flag, param])
 
-
```
LinusU, line 300: Why remove this line?

system/iptables.py
```diff
@@ -296,12 +315,13 @@ def main():
         state=module.params['state'],
     )
     ip_version = module.params['ip_version']
+    insert     = module.params['insert']
```
LinusU: PEP 8 don't like the multiple spaces, personally I would have removed them and also put this line above the previous one...

dvigueras:
It isn't the same behavior.
Suppose the following scenario:

Iptables INPUT rule list:
- Rule A
- Rule B

If we ensure rule A is absent, and then present, the result would be:
- Rule B
- Rule A

But what we can achieve with this new insert feature is to be able to insert the rule A at the top of the chain, so with insert=yes the result would be:
- Rule A
- Rule B.

LinusU:
I had a few nits but the code looks good. I have two questions though.

Why is it named `insert`?

What is the use case?

bcoca:
so basically this is insert=top/bottom with bottom being the default?

LinusU:
Ah, now I understand why :)

Could we have a parameter that says wether to insert it at the top or the bottom instead?

system/iptables.py
dvigueras, line 300: It was a typo, fixed.

system/iptables.py
dvigueras: Reordered vars assignment and removed extra spaces.

dvigueras:
I think there are two options:
- Option A) As it is defined now. There is a boolean insert parameter which controls how iptables should be called. insert=yes means iptables will be called with the -I option, and insert=no (default value) means iptables will be called with the -A option.
- Option B) As @bcoca says. There is a string insert parameter which controls how iptables should be called, the possible values are top and bottom. insert=top means iptables will be called with the -I option, and insert=bottom (default value) means iptables will be called with the -A option.

I think both options are valid. Opinions?

gregdek:
@LinusU thanks for your input so far. Please be sure to review this PR to make sure it adheres to the following guidelines:

LinusU:
Hmm, one more thing that I don't like. `insert` moves the rule to the top should it already exist, `append` on the other hand doesn't move the rule to the bottom if it already exists. That feels very unintuitive to me.

What I propose:

One parameter named `action`, with the possible values: `append`, `insert`.

Possibly an extra parameter that controls wether or not the rule should be relocated if it isn't at the desired location. If we implement this, it needs to output correct change information, i.e. if the rule already was in the correct position, it should not be marked as changed.

Do we need the second feature? Could we split it into two pull requests with two different discussions?

I think that the first one is a no-brainer and it's very easy to implement. The other one I don't know the demand for and it's a bit tougher to implement.

bcoca:
The insert=top/bottom can be made to only be used when state='present' and the rule is not present already. The other possibility is adding the rule at a specific line number, which should 'move' existing rules.

LinusU:
I agree. Specific line number though came up when I first wrote this module and the problem is that this is the global list on the entire machine. It's not often that you know what number to put in and when you do, it's quite fragile since that number is prone to changing.

dvigueras:
Renamed parameter `insert` to `action` as @LinusU commented.
Possible `action` values are `append` or `insert`. Default value is `append` in order to be backward compatible.

dvigueras:
Hi, It's been 9 days since my last comment. Any news about this?

bcoca:
sorry, we are a very busy project so sometimes things take a while. Also we are concentrating on releasing 2.0 which means bugs take priority and features might be delayed for 2.1

dvigueras:
Ok no problem, I can wait :) I was just waiting @LinusU's opinion.

LinusU:
There is still a few problems to be resolved here. The documentation still says `insert` instead of `action`, and the largest problem:

It's very unintuitive that insert _always changes_, even if the rule was previously on top. In my opinion, the rules should _only change_ when the rules _was absent_.

ctlajoie:
Will we be able to insert a rule at a specific index? From what I've read here so far it looks like the only options are top/bottom.

dvigueras:
@LinusU I agree with you, insert shouldn't change the state always. It should check if the rule is present at the top, but I don't know how to implement it in a clean way (i.e. not parsing `iptables -L` command output). 

With `iptables -C` you can check if a rule is present (as you have already done in the module), but with this command it isn't possible to find out which position is the rule at. Any ideas about how to implement it in a clean way?

@ctlajoie that should be part of another PR and would be pretty easy to add.

LinusU:
I think that checking if the rule is present is good. It would be unexpected if `insert` always move is to the top, even if it's present, and `append` adds the rule at the bottom, but doesn't move it if it's already present.

The behaviour should be the same for both commands, and I think that only inserting it if it isn't present is best. Consider what would happen if you are inserting two rules, they would always change since the first one would be moved to under the second one, and then the second one would be moved to be last again.

dvigueras:
@LinusU done, and I also fixed the documentation mistake. 

Now if you insert an already present rule it won't change the chain.

LinusU:
@dvigueras Looks very good :+1:

LinusU:
Hope I'm doing this right now:

'shipit'

gregdek:
Thanks again to @dvigueras for this PR, and thanks @LinusU for reviewing. Marking for inclusion.

LinusU:
@dvigueras Why did you close this? Could you reopen, we would love to get this in 👍

dvigueras:
Maybe I deleted the branch accidentally, sorry. I'll reopen the PR but I think it will require a rebase.

LinusU:
Thanks!

'shipit'

PR #1185, opened 2015-10-30
New module: xbps for managing Void Linux packages

adamruzicka:


packaging/os/xbps.py
```diff
@@ -0,0 +1,227 @@
...
+    if module.check_mode:
+        return CHECK_FLAG
+    else:
+        return ""
+
+def update_flag(state):
+    if state == 'latest':
+        return UPDATE_FLAG
+    else:
+        ""
```
chrisdevilliers: I believe you are missing a return statement on line 98.

adamruzicka: Right, thanks for looking into this

PR #1187, opened 2015-11-01
Revert "Change show_diff to default to yes"

emonty:
This was originally to match what puppet agent --test is, since the
rest of the options defaulted to on are grabbed from --test. However,
some security concerns have since been raised - namely that since this
is not the same invocation as --test but instead a remote orchestration
of puppet, the fact that passwords leak into the diff is a dangerous
default.

This reverts commit b86762c1806aa7f021a4780d06db2d3937910a62.

### DIFF
--- a/system/puppet.py
+++ b/system/puppet.py
@@ -45,9 +45,9 @@ options:
     default: None
   show_diff:
     description:
-      - Should puppet return diffs of changes applied. Defaults to yes, to match puppet agent --test. Change to no to avoid leaking secret changes.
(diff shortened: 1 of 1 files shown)


emonty:
@bcoca @gregdek sorry for the flip-flop on this one - spoke with several people using this over the last few days and everyone was concerned about the default change. I think we should default to conservative here.

Jmainguy:
@emonty Sorry to see you feel this way.

bcoca:
it is configurable, and I will prefer to err on the side of caution here so I'm fine with the default that avoids disclosure.

PR #1189, opened 2015-11-01
The dnf API appears to hang if the user does not have adequate permissions

abadger:
Do our own check for root user privileges here.

### DIFF
--- a/packaging/os/dnf.py
+++ b/packaging/os/dnf.py
@@ -234,9 +234,6 @@ def _mark_package_install(module, base, pkg_spec):
 
 
 def ensure(module, base, state, names):
-    if not util.am_i_root():
-        module.fail_json(msg="This command has to be run under the root user.")
-
     if names == ['*'] and state == 'latest':
         base.upgrade_all()
     else:
@@ -330,12 +327,20 @@ def main():
         mutually_exclusive=[['name', 'list']],
         supports_check_mode=True)
     params = module.params
-    base = _base(
-        module, params['conf_file'], params['disable_gpg_check'],
-        params['disablerepo'], params['enablerepo'])
     if params['list']:
+        base = _base(
+            module, params['conf_file'], params['disable_gpg_check'],
+            params['disablerepo'], params['enablerepo'])
         list_items(module, base, params['list'])
     else:
+        # Note: base takes a long time to run so we want to check for failure
+        # before running it.
+        if not util.am_i_root():
+            module.fail_json(msg="This command has to be run under the root user.")
+        base = _base(
+            module, params['conf_file'], params['disable_gpg_check'],
+            params['disablerepo'], params['enablerepo'])
+
         ensure(module, base, params['state'], params['name'])
 
 


abadger:
/cc @ignatenkobrain @DJMuggs

I was working on a problem in core that used the dnf module as a test case and found that doing something like this as a normal user seemed to hang the dnf module:

```
---
- hosts: localhost
  tasks:
    - name: Install one
      dnf:
        name:
          - bash-completion
          - sssd
        state: present
```

Traced the problem to me not being root when executing it:

```
$ whoami
badger
$ ansible-playbook dnf.yml
Using /etc/ansible/ansible.cfg as config file
1 plays in dnf.yml
Loaded callback default of type stdout, v2.0

PLAY ***************************************************************************

ok: [localhost]

TASK [Install one] *************************************************************
[Appears to hang]
^C
```

This should fix things but I wanted to run it by you all in case:
- dnf/fedora/rhel has an alternate way of elevating permissions for dnf so the checking for root is too strict.  Instead we should check for something else.
- In case there's something in the dnf API that were preform a permission check and we should use that instead.
- this is something that should be reported upstream as well.  The dnf CLI currently performs a check for root so I was surprised that the API doesn't (and just hangs if we aren't root).  Maybe one of those two things is a bug.

I'll go ahead and merge this in a few days if you guys don't object.  Just wanted to put it out there in case you knew something I didn't or wanted to follow up with the dnf upstream.

igor-raits:
CC @jsilhan @mluscon

mluscon:
Hi,

the user of the api is responsible for permission check, since DNF can work without root permission for various use cases. The only unfortunate issue here is that hang where we should rather raise an exception. Are you able to provide us a reproducer?

abadger:
Apparently... not. :-(  And I notice that our code has this a little later than my patch:

``` python
     if not util.am_i_root():
         module.fail_json(msg="This command has to be run under the root user.")
```

And today that's triggering fine.  Hmm...

These are the steps the code is taking prior to getting to that point with my test data:

``` python
    base = dnf.Base()
    conf = base.conf
    conf.debuglevel = 0
    conf.gpgcheck = not disable_gpg_check
    conf.assumeyes = True
    conf.config_file_path = conf_file
    base.read_all_repos()
    base.fill_sack()
```

When I looked at the process with strace, it seemed to be in a select loop but receiving no data (only checked it sporadically.  The process ran for around half an hour and seemed to be receiving no data in the select loop in the last ten minutes of that.)

My guess is that the problem was either some sort of network failure between myself and the repository serving the metadata that dnf was retrieving or maybe a stale lock file for that user account.  Changing to the root user didn't fix that underlying problem but instead worked around it by hitting the currently up to date system dnf cache (or not hitting a per-user lock).

If that's the failure case, then the only thing I could see dnf doing differently is maybe adding a timeout in the select loop where this can occur.  If we get no data back for 5-10 minutes (or hit a lock repeatedly for the same amount of time?  An amount of time specified in a timeout parameter when constructing dnf.Base()?) then throw an error rather than continuing.

For the code in this module, I'll move the check for root before the construction of dnf.Base() for  non-query operations so that the user doesn't have to download their own copy of the metadata only to find that they forgot to become root in order to run the command.

abadger:
Thanks @mluscon and @ignatenkobrain for taking a look and bearing with me as I figured out what the real culprit was!

PR #1196, opened 2015-11-03
Prevent ec2_remote_facts from failing when listing a terminated instance

wimnat:
When an instance has been terminated, aws stops returning the sourceDestCheck attribute.  This fix prevents the module from failing with an AttributeError in this case by setting the attribute to None.


evanccnyc:
Looks good to me.

PR #1201, opened 2015-11-03
add ports support to openbsd_pkg

bonds:
Adds ports support to openbsd_pkg.

### DIFF
--- a/packaging/os/openbsd_pkg.py
+++ b/packaging/os/openbsd_pkg.py
@@ -31,6 +31,7 @@ version_added: "1.1"
 short_description: Manage packages on OpenBSD.
 description:
     - Manage packages on OpenBSD using the pkg tools.
+requirements: [ "python >= 2.5" ]
 options:
     name:
         required: true
@@ -52,12 +53,14 @@ options:
             a binary. Requires that the port source tree is already installed.
             Automatically builds and installs the 'sqlports' package, if it is
             not already installed.
+        version_added: "2.1"
     ports_dir:
         required: false
         default: /usr/ports
         description:
           - When used in combination with the 'build' option, allows overriding
             the default ports source directory.
+        version_added: "2.1"
 '''
 
 EXAMPLES = '''
--- a/packaging/os/openbsd_pkg.py
+++ b/packaging/os/openbsd_pkg.py
@@ -18,8 +18,10 @@
 # You should have received a copy of the GNU General Public License
 # along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
 
+import os
 import re
 import shlex
+import sqlite3
 
 DOCUMENTATION = '''
 ---
@@ -41,6 +43,21 @@ options:
           - C(present) will make sure the package is installed.
             C(latest) will make sure the latest version of the package is installed.
             C(absent) will make sure the specified package is not installed.
+    build:
+        required: false
+        choices: [ yes, no ]
+        default: no
+        description:
+          - Build the package from source instead of downloading and installing
+            a binary. Requires that the port source tree is already installed.
+            Automatically builds and installs the 'sqlports' package, if it is
+            not already installed.
+    ports_dir:
+        required: false
+        default: /usr/ports
+        description:
+          - When used in combination with the 'build' option, allows overriding
+            the default ports source directory.
 '''
 
 EXAMPLES = '''
@@ -52,6 +69,9 @@ EXAMPLES = '''
 
 # Make sure nmap is not installed
 - openbsd_pkg: name=nmap state=absent
+
+# Make sure nmap is installed, build it from source if it is not
+- openbsd_pkg: name=nmap state=present build=yes
 
 # Specify a pkg flavour with '--'
 - openbsd_pkg: name=vim--nox11 state=present
@@ -118,15 +138,33 @@ def get_package_state(name, pkg_spec, module):
 
 # Function used to make sure a package is present.
 def package_present(name, installed_state, pkg_spec, module):
+    build = module.params['build']
+
     if module.check_mode:
         install_cmd = 'pkg_add -Imn'
     else:
-        install_cmd = 'pkg_add -Im'
+        if build is True:
+            port_dir = "%s/%s" % (module.params['ports_dir'], get_package_source_path(name, pkg_spec, module))
+            if os.path.isdir(port_dir):
+                if pkg_spec['flavor']:
+                    flavors = pkg_spec['flavor'].replace('-', ' ')
+                    install_cmd = "cd %s && make clean=depends && FLAVOR=\"%s\" make install && make clean=depends" % (port_dir, flavors)
+                elif pkg_spec['subpackage']:
+                    install_cmd = "cd %s && make clean=depends && SUBPACKAGE=\"%s\" make install && make clean=depends" % (port_dir, pkg_spec['subpackage'])
+                else:
+                    install_cmd = "cd %s && make install && make clean=depends" % (port_dir)
+            else:
+                module.fail_json(msg="the port source directory %s does not exist" % (port_dir))
+        else:
+            install_cmd = 'pkg_add -Im'
 
     if installed_state is False:
 
         # Attempt to install the package
-        (rc, stdout, stderr) = execute_command("%s %s" % (install_cmd, name), module)
+        if build is True and not module.check_mode:
+            (rc, stdout, stderr) = module.run_command(install_cmd, module, use_unsafe_shell=True)
+        else:
+            (rc, stdout, stderr) = execute_command("%s %s" % (install_cmd, name), module)
 
         # The behaviour of pkg_add is a bit different depending on if a
         # specific version is supplied or not.
@@ -134,7 +172,7 @@ def package_present(name, installed_state, pkg_spec, module):
         # When a specific version is supplied the return code will be 0 when
         # a package is found and 1 when it is not, if a version is not
         # supplied the tool will exit 0 in both cases:
-        if pkg_spec['version']:
+        if pkg_spec['version'] or build is True:
             # Depend on the return code.
             module.debug("package_present(): depending on return code")
             if rc:
@@ -177,6 +215,10 @@ def package_present(name, installed_state, pkg_spec, module):
 
 # Function used to make sure a package is the latest available version.
 def package_latest(name, installed_state, pkg_spec, module):
+
+    if module.params['build'] is True:
+        module.fail_json(msg="the combination of build=%s and state=latest is not supported" % module.params['build'])
+
     if module.check_mode:
         upgrade_cmd = 'pkg_add -umn'
     else:
@@ -275,6 +317,7 @@ def parse_package_name(name, pkg_spec, module):
             pkg_spec['version']           = match.group('version')
             pkg_spec['flavor_separator']  = match.group('flavor_separator')
             pkg_spec['flavor']            = match.group('flavor')
+            pkg_spec['style']             = 'version'
         else:
             module.fail_json(msg="Unable to parse package name at version_match: " + name)
 
@@ -287,6 +330,7 @@ def parse_package_name(name, pkg_spec, module):
             pkg_spec['version']           = None
             pkg_spec['flavor_separator']  = '-'
             pkg_spec['flavor']            = match.group('flavor')
+            pkg_spec['style']             = 'versionless'
         else:
             module.fail_json(msg="Unable to parse package name at versionless_match: " + name)
 
@@ -299,6 +343,7 @@ def parse_package_name(name, pkg_spec, module):
             pkg_spec['version']           = None
             pkg_spec['flavor_separator']  = None
             pkg_spec['flavor']            = None
+            pkg_spec['style']             = 'stem'
         else:
             module.fail_json(msg="Unable to parse package name at else: " + name)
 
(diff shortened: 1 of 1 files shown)


eest:
Hello Scott,

I have not yet read through the patch in detail or tested it (though my initial reaction is that it looks nice), but the first question that popped into my mind is "Why do you want this?".

I try to keep as close to the goals of the OpenBSD project itself as possible with this module, and looking at http://www.openbsd.org/faq/faq15.html#PkgVsPorts it is clear the project itself prefers binary packages over ports. This makes me lean towards "you should build your package and make it available somewhere through /etc/pkg.conf" rather than building ports directly with the module.

bonds:
I run OpenBSD on my laptop and I use Ansible to configure my laptop. I recently bought a new laptop which has hardware supported in -current but not in -stable. I've found that a couple weeks after installing a snapshot, installing from the snapshot packages becomes unreliable because the devs update libraries is the OS, so that installing a package on a machine running an old snapshot won't work. That leaves me with 2 choices: upgrade my snapshot and all my packages every couple weeks (or at least, every time I want to install a package and it doesn't work) or build from the ports I downloaded at the same time as my snapshot. I've been doing that latter.

I could just use the shell module and replace all my openbsd_pkg calls with 1) a test to see if the port is already installed and 2) install it if not, but given how often I'm using openbsd_pkg it seems like I'd be better off just writing a module, i.e. openbsd_port. But since I'd mostly be duplicating code from openbsd_pkg, it seemed better to fold it in instead of forking.

bonds:
Additionally, even when running -stable, I find some precompiled packages don't come in the combination of flavors I need, so I have to build from ports myself.

eest:
Thank you for the explanation. The reasoning sounds good to me, the last addition of missing flavors in -stable definitely sold me on the idea. Let's try to get this in :). My evening time is currently pretty limited because I am spending it on infrastructure hacking for an upcoming Dreamhack event.

Give me a few days to try this out and I will try to get back with feedback on the actual code.

bonds:
sounds good!

robynbergeron:
Hi @bonds -- thanks for submitting this PR. And thanks to @eest for your comments and feedback thus far :)

@eest -- ... when you have a moment :D Please review this PR to make sure it adheres to the following guidelines:

Thanks!

eest:
@bonds: just a quick update. Sorry for taking so long to review this. I have not forgotten about it. My estimate of  "a few days" will have to become "a few weeks". This week and the next I am busy with the Dreamhack stuff. After that my schedule will be more forgiving. I hope this is okay :).

I did notice there is a "I'm testing on Ansible 1.9 which doesn't support the new module.debug" comment in the code. This should be worked on against the devel branch of upstream Ansible by checking out the repo and sourcing hacking/env-setup. My stand is that any backwards compatibility stuff should be handled in the port if upgrading to the latest version is not possible.

Looking through the TODO list I don't think "handle non-standard ports base dir" is necessary. Any specific reason you feel this should be supported?

bonds:
@eest no worries at all, take whatever time you need. I'll keep plugging away when I have a moment too. I'll upgrade to the devel version Ansible and takeout the placeholder stuff from this patch. The OpenBSD ports(7) man page references the PORTSDIR environment variable in a few places...I figure the ideal implementation respects that variable, though a minimal version can probably ignore it and work for 99% of users.

bonds:
I tried switching to ansible@devel but I'm blocked by https://github.com/ansible/ansible/issues/13199.

gregdek:
Assuming that @eest is still busy. :)

eest:
Waiting for @bonds to get unblocked so I am sure we are working against the same codebase :).

bonds:
I'm unblocked. Works on ansible@devel. More to do, but probably far enough along for you to take a look at and comment on the approach at least.

eest:
I have started looking closer, here are some notes:
- "port_dir" is set in two places. First of all I would call it "ports_dir", but secondly the variable used in package_present() should depend on a global variable for the "/usr/ports" part.
- You use execute_command("test -d ..." to check if a directory exists, I would rather import os and use os.path.isdir for that.
- You require the use of sqlports, any specific reason for doing this over parsing the output of "make search"? I agree that using SQL is much cleaner in general, but needing an additional package does have a cost.
- What is the expected result of doing "name=\* state=latest build=yes"? Should we error out if name is "*" and build is set until we have support for that? I would rather throw an controlled error than doing "something" :).
- We are currently passing build to package_absent(), but that function does not use it. I would rather not pass the variable until it is actually supported.

bonds:
Ok, I refactored the port_dir stuff, let me know what you think of my
proposed approach. I refactored the test -d thing as you suggested. I
added an error condition for build=True and state=latest. I switched
to leaving the build param in the module, so its no longer being
passed around separately including to functions that don't yet use it.

Regarding sqlports, the make search method is deprecated as best I can
tell, see 'man ports':

"
search     invoked with a key, e.g., make search key=foo, retrieve
information relevant to a given port (obsolescent).
"

and

"Starting in OpenBSD 4.0, there is a port, databases/sqlports that
builds an sqlite database containing most information relevant to
every port in the ports tree..."

eest:
I like your new ports dir approach, guess you can also remove that TODO from the bottom of the file for handling different paths.

There are some typos which currently stops the module from working:
- os.port.isdir(), should be os.path
- if module.param['build'] is True:, should be module.params

I would probably make the message "the combination of build=true and state=latest is not supported" show the contents of the build variable instead of a static "true" string.

It seems reasonable to store the build variable in module, however you forgot to remove the build argument to the package_present() call inside package_latest().

Thanks for clarifying the use of sqlports: having "search" marked as obsolescent in the man page definitely convinced me of going down the sqlports road.

eest:
I had missed that you had fixed what I asked about, sorry for the delay in replying :).

I am now looking more into the sqlports usage and i have one initial question:
- What is supposed to happen when no match is found in the sql lookup?
  Doing "ansible -u root -i hosts -m openbsd_pkg -a 'name=nonexistant build=yes state=installed' 192.168.1.83" appears to me to be doing "make install" directly in /usr/ports, which if am not mistaken would mean we are installing all available ports. It should error out if no match is found.

Some more general points:
- You should add the build argument to the DOCUMENTATION text, here I would like you to point out in the description that enabling the build argument will build and install sqlports if it is not already present.
- You should add an example using the build argument to the EXAMPLES text.

bonds:
No worries @eest, I'm just plugging away as you find stuff that can be improved :) I addressed your latest comments just now.

eest:
Thanks for your patience!

After some testing with switching between absent/present with and without --check there is currently one call that fails with --check:

```
$ ansible --check -u root -i hosts -m openbsd_pkg -a 'name=kea build=yes state=present' 192.168.1.83
192.168.1.83 | FAILED! => {
    "changed": false, 
    "cmd": "pkg_add -Imn", 
    "failed": true, 
    "msg": "pkg_add: Missing pkgname\nUsage: pkg_add [-acinqrsUuvxz] [-A arch] [-B pkg-destdir] [-D name[=value]]\n       [-L localbase] [-l file] [-P type] pkg-name ...", 
    "rc": 1, 
    "stderr": "pkg_add: Missing pkgname\nUsage: pkg_add [-acinqrsUuvxz] [-A arch] [-B pkg-destdir] [-D name[=value]]\n       [-L localbase] [-l file] [-P type] pkg-name ...\n", 
    "stdout": "", 
    "stdout_lines": []
}
```

Here are some additional things, mostly superficial:
- I like your description, small nit: "if it not" -> "if it is not".
- There is some trailing whitespace after ports_dir documentation, I would just remove that line completely.
- The example with the build argument currently has "state=absent" in it, should be "state=present" to fit the description :).
- You seem to have merged the line "# Import module snippets." to the end of the TODO list.

bonds:
fixed @eest

eest:
Thanks, I have tested the current state of the code and feel we can move forward with committing this.

@bonds: do you agree?

bonds:
ok, now I think its ready @eest, I'll squash this PR next

bonds:
squashed

bonds:
BTW, I removed the other TODOs and changed it to use the exit code from the build process rather than the workaround code you have for pkg_add. That info disappeared when I squashed, so I wanted to mention it.

eest:
The return value change makes sense to me, thanks for mentioning it. I am currently doing a final round of testing and if I can not find any gotchas ill ask the ansible people to commit it. Thanks for working on this and being patient with my review.

bonds:
Thank you for your thoughtful and helpful review, its been fun working
on this together.

eest:
My tests have been successful (including running my test playbook: https://github.com/eest/openbsd_pkg-tests). I have even tried the case of filling the disk while building a port which makes the module return failure which is good. Time to get this commited :).

eest:
shipit

robynbergeron:
Thanks again to @bonds for this PR, and thanks @eest for reviewing. Marking for inclusion.

bonds:
@eest subpackages weren't working, e.g. 'jre' would install the 'jdk' with the default ('-main') subpackage instead, not sure what 'mariadb-server' was doing, but it wasn't installing 'mariadb-server', now it is

please take a look

eest:
Nice that you spotted that issue, I'll take a look tonight.

needs_revision

eest:
Here are my thoughts so far:
I think I would prefer if pkg_spec['subpackage']=None is set in a single location instead of doing it inside all the match statements when parsing the package name. My reasoning behind this is that a "subpackage" is not something that can be parsed out of a package name which is the purpose of that function.

I am thinking it is cleaner to set it inside get_package_source_path() just prior to doing the "parts" magic instead.

On a bit of an unrelated note I also happened to notice that the "LIMIT 1" being used in the SQL lookups has a possibly unwanted side effect when looking at the jre case:

```
$ sqlite3 /usr/local/share/sqlports "SELECT fullpkgpath FROM ports WHERE pkgspec = 'jre-*';"                    
devel/jdk/1.7,-jre
devel/jdk/1.8,-jre
```

This would mean we would unconditionally be installing jre 1.7 via ports right? I do recall that one of the TODOs mentioned something about caring about version strings. My general approach is that the module should halt if there are any unclear situations appearing, and getting more than one match feels like one of those situations. The last thing I want is the module to do undefined/unexpected things.

I would probably prefer that we handle the return of more than one match as an error, making it obvious either the user needs to supply more information and/or we need to add more logic to make it clear what we are asking for, rather than limiting the search and hoping for the best.

The error could even include all the different parts[0] matches to make it clear what is being found.

What do you think?

bonds:
I agree with setting pkg_spec['subpackage']=None in one location. I think it would be better to set it at the top of get_package_source_path() so its defined for _all_ packages in the case that build=true, seems less likely to trip some future refactor up.

As for getting rid of LIMIT=1 and erroring when things get ambiguous...I was trying to avoid that because its hard. :) Here's an attempt at it, see what you think.

bonds:
I'm relying an the behavior that versionless specs, e.g. mtr--, are assigned a flavor separator but not a flavor in parse_package_name(). That feels a little fragile to me. Perhaps it would be better to have an explicit pkg_spec['style'] = 'stem_only' or 'versionless' or some such for me to use, or I could create the SQL in parse_package_name and store it in pkg_spec['sql']. What do you think?

bonds:
Also, wondering if I should look into proper escaping to avoid a SQL injection vulnerability, i.e. openpkg: name="mtr--; EXECUTE blah blah"

robynbergeron:
Thanks @bonds. @eest please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

robynbergeron:
Thanks @bonds for this PR. A maintainer of this module has asked for revisions to this PR. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

eest:
I agree with the placement of subpackage=None, it looks good.

As for making it harder not using the "LIMIT 1" approach I understand that. My hope is that being strict straight out of the gates makes it easier to not introduce regressions for existing users when we end up with a feature request to add it later. Even so your code churn seems very fast anyway :).

I think you are on to something with storing a "style" choice in the pkg_spec when parsing the name, rather than depending on the separate items being set in a way that "recreates" that knowledge. Based on the three supported match types I think decent style names would be "stem", "versionless" and "version".

Your final thought regarding SQL injection is interesting, but a bit much for me to reason about simultaneously with the other changes going on. Can you incorporate the above things first and then we can look at that? My initial thought is that the injection protection maby needs to start even earlier given that use_unsafe_shell=True is used when building a port. Not sure what is possible and not at that point yet.

bonds:
Refactored. See what you think.

eest:
First of all I think the error message that is printed when the port building finds more than one match looks very good (and handles n+1 packages nicely). Great work Scott :).

The 'style' setting looks good to me.

I am now trying to figure out ways to inject commands as you mentioned, but I have not had any luck so far. When I try to get creative with the name of the package I get stopped earlier in the code path, like in parse_package_name() or get_package_state(), for example:

```
"msg": "Unable to parse package name at versionless_match: nmap--;touch /file"
```

or

```
 "msg": "failed in get_package_state(): Invalid spec: ;touch\nfile:/ is empty\n"
```

Have you already figured out some clever way of achieving it?

bonds:
I don't have an example attack, I'm just trying to be careful. Here's a version that uses the builtin sqlite3 lib instead of using shell commands, and it uses the db-api substitution which the example in the docs suggests is the way to protect against injection attacks: https://docs.python.org/2/library/sqlite3.html

gregdek:
Thanks @bonds for this PR. A maintainer of this module has asked for revisions to this PR. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

eest:
I have looked through the code and have run a bunch of tests and it seems to work well.

It is very nice that you replaced shell commands with the built in sqlite3 library, much cleaner. I agree that using prepared statements should be done even if there are no known problems since it is safer (and best practice) anyway. I think the changes look great.

A final style nit that I noticed just now: please arrange the module imports in an alphabetical order.

With that fixed I feel like we could have this commited, agree?

bonds:
Ok, sorted the imports and squashed. I think its ready, again. :)

bonds:
ready_for_review

eest:
Thanks! I'm happy you found the subpackage bug because the following rework made things a lot better overall in my opinion :).

eest:
shipit

packaging/os/openbsd_pkg.py
```diff
@@ -18,8 +18,10 @@
 # You should have received a copy of the GNU General Public License
 # along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
 
+import os
 import re
 import shlex
+import sqlite3
```
bcoca, line 24: so this is a new requirement and should be documented, as it is not present in python 2.4

packaging/os/openbsd_pkg.py
```diff
@@ -41,6 +43,21 @@
           - C(present) will make sure the package is installed.
             C(latest) will make sure the latest version of the package is installed.
             C(absent) will make sure the specified package is not installed.
+    build:
```
bcoca, line 47: add version_added to each field (2.1 at this point)

bonds:
@bcoca @eest fixed, ready_for_review

bcoca:
I'm not sure what having python 2.5 min will leave out (openbsd installed base wise), but i'm fine with it if @eest is.

bonds:
The last OpenBSD that shipped Python 2.4.x as the default Python, if
Distrowatch is to believed, was OpenBSD 4.0 in 2006. After that its
Python >= 2.5

http://distrowatch.com/table.php?distribution=openbsd

I don't think most OpenBSD users are in the habit of running 10 year
old versions on the OS, let alone managing said boxes with Ansible.

eest:
@bonds: Thanks for making the requested changes.

@bcoca: Thanks for pointing out the version issue. When developing the OpenBSD modules I mostly care about things working on -current. The reason for this is that any official port of this code will be created against (an even later version of) -current as well.

Focusing on -current allows us to create the cleanest code possible instead of slowly filling the module with backwards compat which will only get more deprecated and in the way as time goes by. If someone feels the need to have features of newer code on older systems backporting the code will be up to them, the same way they are already having to manually backport security patches and such to the system at large.

Just as a side note, the python 2.4 port was removed from CVS at 2011-12-20.

eest:
shipit

bcoca:
@eest yet Ansible tries to cater to users that might not have the latest and greatest deployed in their systems, it does make for 'less clean' code, but we are on the side of helping users manage their systems over the elegance of the code.

Since I don't expect this will impact more than 1 person (probably me) running a very outdated openbsd, I'm fine with the restriction, if we get enough complaints from the community I might have to revise this decision.

PR #1203, opened 2015-11-04
New Route53 facts module

Etherdaemon:
As discussed on the Ansible-Development Google group with @defionscode, here is my initial attempt to create a Route53 facts module that utilises boto3 methods to get and list various Route53 details such as hosted_zones, delegation_sets etc. 

Any feedback gladly welcomed. (First full Ansible/Python module so please be gentle)

Karen

### DIFF
--- a/cloud/amazon/route53_facts.py
+++ b/cloud/amazon/route53_facts.py
@@ -1,0 +1,434 @@
+#!/usr/bin/python
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+module: route53_facts
+short_description: Retrieves route53 details using AWS methods
+description:
+    - Gets various details related to Route53 zone, record set or health check details
+version_added: "2.0"
+options:
+  query:
+    description:
+      - specifies the query action to take
+    required: True
+    choices: [
+            'change',
+            'checker_ip_range',
+            'health_check',
+            'hosted_zone',
+            'record_sets',
+            'reusable_delegation_set',
+            ]
+  change_id:
+    description:
+      - The ID of the change batch request.
+        The value that you specify here is the value that
+        ChangeResourceRecordSets returned in the Id element
+        when you submitted the request.
+    required: false
+  hosted_zone_id:
+    description:
+      - The Hosted Zone ID of the DNS zone
+    required: false
+  max_items:
+    description:
+      - Maximum number of items to return for various get/list requests
+    required: false
+  next_marker:
+    description:
+      - Some requests such as list_command: hosted_zones will return a maximum
+        number of entries - EG 100. If the number of entries exceeds this maximum
+        another request can be sent using the NextMarker entry from the first response
+        to get the next page of results
+    required: false
+  delegation_set_id:
+    description:
+      - The DNS Zone delegation set ID
+    required: false
+  start_record_name:
+    description:
+      - The first name in the lexicographic ordering of domain names that you want
+        the list_command: record_sets to start listing from
+    required: false
+  type:
+    description:
+      - The type of DNS record
+    required: false
+    choices: [ 'A', 'CNAME', 'MX', 'AAAA', 'TXT', 'PTR', 'SRV', 'SPF', 'NS' ]
+  dns_name:
+    description:
+      - The first name in the lexicographic ordering of domain names that you want
+        the list_command to start listing from
+    required: false
+  resource_id:
+    description:
+      - The ID/s of the specified resource/s
+    required: false
+    aliases: ['resource_ids']
+  health_check_id:
+    description:
+      - The ID of the health check
+    required: false
+  hosted_zone_method:
+    description:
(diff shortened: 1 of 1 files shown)


willthames:
Hi, I'm not clear on the need for the list_/get_ distinction (I get that one returns one thing and one returns a list, but still not sure the distinction is necessary). Perhaps it can be simplified? What if you want all of the facts? Is that possible? Is there a need for filtering using a generic approach (using something similar to the filters arguments of other facts modules)?

I also don't like the `MethodExec` class - there are simpler ways of achieving the same thing with a dict. 

```
methods = { 'get_thing': get_thing, 'list_things': list_things }
methods[method_name](args)
```

(e.g. https://github.com/ansible/ansible-modules-core/blob/devel/cloud/amazon/rds.py#L1027-L1049)

Anyway, those are my initial thoughts, it's a great start though.

Etherdaemon:
Hey @willthames - thank you for your feedback. The get and list functions are actually AWS's distinctions - one provides a list of various things while the other provides specific details on a set of objects. The other fact modules in module-extras basically only call one method only and returns the results. While I'm trying to allow for different types of facts to be gathered for Route53 - EG, listing all your resource_record_sets for a given hosted zone, or listing health_checks and then getting health_check last failure reason or get health_check status.

willthames:
The nice thing is we don't have to follow AWS's naming conventions in this case :). 

I wonder if it wouldn't be nicer all around if we abstracted over the `list_`/`get_` distinction. And use filter for all queries rather than specific parameters per query. Might simplify the parameter list

Then you could just have

```
route53_facts:
     query: reusable_delegation_sets
     filter:
         delegation_set_id: {{delegation_id}}
register: delegation_sets
```

Not sure that `query` is the best choice of words but I know I prefer it over `command` (as command should have little place in an idempotent playbook)

Etherdaemon:
Hmm - in terms of using filter - wouldn't that force the user to go figure out which method is being used inside Ansible, then go lookup the AWS API docs to see what input parameters are required for each command they wanted get the details of? 

EG - I want to retrieve all the hosted zones in my AWS account, using filter I would envisage that would mean I would have to first figure out which AWS method is being used in Ansible (in this case it will be client.list_hosted_zones which has these 3 options [Marker='string', MaxItems='string', DelegationSetId='string']. I then need to make sure I pass in a filter string with the exact syntax required by AWS.

Some example Route53 commands that AWS provide are - note the specific syntax has to be like that:

```
response = client.list_hosted_zones(
    Marker='string',
    MaxItems='string',
    DelegationSetId='string'
)
```

Another Route53 example:

```
response = client.list_resource_record_sets(
    HostedZoneId='string',
    StartRecordName='string',
    StartRecordType='SOA'|'A'|'TXT'|'NS'|'CNAME'|'MX'|'PTR'|'SRV'|'SPF'|'AAAA',
    StartRecordIdentifier='string',
    MaxItems='string'
)
```

In terms of other modules that use filters - those filters are actually leveraging AWS's commands that have the filters available - for example the ec2_eni_facts.py module has 1 AWS command which is get_all_network_interfaces(network_interface_ids=None, filters=None, dry_run=False) - so filters is available there from AWS to allow AWS to perform the filtering for you - but Route53 commands don't have this option. 

Example of the AWS command used in ec2_eni_facts.py module:

```
get_all_network_interfaces(network_interface_ids=None, filters=None, dry_run=False)
```

So that function in that module looks like this - and thats the only AWS command that module utilises:

```
def list_eni(connection, module):

    filters = module.params.get("filters")
    interface_dict_array = []

    try:
         all_eni = connection.get_all_network_interfaces(filters=filters)
     except BotoServerError as e:

    for interface in all_eni:
        interface_dict_array.append(get_eni_info(interface))

     module.exit_json(interfaces=interface_dict_array)
```

Same goes for the ec2_vpc_route_table_facts.py module as well with the AWS command 

```
get_all_route_tables(route_table_ids=None, filters=None, dry_run=False)
```

Module code: 

```
try:
        all_route_tables = connection.get_all_route_tables(filters=filters)
    except BotoServerError as e:
```

willthames:
@Etherdaemon I see your point about filters, you may well have to go with what you have already to avoid unnecessarily complicating things for your users.

I still think we can ignore the `get_`/`list_` distinction imposed upon us by `boto3.route53` though (no such distinction is in `boto.route53` but I suspect (well, I hope) that's because boto3 has more functionality.

cloud/amazon/route53_facts.py
```diff
@@ -0,0 +1,401 @@
...
+short_description: list, get route53 details using AWS methods
+description:
+    - Lists or gets Route53 zone, record set or health check details
+options:
+  list_options:
+    description:
+      - specifies the list action to take, can specify get_command
+        or list_command but not both
+    required: false
+    default: null
```
juliedavila: `default: null` isn't necessary

Etherdaemon:
@defionscode @willthames I've refactored the new module after both your feedbacks (thank you both for your time as well). Are you able to take another look please?

cloud/amazon/route53_facts.py
```diff
@@ -0,0 +1,433 @@
...
+    module = AnsibleModule(
+        argument_spec=argument_spec,
+        mutually_exclusive=[
+            ['hosted_zone_method', 'health_check_method'],
+        ],
+    )
+
+    # Validate Requirements
+    if not (HAS_BOTO or HAS_BOTO3):
+        module.fail_json(msg='boto3 or boto is required.')
```
willthames: There is a chance this might fail when there is an ImportError for json. Not that likely but worth considering. Perhaps just change the message to say json and boto/boto3 are required, or handle the json ImportError separately.

willthames:
In terms of the module guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) the only thing I can see that is missing to pass the checklist is `version_added:` which should be 2.0 at this stage. 

The module would likely benefit from a RETURN documentation but it is not mandatory at this time.

gregdek:
Thanks @Etherdaemon for the PR, and thanks @willthames for the thorough review. Marking for inclusion.

Etherdaemon:
Thanks everyone.

Issue #1200, opened 2015-11-03
Create a bucket on Amazon s3 with Ansible results in Location Constraint Error for region us-east-1


danieldbower:
I'm trying to repeat the simplest example for creating an s3 bucket for us-east-1 from the documentation, but I keep getting the following error:

An exception occurred during task execution. To see the full traceback, use -vvv. The error was: 
    &lt;Error>
      &lt;Code>InvalidLocationConstraint&lt;/Code>
      &lt;Message>The specified location-constraint is not valid&lt;/Message>
      &lt;LocationConstraint>us-east-1&lt;/LocationConstraint>
      &lt;RequestId>---Redacted---&lt;/RequestId>&lt;HostId>---Redacted---&lt;/HostId>
    &lt;/Error>

Playbook Target:

```
- s3_bucket:
    aws_access_key: "{{aws_access_key}}"
    aws_secret_key: "{{aws_secret_key}}"
    name: "myapp-{{inst}}"
    region: us-east-1
```

I've tried leaving out region altogether, setting it to blank, and putting quotes around it. Each time it fails.

If I change the region to us-west-1, it works fine.

Using Ansible v2.0.0-0.4.beta2

ryansydnor:
This is fixed by https://github.com/ansible/ansible-modules-extras/pull/1268.

bcoca:
closing via merged #1268

PR #1204, opened 2015-11-04
Create_bucket can use the same logic as the aws connection for location

danieldbower:
Create_bucket can use the same logic as the aws connection to determine the location from the region.  This addresses issue #1200.


gregdek:
@wimnat Please review this PR to make sure it adheres to the following guidelines:

juliedavila:
@gregdek @bcoca this should be closed in favor of https://github.com/ansible/ansible-modules-extras/pull/1268

bcoca:
closing this as per mergefd #1268 mentioned above

PR #1207, opened 2015-11-05
for better match some extension string that has upercase letter

SamLiu79:
convert all extension string to lowercase.


SamLiu79:
Tested in real powershell env, seems we don't need to convert lowercase to match the extension name
PS C:\Windows\Temp> $ext = [System.IO.Path]::GetExtension('c:\Windows\temp\Test.ZIP' )
PS C:\Windows\Temp> $ext
.ZIP
PS C:\Windows\Temp> $ext -eq ".zip"
True

PR #1213, opened 2015-11-06
win_share: add module to manage smb shares

h0nIg:


gregdek:

When a module has both “works_for_me” and “passes_guidelines” tags, we will promote the module for inclusion in Ansible Extras.  At this point, you will be expected to maintain the module by fixing bugs and evaluating pull requests in a timely manner. 

cc: @trondhindenes @petemounce @elventear @smadam813 @jhawkesworth @angstwad @sivel @chrishoffman @cchurch

trondhindenes:
Maybe we should consider moving the UserSearch function into powershell.ps1 since it is used by multiple modules? @gregdek: Any thoughts on how we should plan for "common functionality" in PowerShell-based modules?

bcoca:
it is common enough a pattern, yes, if it is something only a subset of modules use, we might want a nother shared .ps1 file ('users.ps1'?)

trondhindenes:
Yep, we should start looking at breaking powershell.ps1 down into smaller chunks. I can start a thread in Groups regarding this.

h0nIg:
regarding the common functions: the split can be made later, right?

Pinging the windows guys for review: @petemounce @elventear @smadam813 @jhawkesworth @angstwad @sivel @chrishoffman @cchurch

koloros:
i would appreciate the inclusion.

works_for_me

ping: @trondhindenes @petemounce @elventear @smadam813 @jhawkesworth @angstwad @sivel @chrishoffman @cchurch

jhawkesworth:
I notice the module only supports Win 8.1 / S 2012 and later.  
Currently modules are supposed to support powershell 3.0 and later - is there an alternative to Get-SMBShare etc that could be used for older windows versions?

h0nIg:
Since powershell is "new" compared to linux/bash there are some missing configuration options for older OS'es like Windows 7 / 2008. The SMBShare commands target a new WMI namespace (ROOT/Microsoft/Windows/SMB) that does not exist on earlier versions of the OS.

However it is possible to do some dirty hacks with the win_share wmi  (https://msdn.microsoft.com/en-us/library/windows/desktop/aa389393%28v=vs.85%29.aspx). BUT i did not implemented it using these methods, because they are very complex and more C like.

tsintwin:
according to http://docs.ansible.com/ansible/developing_modules.html it looks good
+1
passes_guidelines

jlevis:
works_for_me

Ansible 2.0.0.2 (Ubuntu 14.04)
Windows Server 2012 R2
:+1:

h0nIg:
would be nice if some of the core guys will comment on this @abadger @sivel @gregdek @bcoca

Jmainguy:
@h0nIg If you can squash the git commits into one commit, it will make it easier when this gets merged for people to see what was actually added. I am not very familiar with Windows now a days, so this is about the only value I can add to this review.

h0nIg:
@Jmainguy checkout the "Files changed" tab at the top. there you can see the full changeset

nitzmahone:
@h0nIg Sorry it's taken so long to get to this- looks great! I have a few little tweaks I might make (around exception handling/diagnostic stuff) down the line. Also, it'd be great if you wanted to add support for check mode (now that Windows modules support it in 2.0), but I'm merging as-is. Thanks!

h0nIg:
FYI: stict mode is already supported, development was done based on 2.0. Are there some points i missed?

PR #1216, opened 2015-11-06
fix for win_iis_webapppool.ps1 where 'restarted' state isn't working when apppool is running

nlandais:
Scenario: I have a webappool running but I need to recycle it after
making a config change.

Problem: In the current implementation, a restart does not happen since
my app pool is running in the Start-WebAppPool cmdlet is used to
performed the recycling action.

Proposed fix: Invoke the Restart-WebAppPool cmdlet to restart an App
Pool regardless is App Pool running state


gregdek:
A little confusing here, since two of the files included have copyright notices from other authors... are those being merged from elsewhere?

cc @trondhindenes @petemounce @elventear @smadam813 @jhawkesworth @angstwad @sivel @chrishoffman @cchurch for guidance

nlandais:
Sorry, I did not mean to include the new win_certificate module with this PR. Back on track with the correct change

PR #1232, opened 2015-11-11
better cope with rpm not returning package name

robinro:
if the rpm query is missing a package name (or giving some error): fail soft

before the patch: the module fails because the installed_state dict is missing the package name

after the patch: the missing package is assumed to not be in the correct state and is installed/removed with zypper

also an ugly `print` is removed


gregdek:
@dirtyharrycallahan @alxgu Please review this PR to make sure it adheres to the following guidelines:

alxgu:
tested and looks good

shipit

gregdek:
Thanks @robinro for the PR and thanks @alxgu for the quick review.

Marking for inclusion.

PR #1234, opened 2015-11-11
Add apply_to option to rabbitmq_policy

bradwilson:
This adds support to allow the user to specify what context the policy applies to (exchanges, queues, or all), which was added in RabbitMQ 3.2.0 (when support for federation was added). The code only emits the `--apply-to` command line option when the value is not `all`, so that it doesn't break usage with RabbitMQ < 3.2.0.

/cc @retr0h

### DIFF
--- a/messaging/rabbitmq_policy.py
+++ b/messaging/rabbitmq_policy.py
@@ -38,6 +38,12 @@ options:
       - The name of the vhost to apply to.
     required: false
     default: /
+  apply_to:
+    description:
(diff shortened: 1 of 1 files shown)


evanccnyc:
This looks good to me.

gregdek:
@retr0h Please review this PR to make sure it adheres to the following guidelines:

retr0h:
Change looks good.  Don't see any issues with this.

"shipit"

gregdek:
Thanks again to @bradwilson for this PR, and thanks @retr0h for reviewing. Marking for inclusion.

PR #1238, opened 2015-11-12
IRC module: Added style= and more colors.

Jmainguy:
Used to be part of the commit message, and now I edited it to say this

### DIFF
--- a/notification/irc.py
+++ b/notification/irc.py
@@ -56,9 +56,11 @@ options:
   color:
     description:
       - Text color for the message. ("none" is a valid option in 1.6 or later, in 1.6 and prior, the default color is black, not "none"). 
+        Added 11 more colors in version 2.0.
     required: false
     default: "none"
-    choices: [ "none", "yellow", "red", "green", "blue", "black" ]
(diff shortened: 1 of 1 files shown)


Jmainguy:
Renamed some of the colors, so some people might have different colors displayed based on existing playbooks, but the colors they enjoyed are still there, just under a new name. Grabbed names and codes from https://github.com/myano/jenni/wiki/IRC-String-Formatting.

willthames:
@Jmainguy can you adhere to standard commit message practices? (short <56 char title, fuller description in the commit message body). Then you PRs won't have truncated titles with the rest overflowing into comments.

willthames:
Presumably you can have coloured and formatted messages at the same time.

I wonder if bold, underline and swap should be a separate argument called e.g. format, so you could have `irc: msg="Hello world" color="green" format="bold"`.

Also, can we make colour an alias for color? :)

Perhaps 'swap' would be better named 'reverse'

Jmainguy:
@willthames added your suggestions. colour alias, I went with style instead of format, changed swap to reverse, and added italic even though it doesn't work in my preferred client, it does work in others.

willthames:
It could still use a `git commit --amend` but I approve of the code changes.

PR #1243, opened 2015-11-14
choices for force should be list of ['yes','no'] - according to the d…

cinek810:
…ocumentation and internal check functions


bcoca:
boolean does this already, though it is a bit more permissive and alllows equivalents true/false, etc this is fine for this type of parameter.

so we are not going to apply this change as it is a decrease in functionality.

cinek810:
@bcoca but this doesnt work with current development of ansible:
https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/basic.py#L1145 returns internal error.

PR #1246, opened 2015-11-14
choices of boolean values should probably use predefined BOOLEANS list

cinek810:
Probably (with second attempt) this is right fix for issue I've experienced.

### DIFF
--- a/monitoring/zabbix_host.py
+++ b/monitoring/zabbix_host.py
@@ -377,7 +377,7 @@ def main():
             state=dict(default="present", choices=['present', 'absent']),
             timeout=dict(type='int', default=10),
             interfaces=dict(required=False),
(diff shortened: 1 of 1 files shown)


sivel:
I believe the correct fix is `type='bool'`

cinek810:
choices=BOOLEANS is widely used around the modules and it means that you can use values like:BOOLEANS_TRUE = ['yes', 'on', '1', 'true', 1] + BOOLEANS_FALSE = ['no', 'off', '0', 'false', 0].

It's the functionality mentioned by @bcoca in my previos pull request: #1243

bcoca:
already fixed via 875a0551032e721f06625c237bd7c2a6fecaa7fd,

PR #1263, opened 2015-11-18
Added support for DNS SRV records in 'dnsmadeeasy' module

sysreq0:
This patch implements support for SRV records and catches up with the module's documentation that indicates SRV records are supported, but are actually not per the comment on line 207:

```
# TODO SRV type not yet implemented
```

### DIFF
--- a/network/dnsmadeeasy.py
+++ b/network/dnsmadeeasy.py
@@ -204,7 +204,6 @@ class DME2:
         if not self.all_records:
             self.all_records = self.getRecords()
 
-        # TODO SRV type not yet implemented
         if record_type in ["A", "AAAA", "CNAME", "HTTPRED", "PTR"]:
             for result in self.all_records:
                 if result['name'] == record_name and result['type'] == record_type:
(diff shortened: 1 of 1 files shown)


gregdek:
@briceburg Please review this PR to make sure it adheres to the following guidelines:

briceburg:
shipit

thanks @gregdek and @sysreq0 for the contribution!

gregdek:
Thanks again to @sysreq0 for this PR, and thanks @briceburg for reviewing. Marking for inclusion.

bcoca:
SRV type was already documented, does this fix the handling being broken or does it improve on it?

briceburg:
@bcoca I can't test ATM, but this would fix a broken implementation. From memory,  the srv record requires a special payload.

briceburg:
@sysreq0 could you document and/or add an example for handling arv records to the header comments as well? As this stands, it's a bit unclear.

sysreq0:
@bcoca @briceburg  It is indeed documented in the code, but it wasn't implemented. See line 207 for this comment:

```
# TODO SRV type not yet implemented
```

I should probably remove this comment as part of the pull request, now to think of it. This pull request implements support for SRV records and thus catches up with the documentation that's already there.

PR #1268, opened 2015-11-20
Use boto normalized location for bucket creation

ryansydnor:
If a bucket is being created in us-east-1, the module passed
'us-east-1' to boto's s3.create_bucket method rather than
Location.DEFAULT (an empty string). This caused boto to generate
invalid XML which AWS was unable to interpret.

### DIFF
--- a/cloud/amazon/s3_bucket.py
+++ b/cloud/amazon/s3_bucket.py
@@ -129,11 +129,10 @@ def create_tags_container(tags):
     tags_obj.add_tag_set(tag_set)
     return tags_obj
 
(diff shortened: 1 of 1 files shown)


tima:
Probably a good idea and looks relative harmless to the current function of the module. @wimnat and @defionscode what do you think?

wimnat:
Similar to https://github.com/ansible/ansible-modules-extras/pull/1204 but I think i prefer this one.

shipit

juliedavila:
shipit

wimnat:
Should be in 2.0 rc2

PR #1270, opened 2015-11-21
aws lambda module

steynovich:
This is the first version of the AWS Lambda module, which allows management (creation, updating and deletion) of Lambda functions in AWS. Please review this code.


cloud/amazon/lambda.py
```diff
@@ -0,0 +1,368 @@
...
+
+
+DOCUMENTATION = '''
+---
+module: lambda
+short_description: Manage AWS Lambda functions
+description:
+     - Allows for the management of Lambda functions.
+version_added: '2.0'
+requirements: [ boto, botocore, boto3, base64, hashlib ]
```
wimnat: boto requirements aren't necessary if you're including the aws doc fragment

cloud/amazon/lambda.py
```diff
@@ -0,0 +1,368 @@
...
+    path = module.params.get('path')
+    description = module.params.get('description')
+    timeout = module.params.get('timeout')
+    memory_size = module.params.get('memory_size')
+
+    check_mode = module.check_mode
+    changed = False
+
+    if not HAS_BOTO or not HAS_BOTO3 or not HAS_BASE64 or not HAS_BINASCII or not HAS_HASHLIB:
+        module.fail_json(msg='Some Python modules are missing, please install it')
```
wimnat: You should explicitly check for boto, hashlib etc and specifically fail on what is missing

cloud/amazon/lambda.py
steynovich: ok, I've removed boto/botocore from the requirements, boto3 is still there since it's not included in ec2/aws.

cloud/amazon/lambda.py
steynovich: ok, done.

steynovich:
Thanks for your reply. Can you post your boto configfile (without any sensitive data)?
~/.aws/config

From your output it looks like boto/asible is using the region name as an endpoint, which is of course not a valid endpoint name. I'm also wondering if this happens only with this module or with other modules as well (e.g. route53_facts)?

eladhartog:
Hi,
Regarding to the timeout and memory_size, I don't get any error, but when trying to create a lambda function using the following code the lambda function is created but with the defaults timeout and memory size parameters (3 sec, 128MB):

my code:

```
- name: create lambda function
local_action:
module: lambda
name: 'ansibleTest'
state: present
path: '/home/dev/PythonLambdaDir2/runEc2UsingLambda.zip'
runtime: 'python2.7'
role_arn: 'arn:aws:iam::533188012830:role/lambda_basic_execution'
handler: 'runEc2UsingLambda.lambda_handler'
timeout: 300
memory_size: 256
```

Regarding to the zip file I mean that it creates the lambda function but I do see the code in the code area (in the AWS console), but you said that it loads zip file (it looks like that it open the zip --> take the python scripts and write it to the code area in the lambda AWS console).
Note: my zip contain only 1 python script with no exteranl packages

steynovich:
@eladhartog Thanks for testing and troubleshooting. Three vars were not passed to the creation of a function: timeout, memory size and description. I've pushed commits 83624d2 and 7c8a857 to fix this issue. Can you also test this for me?

Regarding to the code area in the console: I'm afraid I can't change this behaviour, since the only option is to upload a zipfile or refer to an object in S3.

eladhartog:
Thanks,
I checked your new commits and now the lambda function is created properly  (with the right timeout, memory_size values.
I will continue checking it and let you know if I have any issues...

smiller171:
glad to see someone is working on a module for this. I'm sure this will be much needed in the near future. :+1:

cagatay:
Thanks for the effort, I'm looking forward to see this merged :+1:

eladhartog:
Hi,
I downloaded ansible-2.0.0-0.8.rc3.tar.gz but i can't find the lambda.py script inside ansible-2.0.0\lib\ansible\modules\extras\cloud\amazon\ folder.

Did you merge it to this RC?

sivel:
@eladhartog this module has not been merged yet.  As such, it is expected that this module will not appear in ansible 2.0, but be included in ansible 2.1.

s0enke:
Works great for me, already using it here: https://github.com/s0enke/pocketcleaner-ansible/commit/46077abea776601bb31969b165d7a9378f5c6fec :)

manuelmeurer:
[Using this as well](https://github.com/krautcomputing/ansible-aws-lambda-ruby-test/blob/master/library/lambda.py) already and no problems so far.

wimnat:
@steynovich - can you rebase to collapse the commits?

steynovich:
@wimnat Done

xudawei:
hi steynovich, i had created the lambda function via this module , that`s cool . but i found the module can not suppoort create event source for lambda.  am i right?

steynovich:
@xudawei currently not. However I could implement support for event source mapping of pull based models like Kinesis or a DynamoDB streams. Unfortunately not for push based models such as CloudWatch Events (cron) or S3 Events.

xudawei:
@steynovich got it , I am looking forward to next version :+1:   thank you very much ! happy every day

steynovich:
@wimnat is there some idea of whether this module can be shipped with the next release of Ansible?

sgzijl:
I use this module in one of my current projects and heavily depend on it. The module works great :+1: Would love to see this shipped, also.

smiller171:
shipit!

cloud/amazon/lambda.py
```diff
@@ -0,0 +1,394 @@
...
+        'Timeout': 123,
+        'MemorySize': 123,
+        'LastModified': 'string',
+        'CodeSha256': 'string',
+        'Version': 'string',
+    }
+'''
+
+try:
+    import boto
```
wimnat: I can't see where boto is used in your code (and it shouldn't be as new modules must be boto3 only) so is this necessary?

cloud/amazon/lambda.py
```diff
@@ -0,0 +1,394 @@
...
+    if not HAS_HASHLIB:
+        module.fail_json(msg='Python module "hashlib" is missing, please install it')
+
+    if not HAS_BASE64:
+        module.fail_json(msg='Python module "base64" is missing, please install it')
+
+
+    try:
+        region, ec2_url, aws_connect_kwargs = get_aws_connection_info(module, boto3=True)
+        client = boto3_conn(module, conn_type='client', resource='lambda',
```
wimnat, line 255: You should check that region is set here before connection.  See other modules for examples.

steynovich:
@wimnat I followed your advice, so I've removed the boto dependency and check if region is set before setting up the connection.

hvpareja:
+1

gregdek:
Thanks @steynovich for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

steynovich:
Prepared the module for Ansible 2.2, it will now successfully build (version_added 2.1 -> 2.2). 

This module is ready_for_review .

gregdek:
Thanks @steynovich for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

filipenf:
Maybe it would be nice to have a `permissions` attribute to allow the user to specify the lambda function's permissions, doing everything in a single step.

My 2 cents

cloud/amazon/lambda.py
```diff
@@ -0,0 +1,395 @@
...
+
+
+DOCUMENTATION = '''
+---
+module: lambda
+short_description: Manage AWS Lambda functions
+description:
+     - Allows for the management of Lambda functions.
+version_added: '2.2'
+requirements: [ boto3, base64, hashlib ]
```
ryansb: `base64` and `hashlib` are both in the standard library, so you don't need to include them here.

cloud/amazon/lambda.py
```diff
@@ -0,0 +1,395 @@
...
+
+    return hex_digest
+
+
+def main():
+    argument_spec = ec2_argument_spec()
+    argument_spec.update(dict(
+        name=dict(type='str', required=True),
+        state=dict(type='str', default='present', choices=['present','absent']),
+        runtime=dict(type='str', choices=['nodejs','nodejs4.3','java8','python2.7'], default=None),
```
ryansb: I recommend against requiring only those choices. I'd rather rely on catching `ClientError` later, since AWS is likely to update the supported languages and it would be unfortunate to have Ansible prevent folks from using them.

ryansb: Also, why not just make this paramter `required=True` since the module fails without it?

cloud/amazon/lambda.py
```diff
@@ -0,0 +1,395 @@
...
+        if len(code_kwargs) > 1:
+            try:
+                if not check_mode:
+                    response = client.update_function_code(**code_kwargs)
+                changed = True
+            except (botocore.exceptions.ParamValidationError, botocore.exceptions.ClientError) as e:
+                module.fail_json(msg=str(e))
+
+        # We're done
+        module.exit_json(changed=changed, result=response)
```
ryansb: If the function exists but there are no changes (thus `len(code_kwargs)` is just one item), this references a nonexistent response object & causes the task to fail.

cloud/amazon/lambda.py
```diff
@@ -0,0 +1,395 @@
...
+    choices: [ 'nodejs', 'nodejs4.3', 'java8', 'python2.7' ]
+  role_arn:
+    description:
+      - The Amazon Resource Name (ARN) of the IAM role that Lambda assumes when it executes your function to access any other Amazon Web Services (AWS) resources
+    default: null
+  handler:
+    description:
+      - The function within your code that Lambda calls to begin execution
+    default: null
+  path:
```
ryansb: Can this be renamed `zip_path` or similar, so the expected format is part of the parameter name? 

If you set it to be `zip_path` it also leaves open the option of adding a parameter in the future that automatically zips up a single file as `code_path` or `file_path` or something.

ryansb: Or, match the API name with `zip_file`

steynovich:
@ryansb I've processed your feedback. Can you please review the new version?

cloud/amazon/lambda.py
```diff
@@ -0,0 +1,401 @@
...
+        module.fail_json(msg='region must be specified')
+
+    try:
+        client = boto3_conn(module, conn_type='client', resource='lambda',
+                            region=region, endpoint=ec2_url, **aws_connect_kwargs)
+    except (botocore.exceptions.ClientError, botocore.exceptions.ValidationError) as e:
+        module.fail_json(msg=str(e))
+
+    current_function = get_current_function(client, name)
+    if not current_function:
```
ryansb: This conditional makes all runs fail - if I do:

```
  - lambda:
      state: present
      runtime: nodejs4.3
      handler: csv_read.handlers
      role_arn: arn:aws:iam::REDACTED:role/lambda-default-role
      name: myfunc
      profile: ansible-test
      zip_file: /home/ryansb/code/test-lambda/code.zip
```

For both `present` and `absent` state for a function that does not yet exist, I get the error `"msg": "Unable to get function information prior updating"`

ryansb: I removed this conditional and the play works as expected: absent + existing results in deletion of the function, absent + nonexisting results in correct unchanged output, absent+present updates the function, and present+no change does nothing as expected.

tima:
Just came across this pull request.  Out of curiosity how does this module compare to the one in this project: https://github.com/pjodouin/ansible-lambda? I know little about lambda -- just putting it out there as food for thought.

cloud/amazon/lambda.py
```diff
@@ -0,0 +1,401 @@
...
+except ImportError:
+    HAS_BOTOCORE = False
+
+try:
+    import boto3
+    HAS_BOTO3 = True
+except ImportError:
+    HAS_BOTO3 = False
+
+try:
```
ryansb: The `hashlib`, `binascii`, and `base64` conditionals aren't needed - these are in the Python standard library from 2.6 through 3.5, so please remove these conditionals and the checks below (I'll comment there too).

cloud/amazon/lambda.py
```diff
@@ -0,0 +1,401 @@
...
+    check_mode = module.check_mode
+    changed = False
+
+    if not HAS_BOTOCORE:
+        module.fail_json(msg='Python module "botocore" is missing, please install it')
+
+    if not HAS_BOTO3:
+        module.fail_json(msg='Python module "boto3" is missing, please install it')
+
+    if not HAS_BINASCII:
```
ryansb: The `hashlib`, `binascii`, and `base64` conditionals aren't needed - these are in the Python standard library from 2.6 through 3.5, so please remove these checks.

ryansb:
@tima neat, I didn't know about that external module set. I notice they're licensed as GPLv3, and they support more of the Lambda API (VPCs/subnets, alias publishing, etc). I'll see if the author is amenable to filing a pull request to get into modules-core instead of needing to be added to lib/custom plugins.

mattclay:
The lambda module https://github.com/pjodouin/ansible-lambda that @tima mentioned previously had a PR, which is now closed: https://github.com/ansible/ansible-modules-extras/pull/1890

pilgrim2go:
If I only need update lambda function, this snippet failed

```
  - lambda:
      state: present
      name: sb-t0
      profile: lambda
      s3_bucket: test-lambda
      s3_key: billing-1.0.1-41.jar   
      runtime: java8
```

error is

```
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "invocation": {"module_args": {"aws_access_key": null, "aws_secret_key": null, "description": "", "ec2_url": null, "handler": null, "memory_size": 128, "name": "sb-t0", "profile": "lambda", "region": null, "role_arn": null, "runtime": "java8", "s3_bucket": "test-lambda", "s3_key": "billing-1.0.1-41.jar", "s3_object_version": null, "security_token": null, "state": "present", "timeout": 3, "validate_certs": true, "zip_file": null}, "module_name": "lambda"}, "msg": "Parameter validation failed:\nInvalid type for parameter Handler, value: None, type: <type 'NoneType'>, valid types: <type 'basestring'>\nInvalid type for parameter Role, value: None, type: <type 'NoneType'>, valid types: <type 'basestring'>"}
```

I temporarily comment these following lines

```
        # Upload new configuration
        # if len(func_kwargs) > 1:
        #     try:
        #         if not check_mode:
        #             client.update_function_configuration(**func_kwargs)
        #         changed = True
        #     except (botocore.exceptions.ParamValidationError, botocore.exceptions.ClientError) as e:
        #         module.fail_json(msg=str(e))
```

to make it work.

Since time is limited, i didn't dig further.

steynovich:
Hi, I've commited a new version of the module. This should fix the bug as mentioned by @ryansb.

I've also added support for VPC subnet/security groups.

steynovich:
This PR is ready_for_review

sgzijl:
shipit

steynovich:
It looks like the execute Lambda module will be shipped with Ansible 2.2 after being in PR for a few weeks (https://docs.ansible.com/ansible/execute_lambda_module.html). It would be very useful to include a module which actually can manage your Lambda function as you don't want to manually configure your functions...

This PR is now open for months. Please shipit (or reject it).

tima:
@steynovich: https://github.com/ansible/ansible-modules-extras/pull/2829

danvaida:
@steynovich I believe the same applies for your `lambda` module: https://github.com/ansible/ansible-modules-extras/pull/1890#issuecomment-248334785
one mention would be that your module is offering the upload functionality as an option. I would remove the upload part altogether and encourage the user to use the `s3` module for creating the bucket and uploading the object with proper permissions and such.

oh and btw, I have it failing pretty nasty with

```
$ ansible --version
ansible 2.2.0 (devel 1b54d3b6dc) last updated 2016/09/20 17:00:01 (GMT +200)
  lib/ansible/modules/core: (detached HEAD 12a7027c49) last updated 2016/09/20 17:00:06 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD db7a3f48e1) last updated 2016/09/20 17:00:06 (GMT +200)
  config file =
$ sudo pip freeze | grep boto
Password:
boto==2.42.0
boto3==1.4.0
botocore==1.4.55
```

```

    - name: Lambda function code should be updated
      lambda:
        name: test
        description: test
        s3_bucket: 'some-bucket-name'
        s3_key: "/test.zip"
        s3_object_version: "1"
        runtime: 'python2.7'
        timeout: 15
        handler: 'test.run'
        memory_size: 128
        role_arn: 'arn:aws:iam::987654321012:role/lambda_basic_execution'
        region: 'eu-central-1'
        state: present
```

```
 TASK [Lambda function code should be updated] **********************************
 An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ValueError: No JSON object could be decoded
 failed: [localhost] (item={u'name': u'test', u's3_key': u'/test.zip', u'policy_principal': u'events.amazonaws.com', u's3_bkt': u'some-bucket-name', u'publish ': True, u'version_to_delete': 0, u'policy_action': u'lambda:InvokeFunction', u'handler': u'test.run', u'role': u'lambda_basic_exection', u'timeout': 15, u'm emory': 128, u'runtime': u'python2.7', u'region': u'eu-central-1', u'desc': u'test'}) => {"failed": true, "item": {"desc": "test", "handler": "test.run", "me mory": 128, "name": "test", "policy_action": "lambda:InvokeFunction", "policy_principal": "events.amazonaws.com", "publish": true, "region": "eu-central-1",  "role": "lambda_basic_execution", "runtime": "python2.7", "s3_bkt": "some-bucket-name", "s3_key": "/test.zip", "timeout": 15, "version_to_delete": 0}, "modul e_stderr": "Traceback (most recent call last):\n  File \"/var/folders/xz/t667dq_j07z6dp1mmg8ylgq40000gn/T/ansible_jTL2gz/ansible_module_lambda.py\", line 437 , in <module>\n    main()\n  File \"/var/folders/xz/t667dq_j07z6dp1mmg8ylgq40000gn/T/ansible_jTL2gz/ansible_module_lambda.py\", line 411, in main\n    respon se = client.create_function(**func_kwargs)\n  File \"/Library/Python/2.7/site-packages/botocore/client.py\", line 159, in _api_call\n    return self._make_ap i_call(operation_name, kwargs)\n  File \"/Library/Python/2.7/site-packages/botocore/client.py\", line 483, in _make_api_call\n    operation_model, request_di ct)\n  File \"/Library/Python/2.7/site-packages/botocore/endpoint.py\", line 141, in make_request\n    return self._send_request(request_dict, operation_mode l)\n  File \"/Library/Python/2.7/site-packages/botocore/endpoint.py\", line 168, in _send_request\n    request, operation_model, attempts)\n  File \"/Library /Python/2.7/site-packages/botocore/endpoint.py\", line 233, in _get_response\n    response_dict, operation_model.output_shape)\n  File \"/Library/Python/2.7/ site-packages/botocore/parsers.py\", line 209, in parse\n    parsed = self._do_error_parse(response, shape)\n  File \"/Library/Python/2.7/site-packages/botoc ore/parsers.py\", line 687, in _do_error_parse\n    error = super(RestJSONParser, self)._do_error_parse(response, shape)\n  File \"/Library/Python/2.7/site-p ackages/botocore/parsers.py\", line 542, in _do_error_parse\n    body = self._parse_body_as_json(response['body'])\n  File \"/Library/Python/2.7/site-package s/botocore/parsers.py\", line 574, in _parse_body_as_json\n    original_parsed = json.loads(body)\n  File \"/System/Library/Frameworks/Python.framework/Versi ons/2.7/lib/python2.7/json/__init__.py\", line 338, in loads\n    return _default_decoder.decode(s)\n  File \"/System/Library/Frameworks/Python.framework/Ver sions/2.7/lib/python2.7/json/decoder.py\", line 366, in decode\n    obj, end = self.raw_decode(s, idx=_w(s, 0).end())\n  File \"/System/Library/Frameworks/Py thon.framework/Versions/2.7/lib/python2.7/json/decoder.py\", line 384, in raw_decode\n    raise ValueError(\"No JSON object could be decoded\")\nValueError:  No JSON object could be decoded\n", "module_stdout": "", "msg": "MODULE FAILURE"}
```

steynovich:
@danvaida I'm not really seeing the point here. If you want to separate the deployment of the code and the configuration of the code you're free to use the s3 parameters of this module and thus don't use the local zip file parameters. You can have the s3 module to to update your code. It's up to you how you want to manage your Lambda function.

Futhermore your exception is not thrown by the module in this PR. It looks like you're using the module from #1890. Also item parameters in the exception are not mentoined in either the source code of this PR or in your example task. Can you please make sure you don't have any module or playbook conflict?

danvaida:
Regarding wether or not the exception is being thrown by the module in this PR or the one in #1890:
Yes, indeed, for simplicity purposes I left out the `with_items` task parameter because I don't see how this makes any difference with that sort of exception, in this particular situation.
I am not including the module from the other PR in my `./library` directory. If I would, my task wouldn't even make it out of the module's params validation part as that one is requiring the parameter `local_path`, which is not passed in the above task.
Also, the `lambda` module present in the `devel` branch is yours (see `--version` above) and

```
$ ansible-doc lambda | tail -n 1
MAINTAINERS: Steyn Huizinga (@steynovich)
```

Now, I'd be more than happy to create a complete gist if you think that would make sense.

To get back to the operation of your module, my only point is that a certain microservice that you'd wanna run as a lambda function will have a build and a deploy procedure. By having a proper separation between the two, the build procedure should be able to be consisting of a playbook that would call the `lambda` module to create/update the function's code by looking into the provided S3 bucket. The uploading of the code would be the responsibility of the build procedure, which here would mean a couple of tasks invoking the `s3` module.

Hope now my point is clearer. I might have misunderstood your previous comment. Do let me know.

> Think of the following use case:
> 1. your build process generates a zip file and uploads it to an S3 bucket
> 2. you deployment process is running somewhere else and takes this readable bucket object and feeds it to task that runs the lambda module.

Also, what about S3 buckets that don't support versioning? the `s3_object_version` is a required param... _Update: I managed to get away with `s3_object_version: null` for a bucket with suspended versioning. Would be great if this would have a sane default and documentation (maybe in the example section)._

steynovich:
I'm pretty sure you're not using the module from this PR, because the exception throws variables like `s3_bkt` (in this module this could only be `s3_bucket`), `role` (should be `role_arn`), `policy_action` (not part of this module), `memory` (should be `memory_size`), etc etc.

Also I would like to mention that your task definition doesn't match the exception. I'm not talking about the with_items iteration part, but about the naming of variables ie.: `s3_bucket` in your task, but `s3_bkt` in exception, etc etc. `local_path` is not a parameter for this PR and finally `s3_object_version` is not a required parameter for the module in this PR...

Please verify that you don't have any additional lambda modules loaded in your Ansible (e.g. library parameter in cfg file and `grep -r lambda` or `grep -r s3_bkt` in your ansible directory, run ansible with debug parameters, ...).

danvaida:
@steynovich see this [gist](https://gist.github.com/danvaida/91fbf5189bd330093426f69a15bb7d5e). It seems the problem is residing somewhere else. Don't wast your time with it. I haven't had the time to dig deeper yet but hopefully I'll manage to do so this weekend. Indeed, the first exception posted was confusing due to the `with_items` vars mapping.
I got the module working nicely on my personal laptop so now I'm just trying to get it working on this one.

The non-debug related points still stand, though.

danvaida:
@steynovich I've just set up a vanilla wheezy docker image with just the requirements for the module and I'm getting the same error as in the gist mentioned above. running straight from the devel branch.
Plus, now I get the same error on the laptop where I initially got it working. Let's see if running this in a docker container can help us rule out the "works on my machine".
P.S. the documentation could use some updating as per the only works with s3/zip statement and example.

tima:
@danvaida & @steynovich: Once a PR is merged or closed it falls off the core developers radar. They have way to may open issues and PRs on their plate to monitor ones that are closed. If you found a bug or, better yet, have a patch, please open a new issue/PR. It's not going to get spotted and picked up here otherwise. I'm only seeing this because I forgot to turn off my email notifications on this PR once it was merged.

Also @danvaida -- doc PRs are welcome.

PR #1272, opened 2015-11-22
New AWS EC2 Addresses module

Etherdaemon:
Allocate, associate, disassociate and release ec2 addresses.

This module does not support EC2-Classic as its intended for private-addresses which are EC2-VPC only.

Requires Boto3, botocore and json.
- Allocate and deallocate secondary private ip's from an interface
- Associates and if required allocate Elastic IP to private IP
- Disassociate Elastic IPs and release if required


cloud/amazon/ec2_addresses.py
```diff
@@ -0,0 +1,409 @@
...
+    params = dict()
+    changed = False
+
+    if not module.params.get('interface_id'):
+        module.fail_json(msg='interface_id is required')
+    else:
+        params['NetworkInterfaceId'] = module.params.get('interface_id')
+
+    interface_details = get_interface_details(client, module)
+    existing_secondaries = \
```
willthames:
```
existing_secondaries = [i['NetworkInterfaces'][0]['PrivateIpAddresses'] for i in interface_details if not i['Primary']]
```

cloud/amazon/ec2_addresses.py
```diff
@@ -0,0 +1,409 @@
...
+
+    if module.params.get('private_ip'):
+        private_ips = [module.params.get('private_ip')]
+    else:
+        interface_details = get_interface_details(client, module)
+        existing_secondaries = \
+            filter(lambda i:
+                i['Primary'] == False,
+                interface_details['NetworkInterfaces'][0]['PrivateIpAddresses'])
+        private_ips = [n['PrivateIpAddress'] for n in existing_secondaries]
```
willthames: Don't need the intermediate `existing_secondaries`. `filter` isn't very pythonic these days compared to list comprehensions (and Ansible's coding standards say list comprehensions should be kept simple if possible). I think the following expresses the same functionality while keeping it fairly simple:

```
private_ips = [i['NetworkInterfaces'][0]['PrivateIpAddresses']['PrivateIpAddress'] for i in interface_details if not i['Primary']
```

cloud/amazon/ec2_addresses.py
```diff
@@ -0,0 +1,409 @@
...
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+DOCUMENTATION = '''
```
willthames, line 16: I think the DOCUMENTATION string is overly indented compared to other modules (and even the EXAMPLES string in this module)

cloud/amazon/ec2_addresses.py
```diff
@@ -0,0 +1,409 @@
...
+
+
+def date_handler(obj):
+    return obj.isoformat() if hasattr(obj, 'isoformat') else obj
+
+
+def get_interface_details(client, module):
+    params = dict()
+    if not module.params.get('interface_id'):
+        module.fail_json(msg='interface_id is required')
```
willthames, line 163: Should probably say in what circumstance `interface_id` is required (it's probably better to fail before getting to `get_interface_details` as it's not reflecting a function of the module - `return_secondary_address_details` is the only method calling method where `interface_id` isn't checked before the call to `get_interface_details`)

Etherdaemon:
@willthames Thanks for the review Will. I'll update and commit later - cheers.

Etherdaemon:
@willthames @bcoca I've updated the module as per Will's review (thanks again Will).

Updated doco and example indentation to 2 spaces
Updated one of the examples which had 2 states declared
Updated filter/lambda to list comprehension
fixed function associate_eip_to_private_address to check for private addresses that don't match the interface and fail out

wimnat:
Isn't this just duplicating functionality of ec2_eni and ec2_eip?

Etherdaemon:
No.  I needed a module personally for the secondary private ips and attaching and releasing publics to any privates. Neither module was able to do this for me unfortunately.

Etherdaemon:
@wimnat although if you wanted to update the existing modules I'm not opposed to that either.  I do prefer boto3 over boto but as long as something can provide the functionality.  Happy to provide help with porting the relevant parts over if needed.

wimnat:
Could you provide a quick overview of the use case?

Etherdaemon:
Sure.  I currently use my module to assign 2 secondary private ips and relevant EIPs to them.  They come up as virtual interfaces which allows me to route specific requests which require a certificate to that EIP/secondary pair. When I terminate the instance I noticed the EIP module only ever released the first EIP rather than all which left me with a bunch of ips that were now costing money due to being unattached.  Public to secondary privates are recommended over multiple interfaces for ips in the same subnets.

wimnat:
Ah ok makes sense.  I'd certainly like to look at adding the secondary IP functionality to the ec2_eni module.  Not so sure about the public ip association as ec2_eip isn't my module but it could be possible.  Would you be ok with that?  If so, i can probably have the eni changes done by the end of the week.

Etherdaemon:
Sure no worries. Feel free to grab any bits of my code as well.  I'll see if I can do something for the eip module to support it and submit a PR to @rmendes. One thing to keep in mind is that in this module I did have the capability for a user to assign secondary addresses with auto allocated EIPs attached which I found useful and saved me doing 2-3 steps in a play. Also will you be refactoring to boto3 at some point?

wimnat:
OK looking more at ec2_eip it looks like a pretty easy PR to specify the secondary ip to attach to.

I'll submit a PR for both asap unless you get in there first! 

I'd love to get on with refactoring for boto3 but tbh it's pretty low on the list of priorities until it becomes a necessity.

willmcpo:
This module is useful if we wanted to implement a floating ip cloud design pattern. So +1 for this being pulled or ported to existing aws modules

elnur:
Any progress on this? I need to manage secondary private IP addresses of EC2 instances. This PR seems to be targetting this use case. It's been almost half a year since this PR came to life. What's holding it?

wimnat:
The functionality is being ported to other more appropriate modules instead.  Namely, ec2_eni for secondary addresses and ec2_eip for association of secondary addresses

wimnat:
@Etherdaemon support for secondary addresses is now merged in ec2_eni and I have submitted a PR for allowing an EIP to be assigned to a secondary address - https://github.com/ansible/ansible-modules-core/pull/3604.

I think this means this is no longer required?

PR #1277, opened 2015-11-23
Add module for modifying SmartOS image sources

xen0l:
This module implements management of SmartOS image sources.

### DIFF
--- a/cloud/smartos/smartos_image_source.py
+++ b/cloud/smartos/smartos_image_source.py
@@ -1,0 +1,162 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2015, Adam Števko <adam.stevko@gmail.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible. If not, see <http://www.gnu.org/licenses/>.
+#
+
+DOCUMENTATION = '''
+---
+module: smartos_image_source
+short_description: Manage image sources on SmartOS compute nodes.
+description:
+    - Add/delete image source on SmartOS compute nodes.
+version_added: "2.0"
+author: Adam Števko (@xen0l)
+options:
+    url:
+        description:
+            - URL of the image server.
+        required: true
+    type:
+        description:
+            - Specifies the type of the image server.
+        required: false
+        default: imgapi
+        choices: [ "imgapi", "docker", "dsapi" ]
+    state:
+        description:
+            - Set or reset the property value.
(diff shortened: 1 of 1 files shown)


lenw:
+1

xen0l:
Anyone else?

cloud/smartos/smartos_image_source.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+def main():
+    module = AnsibleModule(
+        argument_spec=dict(
+            url=dict(required=True),
+            type=dict(required=False, default='imgapi',
+                      choices=SUPPORTED_IMAGE_SOURCE_TYPES),
+            state=dict(
+                default='present', choices=['absent', 'present', 'reset']),
```
mscherer, line 115: So, what does "reset" do ? This is not handled by the code
( and if there is just present and absent, I think the condiction can later be simplified to just if/else rather than if/elsif)

cloud/smartos/smartos_image_source.py
```diff
@@ -0,0 +1,162 @@
...
+    result['url'] = image_source.url
+
+    if image_source.state == 'absent':
+        if image_source.image_source_exists():
+            if module.check_mode:
+                module.exit_json(changed=True)
+
+            (rc, out, err) = image_source.delete_image_source()
+            if rc != 0:
+                module.fail_json(msg='Failed to delete %s' % image_source.url)
```
mscherer, line 137: It should ideally give the error message if it fail. Might help for debugging

cloud/smartos/smartos_image_source.py
```diff
@@ -0,0 +1,162 @@
...
+        cmd = [self.module.get_bin_path('imgadm')]
+
+        cmd.append('sources')
+
+        if self.module.check_mode:
+            return True
+
+        (rc, out, _) = self.module.run_command(cmd)
+
+        if rc == 0 and self.url in out:
```
mscherer, line 82: So what happen if I try to add as a source http://docker.io while a image source 'http://docker.io.example.org' already exist, wouldn't it it be blocked ( since 'http://docker.io' in 'foo\nhttp://docker.io.example.org\n' would be True ?

cloud/smartos/smartos_image_source.py
```diff
@@ -0,0 +1,162 @@
...
+    out = ''
+    err = ''
+    result = {}
+    result['state'] = image_source.state
+    result['type'] = image_source.type
+    result['url'] = image_source.url
+
+    if image_source.state == 'absent':
+        if image_source.image_source_exists():
+            if module.check_mode:
```
pilou-, line 132: The value of `check_mode` is tested in `image_source_exists` method too.
With `check_mode` enabled, `result['changed']` is:
- always `True` if `state` is `absent`;
- always `False` if `state` is `present`.

This module supports `check_mode` (cf `support_check_mode`), so when `check_mode` is enabled, effective state of the image must be taken in account: the test on `check_mode` in `image_source_exists` should be removed.

cloud/smartos/smartos_image_source.py
```diff
@@ -0,0 +1,162 @@
...
+    elif image_source.state == 'present':
+        if not image_source.image_source_exists():
+            if module.check_mode:
+                module.exit_json(changed=True)
+
+            (rc, out, err) = image_source.add_image_source()
+            if rc != 0:
+                module.fail_json(msg='Failed to add %s' % image_source.url)
+
+    if rc is None:
```
pilou-, line 148: You could use `result['changed'] = rc is not None`.

PR #1283, opened 2015-11-25
Added module to deploy/undeploy JEE applications to the Glassfish application server

nicholasmckinney:


nicholasmckinney:
Ok, I see the error due to the lack of "with" in Python 2.4. I'll get that fixed...

nicholasmckinney:
I ostensibly made the necessary tweaks to get it to work under Python 2.4

gregdek:
Thanks @NickMk for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

web_infrastructure/glassfish_deploy.py
```diff
@@ -0,0 +1,269 @@
...
+    payload = io.BytesIO()
+    needsCLRF = False
+    try:
+        f = open(file, 'rb')
+        try:
+            data = f.read()
+            writeFile(GlassfishFileField, getFilename(file), data, field_boundary, payload, needsCLRF)
+        finally:
+            f.close()
+    except Exception, e:
```
pilou-, line 175: This `try` statement is useless.

0mmariano (changes requested):
Can we hide the password provided?

Update this line --> url_password=   dict(required=True),
To this --> url_password=   dict(required=True, no_log=True),

Also, module doesn't look like it's idempotent. It returns a failure when you run the playbook again, with the application/war already deployed, even if you explicitly set force=false (default value). I would think it'll just skip over the deployment since the app is already deployed.

PR #1284, opened 2015-11-25
removed all import json references 

tronicum:
module_utils/basic.py already takes care of that (and in a robust way).


bcoca:
first thing, removing json from requirements is incorrect.

many of these modules already require python 2.6 or above, in which case we really don't need to remove the json import nor fallback to simplejson

lastly, the preferred solution is not removing the json imports but doing what you've removed from some modules, try json and if this fails fallback to simplejson but rename it to json.

For all of the above, I'm going to close this PR as it won't be merged.

tronicum:
I am fine with the requirement should stay or the pyhton dependancy. But i am a little confused about the json import. You told me at https://github.com/ansible/ansible/issues/13231 that it should not be included and removed it from the puppet module. IMHO it is quit redundant to have such try json in every submodule as the basic class already does the magic.

bcoca:
That was my mistake, after discussing with the rest of core we came to the conclusion that it is better to have it on all modules as this will not break when we redo how the shared code is used, from a copy/paste to an actual import

Issue #1273, opened 2015-11-23
puppet module uses timout binary


tronicum:
ansible version: git latest
issue type: bug/feature request

summary: the puppet modules uses the "timeout" binary (Line 215) from GNU coreutils to set a puppet agent timeout. centos 5.7 don't have a recent coreutils RPM so fails on such hosts. puppet has a built in timeout parameter. it would be much nicer to use native that option.

robynbergeron:
[module: system/puppet.py]

robynbergeron:
@tronicum -- Thanks for submitting this feature idea.

@emonty -- Pinging you to bring this feature request to your attention. Please note that you are under no particular obligation to add this feature.

We encourage all Ansible community members to submit pull requests to satisfy feature requests. Consult the module guidelines (http://docs.ansible.com/ansible/developing_modules.html) for more information.

emonty:
@tronicum we actually use the timeout binary because the timeout parameter in agent does not protect against puppet agent hangs - which we actually had consistent problems with in a large farm. However, it should be simple enough to detect if timeout exists on the host and if not pass timeout to agent anyway.

tronicum:
That is a nice solution, we have an mixed cluster with old and new centos vms so it could use timout if exists

Jmainguy:
With #1285 getting merged, can this be closed now?

bcoca:
closing as per above

PR #1285, opened 2015-11-25
Don't use the timeout command if it's not there

emonty:
The timeout command is there to improve life and workaround puppet
deficiencies. However, it's not working around deficiencies on systems
that do not have the timeout command if we blindly use it.

The puppet specific timeout options are more complex and out of scope of this.

Issue: #1273

### DIFF
--- a/system/puppet.py
+++ b/system/puppet.py
@@ -19,6 +19,11 @@ import os
 import pipes
 import stat
 
+try:
+    import json
+except ImportError:
+    import simplejson as json
+
 DOCUMENTATION = '''
 ---
 module: puppet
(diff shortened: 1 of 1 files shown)


bcoca:
i had removed the import json as we had a bug report about it, but I should really add the whole try/json/except/simplejson ugliness

gregdek:
Thanks @emonty. Since you are the owner of this module, we are marking this PR for immediate inclusion.

gregdek:
Oh, hey, by the way, this was the result of a bot pass!

The bot looked at this PR and said "hey, this is owner_pr, so we're shipping it." Because the assumption is "if you committed it to your own module, it must be good!"

If that's not the right behavior, we should probably figure out what the right behavior is. (I still think it's the right behavior, fwiw.)

robynbergeron:
@gregdek so disappoint that @ansibull is not the commenter :cry::cow:

bcoca:
@emonty could you flatten these commits?

emonty:
On 12/14/2015 09:52 PM, Brian Coca wrote:

> @emonty https://github.com/emonty could you flatten these commits?
> 
> —
> https://github.com/ansible/ansible-modules-extras/pull/1285#issuecomment-164628200.

I surely can!

gregdek:
Thanks @emonty. @emonty please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

Issue #1236, opened 2015-11-12
win_regedit: can't remove registry entries following #1107


jhawkesworth:
##### Issue Type:
- Bug Report
##### Ansible Version:

ansible 2.0.0 (reg_merge_tests 7b83698ac2) last updated 2015/10/27 06:29:59 (GMT +100)

(yes this is a branch but the branch just adds integration tests, so likely the same on latest devel.
##### Ansible Configuration:

config file =
configured module search path = None
##### Environment:

Ansible Controller: Lubuntu 14.04
Targets:
Windows Server 2008 R2 (Powershell 3.0)
Windows Server 2012 R2 (Powershell 4.0)
##### Summary:

Since https://github.com/ansible/ansible-modules-extras/commit/c6a934d57b9d0b171ad6a4c45c14ca5087a941c0

It is not possible to remove registry entries using a key

I think pre-pending 'Registry::' to the registry key is some how causing Test-Path to fail.   Note that the example registry key has a space in the name.
##### Steps To Reproduce:

Add a registry key called 'Cow Corp' to  'HKLM:\SOFTWARE\Wow6432Node'

Run the following task in a playbook:

win_regedit:
  key: 'HKLM:\SOFTWARE\Wow6432Node\Cow Corp'
  state: absent
##### Expected Results:

win_regedit should report changed: true
##### Actual Results:

win_regedit reports changed: false

jhawkesworth:
@smadam813 - can you spare any cycles to investigate?
cc ing @bgeihsgt for info/comment.

robynbergeron:
[module: windows/win_regedit.py]

robynbergeron:
@jhawkesworth -- Thanks for submitting this bug report.

@smadam813 @joshludwig --  thanks in advance for taking a look at this bug report and for your ongoing work in maintaining this module. And re-pinging @bgeihsgt as @jhawkesworth did originally :)

smadam813:
Sorry for not stepping in earlier!  I have been crazy busy lately, so I will take a look as soon as I can.

smadam813:
I submitted a pull request (#1286) which fixes this issue.  @jhawkesworth Please let me know if this fixes the issue for you!

jhawkesworth:
To confirm here as well.. Yes https://github.com/ansible/ansible-modules-extras/pull/1286 fixes this issue.

Once https://github.com/ansible/ansible-modules-extras/pull/1286 has been merged this can be closed.

Many thanks everyone.

Issue #1107, opened 2015-10-14
win_regedit: Failure in parameter validation and in creating registry key


bgeihsgt:
##### Issue Type:

Bugfix Pull Request
##### Ansible Version:

```
ansible 2.0.0 (detached HEAD 5ae74f7d38) last updated 2015/10/13 06:57:44 (GMT +000)
  lib/ansible/modules/core: (detached HEAD d659c79db7) last updated 2015/10/13 06:57:49 (GMT +000)
  lib/ansible/modules/extras: (detached HEAD 0ff1776a84) last updated 2015/10/13 06:58:00 (GMT +000)
  config file = /vagrant/ansible/ansible.cfg
  configured module search path = /vagrant/ansible/library
```
##### Ansible Configuration:

I haven't changed anything that affects this module.
##### Environment:

Running from Ubuntu 14, managing Windows Server 2012 R2.
##### Summary:

Two issues: 
1. win_regedit fails with property error when parsing parameters
2. win_regedit fails to create a new registry key with a NonInteractiveMode error
##### Steps To Reproduce:
- Use win_regedit to create a new registry value in a new key
##### Expected Results:

It successfully creates the key
##### Actual Results:

It fails with strict mode errors when parsing parameters.  After moving the parameter parsing to use Get-Attr, it started failing saying you can't prompt in NonInteractiveMode.  Digging in, I saw that it was prompting for the Type parameter in the New-Item call to create the key.  I fixed this by prepending a "Registry::" before the path so Powershell knows it's a registry path.  Pull request coming shortly.

robynbergeron:
@bgeihsgt -- Thanks for submitting this bug report, and the accompanying PR (#1108).

@smadam813 & @joshludwig --  thanks in advance for taking a look at this bug report and for your ongoing work in maintaining this module. Please note there is a PR addressing this issue.

robynbergeron:
[module: windows/win_regedit.py]

smadam813:
Will take a look!  Thanks for the bug submission!

dreamcat4:
Hi there. I also found problems with this module accepting certain input values. Specifically when trying to set the `data:` or a registry key to be `0` or `false`, `no` or something like that. Then it gets confused and thinks the `data` param doesn't exist or wasn't set to any specific value.

The module is also _very slow_ to process any loop of multiple items using the outside looping of `with_items:`. As it re-uploads a new copy of the same script for setting each individual key. And each winrm file transfers are very slow.

[...]

PR #1286, opened 2015-11-25
Fixing Issue #1236

smadam813:
Regarding Issue #1107 - Appending "Registry::" no longer works correctly.  I think it has to do with updates to "Get-Attr".  The error I receive when appending "Registry::" and attempting to create a key:

```
{"changed": false, "failed": true, "msg": "Cannot process argument because the value of argument \"path\" is not valid. Change the value of the \"path\" argument and run the operation again."}
```

I have tested this on Windows 7, 8.1, 10, and Server 2012.  With "Registry::" appended, the script failed to create registry keys as well as delete them.  If there are errors regarding interactive mode, it is how the account is authenticating / account permissions.

Removing the appended "Registry::" has corrected all functionality when testing locally.

Fix for: #1236 
@jhawkesworth Please let me know if this fixes the issue for you!

### DIFF
--- a/windows/win_regedit.ps1
+++ b/windows/win_regedit.ps1
@@ -31,8 +31,6 @@ $state = Get-Attr -obj $params -name "state" -validateSet "present","absent" -de
 $registryData = Get-Attr -obj $params -name "data" -default $null
 $registryDataType = Get-Attr -obj $params -name "datatype" -validateSet "binary","dword","expandstring","multistring","string","qword" -default "string"
 
-$registryKey = "Registry::" + $registryKey
-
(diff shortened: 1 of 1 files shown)


jhawkesworth:
@smadam813 Thanks for this fix.  I can confirm it works for me.

gregdek:
Thanks @smadam813 for the patch.

Thanks @jhawkesworth for the review, and a reminder to please use the term "shipit" in the comment so our new bots can catch it. :)

jhawkesworth:
Sorry @gregdek  I didn't say "shipit" - I thought it was @smadam813 's call as module owner.
Anyway hope this can be merged soon.

PR #1296, opened 2015-11-29
add openbsd_rcctl, to manage service status on the OpenBSD operating …

sarevok-anchev:
This one is useful when using a recent OpenBSD install. Let me know if anything can be improved. I tested it as thouroughly as I could and everything seems to check out, but quite possibly missed something.

Cheers!

### DIFF
--- a/system/openbsd_rcctl
+++ b/system/openbsd_rcctl
@@ -1,0 +1,157 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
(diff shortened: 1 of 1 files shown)


robynbergeron:
@sarevok-anchev  -- 

eest:
The service module already has rcctl support in it, does it not work for you? Look at the " Subclass: OpenBSD" part of https://github.com/ansible/ansible-modules-core/blob/devel/system/service.py

PR #1298, opened 2015-11-29
safe json importing

tronicum:
exchanged all json imports with a fail safe variant for simplejson
See also pull request #1284.
import code is taken from module_utils/basic.py
redundancy is wanted as stated in above pull request notes.

### DIFF
--- a/cloud/amazon/ecs_cluster.py
+++ b/cloud/amazon/ecs_cluster.py
@@ -101,7 +101,25 @@ status:
     type: string
 '''
 try:
-    import json, time
+    import json
+    # Detect the python-json library which is incompatible
+    # Look for simplejson if that's the case
+    try:
+        if not isinstance(json.loads, types.FunctionType) or not isinstance(json.dumps, types.FunctionType):
+            raise ImportError
+    except AttributeError:
+        raise ImportError
+except ImportError:
+    try:
+        import simplejson as json
+    except ImportError:
+        print('{"msg": "Error: ansible requires the stdlib json or simplejson module, neither was found!", "failed": true}')
+        sys.exit(1)
+    except SyntaxError:
+        print('{"msg": "SyntaxError: probably due to installed simplejson being for a different python version", "failed": true}')
+        sys.exit(1)
+try:
+    time
     import boto
     HAS_BOTO = True
 except ImportError:
--- a/cloud/amazon/ecs_task.py
+++ b/cloud/amazon/ecs_task.py
@@ -99,6 +99,23 @@ task:
 '''
 try:
     import json
+    # Detect the python-json library which is incompatible
+    # Look for simplejson if that's the case
+    try:
+        if not isinstance(json.loads, types.FunctionType) or not isinstance(json.dumps, types.FunctionType):
+            raise ImportError
+    except AttributeError:
+        raise ImportError
+except ImportError:
+    try:
+        import simplejson as json
+    except ImportError:
+        print('{"msg": "Error: ansible requires the stdlib json or simplejson module, neither was found!", "failed": true}')
+        sys.exit(1)
+    except SyntaxError:
+        print('{"msg": "SyntaxError: probably due to installed simplejson being for a different python version", "failed": true}')
+        sys.exit(1)
+try:
     import boto
     import botocore
     HAS_BOTO = True
--- a/cloud/amazon/ecs_taskdefinition.py
+++ b/cloud/amazon/ecs_taskdefinition.py
@@ -96,6 +96,23 @@ taskdefinition:
 '''
 try:
     import json
+    # Detect the python-json library which is incompatible
+    # Look for simplejson if that's the case
+    try:
+        if not isinstance(json.loads, types.FunctionType) or not isinstance(json.dumps, types.FunctionType):
+            raise ImportError
+    except AttributeError:
+        raise ImportError
+except ImportError:
+    try:
+        import simplejson as json
+    except ImportError:
+        print('{"msg": "Error: ansible requires the stdlib json or simplejson module, neither was found!", "failed": true}')
+        sys.exit(1)
+    except SyntaxError:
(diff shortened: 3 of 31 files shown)


bcoca:
First, you are using `print` which will not result in what you want, in modules  errors need to be returned in json format, which is what fail_json function is for.

Also you are using sys.exit, again another thing we don't use in modules as the fail_json and exit_json functions already take care of this, included modules should not terminate any other way.

lastly, most of these modules do not need to attempt simplejson as they already require python 2.6 (mostly through library they depend on) or above, which guarantees the inclusion of the json library.

tronicum:
Well that code was taken from module_utils/basic.py

gregdek:
Thanks @tronicum for this PR. Unfortunately, it is not mergeable in its current state due to merge conflicts. Please rebase your PR. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

tronicum:
ready_for_review

robynbergeron:
Thanks @tronicum. @tarka please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

robynbergeron:
Thanks @tronicum. @tarka please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

tarka:
shipit

robynbergeron:
Thanks again to @tronicum for this PR, and thanks @tarka for reviewing. Marking for inclusion.

tronicum:
so what is the status? shipit or not?

robynbergeron:
Thanks @tronicum. @tarka please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

tarka:
Already marked as "shipit".

robynbergeron:
Thanks again to @tronicum for this PR, and thanks @tarka for reviewing. Marking for inclusion.

abadger:
@tronicum so I guess the first question is -- what problem does this solve for you?  If it's just attempting to cleanup style then we'd rather wait for ziploader to land (https://groups.google.com/d/msg/ansible-devel/Tv_9GXnDJRI/XitYR7erCwAJ ) and then change the import to "from ansible.module_utils.basic import json" if that's deemed necessary.

If you actually have a problem with the json namespace being used by something other than the python stdlib json library then we could consider something to fix that.  @bcoca has made a good case internally for removing the "import json" for now and adding the "from ansible.module_utils.basic import json" once ziploader is in place.  I was championing the case of having "import json" in as that would be closer to the proper thing to do once ziploader is available.  but if you're having actual problems due to having a third-party json module that does the wrong thing then I'd be willing to shift my position.

tronicum:
My problem is quite simple but persistent. I got a bunch of CentOS5 server that wont be upgraded. They have simplejson installed and only work if json is imported with fallback to simple json. If ziploader is coming soon (like in january) I dont care. It looks like a more elegant solution (similar to the solution that the basic class cares about json importing not the module itself, that was subject of an earlier PR but was reject due to internal discussion to not include that code in the basic module).

abadger:
I'm hoping that ziploader will hit in 2.1.0.  However, 2.0 is what we're working on releasing in January so it will be about 4 months before it is in an ansible release.

Looking at your use case, I don't think we need most of this boilerplate.  In the module code we should be able to just get away with this:

``` python
try:
     import json
except ImportError:
    try:
        import simplejson as json
    except:
        # Let snippet from module_utils/basic.py return a proper error in this case
        pass
```

The other portions of the module_utils code deal with two corner cases:
1. A different module with an incompatible API occupies the json namespace. Importing json here shouldn't interfere with the module_utils snippet doing that check later so we don't need to repeat it here.
2. Neither simplejson nor json are present on the system and we want to return a proper ansible error for that.  We have to prevent a traceback in this case but.  We don't have to return the error here, though, as the module_utils snippet will also do that check later and return the error properly.

I think the recipe I give above is simple enough that we can put that in the modules instead of copying the entirety of what module_utils does.

For use on CentOS5 we also shouldn't need to modify all of these modules.  Only the modules that actually can run with python-2.4 need to be changed.  Some of the modules are also importing json but not using the json library at all.  For those, it's best to remove the "import json" altogether.

No need to change:
- sns_topic.py  [requires boto which requires python-2.6+]
- docker_login.py [requires docker-py which requires python-2.6+]
- rabbitmq_binding.py [requires requests which requires python-2.6+]
- rabbitmq_exchange.py [requires requests which requires python-2.6+]
- rabbitmq_policy.py [requires requests which requires python-2.6+]
- rabbitmq_queue.py [requires requests which requires python-2.6+]
- circonus_annotation.py [ requires requests which requires python-2.6+]
- typetalk [ has a proper fallback to simplejson with error handling]

Safe to remove:
- ecs_task.py
- ecs_taskdefinition.py
- route53_facts.py
- vca_nat.py
- consul.py
- consul_kv.py
- pacman.py
- pkgin.py
- pkgng.py
- portinstall.py
- urpmi.py

Needs modification:
- riak.py [falls back to simplejson but will traceback if neither is installed]
- boundary_meter.py
- sensu_check.py [falls back to simplejson but will traceback if neither is installed]
- stackdriver.py [falls back to simplejson but will traceback if neither is installed]
- uptimerobot.py]
- ipify_facts.py [falls back to simplejson but will traceback if neither is installed]
- composer.py  [falls back to simplejson but will traceback if neither is installed]
- npm.py [falls back to simplejson but will traceback if neither is installed]
- github_hooks.py [falls back to simplejson but will traceback if neither is installed]
- jira.py
- puppet.py

I'd be uncomfortable pushing this to 2.0.0 as we're planning on releasing that tomorrow but I can see it going into 2.0.1.  I'll work up a fix.

tronicum:
works like a charm. thanks @abadger

PR #1301, opened 2015-11-29
Adds uncompress module

Jmainguy:
This module allows for uncompressing a single compressed file
using either gzip, bzip2, or xz. I decided not to include zip
or 7zip as they always are archived as well,
and are better suited for the unarchive module.


robynbergeron:
@Jmainguy  -- Thanks for submitting this new module to Ansible Extras! This module is now in community review, a process that is open to all Ansible users. In order for this module to be approved, it must gain the following votes, which can be contributed by anyone other than the author of the new module:

files/uncompress.py
```diff
@@ -0,0 +1,265 @@
...
+
+    return msg
+
+def unxzip(module, src, dest):
+    """
+    Uncompress xz files. Since we must support python 2.4 (EL 5) we cannot import lzma and use native python.
+    Also, xz does not support a destination flag, so we must use > to redirect stdout to file. Really lame.
+    """
+    cmd_path = module.get_bin_path('xz')
+    cmd = '%s -c -d %s > %s' % (cmd_path, src, dest)
```
mscherer: Shouldn't we escape the path for shell metacharacters ? This seems rather unsafe, and while the stuff from a playbook is likely under a sysadmin control, we can't be sure someone will not take data coming from a non admin controlled location.

files/uncompress.py
```diff
@@ -0,0 +1,265 @@
...
+
+    return msg
+
+def filetype(module, src):
+    """
+    Get the filetype from the unix command file, and then pick the correct compression method to use for it
+    """
+    cmd_path = module.get_bin_path('file')
+    cmd = "%s -b -i %s" % (cmd_path, src)
+    ftype = module.run_command(cmd, use_unsafe_shell=True)
```
mscherer: Why do we need use_unsafe_shell here ?

files/uncompress.py
```diff
@@ -0,0 +1,265 @@
...
+                while True:
+                    data = rsp.read(BUFSIZE)
+
+                    if data == "":
+                        break # End of file, break while loop
+
+                    f.write(data)
+                f.close()
+                src = package
+            except Exception, e:
```
mscherer, line 228: There is no cleanup of the file if the download fail partially. IE, let's say that I start to download a 3G file and it break after 2G, I just have 2G of file in a directory which will not be cleaned by default (since that's ~/.ansible/tmp ).

files/uncompress.py
```diff
@@ -0,0 +1,265 @@
...
+        ),
+        add_file_common_args=True,
+    )
+
+    src         = os.path.expanduser(module.params['src'])
+    dest        = os.path.expanduser(module.params['dest'])
+    copy        = module.params['copy']
+    deep_check  = module.params['deep_check']
+    file_args   = module.load_file_common_arguments(module.params)
+    tempdir     = os.path.dirname(__file__)
```
mscherer, line 205: Do we really want to make all download in the user home dir ? This might cause issue if people use it on NFS or have quota ?

But if we switch to /tmp, it should go in a specific tmp directory, and sometime /tmp can have quota too.

files/uncompress.py
mscherer, line 228: So discussing on IRC, it should be cleaned automatically. However, looking at my own server, i see lots of left over in ~/.ansible/tmp, so I am not sure how everything is cleaned and if everything work fine for that.

files/uncompress.py
```diff
@@ -0,0 +1,265 @@
...
+# saving to a tempfile (64k)
+BUFSIZE = 65536
+
+def ungzip(src, dest):
+    """
+    Uncompress gzip files.
+    """
+    try:
+        f_out = open(dest, 'wb')
+        f_in = gzip.open(src, 'rb')
```
pilou-, line 88: `f_out.close()` is not called when `gzip.open` raises an exception. Same problem in the method below (`unbzip`).

files/uncompress.py
```diff
@@ -0,0 +1,265 @@
...
+    return ftype
+
+def copyfile(src, dest, deep_check):
+    """
+    Copy file from tempsrc to final destination. Unless its already at dest, and the same as tempsrc.
+    """
+    changed = False
+    if os.path.isfile(dest):
+        # This takes a long time
+        if deep_check is True:
```
pilou-: use `if deep_check:` instead

Jmainguy:
@pilou- @mscherer Thank you for the comments. I have fixup up the file as recommended. No longer using > or unsafe_shell. Also added .close() in exceptions. The tmp_dir is set by the user, and deleted automatically unless they set it not to, so I am comfortable with the locations used, and lack of extra clean up

sivel:
Here are some additional notes found while performing an automated scan of this new module:

```
uncompress.py
ERROR: version_added should be 2.1. Currently 2.0
ERROR: No RETURN provided
```

Jmainguy:
@sivel What does no RETURN provided mean

bcoca:
document return values with a `RETURN='''` section, check stat module for example

Jmainguy:
@sivel @bcoca and in this module, I am only returning changed, so a RETUN = section is not needed correct?

Jmainguy:
Updated to reflect version 2.1

sivel:
I'm not positive.  Per recent discussions, all new modules must contain a `RETURN` although I am not sure.

Maybe the better question, is whether it would make sense to return more information from the module?

jctanner:
deprecated by http://docs.ansible.com/ansible/unarchive_module.html

jctanner:
Guess I misunderstood the context of this one. @Jmainguy if you still have an interest in this, please migrate it to ansible/ansible

PR #1304, opened 2015-11-30
Fix win_regedit present fails on create new entry of registry

miyamiyaz:
The win_regedit module in its current state fails when specified registry key does not exist.
This was due to `New-ItemProperty` prints extra message like below, and then ansible could not parse json in it.

```
{20D04FE0-3AEA-1069-A2D8-08002B30309D} : 0
PSPath                                 : Microsoft.PowerShell.Core\Registry::HK
                                         CU\Software\Microsoft\Windows\CurrentV
                                         ersion\Explorer\HideDesktopIcons\NewSt
                                         artPanel
PSParentPath                           : Microsoft.PowerShell.Core\Registry::HK
                                         CU\Software\Microsoft\Windows\CurrentV
                                         ersion\Explorer\HideDesktopIcons
PSChildName                            : NewStartPanel
PSProvider                             : Microsoft.PowerShell.Core\Registry

{"changed":true}

```

### DIFF
--- a/windows/win_regedit.ps1
+++ b/windows/win_regedit.ps1
@@ -107,7 +107,7 @@ if($state -eq "present") {
             $result.changed = $true
 
             if($registryValue -ne $null) {
-                $newRegistryKey | New-ItemProperty -Name $registryValue -Value $registryData -Force -PropertyType $registryDataType
+                $newRegistryKey | New-ItemProperty -Name $registryValue -Value $registryData -Force -PropertyType $registryDataType | Out-Null
                 $result.changed = $true
             }
         }


robynbergeron:
Hi @miyazakijunichi -- thanks for submitting this PR. 

@smadam813 @joshludwig -- Please review this PR to make sure it adheres to the following guidelines:

Thanks!

gregdek:
@joshludwig This change is still pending your review; do you have time to take a look and comment? Please comment with text 'shipit' or 'needs_revision' as appropriate.

h-hirokawa:
@miyazakijunichi 
In `win_regedit.ps1`, `New-ItemProperty` is used 3 times.
So it seems that you should add `Out-Null` to all of them.

robynbergeron:
@joshludwig still waiting on your review.  Please comment with text 'shipit' or 'needs_revision' as appropriate. If we don't hear from you within 14 days, we will start to look for additional maintainers for this module.

smadam813:
I am going to work on testing to see if this is actually an issue.  I don't recall every experiencing this issue.

jhawkesworth:
Hi @smadam813 - I haven't seen this behaviour either.
I wonder if it occurs on Powershell 5?  Or when different language defaults are in use (both speculation, I haven't tested).
For what is worth I have an open PR with an integration test which uses win_regedit here which you might be able to tweak to test: https://github.com/ansible/ansible/pull/13481

Jon

smadam813:
Thanks @jhawkesworth !

h-hirokawa:
Thanks for reviewing, @jhawkesworth @smadam813 !
I have the same issue when using Powershell 3 on Japanese localized Windows server 2008r2 sp1.
So this issue maybe caused by the system locale.
I'm going to try @jhawkesworth 's integration test.

jhawkesworth:
Please can @miyazakijunichi or @h-hirokawa share a small playbook that reproduces this?

nitzmahone:
I don't believe this is an issue anymore since we started filtering non-JSON leading lines from the output. There were a number of other instances of this behavior in this module that weren't corrected (cmdlet calls where we didn't capture the output to a var)- I tested a bunch of different ways and we always seem to clean it up. Still not a bad idea to capture or Out-Null every cmdlet instead of letting it flow to the pipeline (performance, noise, "what if it looks enough like JSON to trip us"), but I'm going to close for now and add this behavior to the draft Windows module guidelines.

PR #1237, opened 2015-11-12
added option to specify module version

pileofrogs:
The cpanm module currently won't perform module upgrades because it only tests if a module is installed and not if it is up to date.  This patch adds a version option so the test for _is_package_installed can fail if the installed version is too old.

PR #1306, opened 2015-11-30
added option to specify module version

pileofrogs:
The cpanm module currently won't perform module upgrades because it only tests if a module is installed and not if it is up to date. This patch adds a version option so the test for _is_package_installed can fail if the installed version is too old.

@franckcuny and @robynbergeron
Sorry, I'm a nitwit.  I already posted this PR and then I broke my fork, so it closed the PR.  The closed PR is #1237.  D'oh!

### DIFF
--- a/packaging/language/cpanm.py
+++ b/packaging/language/cpanm.py
@@ -64,6 +64,12 @@ options:
     required: false
     default: false
     version_added: "2.0"
+  version:
+    description:
+      - minimum version of perl module to consider acceptable
+    required: false
+    default: false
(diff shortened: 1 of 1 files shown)


robynbergeron:
@pileofrogs no worries, :hankey: happens! :D 

Since @franckcuny already gave this the thumbs up in the previous PR #1237 -- I will go ahead and mark this for inclusion. 

Thanks to both of you :D

robynbergeron:
@ansible/core team -- put this one on the :ship: :D

PR #1314, opened 2015-12-01
Fix mongodb_user docs typo

joshgachnang:
Bob's last name is [Belcher](http://bobs-burgers.wikia.com/wiki/Bob_Belcher.)

These docs made me chuckle, so thanks :)


bcoca:
idk, Bob Lecher is also funny, just not a cartoon character (it should be!)

joshgachnang:
;)

PR #1319, opened 2015-12-02
vmware_vm_shell: add the ability to start program without network connection

ritzk:
vmware_vm_shell: add the ability to start program without network connection

### DIFF
--- a/cloud/vmware/vmware_vm_shell.py
+++ b/cloud/vmware/vmware_vm_shell.py
@@ -1,0 +1,191 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2015, Ritesh Khadgaray <khadgaray () gmail.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: vmware_vm_shell
+short_description: Execute a process in VM
+description:
+    - Start a program in a VM without the need for network connection
+version_added: 2.0
+author: "Ritesh Khadgaray (@ritzk)"
+notes:
+    - Tested on vSphere 5.5
+    - Only the first match against vm_id is used, even if there are multiple matches
+requirements:
+    - "python >= 2.6"
+    - PyVmomi
+options:
+    hostname:
+        description:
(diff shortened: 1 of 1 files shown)


robynbergeron:
Thanks @ritzk for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

gregdek:
Thanks @ritzk for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

cloud/vmware/vmware_vm_shell.py
```diff
@@ -0,0 +1,191 @@
...
+
+'''
+
+try:
+    from pyVmomi import vim, vmodl
+    HAS_PYVMOMI = True
+except ImportError:
+    HAS_PYVMOMI = False
+
+def find_vm(content, vm_id, vm_id_type="dns_name", datacenter=None):
```
Lujeni, line 122: imho, add a try/exception/finally could be useful.

ritzk, line 122: hi @Lujeni 

  Following the convention of all other modules under ansible-modules-extras/cloud/vmware . none of them use finally.

bcoca, line 122: @Lujeni not sure what you want to do in finally, it is done this way so later you can instantiate the module object and use fail_json method to return an error about the library being missing.

cloud/vmware/vmware_vm_shell.py
```diff
@@ -0,0 +1,191 @@
...
+    
+    if vm_id_type == 'dns_name':
+        vm = si.FindByDnsName(datacenter=datacenter, dnsName=vm_id, vmSearch=True)
+    elif vm_id_type == 'inventory_path':
+        vm = si.FindByInventoryPath(inventoryPath=vm_id)
+        if type(vm) != type(vim.VirtualMachine):
+            vm = None
+    elif vm_id_type == 'uuid':
+        vm = si.FindByUuid(datacenter=datacenter, uuid=vm_id, vmSearch=True)
+
```
chrrrles, line 137: This is a common pattern that we use for finding a vm based on the name:

```
    elif vm_id_type == 'vm_name':
      for machine in get_all_objs(content, [vim.VirtualMachine]):
        if machine.name == vm_id:
          vm = machine
```

 Another thing to consider is getting rid of vm_id_type altogether and iterating through each findBy method until a vm is returned.

chrrrles:
Tested and the code looks good to me.

cloud/vmware/vmware_vm_shell.py
ritzk, line 137: @chrrrles

> consider is getting rid of vm_id_type 
> It is a lot faster to search through a datacenter subset, thant to search through the entire list of VMs which I have access to. 

Using  inventory path ( Datacenter/Cluster/VM ) is the fastest method. In other case, where we are lazy, we use dns_name ( which can be different to vm.name) .

Do I go ahead and add the support for search by vm_name as default ?

chrrrles, line 137: Hi @ritzk  - I went ahead and merged this code, and added support for searching by vm_name - I am keeping it as the default just because that is how the other modules operate.  Good point, however, about the speed for searches using inventory paths...

PR #1325, opened 2015-12-04
Added missing else statement that caused existing rules to be discarded

wimnat:


gregdek:
Thanks @wimnat. Since you are the owner of this module, we are marking this PR for inclusion.

wimnat:
Not part of 2.0 rc2

Issue #1161, opened 2015-10-26
win_nssm issue with no appParameters specified


baudren:
#### Issue Type
- Bug Report
#### Ansible Version

1.9.4
#### Ansible Configuration

Contents of `.ansible.cfg` in the root of my project.

``` ini
[defaults]
hostfile = ./hosts
host_key_checking = False
remote_user = ansible

[ssh_connection]
ssh_args = -o PasswordAuthentication=no -o ControlMaster=auto
pipelining = True
```

Using the ansible extra `win_nssm` (copied the `.ps1` and `.py` file).
#### Environment
- Debian 7.7 Wheezy to run Ansible
- Managing a Windows 7 Enterprise, SP1 (connection via a kerberos ticket)
#### Summary

When defining a service with `win_nssm` based on a simple `.bat` script file, running **without** any `appParameter` option leads to the following error:

```
msg: Cannot bind argument to parameter 'appParameters' because it is an empty string.
```
#### Steps To Reproduce

Example of a task:

``` yaml
- name: create a simple service
  win_nssm: >
    name=toto
    application={{home}}toto.bat
    state=present
```

The file `toto.bat` can be anything. The problem is that the `appParameters` option, when not specified, is set to `$null`. The error

```
msg: Cannot bind argument to parameter 'appParameters' because it is an empty string.
```

Appears when running the task. I found [these lines](https://github.com/ansible/ansible-modules-extras/blob/devel/windows/win_nssm.ps1#L62-L69) where, when unspecified, the value of `appParameters` is set to `$null`. Combined with the fact that this `appParameters` seems later on to be mandatory, this might confuse `NSSM`.
#### Expected Results

I am expecting to be able to launch a service without any `appParameters`.
#### Actual Results

A service without `appParameters` is not successfully initialized through `win_nssm`. By changing the default value of [this line](https://github.com/ansible/ansible-modules-extras/blob/devel/windows/win_nssm.ps1#L68) to `" "` (note the space between the quotes), the problem disappears.

robynbergeron:
[module: windows/win_nssm.py]

@baudren Thanks for submitting this bug report.

@georgefrank @h0nIg --  thanks in advance for taking a look at this bug report and for your ongoing work in maintaining this module.

georgefrank:
Taking a look now. Thanks for the bug report!

nitzmahone:
Should be fixed by #1327, closing

PR #1327, opened 2015-12-04
Use an empty string instead of null for AppParameters

georgefrank:
Apparently running "reset" on AppParameters in nssm crashes nssm (and clears the AppParameters). Setting AppParameters to an empty string clears AppParameters without the crash, so that's what's going on here.

This is for https://github.com/ansible/ansible-modules-extras/issues/1161


gregdek:
Thanks @georgefrank. Since you are a maintainer of this module, we are marking this PR for inclusion.

PR #1328, opened 2015-12-04
AS-2560: First steps: Refactoring RedShift args to Ansible style: snake_case

trinitronx:
I've converted the existing arguments for boto3 into Ansible's "snake_case" style module args.
I tried to add a bunch of aliases for the native boto3 Redshfit::Client.create_cluster params also in snake_case,
so with any luck, those _should_ just work.

Note that I haven't found time to test this yet, so YMMV!

### DIFF
--- a/cloud/amazon/redshift.py
+++ b/cloud/amazon/redshift.py
@@ -1,0 +1,176 @@
+#!/usr/bin/python
+# This file is part of Ansible
+#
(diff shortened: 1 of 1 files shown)


trinitronx:
sorry.. wrong fork / repo! (hub gem always wants `origin` apparently...)

PR #1332, opened 2015-12-05
update current_link without downtime

robberphex:
if `current_path` is a link, `deploy_helper` will unlink it and re-link it. That will cause downtime, such as http 404 error.

`os.rename` can avoid this downtime.

### DIFF
--- a/web_infrastructure/deploy_helper.py
+++ b/web_infrastructure/deploy_helper.py
@@ -308,12 +308,27 @@ class DeployHelper(object):
                 self.module.fail_json(msg="%s exists but is not a symbolic link" % path)
 
     def create_link(self, source, link_name):
-        if not self.module.check_mode:
-            if os.path.islink(link_name):
-                os.unlink(link_name)
-            os.symlink(source, link_name)
+        changed = False
 
-        return True
+        if os.path.islink(link_name):
+            norm_link = os.path.normpath(os.path.realpath(link_name))
+            norm_source = os.path.normpath(os.path.realpath(source))
+            if norm_link == norm_source:
+                changed = False
+            else:
+                changed = True
+                if not self.module.check_mode:
+                    tmp_link_name = link_name + '.' + self.unfinished_filename
+                    if os.path.islink(tmp_link_name):
+                        os.unlink(tmp_link_name)
+                    os.symlink(source, tmp_link_name)
+                    os.rename(tmp_link_name, link_name)
+        else:
+            changed = True
+            if not self.module.check_mode:
+                os.symlink(source, link_name)
+
+        return changed
 
     def remove_unfinished_file(self, new_release_path):
         changed = False
@@ -329,13 +344,23 @@ class DeployHelper(object):
         changes = 0
 
         for release in os.listdir(releases_path):
-            if (os.path.isfile(os.path.join(releases_path, release, self.unfinished_filename))):
+            if os.path.isfile(os.path.join(releases_path, release, self.unfinished_filename)):
                 if self.module.check_mode:
                     changes += 1
                 else:
                     changes += self.delete_path(os.path.join(releases_path, release))
 
         return changes
+
+    def remove_unfinished_link(self, path):
+        changed = False
+
+        tmp_link_name = os.path.join(path, self.release + '.' + self.unfinished_filename)
+        if not self.module.check_mode and os.path.exists(tmp_link_name):
+            changed = True
+            os.remove(tmp_link_name)
+
+        return changed
 
     def cleanup(self, releases_path):
         changes = 0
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @RobberPhex. @ramondelafuente please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

web_infrastructure/deploy_helper.py
```diff
@@ -337,6 +342,16 @@ def remove_unfinished_builds(self, releases_path):
 
         return changes
 
+    def remove_unfinished_link(self, path):
```
ramondelafuente, line 355: Why have a procedure to remove the temporary link? Won't the temp link be removed during rename?

If you have a good reason for it to exist, I do think this could also support the _module.check_mode_
(i.e. not change anything, but know _if_ something will change)

ramondelafuente:
Besides the comment in the source - has this really been causing problems in production? Back-to-back delete/create statements? Just curious :-)

robberphex:
@ramondelafuente in origin code, when current link need update(such as fire a new version), we need unlink it and re-link it.
After unlink, and before re-link, the site is unavailable(because current link doesn't exist).

In my new solution, I will create temp link for new version, and rename temp link to current link. This can avoid the unavailable time.

btw, I met this situation when I fire a new version(written by php), so I create this PR.

As for check_mode, sorry for forget it, I will commit it soon.

robberphex:
@ramondelafuente I fixed check mode in 8361e2a7db67db503acc3854256afb29d8e26dec

---

But, I noticed that when need create_link, it will mark changed to True, even if current link is new version. so I add commit to fix it(a34312e3f3efc566556d7e038e88001f5ed468ae):
When current link point to new release, it will mark changed to False.

Should I split it to new PR?

ramondelafuente:
I was thinking about this a little bit - but having that tmp symlink floating around seems to bring a bit more complexity than needed. How about we do a forced symlink (`ln -s -f`)?

My python-FU is not strong but if os.symlink does not support this we could just run a system command, right?

ramondelafuente:
As for the changes regarding check_mode: nice job. That looks good to me now.

robberphex:
_I am afraid ln -f is not a good idea_ @ramondelafuente 

Let's read the code of `ln`(http://git.savannah.gnu.org/cgit/coreutils.git/tree/src/ln.c#n329). ln also unlink and re-link, it also cause downtime.

So, I look for rename, as [POSIX](http://pubs.opengroup.org/onlinepubs/009695399/functions/rename.html) described, rename will finish link update in one function.
Next, I find the `os.rename`'s implement([2.6](https://github.com/python/cpython/blob/2.6/Modules/posixmodule.c#L2626), [2.7](https://github.com/python/cpython/blob/2.7/Modules/posixmodule.c#L2808), [3.4](https://github.com/python/cpython/blob/1fe0fd9feb6a4472a9a1b186502eb9c0b2366326/Modules/posixmodule.c#L4146)):
- in 2.x, os.rename will call glibc's rename.
- in 3.x, os.rename will call renameat or rename.

`renameat` is a [syscall](http://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/syscalls.list;hb=78bd7499af46d739ce94410eaeea006e874ca9e5#l92), there is no downtime.
in some implement, `rename` also call [`renameat`](http://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/generic/rename.c;h=ef0e1683921bd8bc9cd5e4556055d4a65f9fe03b;hb=78bd7499af46d739ce94410eaeea006e874ca9e5#l26), there is no downtime.
But some implement also use [_unlink and re-link_](http://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/posix/rename.c;h=0f13abbf5d249eabc4c527bbd5ef94185007255e;hb=78bd7499af46d739ce94410eaeea006e874ca9e5#l24) to rename, in low level, I think this is minimal downtime.

robberphex:
Just re-commit for correct my author info(avoid using company's email)

ramondelafuente:
Seems like you did a lot more homework on this than I did... and it seems like os.rename works better indeed.

Final thought: in order to keep everything contained in the build directory, how about we create the tmplink there, and rename it to the current-path? This way you would also not care if a tmplink remained. (although I'm not sure how unhappy some tools would be with a symlink recursively pointing to the directory it's in).

ramondelafuente:
And when it comes to this micro-second downtime during un-link and link, at that level I think moving away from symlinks altogether might be a good idea. Like writing the new path to the webserver config and reloading...

robberphex:
I think recursively link is very unfriendly to some tools, and maybe cause additional complexity in other ansible module. So I think crate temp link in project path is better. :)

ramondelafuente:
In that case @gregdek shipit!  :tada:

gregdek:
Thanks again to @RobberPhex for this PR, and thanks @ramondelafuente for reviewing. Marking for inclusion.

bcoca:
@RobberPhex could you flatten the commits?

robberphex:
I rebase it just now.

@bcoca my three commit in one line( no merge), what's means "flatten"? Did you mean squash three commit to one?

bcoca:
merge the commits down to a single one, to make cherry picking easy

robberphex:
@bcoca done.

PR #1333, opened 2015-12-05
Enable to resize a volume with cs_volume

atsaki:
The patch allow users to change the size and the offering of the volume by changing cs_volume parameters.

This is from resmo/ansible-cloudstack#33.


resmo:
integration tests extended, passed.  Thank you @atsaki 
shipit

PR #1350, opened 2015-12-08
Correct problem with changed:true

daveres:
I have just changed a small piece of this module to avoid to return always that the task is changed

### DIFF
--- a/windows/win_chocolatey.ps1
+++ b/windows/win_chocolatey.ps1
@@ -86,7 +86,7 @@ Function Choco-IsInstalled
         Throw "Error checking installation status for $package" 
     } 
     
-    If ("$results" -match " $package .* (\d+) packages installed.")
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @daveres. @smadam813 please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

daveres:
Hi,

It's the first time I correct some code in Github. I will follow the guide you comment in the future.

Regards,

smadam813:
Thanks @daveres for the contribution.

shipit

nitzmahone:
What's weird is that I can't make the original one break (which is disconcerting, since at first glance the expression should not match), but the fix is fine- thanks!

nitzmahone:
cherry-picked to stable-2.0 as well

PR #1356, opened 2015-12-10
Createparent fix

wishdev:
When using the createparent option when creating a zfs volume there is a bug when you re-run the playbook and the module tries to update the "createparent" option which does not exist.

This also adds the docs for createparent at the same time.

### DIFF
--- a/system/zfs.py
+++ b/system/zfs.py
@@ -76,6 +76,11 @@ options:
       - The copies property.
     required: False
     choices: [1,2,3]
+  createparent:
+    description:
+      - Creates parents, if needed, during creation
+    required: False
+    choices: ['on','off']
   dedup:
     description:
       - The dedup property.
--- a/system/zfs.py
+++ b/system/zfs.py
@@ -250,6 +250,7 @@ class Zfs(object):
         self.changed = False
 
         self.immutable_properties = [ 'casesensitivity', 'normalization', 'utf8only' ]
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @wishdev. @johanwiren please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

johanwiren:
This issue should already be resolved by this commit:
https://github.com/ansible/ansible-modules-extras/commit/5160c4567c6c7cb521cc998c62e18b62ce5ef524

@wishdev Can you verify that you problem still exists when running from the devel branch?

wishdev:
This is a different issue. That commit solved the issue on create that the createparent option wasn't actually being used. The "update" logic (around line 424) doesn't get to the create function so that line doesn't come into play.

johanwiren:
Ah, I see.

I am reworking how this module should deal with zfs properties since they behave differently on different OSes and zpool versions. Could you please try out this branch: https://github.com/johanwiren/ansible-modules-extras/tree/fix_properties_handling and see if that solves your problems (and does not create any other issues)?

wishdev:
This appears to solve my issue and not cause any further issues. I will use that file until it gets merged.

Thanks!

johanwiren:
Great! Thanks for testing.

PR #1360, opened 2015-12-10
Change how to handle zfs properties

johanwiren:
Removes all "options" for zfs properties and uses the check_invalid_args=False to skip verification of "bad" options.

All extra options are passed on to the zfs command that sets properties just as before.

The reason for this big change is that the zfs properties and their possible values are difficult to maintain since they differ on OS'es and zpool versions. If we wanted the options correct we would have to make the module much more complex to try to figure out valid options for different scenarios.

In essence, not much changed in expected behaviour, except fixing some bugs.

### DIFF
--- a/system/zfs.py
+++ b/system/zfs.py
@@ -24,7 +24,7 @@ DOCUMENTATION = '''
 module: zfs
 short_description: Manage zfs
 description:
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @johanwiren. Since you are a maintainer of this module, we are marking this PR for inclusion.

johanwiren:
Ah, did not know that the process was like that.

Due to the nature of the change I really want a few more eyes on it before merging. I will close this one and open a new one when I feel ready.

PR #1366, opened 2015-12-13
Fix up vsphere_copy after open_url change

matburt:
- Remove leading module parameter on open_url call as it's no longer used
  by module_utils.urls.open_url
- Force basic auth otherwise vsphere will just return a 401

---

I think the exception handler below open_url is probably not valid anymore... on bad auth I get a `urlopen error [Errno 32] Broken pipe`  but it seems like it could also raise an HTTPError of some sort under certain circumstances.   At the very least this PR allows vsphere_copy to work where it didn't after the move to open_url.

### DIFF
--- a/cloud/vmware/vsphere_copy.py
+++ b/cloud/vmware/vsphere_copy.py
@@ -137,8 +137,9 @@ def main():
     }
 
     try:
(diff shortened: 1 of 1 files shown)


bcoca:
cp'd into stable-2.0 also

PR #1373, opened 2015-12-15
fix error with misArg not being declared

bcoca:
also fixed test to work on empty string or not for error reporting


gregdek:
Thanks @bcoca. @TimothyVandenbrande please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

t-v:
I'm sorry for the late reply, but I wished to get a full test of the new change.
These came out positive, so you can shipit

PR #1374, opened 2015-12-15
New module to enable the installation of certificates on Windows

nlandais:


gregdek:
Thanks @nlandais for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

jhawkesworth:
Could do with a version bump in the documentation but otherwise this looks like a very clean module.
Not in a position to test this myself at the moment but hope the reporter of https://github.com/ansible/ansible/issues/14387 can test it.

windows/win_certificate.ps1
```diff
@@ -0,0 +1,67 @@
...
+$params = Parse-Args $args;
+
+$result = New-Object PSObject;
+Set-Attr $result "changed" $false;
+
+#pfx_file parameter
+$pfx_file = Get-AnsibleParam -obj $params -name "pfx_file" -failifempty $true -emptyattributefailmessage "Please provide the path to the pfx file as found on the target server";
+
+#password parameter
+$password = Get-AnsibleParam -obj $params -name "password" -failifempty $true -emptyattributefailmessage "Please provide the certifacte password";
```
kwerey, line 31: A minor bit of pedantic copy-editing: this wants to be "the certifacte password", not certifacte.

windows/win_certificate.ps1
```diff
@@ -0,0 +1,67 @@
...
+
+#location parameter
+$location = Get-AnsibleParam -obj $params -name "location" -default "CurrentUser" -ValidateSet "LocalMachine","CurrentUser";
+
+#store_name parameter
+$store_name = Get-AnsibleParam -obj $params -name "certififcate_store_name" -default "MY";
```
kwerey, line 37: A major bit of pedantic copyediting:  you want '$params -name "certificate_store_name"' here, no?

kwerey:
Not the reporter of that bug above, but I'd put together a similar, slightly scruffier module, so here's some thoughts.
- Typo in picking up the 'certificate_store_name' parameter means this can only import into 'My' store currently, that needs fixing
- It's possible, tho perhaps not advisible, to have PFX certs without a password. So perhaps making the password parameter optional, and defaulting to pass an empty string? 

Otherwise, this looks lovely and clean to me. It doesn't fit major main use case for a certificates role, which is to import the internal certificate authority for my organisation onto Windows VMs, but it's a clean way of talking to the store and it seems like it'd be fairly easy to extend by adding something like a type parameter that takes either 'pfx' or 'cert', or something along those lines.

Overall - looks like a good addition to the Windows roleset to me.

windows/win_certificate.ps1
```diff
@@ -0,0 +1,67 @@
...
+	#Get list of certificates in the specified store
+	$store_path=Join-Path -Path cert: -ChildPath $location | Join-Path -Childpath $store_name
+	$thumbprints = Get-ChildItem -Path $store_path | Foreach-Object {$_.Thumbprint}
+	$pfx = new-object System.Security.Cryptography.X509Certificates.X509Certificate2	
+	$pfx.import($pfx_file,$password,"Exportable,PersistKeySet")	 
+	$store = new-object System.Security.Cryptography.X509Certificates.X509Store($store_name,$location)	 
+	$store.open("MaxAllowed")
+	[bool]$certificateisinstalled = $thumbprints -contains $pfx.Thumbprint
+	if (($state -eq "present") -and !($certificateisinstalled)) {
+		$store.add($pfx)
```
ahnooie, line 53: It may just be my environment or a permission issue but I couldn't get IIS to use the cert on Win 2012 R2.  The cert imported properly into the cert store but I couldn't bind it using IIS: "A specified logon session does not exist. It may already have been terminated. (Exception from HRESULT: 0x80070520)" and in the System Event log: "A fatal error occurred when attempting to access the SSL server credential private key. The error code returned from the cryptographic module is 0x8009030D. The internal error state is 10001."

I ended up having to import it this way: `Import-PfxCertificate -FilePath $pfx_file -CertStoreLocation $pfx_cert_location -Exportable -Password $pfx_password`

windows/win_certificate.ps1
twistyWater: A little typo : certificate instead of certifacte

windows/win_certificate.ps1
```diff
@@ -0,0 +1,67 @@
...
+#state parameter
+$state = Get-AnsibleParam -obj $params -name "State" -default "Present" -ValidateSet "present","absent";
+
+try 
+{
+	#Get list of certificates in the specified store
+	$store_path=Join-Path -Path cert: -ChildPath $location | Join-Path -Childpath $store_name
+	$thumbprints = Get-ChildItem -Path $store_path | Foreach-Object {$_.Thumbprint}
+	$pfx = new-object System.Security.Cryptography.X509Certificates.X509Certificate2	
+	$pfx.import($pfx_file,$password,"Exportable,PersistKeySet")
```
twistyWater, line 48: It would be nice to return the thumbprint, as I would use it to assign the certificate hash with win_iis_webbinding, if you want I can do the PR

nlandais: Go ahead and do the PR. Having switched focus away from Windows and webapps, I no longer have the cycles available to maintain this extra module.

PR #1378, opened 2015-12-15
Fix zfs properties handling

johanwiren:
Removes all "options" for zfs properties and uses the check_invalid_args=False to skip verification of "bad" options.

All extra options are passed on to the zfs command that sets properties just as before.

The reason for this big change is that the zfs properties and their possible values are difficult to maintain since they differ on OS'es and zpool versions. If we wanted the options correct we would have to make the module much more complex to try to figure out valid options for different scenarios.

In essence, not much changed in expected behaviour, except fixing some bugs.

### DIFF
--- a/system/zfs.py
+++ b/system/zfs.py
@@ -24,7 +24,7 @@ DOCUMENTATION = '''
 module: zfs
 short_description: Manage zfs
 description:
-  - Manages ZFS file systems on Solaris and FreeBSD. Can manage file systems, volumes and snapshots. See zfs(1M) for more information about the properties.
+  - Manages ZFS file systems, volumes, clones and snapshots.
 version_added: "1.1"
 options:
   name:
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @johanwiren. Since you are a maintainer of this module, we are marking this PR for inclusion.

bcoca:
Knowing the mess that cross ZFS property handling is I understand the change, but this does seem to break backwards compatibility.  Can we keep the old options and just add a 'extra_options' that handles as this new code does?

johanwiren:
In what way does it break backwards compatibility?

bcoca:
you are removing the existing options?

johanwiren:
They are just removed from the argument specification and documentation.

Any option passed to the module that is not part of the argument specification is assumed to be a zfs property and handled accordingly. 

This behaviour is backwards compatible and no playbooks should break because of this.

bcoca:
ah, we just loose the verification and documentation .... or 'defer' the verification to execution

gregdek:
Thanks @johanwiren. Since you are a maintainer of this module, we are marking this PR for inclusion.

abadger:
Merged to devel.  Thanks!

PR #1390, opened 2015-12-17
Add an __init__.py file in openstack/

gpocentek:
The os_project module doesn't get installed if the **init** file doesn't
exist in the repository.


bcoca:
also applied to stable-2.0

PR #1396, opened 2015-12-17
mistake due to GitHub's default PR fork base... please ignore

trinitronx:
Fixed a bunch of bugs in redshift & firehose modules so they work in the playbook


trinitronx:
silly GitHub default PR targets... please ignore...

PR #1398, opened 2015-12-18
Add no_log=True to consul modules' API tokens

joelthompson:
API tokens should be considered sensitive and not logged.


ryansydnor:
Looks good! shipit

copumpkin:
:shipit:

joelthompson:
Thanks for the quick turnaround time, @bcoca :)

PR #1399, opened 2015-12-18
Added update_catalog and some error handling

dermute:
Hello,
pkgutil doesn't update it's catalog from the repository by default. Adding the option 'update_catalog' allows to force an update before installing oder upgrading an package. But it is not enabled by default, to keep the default behaviour of pkgutil.
Moreover I added some errorhandling.

### DIFF
--- a/packaging/os/pkgutil.py
+++ b/packaging/os/pkgutil.py
@@ -155,7 +155,11 @@ def main():
             if len(out) > 75:
                 out = out[:75] + '...'
             if rc != 0:
-                module.fail_json(msg=err if err else out)
+                if err:
+                    msg = err
+                else:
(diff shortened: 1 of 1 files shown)


dermute:
Any hints why the continuous-integration is failing? It works in my environment.

bcoca:
we test in multiple environments and this fails in one of them, just remove the inline if and do something like this:

```
if err:
  msg = err
else:
   msg = out
module.fail_json(msg=msg)
```

robynbergeron:
Thanks @dermute. Since you are a maintainer of this module, we are marking this PR for inclusion.

dermute:
I will fix the bug next week.

PR #1403, opened 2015-12-18
iptables: cleanup

LinusU:
Some of the new code wasn't idiomatic and/or followed the pep8 style guide.

### DIFF
--- a/system/iptables.py
+++ b/system/iptables.py
@@ -54,13 +54,13 @@ options:
     default: filter
     choices: [ "filter", "nat", "mangle", "raw", "security" ]
   state:
-    description: 
+    description:
       - Whether the rule should be absent or present.
     required: false
     default: present
     choices: [ "present", "absent" ]
   ip_version:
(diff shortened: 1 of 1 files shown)


robynbergeron:
Thanks @LinusU. Since you are a maintainer of this module, we are marking this PR for inclusion.

LinusU:
Thank you :)

bcoca:
we don't normally accept cosmetic changes, which is what the comment suggests, but looking at the changes they seem to go farther than that and change functionality, can you explain?

LinusU:
The functionality stays exactly the same. I'm just chancing the code to be more idiomatic.

Surely you must accept refactoring to the code to make it easier to understand?

There was three functions that where very similar, instead of duplicating the code three times, the only difference is now passed as a parameter.

dvigueras:
I agree with @LinusU, functionality is kept, It makes the code cleaner.

PR #1412, opened 2015-12-21
New AWS Managed NAT Gateway module - will resolve #1401

### LINKED ISSUE #1401: Add support for AWS Managed NAT Gateways
##### ISSUE TYPE

Feature Idea
##### COMPONENT NAME

new module
##### ANSIBLE VERSION

N/A
##### SUMMARY

AWS has released support for [Managed NAT Gateways](https://aws.amazon.com/blogs/aws/new-managed-nat-network-address-translation-gateway-for-aws/). It would be nice if one could set them up with Ansible.

Etherdaemon:
New module for creating, deleting AWS Managed NAT gateways as discussed in #1401

### DIFF
--- a/cloud/amazon/nat_gateway.py
+++ b/cloud/amazon/nat_gateway.py
@@ -1,0 +1,401 @@
+#!/usr/bin/python
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+module: nat_gateway
+short_description: Create, delete and describe AWS Managed NAT Gateways. Requires Boto3.
+description:
+  - Creates AWS Managed NAT Gateways with option to provide EIP or
+    allocate new address.
+  - Deletes AWS Managed NAT Gateways with option to release attached EIP.
+  - Describe AWS Managed NAT Gateways, either all or with Filters
+  - This module does not support updates or tagging as it is a current
+    restriction with AWS.
+  - This module support check mode.
+version_added: "2.1"
+options:
+  describe_action:
+    description:
+      - specifies the action to be a get action, to describe existing
+        nat gateways
+    required: false
+  describe_filter:
+    description:
+      - Use in conjunction with describe_action to filter results
+      - See U(http://boto3.readthedocs.org/en/latest/reference/services/ec2.html#EC2.Client.describe_nat_gateways)
+        for possible filters. Only the Filter section is allowed in this module.
+    required: false
+  subnet_id:
+    description:
+      - Required when creating a NAT gateway.
+    required: false
+  eip_address:
+    description:
+      - An elastic IP already allocated in the AWS account to be used
+        by the NAT gateway. This address cannot already by attached to
+        another resource. If this option is not specified, an elastic IP
+        will be allocated and attached for you.
+    required: false
+  token:
+    description:
+      - A unique ASCII string up to 64 characters to identify the request.
+      - Required only for create actions.
+    required: false
+  state:
+    description:
+        - present to ensure resource is created.
+        - absent to remove resource
+    required: false
+    default: present
+    choices: [ "present", "absent"]
+  wait:
+    description:
+      - When specified, will wait for either available status for state present
+        or deleted for state absent. If release_eip is specified during state absent,
+        this will automatically be utilised.
+    required: false
+    default: no
+    choices: ["yes", "no"]
+  wait_timeout:
+    description:
+      - Used in conjunction with wait. Number of seconds to wait for status.
+        Recommended value of over 2 minutes.
+    required: false
+    default: 320
+  release_eip:
+    description:
+      - AWS Managed NAT Gateways do not release the elastic IP by default. Specify
+        this option if releasing eip is required.
+    required: false
+  nat_gateway_id:
+    description:
+      - The ID of the NAT gateway to be removed - used only for state absent
+    required: false
+author: Karen Cheng(@Etherdaemon)
+extends_documentation_fragment: aws
+'''
+
+EXAMPLES = '''
+- name: Get all existing nat gateways
+  nat_gateway:
+    describe_action: yes
+    region: ap-southeast-2
+  register: existing_nat_gateways
+
+
+- name: Get nat gateways with specific filter
+  nat_gateway:
+    describe_action: yes
+    region: ap-southeast-2
+    describe_filter:
+      subnet-id: subnet-12345678
+      state: ['available']
+  register: existing_nat_gateways
+
+
+- name: Create new nat gateway with when condition
+  nat_gateway:
+    state: present
+    subnet_id: subnet-12345678
+    eip_address: 52.1.1.1
+    token: token-12345678
+    region: ap-southeast-2
+  register: new_nat_gateway
+  when: existing_nat_gateways.result == []
+
+
+- name: Create new nat gateway and wait for available status
+  nat_gateway:
+    state: present
+    subnet_id: subnet-12345678
+    eip_address: 52.1.1.1
+    token: uniquetokenstring
+    wait: yes
+    region: ap-southeast-2
+  register: new_nat_gateway
+
+
+- name: Create new nat gateway and allocate new eip
+  nat_gateway:
+    state: present
+    subnet_id: subnet-12345678
+    token: uniquetokenstringyyy
+    wait: yes
+    region: ap-southeast-2
+  register: new_nat_gateway
+
+- name: Describe nat gateways to be removed
+  nat_gateway:
+    describe_action: yes
+    region: ap-southeast-2
+    describe_filter:
+      subnet-id: subnet-12345678
+      state: ['available']
+  register: gateways_to_remove
+
+- name: Delete nat gateway using discovered nat gateways
+  nat_gateway:
+    state: absent
+    region: ap-southeast-2
+    wait: yes
+    nat_gateway_id: "{{ item.NatGatewayId }}"
+    release_eip: yes
+  register: delete_nat_gateway_result
+  with_items: "{{ gateways_to_remove.result }}"
+
+- name: Delete nat gateway and wait for deleted status
+  nat_gateway:
+    state: absent
+    nat_gateway_id: nat-12345678
+    wait: yes
+    wait_timeout: 500
+    region: ap-southeast-2
+
+
+- name: Delete nat gateway and release EIP
+  nat_gateway:
+    state: absent
+    nat_gateway_id: nat-12345678
+    release_eip: yes
+    region: ap-southeast-2
+'''
+
+RETURN = '''
+result:
+  description: The result of the create, delete or describe action.
+  returned: success
+  type: dictionary or a list of dictionaries
+'''
+
+try:
+    import json
+    import botocore
+    import boto3
+    HAS_BOTO3 = True
+except ImportError:
+    HAS_BOTO3 = False
+
+import time
+
+
+def date_handler(obj):
+    return obj.isoformat() if hasattr(obj, 'isoformat') else obj
+
+
+def wait_for_status(client, module, nat_gateway_id, status):
+    polling_increment_secs = 15
+    max_retries = (module.params.get('wait_timeout') / polling_increment_secs)
+    status_achieved = False
+
+    for x in range(0, max_retries):
+        try:
+            nat_gateway = get_nat_gateways(client, module, nat_gateway_id)[0]
+            if nat_gateway['State'] == status:
+                status_achieved = True
+                break
+            else:
+                time.sleep(polling_increment_secs)
+        except botocore.exceptions.ClientError as e:
+            module.fail_json(msg=str(e))
+
+    return status_achieved, nat_gateway
+
+
+def get_nat_gateways(client, module, nat_gateway_id=None):
+    params = dict()
+
+    if module.params.get('describe_filter'):
+        params['Filter'] = []
+        for key, value in module.params.get('describe_filter').iteritems():
+            if isinstance(value, basestring):
+                value = [value]
+            params['Filter'].append({'Name': key, 'Values': value })
+    elif nat_gateway_id:
+        params['NatGatewayIds'] = [nat_gateway_id]
+
+    existing_gateways = json.loads(json.dumps(client.describe_nat_gateways(**params), default=date_handler))
+
+    return existing_gateways['NatGateways']
+
+
+def create_nat_gateway(client, module, allocation_id):
+    params = dict()
+    changed = False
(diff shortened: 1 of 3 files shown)


cloud/amazon/nat_gateway.py
```diff
@@ -0,0 +1,393 @@
...
+      - The ID of the NAT gateway to be removed - used only for state absent
+    required: false
+author: Karen Cheng(@Etherdaemon)
+extends_documentation_fragment: aws
+'''
+
+EXAMPLES = '''
+- name: Get all existing nat gateways
+  nat_gateway:
+    describe_action: yes
```
willthames: I think this should be in nat_gateway_facts

cloud/amazon/nat_gateway.py
```diff
@@ -0,0 +1,393 @@
...
+    existing_gateways = get_nat_gateways(client, module)
+
+    gateway_found = False
+    for gateway in existing_gateways:
+        if gateway['NatGatewayAddresses'][0]['AllocationId'] == allocation_id:
+            gateway_found = True
+            if gateway['SubnetId'] == module.params.get('subnet_id'):
+                result = gateway
+            else:
+                module.fail_json(msg='Update of nat_gateway is not allowed.')
```
willthames: Should we just return changed=False at this point rather than failing?

willthames:
Recommend splitting this into two modules.

Etherdaemon:
Hi @willthames.

Thanks for the review.

I'm not sure I entirely agree with having 2 seperate modules for this particular module. The reason being as the functions to describe are required already in order to provide the wait/wait_timeout options. Having a seperate module just to do a describe would duplicate code unnecessarily.

Due to how AWS have implemented the NAT gateway API as well, it's almost ridiculous to not perform a describe before doing a create as their no real way of ensuring duplication as the client token isn't describable and as ec2_eip is written in boto rather than boto3, it doesnt return an allocation_id and the fact that you can't tag the gateway which rules out any other method of uniquely identifying the nat_gateway and therefore if you were to run it without the describe as a check, everytime you run a play you would get either a token clash or a new gateway. 

There also appears to be other existing Ansible modules that have the "facts" and "create/delete" functions combined now.

I'm happy to stand corrected on this matter, but it would mean unnecessarily duplicating the code across 2 modules.

Regards

Karen

cloud/amazon/nat_gateway.py
bcoca: instead of making this an option, always return the data, as it seems you need it anyways.

cloud/amazon/nat_gateway.py
```diff
@@ -0,0 +1,393 @@
...
+
+
+- name: Delete nat gateway and release EIP
+  nat_gateway:
+    state: absent
+    nat_gateway_id: nat-12345678
+    release_eip: yes
+    region: ap-southeast-2
+'''
+
```
bcoca: missing return docs

cloud/amazon/nat_gateway.py
```diff
@@ -0,0 +1,393 @@
...
+short_description: Create, delete and describe AWS Managed NAT Gateways.
+Requires Boto3.
+description:
+  - Creates AWS Managed NAT Gateways with option to provide EIP or
+    allocate new address.
+  - Deletes AWS Managed NAT Gateways with option to release attached EIP.
+  - Describe AWS Managed NAT Gateways, either all or with Filters
+  - This module does not support updates or tagging as it is a current
+    restriction with AWS.
+  - This module support check mode.
```
bcoca: 2.1 at this point

cloud/amazon/nat_gateway.py
Etherdaemon: Hi @bcoca, it's an option as it allows the user to perform a describe first to then utilise the output in order to make a decision on whether to perform a create action or a delete action.

The particular function is also utilised to allow for the wait option, and also to perform checks before proceeding with the create.

cloud/amazon/nat_gateway.py
Etherdaemon: return added

cloud/amazon/nat_gateway.py
Etherdaemon: version updated

Etherdaemon:
Also this module requires #13166 to update module_utils/ec2.py to be merged

robynbergeron:
Thanks @Etherdaemon for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

sivel:
Here is some feedback based on automated scanning of this new module:

```
  File "ansible_testing/modules.py", line 383, in validate
    doc, examples, ret = get_docstring(self.path, verbose=True)
  File "/Users/matt/python_venvs/ansibledev/ansible/lib/ansible/utils/module_docs.py", line 74, in get_docstring
    doc = yaml.safe_load(child.value.s)
  File "/Users/matt/python_venvs/ansibledev/lib/python2.7/site-packages/yaml/__init__.py", line 93, in safe_load
    return load(stream, SafeLoader)
  File "/Users/matt/python_venvs/ansibledev/lib/python2.7/site-packages/yaml/__init__.py", line 71, in load
    return loader.get_single_data()
  File "/Users/matt/python_venvs/ansibledev/lib/python2.7/site-packages/yaml/constructor.py", line 37, in get_single_data
    node = self.get_single_node()
  File "/Users/matt/python_venvs/ansibledev/lib/python2.7/site-packages/yaml/composer.py", line 36, in get_single_node
    document = self.compose_document()
  File "/Users/matt/python_venvs/ansibledev/lib/python2.7/site-packages/yaml/composer.py", line 55, in compose_document
    node = self.compose_node(None, None)
  File "/Users/matt/python_venvs/ansibledev/lib/python2.7/site-packages/yaml/composer.py", line 84, in compose_node
    node = self.compose_mapping_node(anchor)
  File "/Users/matt/python_venvs/ansibledev/lib/python2.7/site-packages/yaml/composer.py", line 127, in compose_mapping_node
    while not self.check_event(MappingEndEvent):
  File "/Users/matt/python_venvs/ansibledev/lib/python2.7/site-packages/yaml/parser.py", line 98, in check_event
    self.current_event = self.state()
  File "/Users/matt/python_venvs/ansibledev/lib/python2.7/site-packages/yaml/parser.py", line 428, in parse_block_mapping_key
    if self.check_token(KeyToken):
  File "/Users/matt/python_venvs/ansibledev/lib/python2.7/site-packages/yaml/scanner.py", line 115, in check_token
    while self.need_more_tokens():
  File "/Users/matt/python_venvs/ansibledev/lib/python2.7/site-packages/yaml/scanner.py", line 149, in need_more_tokens
    self.stale_possible_simple_keys()
  File "/Users/matt/python_venvs/ansibledev/lib/python2.7/site-packages/yaml/scanner.py", line 289, in stale_possible_simple_keys
    "could not found expected ':'", self.get_mark())
ScannerError: while scanning a simple key
  in "<string>", line 4, column 1:
    Requires Boto3.
    ^
could not found expected ':'
  in "<string>", line 5, column 1:
    description:
    ^

ERROR: Invalid or no DOCUMENTATION provided
```

Etherdaemon:
Thanks @sivel. 

Weirdly I didn't get those errors above when I ran 

```
~/ansible/hacking/test-module -m ./nat_gateway.py -a "profile=personal region=ap-southeast-2 subnet_id=subnet-12345678 token=token-12345678"
```

But I've updated that line in your error above to be on the same first line. Also saw I was letting some exceptions fall through on line 254 so updated that as well.

If using the above test then I got the below output which is correct as I didnt provide it a valid subnet ID:

```
* including generated source, if any, saving to: /Users/Etherdaemon/.ansible_module_generated
* this may offset any line numbers in tracebacks/debuggers!
***********************************
RAW OUTPUT
{"msg": "An error occurred (InvalidSubnet) when calling the CreateNatGateway operation: Subnet subnet-12345678 was not found", "failed": true}


***********************************
PARSED OUTPUT
{
    "failed": true,
    "msg": "An error occurred (InvalidSubnet) when calling the CreateNatGateway operation: Subnet subnet-12345678 was not found"
}
```

Cheers

sivel:
I'm also concerned about this just being called `nat_gateway`.  That is a very generic sounding name, and would probably be better prefixed with something indicated that it is for AWS, such as `aws_nat_gateway.py`.  Even though it is nested in the `amazon` directory, that doesn't prevent things from having name clashes.

Etherdaemon:
@sivel Good point. I've renamed it to ec2_vpc_managed_ngw to be consistent with the existing ec2_vpc_igw (Internet Gateway). I have popped managed in it though as its not the traditional NAT servers people have been using in AWS.

jonhadfield:
I've added some feedback on ideas and my alternative implementation on https://github.com/ansible/ansible-modules-extras/issues/1401.

jonhadfield:
@Etherdaemon I've run some tests against your latest code and found the following issues:
- You are requiring a region parameter is set, but boto will detect from multiple sources if not specified. (http://boto3.readthedocs.org/en/latest/guide/configuration.html)
- You require a token id but one isn't necessary for creating, only to ensure idempotence. I've decided to leave that out of mine as my understanding is that I'd need to keep a copy of the token and later refer to it. I'm not aware of Ansible having a method to keep this between invocations and even if it did, it would need to be available to every machine using this module.
- If the EIP specified is not one in a VPC (standard scope) it throws a stacktrace:
  
  `File "/Users/hadfielj/.ansible/tmp/ansible-tmp-1451653446.48-163231699708505/ec2_vpc_managed_ngw", line 295, in get_eip_address
      allocation_id = client.describe_addresses(**params)['Addresses'][0]['AllocationId']
  KeyError: 'AllocationId'`

This is because a non-VPC EIP doesn't have an allocation id.

What I am going to take from yours is:
- having a region parameter available so it can be overridden (but make it optional so it has a chance to detect)
- add an eip_address parameter as I've found registering the ec2_eip module return (on create) only provides the EIP and not the allocation id it seems. I'm going to support both eip_address and allocation_id but probably make them mutually exclusive (or at least check they refer to the same entity).

We've both obviously spent a lot of time on developing this and can benefit from each other's contribution, so happy to share attribution, i.e. if mine is preferred and you want to test and contribute to it, then we can be co-authors? Maybe the opposite would work also?

Etherdaemon:
Hi @jonhadfield 

Thank you for testing and taking a look. I appreciate we both have spent quite a deal of  time on this.

I'm happy to do the co-author scenario on either module as that seems like the best way forward given the similarities between the 2. I'll update mine first with the below and happy to put you down as a coauthor now for it. I'll see if I can get some time to test yours out as well afterwards.

I'll update my PR with some changes to reflect your feedback and I might actually split out the describe into a separate facts module now as well.

My notes are:

In terms of the eip_address I had found that as well - the existing eip_module didn't return the allocation_id as its not an option in boto. I personally think that the user can provide the eip_address and the module can perform the lookup. I'll see if I can check the type and perform proper handling of it instead. I've also allowed the user to allow the module to auto allocate an IP for them and as such is why I also have the release-eip as it makes the module more user friendly given the fact that it can get expensive if the eip is disassociated but not released.

I agree with the token comment - I found it to be a pain as AWS does not return this value on any describe actions and there is no way to look up the token after the initial creation - so I'll remove that as well - mustn't have read the docs properly and thought it was a required field. I was approaching the idempotency issue by providing the option to describe existing nat gateways so the user could make an informed due whether to perform a create action or not. The reason I hadn't restricted it to a single allocation-id in a given subnet in a VPC is that you can actually create multiple NAT gateways in a single subnet inside a VPC. I know that there wouldn't be many situations where users would go for this option but it seemed a big assumption that this wouldn't be required. As such I went with the facts option instead giving the user the choice of whether they only wanted 1 NAT gateway in that 1 subnet or not using a when statement.

I had the region set as required as I'm trying to remove any dependence on boto itself and unfortunately I am still waiting for https://github.com/ansible/ansible/pull/13166 to be merged. I might have to find another alternative if this region=True is causing grief.

I had an initial glance at the route_table module and I think it'll require a PR as well as I think for initial creation of route tables worked but adding a route for an existing route table pointing to a nat gateway did not.

Etherdaemon:
Hi @joelthompson, if possible could you elaborate on your comment in #1438 regarding feeling more in line with the Ansible general paradigm please? I appreciate any constructive feedback provided but at this stage I'm not 100% sure what you meant and therefore am unable to apply any further updates that could make my modules (existing or future) better.

jimrazmus:
Works for me.  Thanks for writing it!

Running the devel branch as of today on Mac OS X 10.11.2 with boto3.

joelthompson:
Hi @Etherdaemon -- I was mostly referring to the things you cleaned up in your recent commits about the "describe" state option, and the token. I'm also not a huge fan of the name ngw. I understand the comparison with, e.g., an IGW, but AWS rarely refers to an "NGW" whereas they refer to IGWs all the time. For example, if you [search their docs, ](http://docs.aws.amazon.com/search/doc-search.html?searchPath=documentation&searchQuery=ngw&x=0&y=0) there are literally zero results (as of this writing) for ngw, while there are [50 results](http://docs.aws.amazon.com/search/doc-search.html?searchPath=documentation&searchQuery=igw&x=0&y=0) when searching for igw. I think just calling it ec2_vpc_managed_nat would be better.

One other thought -- rather than having an option to release an EIP in this module, what about returning the EIP allocation ID and public IP to the caller so the caller can release it if desired? This would also support the use case of, for example, moving an EIP from one NAT to gateway to another (say, different public subnet in a different AZ, or in a different VPC) by terminating the old gateway and spinning up a new one.

jonhadfield:
FYI - I submitted a PR for ec2_eip to return an allocation_id and the maintainer/author gave it the okay. https://github.com/ansible/ansible-modules-core/pull/2753. I'm currently depending on it for my submission.

I think the cleanest user flow for setup would be:
1) Allocate EIP and Register
2) Create NAT Gateway using the returned allocation_id and subnet_id

For teardown, I'd personally prefer to use a lookup plugin to return the EIP associated with the NAT Gateway id and pass that to ec2_eip and make it absent. Simply because it saves adding code to the module that runs every time regardless and keeps a module doing its job alone. I do expect the majority of users wanting to use the output of the NAT Gateway removal as input for removal of EIP, so will add it to mine also.

Etherdaemon:
That's why I have the facts module.https://github.com/ansible/ansible-modules-extras/pull/1446

jonhadfield:
I've added the private and public ip address output to my PR: https://github.com/jonhadfield/ansible-modules-extras/commit/fe43a2c38503598e087f5d7f6fe8c294176e3348.
@Etherdaemon I see modules more for enforcing state and lookup_plugins for querying state. I do understand that the average user won't be in a position to write their own plugins, so a facts module does make sense for those users.

jonhadfield:
Hi @Etherdaemon, the change to output allocation id for a new EIP in a VPC has been merged: https://github.com/ansible/ansible-modules-core/commit/0320411d036f52e506e009eebb9959fcabec921a. 
Just in case you want to incorporate as well as, or instead of, the eip parameter. I personally prefer it to EIP as it's what the boto3 API expects and therefore future-proofs it somewhat.

Etherdaemon:
Thanks @jonhadfield - I've renamed the module and added allocation_id as a mutually exclusive option.

gregdek:
@Etherdaemon @jonhadfield Help! It looks like this module has the same filename (cloud/amazon/ec2_vpc_nat_gateway.py) as #1438 -- and only one can be merged. Are they doing the same thing? Please advise.

Etherdaemon:
Hey @jonhadfield, how about we try and just get one of these merged? I'll close this pull request and we can get yours merged instead. If both do what needs to be done to get NAT gateways in I see no reason for it to hold up the process.

jonhadfield:
Hi @Etherdaemon, very kind of you. I'm also keen for this functionality to be merged, regardless of original author (we're co-authors anyhow).

PR #1413, opened 2015-12-21
Support the creation of new ACL with supplied ID

bryanwb:
As is, this module quietly does nothing if you supply it is with an ACL ID for an ACL that does not exist

### DIFF
--- a/clustering/consul_acl.py
+++ b/clustering/consul_acl.py
@@ -24,6 +24,9 @@ description:
  - allows the addition, modification and deletion of ACL keys and associated
    rules in a consul cluster via the agent. For more details on using and
    configuring ACLs, see https://www.consul.io/docs/internals/acl.html.
+   In some cases you may want to pre-generate ACL ids. You can
+   use the following shell command to generate one
+   `python -c "import uuid; print(uuid.uuid1())"`
 requirements:
   - "python >= 2.6"
   - python-consul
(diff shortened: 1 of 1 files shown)


bcoca:
I'm not sure if the module should autocreate an id in this manner, it seems it should error out when a non existing ID is used.

cc @sgargan

bryanwb:
@bcoca @sgargan both the python-consul and consul http API explicitly support creation of a new ACL using a pre-generated ACL ID. I think this is a reasonable use of the consul_api. One could argue that a more explicit way to support this behavior would be to add a state "create_if_missing" which i am open to adding.

http://python-consul.readthedocs.org/en/latest/#consul.base.Consul.ACL.create

https://www.consul.io/docs/agent/http/acl.html#acl_create

gregdek:
@sgargan are you available to review?

sgargan:
Being explicit with create_if_missing is probably the way to go. There are a bunch of integration tests around this behaviour so a test for this wouldn't hurt either. The tests are part of the core project though, https://github.com/ansible/ansible/blob/7f8e8ddca9cf103c05bf39d68ebb2e2ded4067f2/test/integration/roles/test_consul_acl/tasks/main.yml

gregdek:
Kicking this old PR.

gregdek:
Thanks @bryanwb. To the current maintainers, @sgargan please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

Issue #1305, opened 2015-11-30
yum_repository module does not support additional settings provided by yum plugins


detiber:
One example off the top of my head is the yum-plugin-priorities module which allows for setting priorities on repos. The yumrepos module should support a way to provide arbitrary additional settings on a repo.

@jtyr I failed to notice your PR before it was merged to bring this up, but on https://github.com/ansible/ansible-modules-extras/pull/411 I proposed the following for supporting this:

> personally, I would add an new param called additional_attributes that takes a dict to allow for setting repo attributes for plugins.
> 
> I'd avoid disabling param checking just because it seems like it could create a confusing user experience.
> 
> An example showing what I'm thinking:
> 
> ```
> yum_repository: 
>   id: example
>   baseurl='http://example.com/repo/'
>   additional_attributes:
>     priority: 1
> ```

robynbergeron:
[module: packaging/os/yum_repository.py]

robynbergeron:
@detiber -- Thanks for submitting this feature idea.

@jtyr -- Pinging you to bring this feature request to your attention. Please note that you are under no particular obligation to add this feature.

We encourage all Ansible community members to submit pull requests to satisfy feature requests. Consult the module guidelines (http://docs.ansible.com/ansible/developing_modules.html) for more information.

jtyr:
@detiber I'm very happy that you brought this feature up as I was thinking about it recently as well. The biggest advantage of this new feature would be that it would allow to override already defined params and also add new params without the need to modify the role where the module is used. For example, some people might need to use a proxy server when fetching packages and if the author of the role did not parametrize it, the user would need to modify the role.

I'm just not sure whether `additional_parameters` is right name for it as it's quite long. What about to name it just `params`?

Here is an example of how people could use it:

```
---

# This is the defaults/main.yaml file of the my_role

# Default set of custom params would be empty hash
my_role_epel_yumrepo_params: {}
```

```
---

# This is the tasks/main.yaml file of the my_role

- name: Add EPEL repo
  yumrepo:
    name: epel
    description: EPEL YUM repo
    baseurl: http://download.fedoraproject.org/pub/epel/$releasever/$basearch/
    # Here we prepare the task to accept the custom params
    params: "{{ my_role_epel_yumrepo_params }}"
```

The user could add the custom params to the module by defining the `my_role_epel_yumrepo_params` variable in the playbook or in an other role:

```
---

# This is the playbook using the my_role

- name: My play
[...]

PR #1415, opened 2015-12-21
Adding params option into the yumrepo module

jtyr:
To address the issue #1305.

### DIFF
--- a/packaging/os/yumrepo.py
+++ b/packaging/os/yumrepo.py
@@ -172,6 +172,12 @@ options:
     required: true
     description:
       - Unique repository ID.
+  params:
+    required: false
+    default: None
+    description:
+      - Option used to allow the user to overwrite any of the other options. To
+        remove an option, set the value of the option to C(null).
   password:
     required: false
     default: None
@@ -284,7 +290,7 @@ options:
       - Username to use for basic authentication to a repo or really any url.
 
 extends_documentation_fragment:
-    - files
+  - files
 
 notes:
   - All comments will be removed if modifying an existing repo file.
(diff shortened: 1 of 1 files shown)


jtyr:
@detiber Please feel free to review it.

robynbergeron:
Hi @jtyr -- thanks for submitting this PR. Since you're the maintainer, we can merge this PR without any other approval -- but if you'd rather wait for @detiber to take a look at it, we can hang on :) I'll check back and see what you'd prefer.

jtyr:
@robynbergeron I would like to wait for @detiber  to test it first.

jtyr:
@detiber Any chance you could test this patch any time soon?

detiber:
@jtyr my apologies for taking so long to reply, I'm not able to give it a run-through right now, but I see no issues with the implementation based on reviewing the diff.

Thanks for adding the feature!

@robynbergeron :+1: from me based on a code review.

jtyr:
@detiber Thanks for the review.

@robynbergeron Please go ahead and merge it. I'm sure @detiber will test it before the Ansible v2.1 is release so we will catch any possible bugs.

jtyr:
@robynbergeron @gregdek @bcoca Please could this be merged?

ke3vin:
I just happened to stumble across this today as it looks like exactly what I need.  I just did a quick test and it looks like it's not quite complete.  I wanted to add the line 'priority=N' to my repo, to use yum_plugin_priorities. The problem is that all of the parameters, including extras defined using this new method, are checked against the 'allowed_params' array, and anything not found there is silently ignored.  So, if your goal is to allow a truly generic list of params, there will need to be additional work done to augment/circumvent the allowed_params check.  If you're merely trying to allow folks to specify only params that are already in the allowed_params array, then your changes are fine as is.
If I come up with something, I'll send an update.

jtyr:
Thanks for testing the patch. The priority parameter is not included as it's not listed in the man page of [yum.conf](http://man7.org/linux/man-pages/man5/yum.conf.5.html). Please could you find where the priority parameter is documented and open a new issue to add it into the `yumrepo` module? Once added, the `params` overrides should allow you to use it.

ke3vin:
Yep, I can certainly do that- the 'priority' parameter is specific to the yum-plugin-priorities plugin.  I was hoping that your intent with this patch was to enable the specification of generic parameters, so that it could support any of the yum plugins out there ( or any to be developed in the future ), without having to update the base yumrepo module each time.

jtyr:
I see. Please let me review it and I will comment here with the result.

jtyr:
I think we should leave this patch as it is. The strict list of parameters have one big advantage - it automatically checks for the type of the input value. If there are plugins which require to add parameters into the `[repository]` section, I'm keen to add them on demand.

@ke3vin please go ahead and create a new issue requesting to add the `priority` parameter.

abadger:
Merged to devel thanks!

PR #1330, opened 2015-12-05
Add new ec2_vpc_dhcp_options module

joelthompson:
This module manages EC2 DHCP options for a given VPC. It allows you to
specify all the options which AWS allows you to set in a DHCP option
set.

PR #1423, opened 2015-12-24
New AWS DHCP Options Sets module

mmochan:
This module will create delete and update dhcp_options_sets.
Any updates to the params require the option_set to be deleted and re-created.
If an update is required and there are VPCs already attached, the newly created options sets will be re-attached to the existing VPCs.


wimnat:
There are already two PRs pending for this functionality - 1311 and 1330. I'll try to review all 3 and choose one to push for inclusion so we can get this functionality in

robynbergeron:
Thanks @mmochan for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

gregdek:
Thanks for this PR @mmochan but #1330 was merged. If the current version of the module does not do what you need, please feel free to submit PRs against the new module.

PR #1424, opened 2015-12-24
Update f5 validate_certs functionality to do the right thing on multi…

abadger:
…ple python versions

This requires the implementation in the module_utils code here
https://github.com/ansible/ansible/pull/13667 to funciton

### DIFF
--- a/network/f5/bigip_facts.py
+++ b/network/f5/bigip_facts.py
@@ -59,7 +59,8 @@ options:
     validate_certs:
         description:
             - If C(no), SSL certificates will not be validated. This should only be used
-              on personally controlled sites using self-signed certificates.
+              on personally controlled sites.  Prior to 2.0, this module would always
+              validate on python >= 2.7.9 and never validate on python <= 2.7.8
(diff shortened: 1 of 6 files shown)


abadger:
\cc @Etienne-Carriere @mhite @perzizzle @srvg  -- Since validate_certs is being added in 2.0, I think we want this to go into 2.0.  if we don't do this then it's a security bug (asking to validate certificates but silently not doing so).  If we take it in 2.0.1 then it's a playbook-breaking change in a bugfix release.  If anyone of you would like to review this change and the one for the module_utils/f5.py module that would be great!

abadger:
Giving it a high priority just to keep it on the radar until we decide for certain whether it must go into 2.0.0.

caphrim007:
These changes look fine to me. I'm ok with the bigip_facts lines 1601 to 1606 changes until such time as we can use the methods in the f5 module_utils. +1

abadger:
Pushed to devel and stable-2.0

PR #1429, opened 2015-12-27
Ensure ec2_win_password doesn't leak file handle

joelthompson:
Currently the module doesn't explicitly close the file handle. This
wraps the reading of the private key in the with context manager to
ensure the file is properly closed.

### DIFF
--- a/cloud/amazon/ec2_win_password.py
+++ b/cloud/amazon/ec2_win_password.py
@@ -140,8 +140,11 @@ def main():
     if wait and datetime.datetime.now() >= end:
         module.fail_json(msg = "wait for password timeout after %d seconds" % wait_timeout)
 
-    f = open(key_file, 'r')
-    key = RSA.importKey(f.read(), key_passphrase)
+    try:
+        f = open(key_file, 'r')
+        key = RSA.importKey(f.read(), key_passphrase)
+    finally:
+        f.close()
     cipher = PKCS1_v1_5.new(key)
     sentinel = 'password decryption failed!!!'
 


ryansydnor:
Nice! Shipit

robynbergeron:
Thanks @joelthompson. @rickmendes please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

ghost:
@joelthompson Thanks for finding this and fixing it.

@robynbergeron shipit

robynbergeron:
Thanks again to @joelthompson for this PR, and thanks @rickmendes for reviewing. Marking for inclusion.

bcoca:
`with`  does not work on python 2.4, in any case handles should not be leaked as once the task ends the process dies.

gregdek:
@bcoca if you put something directly into needs_revision, please notify the module maintainer. Otherwise, feel free to just put "needs_revision" into the text, so that our bot can automatically detect and do the right thing. Thanks!

@joelthompson please note @bcoca's request. When you've made the necessary edits, indicate that you've done so by putting "ready_for_review" in the comment field.

joelthompson:
@bcoca -- understood, but the Python 2.4 compatibility checker explicitly excludes the cloud modules directory: `$ python2.4 -m compileall -fq -x 'cloud/|monitoring/zabbix.*\.py|/dnf\.py|/layman\.py|/maven_artifact\.py|clustering/(consul.*|znode)\.py|notification/pushbullet\.py' .`

It seems like an explicit exception for the cloud modules was made at some point in the past?

Also, I understand that the OS will close the file handle once the process ends, but in general that's considered poor programming style to depend on the OS to do that.

joelthompson:
@bcoca -- just wanted to follow up on this question. #1579 is another example of somebody using code not compatible with Python 2.4 and it not being caught by the syntax checker, so it would be really helpful to not exclude these modules from the Python 2.4 syntax checker.

However, is this even a real issue? This module relies on boto, which seems to only support Python 2.6, and the [auto-generated documentation](http://docs.ansible.com/ansible/ec2_win_password_module.html#requirements) lists Python 2.6 as a requirement -- is that why cloud is excluded from the 2.4 compatibility check?

Thanks!

sivel:
Cloud modules do not _need_ to support Python 2.4, largely because their dependencies usually require Python 2.6 or newer.

Our developer docs state this as well.  https://github.com/ansible/ansible/blob/devel/docsite/rst/developing_modules.rst#module-checklist

However, we prefer the code of modules to adhere to our typical standards, such as not using the `with` context manager and not using `as` syntax in exception handling.

We do not claim to support Python 3 in modules at this time, however should you want to use syntax compatible with Python 3 you should use the new `get_exception()` function provided by `ansible.module_utils.basic`.

As for `with` the recommendation is to use a `try`/`finally` block for handling file closure.

joelthompson:
Thanks @sivel -- that makes a lot more sense!

joelthompson:
ready_for_review

gregdek:
Thanks @joelthompson. @rickmendes please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

ghost:
Looks clean to me. Shipit

PR #1437, opened 2015-12-30
Fix documentation os_project module.

jmferrer:
os_project returns this fatal error:

```
fatal: [openstack-vcenter]: FAILED! => {"changed": false, "failed": true, "msg": "unsupported parameter for module: domain_id"}
```

The documentation is wrong because the correct parameter must be "domain" instead "domain_id".
See line 173

### DIFF
--- a/cloud/openstack/os_project.py
+++ b/cloud/openstack/os_project.py
@@ -44,7 +44,7 @@ options:
         - Description for the project
      required: false
      default: None
-   domain_id:
+   domain:
      description:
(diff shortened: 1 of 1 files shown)


jmferrer:
This is also wrong:
http://docs.ansible.com/ansible/os_project_module.html

bcoca:
the page gets generated from the module docs you update here. I've pushed and alternate fix (makes both domain and domain_id work) via fa11718c0f6632dc9b1117c6887f43df6dd60186

PR #1438, opened 2015-12-30
add ec2 vpc nat gateway module.

jonhadfield:
This PR introduces a new module that ensures state of AWS VPC NAT Gateways (announced 17th Dec 2015).
It depends on boto3 and a recent version of botocore that is checked for within the code.

Jon


joelthompson:
At first blush, I like this better than #1412 -- it feels more in line with Ansible's general paradigm. I'm adding some in-line comments, including a number based upon Ansible's module checklist: http://docs.ansible.com/ansible/developing_modules.html#module-checklist

cloud/amazon/ec2_vpc_nat_gateway.py
```diff
@@ -0,0 +1,251 @@
...
+      - Wait for operation to complete before returning
+    required: false
+    default: true
+  wait_timeout:
+    description:
+      - How long to wait for an operation to complete before timing out
+    required: false
+    default: 300
+
+author: Jon Hadfield
```
joelthompson: This should also include your GitHub username

cloud/amazon/ec2_vpc_nat_gateway.py
```diff
@@ -0,0 +1,251 @@
...
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+import time
+import datetime
+try:
+    import boto3
+    import botocore
+except ImportError:
+    module.fail_json(msg='boto3 and botocore are required.')
```
joelthompson: You shouldn't call module.fail_json here. In fact, module isn't even defined at this point. You should set a boolean value here indicating that boto3 and botocare aren't available, and then in your main method call module.fail_json

cloud/amazon/ec2_vpc_nat_gateway.py
```diff
@@ -0,0 +1,251 @@
...
+from distutils.version import LooseVersion
+if LooseVersion(botocore.__version__) < LooseVersion("1.3.14"):
+    module.fail_json(msg='botocore version 1.3.14 or above is required.')
```
joelthompson: Same as above...

cloud/amazon/ec2_vpc_nat_gateway.py
```diff
@@ -0,0 +1,251 @@
...
+  state:
+    description:
+      - Ensure NAT Gateway is present or absent
+    required: false
+    default: present
+    choices: ['present', 'absent']
+  nat_gateway_id:
+    description:
+      - The id AWS dynamically allocates to the NAT Gateway on creation
+    required: false
```
joelthompson, line 34: "If required is false you need to document default, even if the default is ‘None’"

cloud/amazon/ec2_vpc_nat_gateway.py
```diff
@@ -0,0 +1,251 @@
...
+  subnet_id:
+    description:
+      - The id of the subnet to create the NAT Gateway in
+    required: false
```
joelthompson, line 39: Same...

cloud/amazon/ec2_vpc_nat_gateway.py
```diff
@@ -0,0 +1,251 @@
...
+  allocation_id:
+    description:
+      - The id of the elastic IP allocation
+    required: false
```
joelthompson, line 44: Same...

cloud/amazon/ec2_vpc_nat_gateway.py
```diff
@@ -0,0 +1,251 @@
...
+  wait:
+    description:
+      - Wait for operation to complete before returning
+    required: false
+    default: true
+  wait_timeout:
+    description:
+      - How long to wait for an operation to complete before timing out
```
joelthompson: You should include units in the description (i.e., seconds).

cloud/amazon/ec2_vpc_nat_gateway.py
```diff
@@ -0,0 +1,251 @@
...
+    type: string
+    sample: "nat-0d1e3a878585988f8"
+'''
+
+
+def get_ec2_client(module=None):
+    """ Return an ec2 client """
+    region, ec2_url, aws_connect_kwargs = get_aws_connection_info(module, boto3=True)
+    if not region:
+        module.fail_json(msg="Region must be specified as a parameter, in EC2_REGION or AWS_REGION "
```
joelthompson: I don't like this particular message, e.g., per ansible/ansible#12690 it could be the AWS_DEFAULT_REGION environment variable.

cloud/amazon/ec2_vpc_nat_gateway.py
```diff
@@ -0,0 +1,251 @@
...
+            ensure_nat_gateway_present(module=module)
+    elif 'nat_gateway_id' not in module.params or not isinstance(module.params.get('nat_gateway_id'), basestring):
+        module.fail_json(msg="nat_gateway_id must be specified when ensuring a NAT gateway is absent")
+    else:
+        ensure_nat_gateway_absent(module=module)
+
+from ansible.module_utils.basic import *
+from ansible.module_utils.ec2 import *
+
+main()
```
joelthompson: You should wrap this in the standard

``` python
if __name__ == "__main__":
    main()
```

cloud/amazon/ec2_vpc_nat_gateway.py
```diff
@@ -0,0 +1,251 @@
...
+    argument_spec.update(dict(
+        state=dict(default='present', choices=['present', 'absent']),
+        subnet_id=dict(required=False, type='str'),
+        allocation_id=dict(required=False, type='str'),
+        nat_gateway_id=dict(required=False, type='str'),
+        wait=dict(required=False, type='bool', default=True),
+        wait_timeout=dict(required=False, default=300)))
+
+    module = AnsibleModule(argument_spec=argument_spec, supports_check_mode=True)
+    if module.params['state'] == 'present':
```
joelthompson: Consider using the (poorly documented, as far as I can tell) required_if keyword in the module rather than this. I suggest reading [other examples](https://github.com/ansible/ansible-modules-core/search?utf8=%E2%9C%93&q=required_if) and [the source](https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/basic.py#L1155) to see how it works.

cloud/amazon/ec2_vpc_nat_gateway.py
```diff
@@ -0,0 +1,251 @@
...
+        if not all((isinstance(module.params.get('subnet_id'), basestring),
+                   isinstance(module.params.get('allocation_id'), basestring))):
+            module.fail_json(msg="subnet_id and allocation_id must be specified when ensuring a NAT gateway is present")
+        else:
+            ensure_nat_gateway_present(module=module)
+    elif 'nat_gateway_id' not in module.params or not isinstance(module.params.get('nat_gateway_id'), basestring):
```
joelthompson: Similar comment about required_if

jonhadfield:
Thanks for all the useful feedback. I'll fix this up today.

jonhadfield:
I've incorporated all the suggested changes and decided to use a try/except for the region issue with a generic message and fall back to ec2_utils for anything else.

I put together a couple of playbooks with different test cases here:
failures: http://pastebin.com/794c4GiF
successes: http://pastebin.com/5b4jq76y

Etherdaemon:
Hi @jonhadfield,

I've updated my module incorporating your feedback and also listed yourself as a coauthor.

I've also taken a look at your PR and have the following (hopefully constructive comments):
- From what I can tell the convention is DOCUMENTATION, EXAMPLES, RETURN and then the import statements rather than having the import statements above the documentation etc
- I named my module managed_ngw as a way to ensure it's not confused with the traditional nat gateways that are ec2 instances that most people current utilise - this may/may not be something you want to do
- Most modules tend to instantiate their client connection in the main function rather than declaring it inside other functions - from what I can tell it seems to be the accepted place. 
- Mistyping subnet_id does return a failed status straight away but has no message for the user to see what went wrong - I got msg: "" with no actual error message
- Providing an incorrect/non-existent allocation_id results in the nat gateway being attempted to create but will fail - maybe a quick check for the allocation_id as a way to handle it? That way the user isn't sitting there for the wait timeout to expire to figure out something went wrong? - EG I'm currently returning the botocore error for InvalidAddress.NotFound if an incorrect eip_address is provided
- Your module currently returns just the nat-id which is fine, except currently for route tables I can only add the nat gateway via the eni interface ID which is normally passed back from AWS. Other information such as the attached public IP, eni, private IP could be useful to the user potentially? I guess users could use the facts module instead (submitted a separate PR #1446 for that just now so that its now separate from my existing module)
- I noticed that boto is still relied upon in your module as well - I have comments in the past  from @defionscode that "the preference at the moment is for boto3/botocore only for new modules"
- Further on the allocation_id - I'm not sure theres a way for me as a user to get the allocation_id atm using the existing eip modules so this would require the user to allocate the address and then either use a command statement to use the awscli to lookup the allocation id or provide it manually - hence why I went with the eip_address which is easier to get, provide from the user and I do the describe to get the allocation id for the user instead - or the user could choose to allow me to allocate an address for them

jonhadfield:
@Etherdaemon Thanks for the very constructive feedback. I've also listed you as a co-author in new commit that incorporates changes based on your comments.
- I've re-ordered documentation, examples, return and imports to follow convention.
- Module name - I prefer the explicit 'gateway' and not sure it would be confused with the EC2 NAT instances. Not precious about changing it though.
- Missing/incorrect subnet and allocation ids - I've added checks for these.
- Client connection - I was originally potentially failing in a function before making use of the connection, so decided to leave creation until needed. That's no longer the case, so makes sense to follow your suggestion.
- Return values - You can specify the destination gateway as the nat gateway id that's returned, so no need to return eni. I'm not convinced the module should allow the creation of a NAT GW without specifying an allocation-id. I'll submit a PR for the ec2_eip module to return an allocation-id if it's created in a VPC so that the registered variable could serve as input. I think that's cleaner than duplicating what the ec2_eip is meant for.
- Boto import - Can't remember why I previously needed it. Removing it doesn't break anything, so removed.
- Getting allocation-id - As above, in trying to keep the module as simple as possible I think ec2_eip should return it. I was tempted to add the option of allocation-id and/or EIP, but i think it just complicates it as you'd be checking for valid EIP, valid allocation-id, check if both refer to the same entity, etc. I do agree it would be required to have an EIP option if it doesn't get merged before ec2_eip returns an allocation-id.

jonhadfield:
I've submitted https://github.com/ansible/ansible-modules-core/pull/2753 to get ec2_eip module to return allocation-id.
As above, I think it's cleaner to rely on boto3 canonical identifier (allocation-id) than having to look it up from an IP address.

robynbergeron:
Thanks @jonhadfield for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

jonhadfield:
New test success and failure playbook examples:
[Successes](http://pastebin.com/cLtKrFqb)
[Failures](http://pastebin.com/EgYKSJf5)

Constantin07:
+1

caseylucas:
Thanks @jonhadfield. I have this module working to create nat gateways.

jonhadfield:
Thanks for testing @caseylucas

caseylucas:
Should the module return information about the nat gateway if it already exists?  The other ec2_ modules return enough information (like ids, etc.) so that they can be idempotent.  This module only returns the nat_gateway_id if the nat gateway was actually created.  Caveat: I'm an ansible noob.

Constantin07:
@caseylucas You are right, it would be better to return details of existing NAT gateway even if it already exists.

jonhadfield:
Not sure I understand. It's idempotent in that you can repeatedly enforce the absent or present state without any causing change after the first run.
Have you got an example of a module that returns information if state is already present and I'll take a look?

Constantin07:
Have a look at cloudtrail.py.

Etherdaemon:
@Constantin07 do you want to see if https://github.com/ansible/ansible-modules-extras/pull/1412 has the result you are after?

caseylucas:
When the nat gateway already exists, the module does not return enough data for subsequent use in the playbook.  Specifically, I need the nat gateway id that was either created or already existed. Regardless of create or update, the nat gateway related information returned should be the same.  Of course the "changed" state may be different.

jonhadfield:
@caseylucas @Constantin07 I've added details depending on operation.
Updated test playbooks to prove idempotence and show example output (please tweak subnet and allocation ids accordingly):
Successes: http://pastebin.com/qWMmtaLV
Failures: http://pastebin.com/S6KNFabt

caseylucas:
@jonhadfield thanks. Code at https://github.com/jonhadfield/ansible-modules-extras/commit/2846d1a589950cd3ec1306d18c2be33b8cb92248 is pretty much what I was thinking.  I'll test out and let you know.

caseylucas:
@jonhadfield new code works well - nat_gateway_id is returned regardless of existing or new nat gateway.  Thanks.

Constantin07:
@jonhadfield Thanks.

gregdek:
@Etherdaemon @jonhadfield Help! It looks like this module has the same filename (cloud/amazon/ec2_vpc_nat_gateway.py) as #1412 -- and only one can be merged. Are they doing the same thing? Please advise.

jonhadfield:
@gregdek I wrote this one and then discovered @Etherdaemon had also written one (https://github.com/ansible/ansible-modules-extras/pull/1412). We've both handled the task a bit differently and helped each other out in testing and improvements, so I guess it's now up to the community/core devs to review and choose one.

jonhadfield:
@gregdek, @Etherdaemon has closed https://github.com/ansible/ansible-modules-extras/pull/1412 in the hope of expediting the inclusion of a module to manage NAT gateways.

wimnat:
@jonhadfield can you squash the commits?

cloud/amazon/ec2_vpc_nat_gateway.py
```diff
@@ -0,0 +1,290 @@
...
+short_description: Manage AWS VPC NAT Gateways
+description:
+  - Ensure the state of AWS VPC NAT Gateways based on their id, allocation and subnet ids.
+version_added: "2.1"
+options:
+  state:
+    description:
+      - Ensure NAT Gateway is present or absent
+    required: false
+    default: "present"
```
wimnat, line 29: There shouldn't be a default for state. Ansible core team have told me in the past that state must be explicit.

cloud/amazon/ec2_vpc_nat_gateway.py
```diff
@@ -0,0 +1,290 @@
...
+  wait_timeout:
+    description:
+      - How many seconds to wait for an operation to complete before timing out
+    required: false
+    default: 300
+
+author:
+  - "Jon Hadfield (@jonhadfield)"
+  - "Karen Cheng(@Etherdaemon)"
+extends_documentation_fragment: aws
```
wimnat, line 67: You should also include the ec2 doc fragment for 'region' parameter

cloud/amazon/ec2_vpc_nat_gateway.py
```diff
@@ -0,0 +1,290 @@
...
+ec2_vpc_nat_gateway:
+  state: present
+  subnet_id: subnet-a67643d1
+  allocation_id: eipalloc-a32baec6
+  wait: false
+
+# Ensure that a VPC NAT Gateway identified by id does not exist
+ec2_vpc_nat_gateway:
+  state: absent
+  nat_gateway_id=nat-0d1e3a878585988f8
```
wimnat: The '=' here should be a ':'

cloud/amazon/ec2_vpc_nat_gateway.py
```diff
@@ -0,0 +1,290 @@
...
+                    for nat_gateway_address in nat_gateway.get('NatGatewayAddresses'):
+                        if nat_gateway_address.get('AllocationId') == allocation_id:
+                            gateway_list.append({'state': nat_gateway.get('State'),
+                                                 'created': nat_gateway.get('CreateTime'),
+                                                 'nat_gateway_id': nat_gateway.get('NatGatewayId')})
+                return gateway_list
+        return [{'state': 'absent'}]
+    except botocore.exceptions.ClientError as ce:
+        if "NatGatewayNotFound" in ce.message:
+            return [{'state': 'absent'}]
```
wimnat, line 146: There should be an else here to handle other types of exception

cloud/amazon/ec2_vpc_nat_gateway.py
```diff
@@ -0,0 +1,290 @@
...
+                time.sleep(5)
+        module.fail_json(msg="Waited too long for VPC NAT Gateway to be deleted.")
+    else:
+        module.exit_json(**results)
+
+
+def create_nat_gateway(ec2_client=None, module=None):
+    """ Create the NAT gateway """
+    result = ec2_client.create_nat_gateway(SubnetId=module.params.get('subnet_id'),
+                                           AllocationId=module.params.get('allocation_id'))
```
wimnat: You should wrap the create_nat_gateway in a try:

jonhadfield:
Thanks for the review @wimnat . Will sort it out today.

jonhadfield:
github decided to close as I made the changes and did a force push rather than squash previous commits (worked previously!).
Not sure if this can be re-opened now - so raised https://github.com/ansible/ansible-modules-extras/pull/1687 to replace.

@wimnat The only change I didn't incorporate was to remove the default as 'present'. I've had a look through a number of other modules and documentation and can't see a reason for doing this. Have you got any more info on this being a new standard/requirement?

PR #1454, opened 2016-01-05
Feature/tag elb

welex91:
Tag ELBs with boto 2. Uses a lot of core/cloud/amazon/ec2_tag code for consistent look and feel.

### DIFF
--- a/cloud/amazon/ec2_elb_facts.py
+++ b/cloud/amazon/ec2_elb_facts.py
@@ -112,7 +112,7 @@ def get_elb_listeners(listeners):
 def get_health_check(health_check):
     protocol, port_path = health_check.target.split(':')
     try:
-        port, path = port_path.split('/')
+        port, path = port_path.split('/', 1)
         path = '/{}'.format(path)
     except ValueError:
         port = port_path
--- a/cloud/amazon/ec2_elb_facts.py
+++ b/cloud/amazon/ec2_elb_facts.py
@@ -112,7 +112,7 @@ def get_elb_listeners(listeners):
 def get_health_check(health_check):
     protocol, port_path = health_check.target.split(':')
     try:
-        port, path = port_path.split('/', 1)
+        port, path = port_path.split('/')
         path = '/{}'.format(path)
     except ValueError:
         port = port_path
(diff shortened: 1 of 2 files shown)


jonhadfield:
@welex91 Hoping for this functionality to be added soon, so thought I'd respond with some very opinionated, but hopefully helpful, feedback:
- Why not integrate with ec2_elb_lb module?
- When declaring state as present it should set the tags to those defined, removing any not specified. It currently appends without removing existing which isn't the declarative behaviour I'd expect.
- When declaring state absent, I'd expect it to remove all tags from the elb. Maybe the only parameters should be the name of the load balancer and the tags param for the dict? In that case, if you want to remove tags you leave an empty value.
- Not sure of purpose of state 'list'? Why not just set a parameter called 'elb_tags' to contain the dict when calling module.exit_json so it can be accessed after Register?
- As recently pointed out to me by @joelthompson you can tidy up the module definition with [required_if](https://github.com/ansible/ansible-modules-core/search?utf8=%E2%9C%93&q=required_if). So if you are going to go with state: absent and require tags, then it saves the additional checks.

welex91:
**@jonhadfield** Thanks for the feedback, here's my thinking:
- It's a standalone extras module right now, since I'm not sure whether the right place is ec2_elb_lb or ec2_tag. 
- the state logic is based on ec2_tag, I agree it is not quite declarative. Not sure what the right solution is here. The _list_ state is also from ec2_tag, my understanding is that it should go into ec2_elb_facts instead ?
- thanks for the pointer to required_if, I'll incorporate that.

Constantin07:
I agree with @jonhadfield that it's better to integrate into ec2_elb_lb. There are other modules which have 'tags' option too.

robynbergeron:
Thanks @welex91 for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

welex91:
@jonhadfield and @Constantin07 : I'll add the list part to ec2_elb_facts and the present/absent parts to ec2_elb_lb. Thanks for the feedback.

Constantin07:
Thanks @welex91. Meanwhile I went for AWS CLI equivalent:

```
- name: Tag ELB
  command: >
    aws elb add-tags --load-balancer-names "{{ elb_lb_res.elb.name }}"
    --tags "Key=Env,Value={{ env }}" "Key=Name,Value={{ elb.name }}"
    --region "{{ aws_region }}"
  tags: elb
```

PR #1455, opened 2016-01-05
New module to enable editing the HOSTS file on Windows

alxwest:
New module to configure the HOSTS file on Windows.

Moved to a separate branch as I've just learnt pull requests track branches not commits.


jhawkesworth:
Hi @alxwest thanks for submitting this.  

I like the idea of this module.  My first thought is that perhaps a similar result could be achieved using the existing (in Ansible 2.0) win_lineinfile module, but I can see how dealing with hostnames and ips, rather than just strings is useful in the context of what Ansible does.

Does this module need to run as a specific user in order to have permissions to edit the hosts file.  If so, it should say so in the documentation?

I have some other comments which I'll put in line.

windows/win_hosts.ps1
```diff
@@ -0,0 +1,100 @@
...
+                } else {
+                        $newLines += $line
+                }
+        }
+
+        # Write file
+        Clear-Content $filename
+        foreach ($line in $newLines) {
+                $line | Out-File -encoding ASCII -append $filename
+        }
```
jhawkesworth, line 64: I think as a minimum
 $result.changed = $true 
should be set here.
However this means that any run of the module will always modify the file and report true.

I think really the module should detect if the change it is about to make needs to be made and then on subsequent runs of the same playbook ansible would report 'ok' rather than changed if it didn't actually have to change the hosts file.  This would be consistent with how other modules work in Ansible.

windows/win_hosts.ps1
```diff
@@ -0,0 +1,100 @@
...
+}
+
+# name
+$name = Get-Attr $params "name" $FALSE
```
jhawkesworth, line 68: You could save a few lines of code here by using Get-AnsibleParam function - some examples here: http://docs.ansible.com/ansible/developing_modules.html#windows-modules-checklist

windows/win_hosts.py
```diff
@@ -0,0 +1,71 @@
...
+    description:
+      - The name of the Hostname
+    required: true
+  ip:
+    description:
+      - The IP address
+    required: true
+  state:
+    description:
+      - present to ensure environment variable is set, or absent to ensure it is removed
```
jhawkesworth: Do you mean 'host entry' instead of 'environment variable' here?

windows/win_hosts.ps1
```diff
@@ -0,0 +1,100 @@
...
+                                $result.changed = $true
+                        }
+                } else {
+                        $newLines += $line
+                }
+        }
+
+        # Write file
+        Clear-Content $filename
+        foreach ($line in $newLines) {
```
jhawkesworth, line 62: Might be quicker to append all the lines in one go rather than for each member of $newLines

alxwest:
Hi @jhawkesworth, thanks for the feedback.
You're right, it needs administration rights, I'll add that to the documentation.
I'm using Ansible because it abstracts tasks into sensible modules that the layman can understand and edit. You're right the same task can be achieved by using win_lineinfile but means the user needs to understand the concept of the hosts file. 
The main reason I chose Ansible was of the low entry barrier when configuring servers. It uses configuration files not code. It overcomes the "I'm not a coder, I can't do this, give it to the coders" mentality that can come with Puppet or Chef, then I find for loops in my configuration files.

jhawkesworth:
I too like ansible because its not a programming language.

Looking forward to seeing your changes.

sivel:
Feedback from automated scanning of this module:

```
./t
ERROR: Official Ansible modules must have a .py extension for python modules or a .ps1 for powershell modules
ERROR: Python SyntaxError while parsing module
./windows/win_hosts.ps1
ERROR: GPLv3 license header not found
./windows/win_hosts.py
```

It appears as though you have accidentally included an errant file named `t`.

Constantin07:
+1

tommywo:
+1

Issue #1452, opened 2016-01-05
Maven artifact module doesn't understand ~ for home directory


danielcompton:
##### Ansible Version:

```
ansible 2.0.0
  config file =
  configured module search path = Default w/o overrides
```

```
brew info ansible
... devel 2.0.0-0.8.rc3, HEAD
```
##### Ansible Configuration:

N/A
##### Environment:

From Mac OS X, to Ubuntu 14.04
##### Summary:

Maven artifact module doesn't understand `~` in `dest` path.
##### Steps To Reproduce:

The first module succeeds, the second one fails

```
- maven_artifact: group_id=junit artifact_id=junit dest=/home/vagrant/junit-latest.jar
- maven_artifact: group_id=junit artifact_id=junit dest=~/junit-latest2.jar
```
##### Expected Results:

I expect there to be two JARs in my home directory, `junit-latest.jar`, and `junit-latest2.jar`.
##### Actual Results:

```
TASK [myrole : maven_artifact] ************************************************
ok: [127.0.0.1]

TASK [myrole : maven_artifact] ************************************************
[...]

PR #1457, opened 2016-01-05
maven_artifact - Use expanduser for dest path

chrrrles:
Fix for #1452


abadger:
Instead see if using `dest = dict(default=None, type="path")` in the argument spec does the job.

bcoca:
using type=path also expands vars ($HOME)

chrrrles:
Ahh - good point.

abadger:
Merged!

PR #1458, opened 2016-01-06
Fix puppet module formatting issue

caseylang:
The `->` in the `show_diff` option doc seemed to be causing the docs page to break. Not sure why, since it was still valid YAML. Please let me know if there is anything else I can do to help.

Image of issue from docs site:

image

cc: @emonty

### DIFF
--- a/system/puppet.py
+++ b/system/puppet.py
@@ -49,9 +49,7 @@ options:
     default: None
   show_diff:
     description:
-      - >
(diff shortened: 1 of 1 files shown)


Jmainguy:
Yup site looks busted to me, this change looks good.

bcoca:
hmm, i really don't like the show_diff option ... as --diff on the command line should take care of this

PR #1460, opened 2016-01-06
Remove advertising show_diff feature

Jmainguy:
--diff will be added to more modules soon, and we want puppet module to utilize this instead of show_diff

### DIFF
--- a/system/puppet.py
+++ b/system/puppet.py
@@ -47,12 +47,6 @@ options:
       - Path to the manifest file to run puppet apply on.
     required: false
     default: None
-  show_diff:
-    description:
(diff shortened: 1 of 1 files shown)


Jmainguy:
see @bcoca comment in #1458

Jmainguy:
After talking with the @emonty we decided it was best to remove the documentation but leave the feature, since some people might already be using it. However, we do not recommend anybody use this feature going forward, as the correct way to have show_diff on will be to use --diff. That feature will be added shortly, but not before the 2.0 window drops, for this reason, we wanted to get this documentation removed as soon as possible before people start writing playbooks.

PR #1223, opened 2015-11-09
New AWS Cloudfront module

Etherdaemon:
Create, delete and update actions for the main AWS Cloudfront actions. 
- Requires Boto3 and botocore and json.
- Refer to the AWS documentation for Cloudfront for the correct json values as there are too many to list here.
- Creates distributions, streaming_distributions, invalidations, origin_access_ids
- Deletes distributions, streaming_distributions, origin_access_ids
- Updates distributions, streaming_distributions

PR #1467, opened 2016-01-08
Updated Amazon module guidelines regarding boto3

wimnat:
Updated Amazon module guidelines as per https://groups.google.com/forum/#!searchin/ansible-devel/boto3/ansible-devel/BlruiipCVxk/KdvQI-wSSYUJ


Etherdaemon:
Hi @wimnat - can you help clarify the boto3/boto/botocore requirements for new modules?

When I had boto3 with boto in https://github.com/ansible/ansible-modules-extras/pull/1223#issuecomment-156488951 @defionscode seemed to indicate that boto should be culled out which is what I've been trying to do with any new PR's I've been raising. But the module guidelines here seem to suggest that we should be importing boto and boto3? Or have I interpreted that incorrectly?

wimnat:
The example shows importing boto and boto3 however, the guidelines change is supposed to highlight the fact that any new modules for aws must use boto3. So if you've been culling boto in any new modules then that's a good thing. However, we want to maintain backward compatibility with existing modules so in those cases, boto should remain.

Etherdaemon:
@wimnat Ah that makes sense - thanks heaps for that.

robynbergeron:
Thanks @wimnat. Since you are a maintainer of this module, we are marking this PR for inclusion.

wyaeld:
@wimnat what about how the connection and resource loading happens?  I'm pretty new to boto, so might be reading wrong, but it seems that these lines below change quite a bit, which makes co-existence of features within a single module trickier

Using SQS module example:

``` python
#boto
connection = connect_to_aws(boto.sqs, region, **aws_connect_params)

#boto3
sqs = boto3.client('sqs')
```

Should your guidelines take a stance on this so newbies (like me) understand what to do?

juliedavila:
Yeah a good bit on how to use the boto3 piece in module_utils/ec2.py would be good

wimnat:
@defionscode can you elaborate on what else you'd like to see?  I'd like to get this merged asap as i've seen a few new modules coming through using old boto.

juliedavila:
@wimnat I'd like to see an update to the `Connecting to AWS` portion that provides the boto3 way of connecting via ec2.py from module_utils

bcoca:
Also I'd like to leave very clear:
- new stuff should be boto3
- existing functionality can be 'fixed' to also support boto3
- existing functionality MUST still be boto compatible.

wimnat:
@bcoca @defionscode updated. let me know what you think

cloud/amazon/GUIDELINES.md
```diff
@@ -13,76 +26,197 @@ known abbreviation due to it being a major component of
...
+### Connecting to AWS
+
+To connect to AWS, you should use `get_aws_connection_info` and then
+`connect_to_aws`.
+
+The reason for using `get_aws_connection_info` and `connect_to_aws` rather than doing it
+yourself is that they handle some of the more esoteric connection
+options such as security tokens and boto profiles.
+
+Some boto services require region to be specified. You should for check for the region parameter if required.
```
juliedavila: Small grammar error. `for` is extra here and should be removed.

juliedavila:
@wimnat alright so just fix up the small grammar error then just to confirm, your module_util functions, they only apply to boto3? Perhaps I'm having a brain fart but I thought boto also returned camelCase values as it's really just a return from the AWS API. If I'm mistaken, disregard, otherwise it would be good to make the language around the description of the functions to read as agnostic to boto/boto3.

wimnat:
@defionscode I fixed the extra for.  With boto, it usually returns just the object with some attributes so it was up to the developer to assign to a dict key e.g. mydict['some_id'] = object.some_id so i think the module_utils functions being boto3 specific is valid.

juliedavila:
as always, thanks @wimnat

PR #1478, opened 2016-01-10
Extends firewalld module with ability to add/remove interfaces to/from zones

m0:


gregdek:
Thanks @m0. @maxamillion please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

mscherer:
Shouldn't we verify that people do not give incompatible arguments ? Like if someone use service= and interface=, should we accept it, or not ?

mscherer:
Ok, forget, just did read the code, seems to be taken in account already.

system/firewalld.py
```diff
@@ -47,6 +47,11 @@
     required: false
     default: null
     version_added: "2.0"
+  interface:
+    description:
+      - 'The interface you would like to add/remove to/from a zone in firewalld - zone must be specified'
```
mscherer: Not sure about the warning about the zone, the others docstring do not have it.

system/firewalld.py
```diff
@@ -47,6 +47,11 @@
...
+    required: false
+    default: null
```
mscherer, line 54: you should also add a version_added variable.

system/firewalld.py
```diff
@@ -162,6 +168,29 @@ def remove_source(zone, source):
     fw_zone.update(fw_settings)
 
 ####################
+# interface handling
+#
+def get_interface(zone, interface):
+    fw_zone = fw.config().getZoneByName(zone)
+    fw_settings = fw_zone.getSettings()
+    if interface in fw_settings.getInterfaces():
+       return True
```
mscherer, line 178: why not "return interface in fw_settings.getInterfaces()" directly ?

system/firewalld.py
maxamillion, line 54: I agree with the addition of version_added here for sake of posterity

m0, line 54: When I was thinking about that, I came to the point that the version might be added by the maintainer on release. Is "2.0" the version that should be used here?

m0:
Thanks for the feedback, I'll apply the proposed changes.
@maxamillion Could you give me a hint about the right version number to be used?

maxamillion:
@gregdek any guidance there? I was thinking 2.1

system/firewalld.py
m0, line 178: I agree, that would be less code, but the "if .. in ...: return True else return False" pattern is used in the other methods, too. Personally I would sacrifice "less code" for "keep existing patterns" for better readability and I didn't want to introduce too many changes by also changing the other methods. @maxamillion, what are your preferences?

r1k0:
any status on this? I'd be glad to add interfaces with firewalld.
Thanks

mscherer:
yeah, 2.1 is the current, so I think that's ok.

system/firewalld.py
maxamillion, line 178: Wow, I somehow completely missed the comment thread here. Apologies.

I'm fine either way, there's a major module refactor effort happening elsewhere anyways so I'm +1 for keeping this pattern for now and the refactor can update everything together for consistency.

maxamillion:
:+1: shipit

gregdek:
Thanks again to @m0 for this PR, and thanks @maxamillion for reviewing. Marking for inclusion.

PR #1482, opened 2016-01-11
Update for modules which import json.

abadger:
Some do not use the json module directly so don't need import json.
Some needed to fallback to simplejson with no traceback if neither was installed

Fixes #1298

### DIFF
--- a/cloud/amazon/ecs_task.py
+++ b/cloud/amazon/ecs_task.py
@@ -98,7 +98,6 @@ task:
     sample: "TODO: include sample"
 '''
 try:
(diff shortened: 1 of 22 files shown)


abadger:
@tronicum Please take a look and see if this addresses your issue with using these modules on CentOS5.

robynbergeron:
Thanks @abadger. @tarka please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

PR #1420, opened 2015-12-23
Ensure that port_path is split into at most 2 components.

welex91:
Without this fix, ec2_elb_facts will fail with:
ValueError: invalid literal for int() with base 10: '80/api/v1/health_check/'

if the Ping Target is HTTP:80/api/v1/health_check/

PR #1489, opened 2016-01-12
a ping_path may have more than one path separator

robbwagoner:
@mjschultz This is a very small change to handle `ping_path`s which have more than one occurrence of the path separator. E.g. `/path/health`


mjschultz:
Thanks, I think https://github.com/ansible/ansible-modules-extras/pull/1420 already plans to cover this case though!

robbwagoner:
@mjschultz Doh! ;-) Closing this.

Issue #1472, opened 2016-01-09
Filesystem module doesn't allow re-initialization


bsanders:
So I've hit what is maybe a corner case.  Using the filesystem module, it appears you can't overwrite an existing filesystem with a new one, if the filesystem would be the same fstype and size.  I can certainly see why that is sensible, but it seems like maybe with the force=yes option, it could be viable?  I could put together a pull request, if there's interest.

robynbergeron:
[module: system/filesystem.py]

robynbergeron:
@bsanders Thanks for submitting this bug report.

@abulimov -- thanks in advance for taking a look at this bug report and for your ongoing work in maintaining this module.

abulimov:
Thanks for the report, I've made a pull-request #1493 to fix it.

freels:
FYI, this change almost caused us to lose data in one of our ansible-managed clusters. This behavior should have been introduced via a separate option, or at the very least come with a lot more warning.

ansibot:
@abulimov, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@abulimov, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@abulimov, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

bsanders:
@robynbergeron looks like @abulimov has been AFK for a while.

Wondering if this should be closed, as it's been merged?  Or should it be better documented as @freels suggested?

ansibot:
@abulimov, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@abulimov, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@abulimov, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
[...]

PR #1493, opened 2016-01-13
Allow recreation of same FS with force=yes in filesystem module

abulimov:
This small change fixes issue #1472.


gregdek:
Thanks @abulimov. Since you are a maintainer of this module, we are marking this PR for inclusion.

PR #1495, opened 2016-01-13
Allow numeric npm package versions

nwjsmith:
When passing a package version that parses as a number (e.g. `1.9`), the version should be converted to a string before being concatenated to the package name.


resmo:
This looks good to me and I am about to merge it, but also like to ping @chrishoffman, the maintainer which hasn't been informed about this bug fix PR in the first place.

chrishoffman:
looks good to me

PR #1497, opened 2016-01-13
Fix documentation default value for gluster_volume start_on_create

jgeusebroek:


jgeusebroek:
The failed Travis test is totally unrelated to this commit.

PR #1502, opened 2016-01-15
New AWS module for managing VPC Network ACLs

mmochan:
Creates and deletes VPC Network ACLs and manages the subnet associations. Requires BOTO3.

### DIFF
--- a/cloud/amazon/ec2_vpc_nacl.py
+++ b/cloud/amazon/ec2_vpc_nacl.py
@@ -162,15 +162,15 @@ def load_tags(module):
     return tags
 
 
-def subnets_removed(nacl_id, subnets, client):
-    results = find_acl_by_id(nacl_id, client)
+def subnets_removed(nacl_id, subnets, client, module):
+    results = find_acl_by_id(nacl_id, client, module)
     associations = results['NetworkAcls'][0]['Associations']
     subnet_ids = [assoc['SubnetId'] for assoc in associations]
     return [subnet for subnet in subnet_ids if subnet not in subnets]
 
 
-def subnets_added(nacl_id, subnets, client):
-    results = find_acl_by_id(nacl_id, client)
+def subnets_added(nacl_id, subnets, client, module):
+    results = find_acl_by_id(nacl_id, client, module)
     associations = results['NetworkAcls'][0]['Associations']
     subnet_ids = [assoc['SubnetId'] for assoc in associations]
     return [subnet for subnet in subnets if subnet not in subnet_ids]
@@ -183,22 +183,22 @@ def subnets_changed(nacl, client, module):
     nacl_id = nacl['NetworkAcls'][0]['NetworkAclId']
     subnets = subnets_to_associate(nacl, client, module)
     if not subnets:
-        default_nacl_id = find_default_vpc_nacl(vpc_id, client)[0]
-        subnets = find_subnet_ids_by_nacl_id(nacl_id, client)
+        default_nacl_id = find_default_vpc_nacl(vpc_id, client, module)[0]
+        subnets = find_subnet_ids_by_nacl_id(nacl_id, client, module)
         if subnets:
-            replace_network_acl_association(default_nacl_id, subnets, client)
+            replace_network_acl_association(default_nacl_id, subnets, client, module)
             changed = True
             return changed
         changed = False
         return changed
-    subs_added = subnets_added(nacl_id, subnets, client)
+    subs_added = subnets_added(nacl_id, subnets, client, module)
     if subs_added:
-        replace_network_acl_association(nacl_id, subs_added, client)
+        replace_network_acl_association(nacl_id, subs_added, client, module)
         changed = True
-    subs_removed = subnets_removed(nacl_id, subnets, client)
+    subs_removed = subnets_removed(nacl_id, subnets, client, module)
     if subs_removed:
-        default_nacl_id = find_default_vpc_nacl(vpc_id, client)[0]
-        replace_network_acl_association(default_nacl_id, subs_removed, client)
+        default_nacl_id = find_default_vpc_nacl(vpc_id, client, module)[0]
+        replace_network_acl_association(default_nacl_id, subs_removed, client, module)
         changed = True
     return changed
 
@@ -216,9 +216,9 @@ def nacls_changed(nacl, client, module):
     tmp_ingress = [entry for entry in entries if entry['Egress'] is False]
     egress = [rule for rule in tmp_egress if DEFAULT_EGRESS != rule]
     ingress = [rule for rule in tmp_ingress if DEFAULT_INGRESS != rule]
-    if rules_changed(egress, params['egress'], True, nacl_id, client):
+    if rules_changed(egress, params['egress'], True, nacl_id, client, module):
         changed = True
-    if rules_changed(ingress, params['ingress'], False, nacl_id, client):
+    if rules_changed(ingress, params['ingress'], False, nacl_id, client, module):
         changed = True
     return changed
 
@@ -229,7 +229,7 @@ def tags_changed(nacl_id, client, module):
     if module.params.get('tags'):
         tags = module.params.get('tags')
     tags['Name'] = module.params.get('nacl')
-    nacl = find_acl_by_id(nacl_id, client)
+    nacl = find_acl_by_id(nacl_id, client, module)
     if nacl['NetworkAcls']:
         nacl_values = [t.values() for t in nacl['NetworkAcls'][0]['Tags']]
         nacl_tags = [item for sublist in nacl_values for item in sublist]
@@ -246,7 +246,7 @@ def tags_changed(nacl_id, client, module):
     return changed
 
 
-def rules_changed(aws_rules, param_rules, Egress, nacl_id, client):
+def rules_changed(aws_rules, param_rules, Egress, nacl_id, client, module):
     changed = False
     rules = list()
     for entry in param_rules:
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @mmochan for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

sivel:
This PR seems to have the `ec2_vpc_nacl.py` duplicated in 2 paths:
- `./cloud/amazon/ec2_vpc_nacl.py`
- .`/ec2_vpc_nacl.py`

Please remove the one in the root.

Additionally please squash all commits, that way we don't have any commits with the duplicate file.

sivel:
Also, after you remove the file in the root, and just keep the one nested in `cloud/amazon` the travis failure should go away, as we exclude all cloud modules from the python2.4 syntax check.

mmochan:
Thanks @sivel

naslanidis:
+1  I have been using this module today for NACL creates and deletes and it works well.

phumpal:
@mmochan Using your example as a template I'm also passing `security_token`, `aws_access_key` and `aws_secret_key` but am seeing

```
fatal: [localhost -> localhost]: FAILED! => {"changed": false, "failed": true, "msg": "An error occurred (AuthFailure) when calling the DescribeNetworkAcls operation: AWS was not able to validate the provided access credentials"}
```

This is the same approach I use for other aws/ec2 core/extra modules and those work ;)

Package versions

```
ansible==2.0.0.2
boto==2.39.0
boto3==1.2.4
botocore==1.3.28
```

mmochan:
@phumpal - Could you paste in your example and I'll have a look.

phumpal:
@mmochan - thanks!

It's essentially the same as the example [here](https://github.com/mmochan/ansible-modules-extras/blob/new_aws_module_ec2_vpc_nacl/cloud/amazon/ec2_vpc_nacl.py#L67-L87) but with an sts component.

Here's a [gist](https://gist.github.com/phumpal/c7f63b51cf1d3f0348e6) which creates a vpc and attempts to assign acls using this module.

mmochan:
@phumpal - Can you confirm that the role has access permissions to do a  DescribeNetworkAcls. The module itself is does not handle the AWS authentication as this is handed off to Ansible.

phumpal:
@mmochan - the role is assigned the `AmazonVPCFullAccess` policy which includes `DescribeNetworkAcls`

mmochan:
@phumpal - Ok leave it with me I will test this.

phumpal:
Thanks @mmochan - much appreciated.

mmochan:
@phumpal - The problem has been identified. A previous commit to the ec2.py module had been overwritten.  
You can test this by manually removing the 2 lines below in the `boto3_conn` function of the ~/ansible/lib/ansible/module_utils/ec2.py  module. 

> ```
> params['aws_session_token'] = params.pop('security_token', None)
> params['verify'] = params.pop('validate_certs', None)
> ```

(Around line 46 of the module)

Thanks @Etherdaemon for identifying and raising the fix  [https://github.com/ansible/ansible/pull/14633]

phumpal:
:+1: thanks @mmochan - looking forward to see this released. NACLs creation is working now! :+1: 

Great work on the module btw.

and many thanks to @Etherdaemon for tacking the issue down.

phumpal:
shipit

dennisconrad:
Python: 2.7.11
botocore: 1.4.0
boto3: 1.2.6

@mmochan Thank you for your work on this module!  However, in its current state, it unfortunately fails when trying to create a Network ACL without assigning the ACL to a subnet; I can see two problems.

First:

If no subnets are provided, `subnets_to_associate` changes the type of `subnets` from an empty list to a boolean `False`:

``` python
def subnets_to_associate(nacl, client, module):
    params = list(module.params.get('subnets'))
    if not params:
        return False
```

This eventually breaks the boto3 call in `describe_acl_associations`, as `Values` should be either a list or a tuple:

``` python
def describe_acl_associations(subnets, client, module):
    try:
        results = client.describe_network_acls(Filters=[
            {'Name': 'association.subnet-id', 'Values': subnets}
        ])
```

Changing the`return False` to `return []` in `subnets_to_associate` (see above) fixes this problem.

Second:

With the above fixed, the boto3 call in `describe_acl_associations` works, but returns no associations, which leads to an `IndexError` in

``` python
    associations = results['NetworkAcls'][0]['Associations']
```

Since the result would be an empty list anyway,  just return and empty list if `subnets` is an empty list:

``` python
def describe_acl_associations(subnets, client, module):
    if not subnets:
        return []
    try:
```

This should fix the creation of Network ACLs without assigning the ACLs to a subnet or subnets.  Thank you!

mmochan:
@dennisconrad - You are absolutely right and I'm not sure how missed that. I have updated the module with your suggestions, which as you state allows the NACL to be created when no subnets are given. 

Thank you for taking the time to review and finding the bug. Very much appreciated.

gregdek:
@mmochan A friendly reminder: this pull request has been marked as needing your action. If you still believe that this PR applies, and you intend to address the issues with this PR, just let us know in the PR itself and we will keep it open pending your changes.

mmochan:
@gregdek I'm not sure what action is needed?

wholroyd:
@mmochan The only thing I would suggest changing about this module is the `nacl` property which is just a name. The thing that drives a bunch of people nuts I work with on all of the AWS modules (core and extras) is the lack of consistency - sometimes it's `tags` other times it's `resource_tags`, sometimes it's `[property]` other times it's `[property]_id`. Could you change it to `name` given that the way it is now, it lends itself to possibly wanting an ID?

gregdek:
@mmochan looks like we never retagged as "ready_for_review". Sorry about that. Looks like we just need two shipits.

mmochan:
@wholroyd The nacl property has now been changed to name. Testing that change uncovered a bug where a nacl would not be deleted if there were no subnet associations present, this has also been fixed.

Thanks for highlighting this.

wholroyd:
Thanks @mmochan , works perfectly!

shipit :shipit:

MichaelBaydoun:
@gregdek I tested this module as promised during new module review meeting today, and it worked great for me in all tested cases.  

shipit

mmochan:
@wholroyd @gregdek Thanks Guys

mmochan:
ready_for_review

cloud/amazon/ec2_vpc_nacl.py
```diff
@@ -0,0 +1,545 @@
...
+  state:
+    description:
+      - Creates or modifies an existing NACL
+      - Deletes a NACL and reassociates subnets to the default NACL
+    required: false
+    choices: ['present', 'absent']
+    default: present
+author: Mike Mochan(@mmochan)
+extends_documentation_fragment: aws
+'''
```
gundalow, line 62: Please add

`requirements: [ botocore, boto3, json ]`

cloud/amazon/ec2_vpc_nacl.py
```diff
@@ -0,0 +1,545 @@
...
+        tags=dict(required=False, type='dict'),
+        ingress=dict(required=False, type='list', default=list()),
+        egress=dict(required=False, type='list', default=list(),),
+        state=dict(default='present', choices=['present', 'absent']),
+        ),
+    )
+    module = AnsibleModule(argument_spec=argument_spec)
+
+    if not HAS_BOTO3:
+        module.fail_json(msg='json and boto3 is required.')
```
gundalow: - botocore

are (not is) required

cloud/amazon/ec2_vpc_nacl.py
```diff
@@ -0,0 +1,545 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+DOCUMENTATION = '''
+module: ec2_vpc_nacl
+short_description: create and delete Network ACLs.
+description:
+  - Read the AWS documentation for Network ACLS
+    U(http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_ACLs.html)
+version_added: "2.1"
```
gundalow: Please can you update this to 2.2

gundalow:
Hi,

Thanks for this PR, I've raised just a few very minor issues. Once those are fixed I'll give this the stamp of approval

needs_revision

gregdek:
@mmochan looks like you're quite close! If you can make the changes suggested by @gundalow and then mark this "ready_for_review", I think we can put this into shipit very shortly after that. Thanks for your hard work and patience!

cloud/amazon/ec2_vpc_nacl.py
mmochan, line 62: done

cloud/amazon/ec2_vpc_nacl.py
mmochan: done

cloud/amazon/ec2_vpc_nacl.py
mmochan: and done

mmochan:
@gundalow I've made those changes as requested

mmochan:
@gregdek ready_for_review

gregdek:
Thanks @mmochan for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

Etherdaemon:
@gregdek - can this be shipped? It has 2 shipit comments already - and I would like to add my own shipit as well.

gundalow:
shipit

wholroyd:
Shipit

gregdek:
That makes it 4 shipits. Hard to argue with that. :)

mmochan:
Thanks guys

brandond:
One nit before this is merged - the VPC ID parameter documentation and examples do not match, and the code (as written) does not match the convention of using 'vpc_id' as the option when an actual ID is required. I have created https://github.com/mmochan/ansible-modules-extras/pull/2 to fix this.

mmochan:
@brandond That makes sense and I've merged that  PR.

Thanks

cpcundill:
Can't wait for this!  Needed in our org!

brandond:
@mmochan shipit!

brandond:
@cpcundill if you like this you might like https://github.com/ansible/ansible-modules-extras/pull/2209

mmochan:
@brandond Thanks

Issue #939, opened 2015-09-09
zypper_repository dont changes repo url


martingabelmann:
##### Summary:

The module zypper_repository dont changes repositories if the given URL has changed.
##### Steps To Reproduce:

First I added a repo with

```
- name: add repository
  zypper_repository: name='OurRepo' repo='file://path/to/repo' state=present
```

which worked as expected. On the second run nothing changes.

Then we switched  to a http-repo, so I just changed the repo-url:

```
- name: add repository 
  zypper_repository: name='OurRepo' repo='http://rpm.exampe.org/repo/' state=present
```

but ansible didnt changed something until I removed the repo by hand and run it again.

Maybe there is something wrong with [L150](https://github.com/ansible/ansible-modules-extras/blob/devel/packaging/os/zypper_repository.py#L150) (I guess this part was for comparing if the repo-url/-name has changed) , or maybe i missunderstand what the 'expected behavior' should be.
##### Expected Results:

TASK: changed
on first run after edit.
##### Actual Results:

TASK: ok

robynbergeron:
Hi @martingabelmann - thanks for submitting this bug report.

@matze -- pinging you to let you know about this bug report for zypper_repository.py.

robynbergeron:
[module: packaging/os/zypper_repository.py]

robinro:
2 scenarios that should work:

```
tasks:
- zypper_repository: alias=myoldalias url=mynewurl
- zypper_repository: alias=mynewalias url=mynewurl
```

and

```
tasks:
- zypper_repository: alias=mynewalias url=myoldurl
- zypper_repository: alias=mynewalias url=mynewurl
```

In both cases, the second task should be changed and the state afterwards should be `alias=mynewalias url=mynewurl`

agx:
This is fixed by #1515. I'm happy to provide integration test once we've merged this to make sure we don't regress.

PR #1515, opened 2016-01-18
zypper_repository: don't silently ignore repo changes

agx:
So far when a repo URL changes this got silently ignored (leading to
incorrect package installations) due to this code:

```
elif 'already exists. Please use another alias' in stderr:
    changed = False
```

Removing this reveals that we correctly detect that a repo definition
has changes (via repo_subset) but don't indicate this as change but as a
nonexistent repo. This makes us currenlty bail out silently in the above
statement.

To fix this distinguish between non existent and modified repos and
remove the repo first in case of modifications (since there is no force
option in zypper to overwrite it and 'zypper mr' uses different
arguments). Recreate it as usual. Only do this if we have an alias name
since with a different name a new repo is o.k.

### DIFF
--- a/packaging/os/zypper_repository.py
+++ b/packaging/os/zypper_repository.py
@@ -83,15 +83,6 @@ EXAMPLES = '''
 
 REPO_OPTS = ['alias', 'name', 'priority', 'enabled', 'autorefresh', 'gpgcheck']
 
-def zypper_version(module):
-    """Return (rc, message) tuple"""
-    cmd = ['/usr/bin/zypper', '-V']
-    rc, stdout, stderr = module.run_command(cmd, check_rc=False)
-    if rc == 0:
-        return rc, stdout
-    else:
-        return rc, stderr
-
 def _parse_repos(module):
     """parses the output of zypper -x lr and returns a parse repo dictionary"""
     cmd = ['/usr/bin/zypper', '-x', 'lr']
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @agx. @matze please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

matze:
shipit :+1:

agx:
@matze thanks a lot for review! 

I needed to do some major changes since the detection of changes in the repo did not work as intended with multiple repos (marking repos as changed although they weren't). 

Since I have no means to test zypper 0.x I removed support for that ancient version (SLES11 already has 1.x and all currently supported OpenSUSE versions as well).

I'm happy to provide integration tests in ansible proper but would like to avoid the situation with

https://github.com/ansible/ansible/commit/0821d251c8878990326ffee567ab2a5baafd089a

where we have the tests merged in ansible but the fixes still  pending in ansible-modules-extras.

robinro:
this might also fix #939

gregdek:
Thanks again to @agx for this PR, and thanks @matze for reviewing. Marking for inclusion.

agx:
I fixed an issue when deleting repositores (state=absent) with a separat patch on top of the current one. I'm happy to squash the patches but wanted to ease review.

This makes the just submitted integration tests

```
https://github.com/ansible/ansible/pull/15004
```

pass so hopefully we won't regress in the future.

agx:
This one has a four months old. Can it be merged?

gregdek:
@ansible/core team, looks like this has been in shipit for a while... help?

robinro:
https://github.com/ansible/ansible-modules-extras/pull/1990#issuecomment-209676569 raises an issue also for this PR. We drop support for old zypper versions. I added some requirements statement over there but maybe @agx knows more about which SLES version we need to have the new zypper?

https://github.com/ansible/ansible-modules-extras/pull/1990 is based on this PR and extends it. Maybe they can be considered together?

robinro:
@agx
I hope you agree with the merge and close. Please let me know if not. I only now realized you didn't comment on them.
I kept your commits, so your contributions are visible. Thanks for your work!

agx:
Hi,
On Mon, Apr 18, 2016 at 09:12:07AM -0700, Robin Roth wrote:

> @agx
> I hope you agree with the merge and close. Please let me know if not. I only now realized you didn't comment on them.
> I kept your commits, so your contributions are visible. Thanks for your work!

I'm fine with anything that gets ansible usable on SLES. That one of
these. Thanks a log!
 -- Guido

PR #1521, opened 2016-01-19
Fix exception output for os_keystone_domain.

Shrews:
The message attribute of a shade exception is not very helpful.
Converting to a full string will contain many more details.


Shrews:
ping @emonty @j2sol

emonty:
shipit

sivel:
Merged, thanks!

PR #1518, opened 2016-01-19
Allow a list of successive responses for a pattern

quinot:
Allow the user to specify successive responses to be specified for
a given pattern, e.g. to send a succession of commands to a prompt:

```
    responses:
      "assword:": MySecret
      ">": enable
      "#":
        - skip-page-display
        - show conf
        - logout
```

PR #1522, opened 2016-01-19
Add functionality to give multiple iterative responses for a question

sivel:
This PR is an evolution of #1518

It allows for a list of responses for a single question that are provided in order.

In the event that the question is asked more times than we have answers, a proper failure is reported.

Doc updates included.

### DIFF
--- a/commands/expect.py
+++ b/commands/expect.py
@@ -56,7 +56,9 @@ options:
     required: false
   responses:
     description:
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @sivel. Since you are a maintainer of this module, we are marking this PR for inclusion.

quinot:
thanks @sivel, I verified that your version works for me as well.

PR #1526, opened 2016-01-20
Add Infiniband support to nmcli module

ke3vin:
This change adds support for Infiniband interfaces to the nmcli module.  It is basically a copy of the ethernet interface type with a few additional parameters.


ke3vin:
Let me resubmit this - found some other issues

PR #1532, opened 2016-01-21
Add custom parameter for a sensu_check

mbobakov:


gregdek:
Thanks @mbobakov. @andsens please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

monitoring/sensu_check.py
```diff
@@ -149,6 +149,11 @@
       - The low threshhold for flap detection
     required: false
     default: null
+  custom:
+    description:
+      - JSON mixin to add to the configuration
```
andsens: You should specify whether other existing keys are overwritten (you throw an error, I know, but the user should be able to know this by reading the docs).

monitoring/sensu_check.py
```diff
@@ -183,6 +188,7 @@
         # Let snippet from module_utils/basic.py return a proper error in this case
         pass
 
+import ast
```
andsens: ~~Why not use the json module that was imported 4 lines prior?~~

Ignore this, just use the `dict` type for the ansible param.

monitoring/sensu_check.py
```diff
@@ -257,6 +263,36 @@ def sensu_check(module, path, name, state='present', backup=False):
                     changed = True
                     reasons.append('`{opt}\' was removed'.format(opt=opt))
 
+        if module.params['custom']:
+          # Convert to json
+          try:
+            custom_params = ast.literal_eval(module.params['custom'])
```
andsens: ~~The python `dict` notation is not the same as the JSON notation. You should really use the JSON module here.~~

Ignore this, just use the `dict` type for the ansible param.

monitoring/sensu_check.py
```diff
@@ -257,6 +263,36 @@ def sensu_check(module, path, name, state='present', backup=False):
...
+          except:
+            msg = 'Module parameter "custom" contains invalid JSON. Example: custom=\'{"JSON": "here"}\''
+            module.fail_json(msg=msg)
+
+          overwrited_fields = set(custom_params.keys()) & set(simple_opts + ['type','subdue'])
```
andsens: You're missing `subdue_end` and `subdue_begin` (`subdue` is the name of the param in sensu and is an object with `begin` and `end`, in ansible we simply split the two subelements up into two options)

monitoring/sensu_check.py
```diff
@@ -257,6 +263,36 @@ def sensu_check(module, path, name, state='present', backup=False):
...
+          if overwrited_fields:
+            msg = 'You can\'t overwriting standard module parameters via "custom". You are trying overwrite: {of}'.format(of=list(overwrited_fields))
+            module.fail_json(msg=msg)
+
+          for k,v in custom_params.items():
```
andsens, line 274: Use `opt, value` like above, for consistency.

monitoring/sensu_check.py
```diff
@@ -257,6 +263,36 @@ def sensu_check(module, path, name, state='present', backup=False):
...
+            if k in config['checks'][name].keys():
```
andsens: The `in` operator checks the keys when given a dict, so there\s no need for `.keys()`

monitoring/sensu_check.py
```diff
@@ -320,6 +356,7 @@ def main():
                 'aggregate':    {'type': 'bool'},
                 'low_flap_threshold':  {'type': 'int'},
                 'high_flap_threshold': {'type': 'int'},
+                'custom':   {'type': 'str'},
```
andsens: You can just use `dict` as the type here. In fact, that would make the code a lot simpler. Ansible will then make sure that your param is properly formatted :-)

andsens:
Great job @mbobakov! Kudos for following the style of the existing code. I added a few comments. If you could fix those, we can get this merged :-)

needs_revision

monitoring/sensu_check.py
```diff
@@ -257,6 +263,36 @@ def sensu_check(module, path, name, state='present', backup=False):
...
+                changed = True
+                reasons.append('`custom param {k}\' was changed'.format(k=k))
+            else:
+              changed = True
+              reasons.append('`custom param {k}\' was added'.format(k=k))
+            check[k] = v
+          simple_opts += custom_params.keys()
+
+        # Remove obsolete custom params
+        for opt in set(config['checks'][name].keys()) - set(simple_opts + ['type','subdue']):
```
andsens: Don't forget `subdue_begin` and `subdue_end` here as well.

mbobakov:
Hello.
Dictionary, exactly! Thank you for a tip. I just did not think about it.
I have fixed a code, can you check it again?

andsens:
Ah :-) That made the code a lot simpler. 
Looks great!

shipit

gregdek:
Thanks again to @mbobakov for this PR, and thanks @andsens for reviewing. Marking for inclusion.

PR #1548, opened 2016-01-24
Do not fail when action is delete and volume is not present

ghost:
This prevents failing when a playbook describes a volume deletion and
is launched more that once.

Without this fix, if you run the playbook a second time, it will fail.

### DIFF
--- a/system/gluster_volume.py
+++ b/system/gluster_volume.py
@@ -433,7 +433,7 @@ def main():
         else:
             module.fail_json(msg='failed to create volume %s' % volume_name)
 
-    if volume_name not in volumes:
+    if action != 'delete' and volume_name not in volumes:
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @vvanholl. @rosmo please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

rosmo:
shipit

gregdek:
Thanks again to @vvanholl for this PR, and thanks @rosmo for reviewing. Marking for inclusion.

turmalinow:
Actually this doesn't prevent from fail as action for deleting volume is 'absent'

PR #1565, opened 2016-01-28
Ensure that port_path is split into at most 2 components.

mjschultz:
substitutes https://github.com/ansible/ansible-modules-extras/pull/1420


gregdek:
Thanks @mjschultz. Since you are a maintainer of this module, we are marking this PR for inclusion.

PR #1577, opened 2016-01-31
fix typo

bmillemathias:
conainerization => containerization


bcoca:
no fun removing stealth movie references :-(

Issue #1573, opened 2016-01-29
sns_topics tasks fails with "TypeError: 'NoneType' object is not iterable"


gb-klein:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 2.0.0.2
##### Ansible Configuration:

only changed 'host_key_checking = False' in /etc/ansible/ansible.cfg
##### Environment:

DISTRIB_ID=LinuxMint
DISTRIB_RELEASE=17.1
DISTRIB_CODENAME=rebecca
DISTRIB_DESCRIPTION="Linux Mint 17.1 Rebecca"
NAME="Ubuntu"
VERSION="14.04.3 LTS, Trusty Tahr"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 14.04.3 LTS"
VERSION_ID="14.04"
##### Summary:

Simple playbook with sns_topic task fails with message: "TypeError: 'NoneType' object is not iterable"
##### Steps To Reproduce:

define playbook setup-sns.yml with content:

```
- name: Setup required SNS topics and subscriptions for our lambda functions
  hosts: localhost
  connection: local
  tasks:  
    - name: setup simpleSnsTopic
      sns_topic:
        name: "simpleSnsTopic"
        state: present
        region: eu-central-1
```

execute with _ansible-playbook -vvv setup-sns.yml_

if i however append a subscriptions defintion

```
subscriptions:
          - endpoint: "my_email_address@example.com"
            protocol: "email"
```

the playbook succeeds. But that is not a workaround for us as we have sns topics without subscriptions
##### Expected Results:

simpleSnsTopic should be created without any error and tasks should not fail
##### Actual Results:

simpleSnsTopic is created and task fails with error message:

> An exception occurred during task execution. The full traceback is:
> Traceback (most recent call last):
>   File "/home/user/.ansible/tmp/ansible-tmp-1454082395.19-227895112884618/sns_topic", line 2421, in <module>
>     main()
>   File "/home/user/.ansible/tmp/ansible-tmp-1454082395.19-227895112884618/sns_topic", line 234, in main
>     subscriptions]
> TypeError: 'NoneType' object is not iterable
> fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "invocation": {"module_name": "sns_topic"}, "parsed": false}

wimnat:
@f24-gklein I've submitted a PR above to address this issue. Let me know if it solves your problem.

[...]

PR #1579, opened 2016-02-01
Allow SNS topics to be created without subscriptions.

wimnat:
Fix for https://github.com/ansible/ansible-modules-extras/issues/1573

### DIFF
--- a/cloud/amazon/sns_topic.py
+++ b/cloud/amazon/sns_topic.py
@@ -132,8 +132,8 @@ import json
 import re
 
 try:
-    import boto
     import boto.sns
+    from boto.exception import BotoServerError
     HAS_BOTO = True
 except ImportError:
     HAS_BOTO = False
@@ -146,12 +146,15 @@ def canonicalize_endpoint(protocol, endpoint):
     return endpoint
 
 
-
-def get_all_topics(connection):
+def get_all_topics(connection, module):
     next_token = None
     topics = []
     while True:
-        response = connection.get_all_topics(next_token)
+        try:
+            response = connection.get_all_topics(next_token)
+        except BotoServerError, e:
+            module.fail_json(msg=e.message)
+
         topics.extend(response['ListTopicsResponse']['ListTopicsResult']['Topics'])
         next_token = \
                 response['ListTopicsResponse']['ListTopicsResult']['NextToken']
@@ -160,14 +163,15 @@ def get_all_topics(connection):
     return [t['TopicArn'] for t in topics]
 
 
-def arn_topic_lookup(connection, short_topic):
+def arn_topic_lookup(connection, short_topic, module):
     # topic names cannot have colons, so this captures the full topic name
-    all_topics = get_all_topics(connection)
+    all_topics = get_all_topics(connection, module)
     lookup_topic = ':%s' % short_topic
     for topic in all_topics:
         if topic.endswith(lookup_topic):
             return topic
     return None
+
 
 def main():
     argument_spec = ec2_argument_spec()
@@ -179,7 +183,7 @@ def main():
             display_name=dict(type='str', required=False),
             policy=dict(type='dict', required=False),
             delivery_policy=dict(type='dict', required=False),
-            subscriptions=dict(type='list', required=False),
+            subscriptions=dict(default=[], type='list', required=False),
             purge_subscriptions=dict(type='bool', default=True),
         )
     )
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @wimnat. @joelthompson please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

cloud/amazon/sns_topic.py
```diff
@@ -95,11 +94,15 @@ def canonicalize_endpoint(protocol, endpoint):
...
 
-def get_all_topics(connection):
+def get_all_topics(connection, module):
     next_token = None
     topics = []
     while True:
-        response = connection.get_all_topics(next_token)
+        try:
+            response = connection.get_all_topics(next_token)
+        except BotoServerError as e:
```
joelthompson: This syntax is unfortunately not Python 2.4 compatible, and while Ansible doesn't check the cloud directory for Python 2.4 compatibility, I'm still being asked to keep my contributions to cloud/ compatible with Python 2.4 (see #1429). `sys.exc_info()[1]` is, I believe, the best way to do it which has the broadest Python compatibility (2.4 through 3.x).

cloud/amazon/sns_topic.py
```diff
@@ -228,9 +252,14 @@ def main():
         if not next_token:
             break
 
-    desired_subscriptions = [(sub['protocol'],
```
joelthompson, line 286: Wouldn't it be simpler to just set the default for subscriptions to be [] instead of None? Then you wouldn't need to check for `subscriptions is not None`.

joelthompson:
Thanks @wimnat for this and for cleaning it up! :) Unfortunately, this isn't compatible with Python 2.4, can you tweak the syntax as suggested?

Also, wouldn't a simpler approach be to just set the default for `subscriptions` to `[]` instead of `None`? E.g.,

``` python
foo = None
[x for x in foo]
```

gives "TypeError: 'NoneType' object is not iterable" while

``` python
foo = []
[x for x in foo]
```

returns `[]` as desired.

joelthompson:
needs_revision

cloud/amazon/sns_topic.py
wimnat: That is interesting. I've used 'as' for exception handling all over my cloud modules which have been merged without any comments from the core team.

I don't see the point in writing a module as 2.4 compatible when we know that it'll only ever work with 2.6+ because of dependencies such as boto.

cloud/amazon/sns_topic.py
wimnat, line 286: Yes. Yes it would. This is why i shouldn't write my PRs while also watching Game of Thrones.

gregdek:
Thanks @wimnat for this PR. A maintainer of this module has asked for revisions to this PR. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

cloud/amazon/sns_topic.py
joelthompson: I think @sivel's explanation in #1429 makes sense -- even if the module doesn't need to be Python 2.4 compatible, I agree with the desire to maintain consistent style throughout the codebase.

wimnat:
@joelthompson added 2.4 exception handling

joelthompson:
Thanks @wimnat !

If I'm reading it correctly, it looks to me like the CI error is due to unrelated changes in a different module, so :shipit: (If I'm missing something, let me know.)

I'll follow up with a separate PR to improve the documentation to conform to the current standards, but I don't think this bugfix PR needs to wait for that.

gregdek:
@wimnat Thanks for submitting this pull request. Unfortunately, it appears that the underlying code has changed since your submission, and this PR now needs to be rebased. 

If you believe that this PR still applies, please rebase and we will put it into our community review process as soon as possible. 

Once you have rebased this PR, please add a comment with the text "ready_for_review" and we will continue with the review process.  Thanks!

wimnat:
ready_for_review

gregdek:
Thanks @wimnat. @joelthompson please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

gregdek:
Thanks @wimnat. @joelthompson please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

joelthompson:
Thanks @wimnat :)

shipit

gregdek:
Thanks again to @wimnat for this PR, and thanks @joelthompson for reviewing. Marking for inclusion.

resmo:
As there is a workaround for #1573 I don't plan to backport it to stable-2.0.

PR #1582, opened 2016-02-02
Fixes missing validate_certs parameter

caphrim007:
The bigip_api method was changed in the module_utils function definition
to include the validate_certs option but the bigip_virtual_server module
was not updated accordingly. This patch updates the method so that the
error message below is not returned to the user

received exception: bigip_api() takes exactly 4 arguments (3 given)


hpiard:
I tested this fix on my local installation and it is working fine.

abadger:
Merged to devel.  Thanks!

PR #1584, opened 2016-02-02
DOCUMENTATION and RETURN fixes for modules

sivel:
This PR fixes the `DOCUMENTATION` (specifically `options`) for `ec2_vpc_dhcp_options.py` as well as adds an empty `RETURN` to `ec2_vol_facts.py`

Both of these issues were causing failures in travis.


bcoca:
thanks!

Issue #980, opened 2015-09-15
Add a mysql_query module


tersmitten:
I would be nice to have a module to execute queries. For instance:
- From a variable
- From a local file (same as `archive` module)
- From a remote file (same as `archive` module)

robynbergeron:
@tersmitten --  Thanks for submitting this feature idea. Much appreciated!

We encourage all Ansible community members to submit pull requests to satisfy feature requests. If anyone is interested in contributing this as a new module, you can consult the module guidelines (http://docs.ansible.com/ansible/developing_modules.html) for more information.

kiwifellows:
Hi,
I'll be working on this as a new module. 
Proposed approach is outlined below.
**New module called 'mysql_query'**
Features:
- select, insert, delete, update
- returns result set or error if can't connect

Parameters:
- sqlstring (from a variable), database (host, credentials etc...)

Behaviour:
SELECT returns result set
INSERT returns new ID if an auto increment, or returns true if affected rows=1
DELETE returns number of affected rows
UPDATE returns number of affected rows

Out of scope:
Doesn't use a local file or remote file. Reason: The existing mysql module does this quite nicely using the "import" feature

kiwifellows:
Hi @tersmitten and @robynbergeron I've currently started a dev version of this. However it is currently sitting in another repository. Once I get some feedback etc.. then Ill do a PR to put it into ansible-modules-extra repo.
Let me know what you think below:
https://github.com/teemops/ansible-modules/blob/master/database/mysql_query.py

It also needs some work on doco.

Cheers,

Ben
(Currently building up a repo of special ansible modules to include in extras for features such as mysql queries, SQL Server tasks, azure) https://github.com/teemops/ansible-modules

Jmainguy:
@kiwifellows Have a look at https://github.com/ansible/ansible-modules-extras/pull/1593.

PR #1593, opened 2016-02-03
Add mysql_query module

nerzhul:
This module permit to run single query or file to a remote MySQL server.
All syntax is included in the PR :)

### DIFF
--- a/database/mysql/mysql_query.py
+++ b/database/mysql/mysql_query.py
@@ -30,7 +30,7 @@ module: mysql_query
 short_description: Execute SQL requests
 description:
     - Execute a single SQL query or single SQL file on a MySQL server
-version_added: "2.1"
+version_added: "2.2"
 author: "Loic Blot (@nerzhul)"
 options:
     db:
--- a/database/mysql/mysql_query.py
+++ b/database/mysql/mysql_query.py
@@ -1,0 +1,185 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+"""
+
+Ansible module to run SQL requests and files to a running MySQL instance
+(c) 2016, Loic Blot <loic.blot@unix-experience.fr>
+Sponsored by Infopro Digital. http://www.infopro-digital.com/
+Sponsored by E.T.A.I. http://www.etai.fr/
+
+This file is part of Ansible
+
+Ansible is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+Ansible is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+You should have received a copy of the GNU General Public License
+along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+"""
+
+DOCUMENTATION = '''
+---
+module: mysql_query
+
+short_description: Execute SQL requests
+description:
+    - Execute a single SQL query or single SQL file on a MySQL server
+version_added: "2.1"
+author: "Loic Blot (@nerzhul)"
+options:
+    db:
+        description:
+            - MySQL database where queries will be executed
+        default: mysql
+    query:
+        description:
+            - SQL query to execute
+        default: None
+    query_file:
+        description:
+            - SQL filepath to execute
+        default: None
+
+extends_documentation_fragment: mysql
+'''
+EXAMPLES = '''
+# Execute a single query
+- mysql_query: db=concepts query="INSERT INTO `amazing_things` (id,description) VALUES (1,'test');"
+
+# Execute a SQL file
+- mysql_query: db=cars query_file=/usr/share/cars/0001_initdb.sql
+'''
+
+RETURN = '''
+changed:
+    description: If module has executed a query
+    returned: success
+    type: string
+rows_affected:
+    description: number of rows affected by the query or query_file
+    returned: success
+    type: int
+'''
+
+import os
+import warnings
+from re import match
+
+try:
+    import MySQLdb
+except ImportError:
+    mysqldb_found = False
+else:
+    mysqldb_found = True
+
+
+def execute_query(cursor, query, check_mode):
+    # execute a specific query
+    try:
+        cursor.execute(query)
+        if check_mode:
+            cursor.connection.rollback()
+        else:
+            cursor.connection.commit()
+        result = (True, cursor.rowcount)
+    except Exception, e:
+        result = (str(e), 0)
+    return result
+
+
+def execute_query_file(cursor, query_file, check_mode):
+    try:
+        row_affected = 0
+        for line in open(query_file):
+            # Ignore empty lines and comments
+            if re.match("^\s*$", line) or re.match("^\s*-{2,}", line):
+                continue
+            cursor.execute(line)
+            row_affected += cursor.rowcount
+        if check_mode:
+            cursor.connection.rollback()
+        else:
+            cursor.connection.commit()
+        result = (True, row_affected)
+    except Exception, e:
+        result = (str(e), 0)
+    return result
+
+
+def main():
+    module = AnsibleModule(
+        argument_spec=dict(
+            login_user=dict(default=None),
+            login_password=dict(default=None, no_log=True),
+            login_host=dict(default="127.0.0.1"),
+            login_port=dict(default="3306", type='int'),
+            login_unix_socket=dict(default=None),
+            query=dict(default=None),
+            query_file=dict(default=None),
+            db=dict(default="mysql"),
+            ssl_cert=dict(default=None),
+            ssl_key=dict(default=None),
+            ssl_ca=dict(default=None),
+            config_file=dict(default="~/.my.cnf")
+        ),
+        supports_check_mode=True
+    )
+    user = module.params["login_user"]
+    password = module.params["login_password"]
+    host = module.params["login_host"]
+    port = module.params["login_port"]
+    ssl_cert = module.params["ssl_cert"]
+    ssl_key = module.params["ssl_key"]
+    ssl_ca = module.params["ssl_ca"]
+    config_file = module.params['config_file']
+    config_file = os.path.expanduser(os.path.expandvars(config_file))
+    db = module.params["db"]
+
+    query = module.params["query"]
+    query_file = module.params["query_file"]
+
+    if query is None and query_file is None:
+        module.fail_json(msg="Cannot run without query or query_file to operate with")
+    if query is not None and query_file is not None:
+        module.fail_json(msg="Cannon run with query and query_file both specified.")
+    if not mysqldb_found:
+        module.fail_json(msg="the python mysqldb module is required")
+    else:
+        warnings.filterwarnings('ignore', category=MySQLdb.Warning)
+
+    try:
+        cursor = mysql_connect(module, user, password, config_file, ssl_cert, ssl_key, ssl_ca, db)
+    except Exception, e:
+        if os.path.exists(config_file):
+            module.fail_json(
+                msg="unable to connect to database, check login_user and login_password "
+                    "are correct or %s has the credentials. Exception message: %s" % (
+                        config_file, e))
+        else:
+            module.fail_json(msg="unable to find %s. Exception message: %s" % (config_file, e))
+
+    if query is not None:
+        result = execute_query(cursor, query, module.check_mode)
+        if result[0] is True:
+            module.exit_json(msg="Query execution succeed", rows_affected=result[1], changed=(result[1] > 0))
+    # if not query, we will execute a query file
+    else:
+        result = execute_query_file(cursor, query_file, module.check_mode)
+        if result[0] is True:
+            module.exit_json(msg="Query file execution succeed", rows_affected=result[1], changed=(result[1] > 0))
+
+    module.fail_json(msg=result[0], changed=(result[1] > 0))
+
+
+# import module snippets
+from ansible.module_utils.basic import *
+from ansible.module_utils.database import *
+from ansible.module_utils.mysql import *
+
+main()


sivel:
This pull request has failed Travis checks due to an incorrect `version_added`.  This should be updated to reflect `2.1`

nerzhul:
Fixed

nerzhul:
@sivel fixed :)

tersmitten:
See #980

tersmitten:
Great addition!

nerzhul:
Thanks :) I hope this will be approved :) i use it in production since today to replace my old shell: mysql < {{ file }} or shell: mysql -e 'FLUSH TABLES' ...

gregdek:
Thanks @nerzhul for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

nerzhul:
@gregdek nobody wants to review it ? :)

tersmitten:
I'll try to find some time to review it

tersmitten:
@Jmainguy Can you also have a look at this?

database/mysql/mysql_query.py
```diff
@@ -0,0 +1,170 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+"""
+
+Ansible module to manage mysql variables
```
Jmainguy: This module is actually to execute query's right?

Jmainguy:
This is a great module to add, code looks good from a quick review. I will run some tests later today for good measure and add a 'shipit' then, nice work.

database/mysql/mysql_query.py
nerzhul: You note that i copy from another mysql module, i fix it now :)

nerzhul:
@Jmainguy fixed !

Jmainguy:
@nerzhul I tested it, and it works very well. (Tested on centos7 with mariadb).

A feature request I would love to see, would be to add --diff support to show the result of the query (for instance if a select was used), and storing that variable so that a separate task could then make use of that value.

Great work man. @gregdek 'shipit'

nerzhul:
Thanks @Jmainguy , this addition could be good, but maybe in a second time, i think :)

Another voter present ? :)

tersmitten:
Somehow I have trouble with INSERT queries:

```
---
- hosts: foo.dev
  tasks:
    - name: mysql query
      mysql_query:
        db: test
        query: >
          CREATE TABLE IF NOT EXISTS `foo` (
            `id` INT(6) UNSIGNED AUTO_INCREMENT PRIMARY KEY,
            `firstname` VARCHAR(30) NOT NULL,
            `lastname` VARCHAR(30) NOT NULL,
            `email` VARCHAR(50),
            `reg_date` TIMESTAMP
          );
      ignore_errors: true

    - name: mysql query
      mysql_query:
        db: test
        query: >
          INSERT INTO `foo` (`id`, `firstname`, `lastname`, `email`, `reg_date`) VALUES (NULL, 'foo', 'bar', 'example.com', NOW());

    - name: mysql query
      mysql_query:
        db: mysql
        query: >
          FLUSH TABLES;

    - name: mysql query file
      mysql_query:
        db: test
        query_file: /tmp/foo.sql
      ignore_errors: true
```

```
# cat /tmp/foo.sql
CREATE TABLE IF NOT EXISTS `bar` LIKE `foo`;
INSERT INTO `bar` (`id`, `firstname`, `lastname`, `email`, `reg_date`) VALUES (NULL, 'foo', 'bar', 'example.com', NOW());
```

Create table statements seem to work, but tables `foo` and `bar` remain empty. Is it possible that a `.commit` is missing?

Also I have trouble with queries that generate a warning (e.g. `CREATE TABLE IF NOT EXISTS foo`). Is that the expected behavior?

```
PLAY ***************************************************************************

ok: [foo.dev]

TASK [mysql query] *************************************************************
fatal: [foo.dev]: FAILED! => {"changed": false, "failed": true, "msg": "Table 'foo' already exists"}
...ignoring

TASK [mysql query] *************************************************************
changed: [foo.dev]

TASK [mysql query] *************************************************************
ok: [foo.dev]

TASK [mysql query file] ********************************************************
fatal: [foo.dev]: FAILED! => {"changed": false, "failed": true, "msg": "(1065, 'Query was empty')"}
...ignoring

foo.dev               : ok=5    changed=1    unreachable=0    failed=0   
```

nerzhul:
@tersmitten it's the standard mysqlDB behaviour, not module behaviour. I will look at your tests to see what could be wrong there to improve error handling

tersmitten:
I know, but in order to let the INSERT (UPDATE et cetera)'s succeed you need to commit the cursor.

In order to ignore warnings you can do this:

```
warnings.filterwarnings('ignore', category=MySQLdb.Warning)
```

[See](https://github.com/nadley/ansible-modules-extras/commit/290254d607268439c573722cd4cae7fe8a985b53)

nerzhul:
@tersmitten you are right, sorry i was tired yesterday :)

I fixed the warning issue

nerzhul:
@tersmitten all issues are now fixed ! thanks for your tests

nerzhul:
And to finish i tested all your playbook and it works like a charm

tersmitten:
Indeed, nice work.

@gregdek 'shipit'

nerzhul:
@gregdek @tersmitten when a module has two shitip, when will it be ship into the main repository ? Before a release ?

Jmainguy:
@gregdek @robynbergeron this is ready for inclusion/core review, can one of you update this.

nerzhul:
Don't forget to tag for inclusion !

gregdek:
Thanks @nerzhul and reviewers. Putting into shipit for final review.

nerzhul:
Thanks @gregdek

abadger:
We discussed this at today's meeting and it turned into a discussion of the very fundamental question of where to draw the line between creating a module versus just using the command or shell module.  We'll have to continue this discussion at next week's meeting as there were a plethora of thoughts on the matter but no consensus.  Some of the view points would favor this module going in and other viewpoints would see this module excluded from the ansible tarball (an open question is how we would like to see it distributed if that is the case).  Some of the ideas that were floated today (note -- some of these ideas were popular, others were unpopular.  Many of them are mutually exclusive.):
- Modules should only be written if the module can be idempotent or support check-mode (neither of which are well supported by the command module)
- Modules are not for returning information
- Modules are for simplifying the user experience.
- If it can be done in the shell module then it shouldn't be its own module
- Modules should not be thin wrappers of cli tools but can wrap more complex interactions
- Users should be able to clearly tell which modules are idempotent and which aren't -- Idempotence is desired for playbooks but non-idempotent modules are still good for new users using /usr/bin/ansible to simplify common ad hoc tasks.
- There needs to be a place for modules which would be kept out by these rules
- The extras repository is a place for modules which do not fit these rules (we should apply these rules to core but we need extras to be a place where people can experiment and come up with things that are needed even though they don't fit the current rules).
- Having a module that can shoot you in the foot (drop tables, insert garbage) is not good.
- Many modules allow you to shoot yourself in the foot.
- Having specific modules for each database flavour instead of a generic one that can apply to many sql databases is the wrong design.
- Query functionality should be a lookup plugin or a facts module.

This is kinda a collective brain dump.  It shows that we have a lot of discussion to do in the next meeting to reconcile the major themes here.

nerzhul:
@abadger thanks for your big reply.
This module is not for querying but for executing. It permit to have a scenario with ansible well played commands when you are putting things to production. The problem with shell is it's a generic crap wrapper for non handled ansible things. Shell permit to do everything, but users may also want (and it's some enterprise goal) drop all those shell craps to have a common ansible interface with modules.

thaumos:
@nerzhul, when you say executing, are you inferring INSERT only, or SELECT as well?  I need clarity for myself. 

Thanks for responding back on this!  We really appreciate your continued involvement in this!

nerzhul:
@thaumos the objective of this module is to do modifications on database, not querying, this is not the purpose of this module.
This module is for orchestration & deployment not for idem potence. If you want idem potence into BDD stats, i recommend writing a mysql_record module to maintain a specific record into DB

thaumos:
@nerzhul then it could be argued the name of the module is misleading... If you intend to only do INSERTS that is.  In my mind, this module enables a user to do a mysql_query (INSERT or SELECT). 😄 

I suggest, and we would love, if you joined our meeting next week to discuss this with us!  

When @abadger mentioned the list of things, as he said, it is not all inclusive, but one of the things mentioned was "idempotency."  Knowing databases, there isn't a way to make any action idempotent.  So that point does and does not apply.  Like he said, it was a collective brain dump.

nerzhul:
@thaumos the problem is INSERT, SELECT, UPDATE, DELETE and more are all queries, it's the database noun it's why i choose it. 
This module is not for doing select-only queries, it's for database modifications.

@abadger as you said check mode is missing, i will read the doc and add it, it could be easy to implement because of MySQL transactions

Jmainguy:
@nerzhul I am thinking check mode would basically just be testing the connection to mysql, and then returning changed if it can connect, but I will leave that for you to decide of course.

@abadger @thaumos I can appreciate the discussion being had, In that big list of info dump, I would say we have many modules that are currently merged that do not follow those conventions. I think ansible deciding on what modules should and should not be is a good idea, however once we make those rules we would have to go back and boot a bunch of modules out, hopefully I can make the next meeting as this discussion is very interesting.

bcoca:
@Jmainguy  there are many different opinions reflected in that dump. 

That existing modules violate newer rulesmostly means that the experience of having those modules are the reason for theses rules. They are in place to stop the problem from growing.
Instead of booting them I prefer to update them to reflect the newer rules, as we have been doing in many cases already.  Some modules cannot be updated so we are trying to create new ones that do follow the rules and provide the same functionality in views of deprecating/removing the offenders.

As for the decision on which modules go in or not, I would argue that those that assume the support burden should have a bigger say. As the expectations are that we support anything shipped, even though that might be unreasonable, no matter how much we point at extras being more community driven.

I'm all for giving a lot of latitude and leeway for extras, but there are things I think we should not ship ourselves and I'm fine with others publishing in galaxy/github/anywhere else where it is clearly and unequivocally unrelated to our code base. If we ever push 'extras' to the point of not being shipped or packaged by Ansible, I'm happy to not give any restrictions.

Jmainguy:
@bcoca Understood, and yeah I definitely think anything shipped under the ansible name is supported by ansible/redhat including extras. There is precedent for non officially supported modules out there, https://github.com/cmprescott/ansible-xml being one I am aware of. 

I think ansible defining these rules on modules will really help with future development.

bcoca:
most are already official http://docs.ansible.com/ansible/developing_modules.html, we are still 'fine tunnig' some of the newer ones.

nerzhul:
@Jmainguy @abadger i added check mode to the module, whichonly connects to database without performing requests and return changed=True. Is this good for you ?

tima:
@nerzhul I would expect the check mode on a module like this to check if an INSERT/UPDATE/DELETE is needed (do these rows exists? are those column values different from the ones I want update?) and report that status without performing that actual operation.

nerzhul:
@tima yes, i only refresh the PR to show it's active :). It's the goal, but i didn't know if it could be possible or not, i need to do more tests

nerzhul:
@tima @Jmainguy i improved the module this morning, as i know feature freeze happened.
- It don't log password anymore
- check mode returns rows affected by executing queries but not commiting them
- query_file ignore empty lines and comment lines

nerzhul:
Then another review is needed and @abadger i hope you will prefer this version :)

nerzhul:
I added a little fix provided by @Farzy to fix Dos lines and comments starting with whitespaces

gregdek:
Thanks @nerzhul for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

nerzhul:
ready_for_review @gregdek

gregdek:
Thanks @nerzhul for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

tima:
gun_meet_foot

This module still does not support idempotency and makes doing tremendous harm to a database way too easy. It does little more than calling command/shell. Shipping this module with Ansible communicates an endorsement I don't think should be made. Like @bcoca  I'm fine with whatever being published anywhere else where it is clearly and unequivocally unrelated to the code base here.

Jmainguy:
There are many modules currently in Core/Extras that do not support idempotency, and allow destruction of the system if used in a certain way.

I am in favor of this module.

tima:
@Jmainguy: Two wrongs don't make a right.

nerzhul:
@tima this module is for doing sql queries in ansible way, idempotency is not possible every time, but it's useful because in a playbook you could have a need to do, for example, a delete on a table everytime your play it. This delete guarentee idempotency over the result, you will always delete all results matching the query.

tima:
I want to clarify my previous comments. 

I understand the desire to automate database table operations and I'm not objecting to or discounting the desire for that functionality. My biggest issue is that this module is so wide open to doing anything/everything you throw at it with no safeties or other logic. Idempotentcy helps to an extent, but is not the only way to that end. It's still a really good idea and I highly encourage it, but its not _the_ issue.

So I understand the desire to drop (delete) a table as part of a playbook run. I am not objecting to that need  -- it's that there should be a module that specifically handles table add/drops and only that, not any SQL statement one sends it.

nerzhul:
@tima i agree with idempotency and i agree that should be used everytime it's possible, but for this goal i don't think it's possible, because a database is not really an idempotent system.

For a single table yes we could have a thing, for a single record it's possible too, but for a SQL incoming from external source or which should be played at each playbook run and doesn't give exactly same situation on two runs (for example SQL with updated datas to run on a table every day)

ryansb:
Strictly a database can be used in an idempotent way, but it requires a certain data model that I don't know if we can guarantee. For example:

```
INSERT INTO `foo` (`id`, `firstname`, `lastname`, `email`, `reg_date`) 
    VALUES (NULL, 'foo', 'bar', 'example.com', NOW())
    ON DUPLICATE KEY UPDATE id=id;
```

That will insert the record, unless the `id` column conflicts in which case it will update `id=id` which will be optimized out by the MySQL query engine. Of course:
1. This only works if `id` has a primary key constraint
2. And ON DUPLICATE is supported
3. And the user's data model makes sense for this use case

PR #1598, opened 2016-02-03
Fix enable in win_firewall_rule module

ilpianista:
The `enable` parameter is a string, then compare it with strings. I suppose this
never worked before.

While I'm on it, I also fixed a typo in the example and added a required
parameter.


ilpianista:
Damn. Another commit ended in this PR. I'll split them

PR #1606, opened 2016-02-04
win_environment: never unset a variable when state=present

ilpianista:
`value` is an optional parameter, then I suppose I can use this module
to check if a variable is present or not by doing:
`win_environment: name=MYENV state=present`
Unfortunately the above will unset the MYENV variable. I fixed this by
setting the variable only when a `value` parameter is used.

While I was on it I also used the Get-Attr function to read parameters.


gregdek:
Thanks @ascarpino. @jhawkesworth please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

jhawkesworth:
Thanks for this, and thanks for updating the parameter validatin to use Get-Attr too.

It wasn't really my intention to allow this - value is only optional so that it wouldn't be required in order to remove an environment variable when running in 'state=absent' mode.

Personally I think I wouldn't use this as to me, its not explicit enough to just say 'make sure an env var of this name exists' without specifying what the value should be.  But this change doesn't stop me from being explicit about values so I don't want to stop it from being merged.

shipit

gregdek:
Thanks again to @ilpianista for this PR, and thanks @jhawkesworth for reviewing. Marking for inclusion.

bcoca:
it makes more sense to just add environment vars to the windows fact gathering, this seems like an odd hack to push querying into the present state. It seems to conflict with the ability to 'blank out a value'.

cc @nitzmahone @trondhindenes

jhawkesworth:
@bcoca makes a good point here.  You wouldn't be able to blank out a value.

2.1 will have environment variables in facts - already merged under https://github.com/ansible/ansible-modules-core/pull/2692

Sorry @ilpianista I know I said that this should be shipped but I realize now this would be making things unnecessarily different from the way the rest of ansible operates, so I no longer think this should be merged.

ilpianista:
Sure. No problem, let's wait 2.1!

PR #1619, opened 2016-02-05
amazon/ec2_vpc_igw: Allow internet gateways to be tagged

gechr:
This PR makes it possible to tag Internet Gateways, heavily based on [cloud/amazon/ec2_vpc_subnet.py](https://github.com/ansible/ansible-modules-extras/blob/devel/cloud/amazon/ec2_vpc_subnet.py).

Would be a good idea to factor out the common code (e.g. tagging) in the Amazon modules in future to avoid duplication, but that's probably best picked up as a separate PR.

### DIFF
--- a/cloud/amazon/ec2_vpc_igw.py
+++ b/cloud/amazon/ec2_vpc_igw.py
@@ -27,6 +27,13 @@ options:
       - The VPC ID for the VPC in which to manage the Internet Gateway.
     required: true
     default: null
+  tags:
+    description:
+      - "A dict of tags to apply to the internet gateway. Any tags currently applied to the internet gateway and not present here will be removed."
+    required: false
+    default: null
+    aliases: [ 'resource_tags' ]
+    version_added: "2.2"
   state:
     description:
       - Create or terminate the IGW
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @gechr. @erydo please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

gechr:
Oops, accidentally hit close, re-opening!

aioue:
+1

brandond:
+1, with my PR to fix state: absent

gechr:
@brandond Nice catch, merged. Would have been more correct to not pass `tags` in the first place, since the parameter isn't used, but I'm not too fussed either way!

brandond:
@gechr yeah, I considered pulling the argument out of the call, but I kind of liked the idea of present and absent having the same signature, even if it's not used at the moment. Also, looks like the build is now failing since 'tags' requires version_added: "2.2"

gechr:
@brandond Fixed & squashed everything into a single commit.

aturetta:
+1

aioue:
Any movement with this?

aioue:
@gechr did this get raised against ansible/ansible?

PR #1640, opened 2016-02-09
Updated ec2_vpc_dhcp_options

andynelson:
This PR provides updates to ec2_vpc_dhcp_options module by  @joelthompson

Additional functionality is:
1. Remove specified dhcp_options with state: absent
2. Allow existing dhcp_options to be referenced by aws id (for associating, tagging or removal)
3. Match existing dhcp_options by tags/options (for associating or removal)
4. Tag dhcp options when created if tags are supplied
5. Return the dhcp_options_id when options are created or matched

There's a lot of refactoring but the only change to existing behaviour is the removal of the default on 'netbios_node_type' (this was causing options not to be matched)

### DIFF
--- a/cloud/amazon/ec2_vpc_dhcp_options.py
+++ b/cloud/amazon/ec2_vpc_dhcp_options.py
@@ -16,13 +16,21 @@
 DOCUMENTATION = """
 ---
 module: ec2_vpc_dhcp_options
-short_description: Ensures the DHCP options for the given VPC match what's
+short_description: Manages DHCP Options, and can ensure the DHCP options for the given VPC match what's
   requested
 description:
-  - Converges the DHCP option set for the given VPC to the variables requested.
+  - This module removes, or creates DHCP option sets, and can associate them to a VPC.
+    Optionally, a new DHCP Options set can be created that converges a VPC's existing
+    DHCP option set with values provided.
+    When dhcp_options_id is provided, the module will
+    1. remove (with state='absent')
+    2. ensure tags are applied (if state='present' and tags are provided
+    3. attach it to a VPC (if state='present' and a vpc_id is provided.
     If any of the optional values are missing, they will either be treated
-    as a no-op (i.e., inherit what already exists for the VPC) or a purge of
-    existing options. Most of the options should be self-explanatory.
+    as a no-op (i.e., inherit what already exists for the VPC)
+    To remove existing options while inheriting, supply an empty value
+    (e.g. set ntp_servers to [] if you want to remove them from the VPC's options)
+    Most of the options should be self-explanatory.
 author: "Joel Thompson (@joelthompson)"
 version_added: 2.1
 options:
@@ -30,37 +38,40 @@ options:
     description:
       - The domain name to set in the DHCP option sets
     required: false
-    default: ""
+    default: None
   dns_servers:
     description:
       - A list of hosts to set the DNS servers for the VPC to. (Should be a
         list of IP addresses rather than host names.)
     required: false
-    default: []
+    default: None
   ntp_servers:
     description:
       - List of hosts to advertise as NTP servers for the VPC.
     required: false
-    default: []
+    default: None
   netbios_name_servers:
     description:
       - List of hosts to advertise as NetBIOS servers.
     required: false
-    default: []
+    default: None
   netbios_node_type:
     description:
-      - NetBIOS node type to advertise in the DHCP options. The
-        default is 2, per AWS recommendation
+      - NetBIOS node type to advertise in the DHCP options.
+        The AWS recommendation is to use 2 (when using netbios name services)
         http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_DHCP_Options.html
     required: false
-    default: 2
+    default: None
   vpc_id:
     description:
-      - VPC ID to associate with the requested DHCP option set
-    required: true
+      - VPC ID to associate with the requested DHCP option set.
+        If no vpc id is provided, and no matching option set is found then a new
+        DHCP option set is created.
+    required: false
+    default: None
   delete_old:
     description:
-      - Whether to delete the old VPC DHCP option set when creating a new one.
+      - Whether to delete the old VPC DHCP option set when associating a new one.
         This is primarily useful for debugging/development purposes when you
         want to quickly roll back to the old option set. Note that this setting
         will be ignored, and the old DHCP option set will be preserved, if it
@@ -74,6 +85,28 @@ options:
         reset them to be empty.
     required: false
     default: false
+  tags:
+    description:
+      - Tags to be applied to a VPC options set if a new one is created, or
+        if the resource_id is provided. (options must match)
+    required: False
+    default: None
+    aliases: [ 'resource_tags']
+  dhcp_options_id:
+    description:
+      - The resource_id of an existing DHCP options set.
+        If this is specified, then it will override other settings, except tags
+        (which will be updated to match) 
+    required: False
+    default: None
+  state:
+    description:
+      - create/assign or remove the DHCP options.
+        If state is set to absent, then a DHCP options set matched either
+        by id, or tags and options will be removed if possible.
+    required: False
+    default: present
+    choices: [ 'absent', 'present' ]
 extends_documentation_fragment: aws
 requirements:
     - boto
@@ -81,19 +114,26 @@ requirements:
 
 RETURN = """
 new_options:
-  description: The new DHCP options associated with your VPC
-  returned: changed
-  type: dict
-  sample:
-    domain-name-servers:
-      - 10.0.0.1
-      - 10.0.1.1
-    netbois-name-servers:
-      - 10.0.0.1
-      - 10.0.1.1
-    ntp-servers: None
-    netbios-node-type: 2
-    domain-name: "my.example.com"
+    description: The DHCP options created, associated or found
+    returned: when appropriate
+    type: dict
+    sample:
+      domain-name-servers:
+        - 10.0.0.1
+        - 10.0.1.1
+      netbois-name-servers:
+        - 10.0.0.1
+        - 10.0.1.1
+      netbios-node-type: 2
+      domain-name: "my.example.com"
+dhcp_options_id:
+    description: The aws resource id of the primary DCHP options set created, found or removed
+    type: string
+    returned: when available
+changed:
+    description: Whether the dhcp options were changed
+    type: bool
+    returned: always
 """
 
 EXAMPLES = """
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @andynelson. @joelthompson please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

cloud/amazon/ec2_vpc_dhcp_options.py
```diff
@@ -75,6 +77,26 @@
         reset them to be empty.
     required: false
     default: false
+  tags:
+    description:
+      - Tags to be applied to a VPC options set if a new one is created, or
+        if the resource_id is provided. (options must match)
+    required: False
+    aliases: [ 'resource_tags']
```
joelthompson, line 94: Per the module checklist, "If required is false you need to document default, even if the default is ‘null’ (which is the default if no parameter is supplied). Make sure default parameter in docs matches default parameter in code."

Can you add the default here as well as for the dhcp_options_id option?

cloud/amazon/ec2_vpc_dhcp_options.py
```diff
@@ -128,105 +157,204 @@
...
+        if to_delete and not add_only:
+            vpc_conn.delete_tags(resource_id, to_delete, dry_run=check_mode)
+
+        to_add = dict((k, tags[k]) for k in tags if k not in cur_tags)
+        if to_add:
+            vpc_conn.create_tags(resource_id, to_add, dry_run=check_mode)
+
+        latest_tags = get_resource_tags(vpc_conn, resource_id)
+        return {'changed': True, 'tags': latest_tags}
+    except EC2ResponseError as e:
```
joelthompson, line 227: Minor style note, the Ansible team would prefer to avoid using the "as" syntax.

cloud/amazon/ec2_vpc_dhcp_options.py
```diff
@@ -128,105 +157,204 @@
...
-        dns_servers=dict(type='list', default=[]),
-        ntp_servers=dict(type='list', default=[]),
-        netbios_name_servers=dict(type='list', default=[]),
-        netbios_node_type=dict(type='int', default=2),
-        vpc_id=dict(type='str', required=True),
+        dhcp_options_id=dict(type='str', default=None),
+        domain_name=dict(type='str', default=None),
+        dns_servers=dict(type='list', default=None),
+        ntp_servers=dict(type='list', default=None),
+        netbios_name_servers=dict(type='list', default=None),
```
joelthompson, line 269: Since you're changing the default from `[]` to `None`, can you update the `DOCUMENTATION` string to reflect the updated default, for dns_servers, ntp_servers, and netbios_name_servers?

cloud/amazon/ec2_vpc_dhcp_options.py
```diff
@@ -128,105 +157,204 @@
...
+
+    # If we were given a vpc_id, then attach the options we now have to that before we finish
+    if params['vpc_id'] and not module.check_mode:
         connection.associate_dhcp_options(dhcp_option.id, params['vpc_id'])
-        if params['delete_old'] and existing_options:
-            other_vpcs = _get_vpcs_by_dhcp_options(existing_options.id, connection)
-            if len(other_vpcs) == 0 or (len(other_vpcs) == 1 and other_vpcs[0].id == params['vpc_id']):
-                connection.delete_dhcp_options(existing_options.id)
+        # and remove old ones if that was requested
+        if params['delete_old'] and params['vpc_id'] and existing_options:
```
joelthompson: The check for `params['vpc_id']` is redundant here since it's in the parent if clause.

cloud/amazon/ec2_vpc_dhcp_options.py
```diff
@@ -128,105 +157,204 @@
...
+        dhcp_option = supplied_options[0]
+        if params['state'] != 'absent' and options['tags']:
+            ensure_tags(connection, options['tags'], False, module.check_mode)
+
+    # Now we have the dhcp options set, let's do the necessary
+
+    # if we found options we were asked to remove then try to do so
+    if found and params['state'] == 'absent' and not module.check_mode:
+        changed = remove_dhcp_options_by_id(connection, dhcp_option.id)
+        module.exit_json(changed=changed, new_options={})
```
joelthompson, line 343: I think your handling of module.check_mode here is off. You should do something like:

``` python
if found and params['state'] == 'absent':
    if not module.check_mode:
        changed = remove_dhcp_options_by_id(connection, dhcp_option.id)
    module.exit_json(changed=changed, new_options={})
```

cloud/amazon/ec2_vpc_dhcp_options.py
```diff
@@ -128,105 +157,204 @@
...
+            dhcp_option = existing_options
+            module.exit_json(changed=changed, new_options=new_options, dhcp_options_id=existing_options.id)
+
+        # If no vpc_id was given, or the options don't match then look for an existing set using tags
+        found, dhcp_option = match_dhcp_options(connection, params['tags'], new_options)
+
+    # Now let's cover the case where there are existing options that we were told about by id
+    # If a dhcp_options_id was supplied we don't look at options inside, just set tags (if given)
+    else:
+        supplied_options = connection.get_all_dhcp_options(dhcp_options_ids=params['dhcp_options_id'])
```
joelthompson: This will fail and raise an exception if `dhcp_options_id` either wasn't specified or is invalid. It's probably right to fail if an invalid `dhcp_options_id` is passed in, but that needs to be done gracefully. You probably also don't want to fail when no dhcp_options_id is passed in.

cloud/amazon/ec2_vpc_dhcp_options.py
```diff
@@ -128,105 +157,204 @@
...
+                if existing_options.options.get(option) and (not new_options[option] or [''] == new_options[option]):
+                    new_options[option] = existing_options.options.get(option)
+
+        # Do the vpc's dhcp options already match what we're asked for? if so we are done
+        if existing_options and new_options == existing_options.options:
+            found = True
+            dhcp_option = existing_options
+            module.exit_json(changed=changed, new_options=new_options, dhcp_options_id=existing_options.id)
+
+        # If no vpc_id was given, or the options don't match then look for an existing set using tags
```
joelthompson, line 320: With regards to the comment, a `vpc_id` has been given, since you're inside the `if` block.

cloud/amazon/ec2_vpc_dhcp_options.py
```diff
@@ -128,105 +157,204 @@
...
+        # if we've been asked to inherit existing options, do that now
+        if params['inherit_existing']:
+            for option in [ 'domain-name-servers', 'netbios-name-servers', 'ntp-servers', 'domain-name', 'netbios-node-type']:
+                if existing_options.options.get(option) and (not new_options[option] or [''] == new_options[option]):
+                    new_options[option] = existing_options.options.get(option)
+
+        # Do the vpc's dhcp options already match what we're asked for? if so we are done
+        if existing_options and new_options == existing_options.options:
+            found = True
+            dhcp_option = existing_options
```
joelthompson: These two assignments are unneeded as the variables are never used before the exit_json.

cloud/amazon/ec2_vpc_dhcp_options.py
```diff
@@ -128,105 +157,204 @@
...
+        if params['dns_servers'] != None:
+            new_options['domain-name-servers'] = params['dns_servers']
+        if params['netbios_name_servers'] != None:
+            new_options['netbios-name-servers'] = params['netbios_name_servers']
+        if params['ntp_servers'] != None:
+            new_options['ntp-servers'] = params['ntp_servers']
+        if params['domain_name'] != None:
+            # needs to be a list for comparison with boto objects later
+            new_options['domain-name'] = [ params['domain_name'] ]
+        if params['netbios_node_type'] != None:
```
joelthompson, line 303: I don't think the behavior is _quite_ right here, but I'd love to hear feedback about your use cases.

In particular, what happens if a user explicitly passes in an empty list for something like dns_servers? `[] != None` will return `True` and then `new_options['domain-name-servers']` will get set to `[]` which will then cause the `new_options == existing_options.options` to potentially return a false negative.

Do you have any use cases where you would want different behavior for `None` and `[]`? I can't think of any reason why that different behavior would be needed, so I'd love to hear your thoughts on this!

If you don't need that different behavior, I'd suggest simplifying these tests and just doing something like:

``` python
if params['dns_servers']:
    new_options['domain-name-servers'] = params['dns_servers']
```

cloud/amazon/ec2_vpc_dhcp_options.py
```diff
@@ -128,105 +157,204 @@
...
-        inherit_existing=dict(type='bool', default=False)
+        inherit_existing=dict(type='bool', default=False),
+        tags=dict(type='dict', default=None, aliases=['resource_tags']),
+        state=dict(type='str', default='present', choices=['present', 'absent'])
         )
     )
 
     module = AnsibleModule(argument_spec=argument_spec, supports_check_mode=True)
     params = module.params
+    found = False
```
joelthompson, line 281: I think your use of the `found` variable is unnecessary. Instead of checking `found` can't you just check `dhcp_option`?

cloud/amazon/ec2_vpc_dhcp_options.py
```diff
@@ -128,105 +157,204 @@
...
     return dhcp_options[0]
 
+def match_dhcp_options(vpc_conn, tags=None, options=None):
+    """
+    Finds a DHCP Options object that optionally matches the tags and options provided
+    """
+    dhcp_options = vpc_conn.get_all_dhcp_options()
+    for dopts in dhcp_options:
+        if tags and get_resource_tags(vpc_conn, dopts.id) == tags:
+            if options and dopts.options == options:
```
joelthompson: Your comment says that it "optionally matches the tags and options provided." However, your code actually matches on both of these, there's no optionality. I think your behavior is right and the comment is off, but I'm not 100% sure. What are your thoughts?

Also, note that multiple DHCP option sets can match both the tags and options. I'm not sure what the right behavior is in this circumstance -- I think the right behavior is to error out since you don't know which one to use. Right now, you're just picking the first one you come across, and that feels like it could lead to many frustrating bugs down the road.

Can you update this to check if multiple match, and, if so, error out?

cloud/amazon/ec2_vpc_dhcp_options.py
```diff
@@ -128,105 +157,204 @@
...
+        to_delete = dict((k, cur_tags[k]) for k in cur_tags if k not in tags)
+        if to_delete and not add_only:
+            vpc_conn.delete_tags(resource_id, to_delete, dry_run=check_mode)
+
+        to_add = dict((k, tags[k]) for k in tags if k not in cur_tags)
+        if to_add:
+            vpc_conn.create_tags(resource_id, to_add, dry_run=check_mode)
+
+        latest_tags = get_resource_tags(vpc_conn, resource_id)
+        return {'changed': True, 'tags': latest_tags}
```
joelthompson, line 226: Both places which call ensure_tags ignore the return value, is this needed?

joelthompson:
Thanks @andynelson for taking the time to put this together! I've added some specific in-line comments, but more broadly, I'm still a bit confused about the semantic changes that you're proposing -- apologies if I'm just being a bit dense! Some specific questions:
1. What happens if I specify both a vpc_id and a dhcp_options_id? A reasonable argument could be made that I want to ensure that dhcp_options_id is attached to vpc_id (e.g., if I create a DHCP options set and attach to a given VPC in one play, and then want to attach the same option set to a different VPC in another play). I think what will end up happening is:
   1. All requested options will be ignored
   2. If inherit_existing is False, then this will turn in to a no-op, ignoring any requested tags (which I think is incorrect behavior)
   3. If inherit_existing is True, then it will try to create an empty DHCP options set (because all requested options are ignored) and that will error out later in an ungraceful fashion (which is a bug).
2. What if I specify neither vpc_id nor dhcp_options_id? I think it will error out ungracefully (which is a bug).

Here's my understanding of what I think you intend and what I think the right semantics are:
1. If dhcp_options_id is specified:
   1. Error if any DHCP options are requested (since they would otherwise be ignored -- I'd rather yell loudly instead of silently ignore them)
   2. If dhcp_options_id does not exist:
      1. If state == absent, exit with success, changed=False
      2. If state != absent, then error
   3. If dhcp_options_id does exist:
      1. If state == absent, then delete if it is not associated with any VPCs, else no-op.
      2. If state == present:
         1. Ensure it has the right tags
         2. Associate with the vpc_id if requested
2. If dhcp_options_id is NOT specified:
   1. Search all existing DHCP options to see if any match **both** the tags **and** requested options.
      1. If no matches, then start with a candidate option set of what is requested.
      2. If >1 match, then error out
      3. If == 1 match, then use that.
   2. If state == absent, then: if associated with no VPCs, delete; else, no-op.
   3. If vpc_id is specified and inherit_existing is True, then update the candidate option set with the options that are attached to the VPC.
   4. Create DHCP option set if it doesn't exist
   5. Ensure DHCP option set has desired tags
   6. If vpc_id is specified, ensure the DHCP option set is associated with the VPC

I don't think your code properly implements these semantics. However, this is quite complex, so I'm probably missing something or misunderstanding something or just over-complicating things! Does this match what you're trying to accomplish? (And, to be clear, I'm not saying that all of these have to be implemented in this PR, I personally think it's OK to document the use cases not yet supported, and then test for them and raise an error. I just want to understand what you're going for.)

Thanks again, @andynelson!

joelthompson:
needs_revision

cloud/amazon/ec2_vpc_dhcp_options.py
andynelson, line 303: Because there's a possibility that the user chose to inherit, this is needed...
otherwise explicitly providing an empty list (intending to replace a populated one) would be ignored,and the list on the old dhcp options would be reapplied.

(e.g. the existing dhcp options contains an ntp server list but we want to remove that)

andynelson, line 303: This would be significantly simpler if we weren't trying to implement inheritance, which feels kind of non declarative.
(since it effectively tries to modify an unknown, rather than just create a known condition)

cloud/amazon/ec2_vpc_dhcp_options.py
andynelson: If no tags are provided, then this matches only on options, if no options were provided it matches only on tags

imo, from a 'dwim' perspective, matching the first found is the right behaviour - the request was for a dhcp_options that matches, and one was found... if they're looking to hit a specific one they could specify the id.

(I can see it being tricky in the case of stat=absent, rather than creation/assignment - but since dhcp options can't be deleted if they are assigned to a vpc, it isn't dangerous per se... worst case is that something _doesn't_ get deleted)

cloud/amazon/ec2_vpc_dhcp_options.py
andynelson, line 226: not at the moment - I dropped in a stock function in case we later want to return a value based on whether tags were changed.

cloud/amazon/ec2_vpc_dhcp_options.py
andynelson, line 320: good catch...
I'd somehow bounced lines 294-307 to the left - this should be outside of that if

cloud/amazon/ec2_vpc_dhcp_options.py
andynelson, line 227: what's the preferred way? I'm seeing other modules doing this (will change to use module.fail_json)

cloud/amazon/ec2_vpc_dhcp_options.py
andynelson, line 281: no - found is set if we found an existing dhcp options set, while dhcp_option might contain a newly created one

cloud/amazon/ec2_vpc_dhcp_options.py
andynelson: thanks, yes - this should fail if `dhcp_options_id` is supplied and it doesn't exist (unless we were removing it anyway)

andynelson:
Thanks for the feedback Joel... a fresh pair of eyes always helps :-)

I _think_ I've covered most everything in your review everything with the new commit.

You're close on my intended semantics:

here's what I mean it to do:

```
1. If dhcp_options_id is provided
    i: If state == 'absent' try to remove it (return changed:True if successful
    ii: If state != 'absent'
        a: tag correctly if tags were supplied
        b: attempt to assign to a vpc if vpc_id was supplied
        c: attempt delete the vpc's old options if we were asked to*
   (note - dhcp options are immutable, so if options are supplied, they are ignored)
2. If no dhcp_options_id is provided
    i: start by trying to find one that matches what we were asked for
        a: use combined options from the vpc's currently assigned options if that was requested*
        b: if we have options at this point, they have to match
        c: if we were given tags, they have to match
    ii: if state == absent
        a: if we found an option set,  try to remove it (return changed:True if successful)
        b: if we didn't, then there's nothing to do (return changed: False)
    iii: if state != absent
        a: if we didn't find matching options, create them (tagging if asked)
        b: if we were given a vpc_id then attach the options to it (otherwise we're done)
        c: if we were asked to delete old options, then try to do that
        (return changed: True if we created and/or attached)
```

so in answer to your two scenarios
1. Provide both dhcp_options_id and vpc_id
   If state == 'absent' then the dhcp options will be removed if possible
   If state != 'absent' the dhcp options will be tagged (if tags were provided), and assigned to the vpc
2. provide neither dhcp_options_id or vpc_id
   if state == 'absent' then if a matching set of options is found, it will be removed (if possible)
   If state != 'absent' then options will be created, and the dhcp_options_id returned

Having been away from the code for a day, I had a lightbulb moment btw...
Your initial role comes at things from a 'vpc centric' perspective - dhcp options are an attribute of a vpc

My starting point has been 'dhcp options centric' - they are an object in their own right

so that may explain the complexity - the role now tries to handle two different world-views...
had I started clean, then 'inherit_existing' and 'delete_old' wouldn't have been considered - and possibly associating to a vpc would have been left to the vpc_net module
I marked the things that wouldn't be done with \* above

(from that perspective, I'm wondering whether simply having two modules, your existing one, and a new one might have been a better plan)

Andy

andynelson:
just realised I didn't tag you in my response @joelthompson

cloud/amazon/ec2_vpc_dhcp_options.py
joelthompson, line 227: Either use a `,` instead of `as`, or (if you wish to also be Python 3 compatible) use the new `get_exception` method provided by `ansible.module_utils.basic`. See, e.g., #1429.

As I said, it's a minor style note.

cloud/amazon/ec2_vpc_dhcp_options.py
joelthompson: Thanks for adding the additional check, but I wasn't clear in my original response, sorry. If you pass in a `dhcp_options_ids` parameter to `get_all_dhcp_options` then it will throw an exception:

```
$ python
>>> import boto.vpc
>>> c = boto.vpc.connect_to_region('us-east-1')
>>> c.get_all_dhcp_options(dhcp_options_ids='dopt-12345678')
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/boto/vpc/__init__.py", line 1228, in get_all_dhcp_options
    [('item', DhcpOptions)])
  File "/usr/local/lib/python2.7/dist-packages/boto/connection.py", line 1186, in get_list
    raise self.ResponseError(response.status, response.reason, body)
boto.exception.EC2ResponseError: EC2ResponseError: 400 Bad Request
<?xml version="1.0" encoding="UTF-8"?>
<Response><Errors><Error><Code>InvalidDhcpOptionID.NotFound</Code><Message>The dhcpOption ID 'dopt-12345678' does not exist</Message></Error></Errors><RequestID>5b749029-e7d2-449e-9960-f0a928eab648</RequestID></Response>
```

I think changing the code to this would solve the problems:

`supplied_options = [d for d in connection.get_all_dhcp_options() if d.id == params['dhcp_options_id']`

(I've found myself writing this type of boilerplate a lot, unfortunately...)

cloud/amazon/ec2_vpc_dhcp_options.py
joelthompson, line 320: Ahh, that makes things a lot clearer now, this led to a lot of my confusion earlier (and the complicated semantics flow I had described -- this significantly alters the flow logic [thanks Python for making whitespace semantically meaningful...])! Thanks for the clarification!

cloud/amazon/ec2_vpc_dhcp_options.py
joelthompson, line 303: Ahh, good point on the need to support `inherit_existing`. I didn't think about the case where I wanted to `inherit_existing` and simultaneously remove options for something. Can I suggest tweaking the documentation to clarify this difference? E.g., something like: "If you choose to inherit_existing, omitting a DHCP option will result in inheriting the old option; if you wish to remove that option, you need to explicitly pass in an empty list."

I agree that at some level this feels non-declarative, though at another level, it is. If my fundamental unit that I can declare is an entire AWS DHCP options set, then yes, you're right. If the fundamental unit I want to declare is particular options (e.g., "I want to declare that the DNS servers for this VPC are 4.4.4.4 and 8.8.8.8"), then I think this strikes a good compromise relative to what AWS offers. I _think_ the latter is more natural to think about in terms of what I want to declare (e.g., I'm standing up NTP servers for my VPC and now I want to declare that DHCP should advertise my new NTP servers -- I don't care what the other things are).

cloud/amazon/ec2_vpc_dhcp_options.py
joelthompson, line 281: Ah, OK. It looks like the only time that that distinction is relevant is after the last usage of `found` which is what I was looking at, but I certainly see the benefit of future-proofing the code.

cloud/amazon/ec2_vpc_dhcp_options.py
joelthompson: > If no tags are provided, then this matches only on options, if no options were provided it matches only on tags

Sorry if I'm just being dense, but I can't see this control flow. If `tags` is `None` then your first conditional (`if tags and get_resource_tags(vpc_conn, dopts.id) == tags:`) will always be false, and no single DHCP option will match. Thus, "if no tags are provided" then nothing will ever match. Similar for `options` -- if `options` is empty (which would be achieved by not specifying any options to the module and setting `inherit_existing` to be `True`) then it will likewise never match anything.

I think what you want instead is a 'if not tags or get_resource_tags(vpc_conn, dopts.id) == tags` and similar.

Am I missing something?

cloud/amazon/ec2_vpc_dhcp_options.py
joelthompson, line 226: Cool, thanks.

cloud/amazon/ec2_vpc_dhcp_options.py
```diff
@@ -128,105 +159,207 @@
...
-    if new_options['domain-name']:
-        new_options['domain-name'] = new_options['domain-name'][0]
-  
-    if not module.check_mode:
-        dhcp_option = connection.create_dhcp_options(new_options['domain-name'],
-                new_options['domain-name-servers'], new_options['ntp-servers'],
-                new_options['netbios-name-servers'], new_options['netbios-node-type'])
+
+    # First check if we were given a dhcp_options_id
+    if not params['dhcp_options_id']:
```
joelthompson, line 292: This is an extremely small style nit...

In general, I dislike these types of conditionals:

``` python
if not some_condition:
    # something if some_condition is False
else:
    # something if some_condition is True
```

It's much harder for me to reason about that, as opposed to something like:

``` python
if some_condition:
    # something if some_condition is True
else:
    # something if some_condition is False
```

Again, this is a minor point. If you happen to get around to it, great! If not, no worries.

cloud/amazon/ec2_vpc_dhcp_options.py
```diff
@@ -128,105 +159,207 @@
...
+            # if we've been asked to inherit existing options, do that now
+            if params['inherit_existing']:
+                if not existing_options:
+                    module.fail_json(msg="a vpc_id was provided, but that vpc does not exist")
+                for option in [ 'domain-name-servers', 'netbios-name-servers', 'ntp-servers', 'domain-name', 'netbios-node-type']:
+                    if existing_options.options.get(option) and (not new_options[option] or [''] == new_options[option]):
+                        new_options[option] = existing_options.options.get(option)
+    
+        # Do the vpc's dhcp options already match what we're asked for? if so we are done
+        if existing_options and new_options == existing_options.options:
```
joelthompson: I think this has the potential to raise an exception -- `existing_options` is only defined if `vpc_id` is passed in as a parameter. If nether dhcp_options_id nor vpc_id is passed in, then this will error out.

joelthompson:
Thanks for your patience with my comments, @andynelson :)

Your intended semantics seem pretty close to what I would want, except a few possible corner cases that  I don't think matter that much. The two issues of potentially unhandled exceptions, and the question of match_dhcp_options, are the most important ones to resolve. The rest are relatively minor, and if you got to them, great; if not, then I wouldn't want to block this great addition on those minor tweaks.

And good insight on the way we're thinking differently :) I suspect we're trying to solve related but different problems. 

--Joel

cloud/amazon/ec2_vpc_dhcp_options.py
andynelson: No, I was... your logic works exactly as I was intending - thanks for spotting that :-)

cloud/amazon/ec2_vpc_dhcp_options.py
andynelson: yep, I was unconsciously reading this as having a filter, rather than hardcoding the request - changing to a filter will solve that

cloud/amazon/ec2_vpc_dhcp_options.py
andynelson, line 303: yep, I can see where you're coming from here for sure...
I'm in an environment where we have to care, so we use detailed inventory, and then make reality match it.
(so I'd already have the existing dhcp options in inventory, and would be updating there then recreating)

cloud/amazon/ec2_vpc_dhcp_options.py
andynelson, line 292: In general I agree... I have a sense that it may take significant refactoring to achieve that though
(because there's a bunch of things we _only_ do if dhcp_options_id wasn't provided, and they happen first)

andynelson:
sorry for the delay in getting back to this @joelthompson - I've been travelling this week.
This should now cover your major points

Travis checks now seem to be failing because
ERROR: version_added for new option (tags) should be 2.1. Currently 0.0
ERROR: version_added for new option (state) should be 2.1. Currently 0.0
ERROR: version_added for new option (dhcp_options_id) should be 2.1. Currently 0.0

although since the version_added for the module is 2.1 it feels like those should be ok?

Andy

sivel:
I've fixed the linting checks, everything should be green as far as that is concerned.

joelthompson:
Hey @andynelson,

Now that I'm (hopefully) being a little less dense and finally understand what you're going for here, I tested a little more thoroughly and found a couple bugs:
- If a VPC currently had no DHCP option set associated with it, it would exception out. Now it gets treated a little more sensibly.
- If I specified both `dhcp_options_id` and `vpc_id` then the use of `existing_options` at the very end would result in an uninitialized variable error.
- You referred to `options['tags']` instead of `params['tags']` in a couple spots
- `changed` wasn't getting set to `True` when an existing DHCP Options Set was being newly applied to a VPC
- You were missing a parameter in a call to `ensure_tags`

I posted a patch that should fix these issues at https://gist.github.com/joelthompson/fefc5a158ad1abc7c606

Thanks!

andynelson:
Thanks @joelthompson for the exhaustive testing, and the fixes!
for some reason the patch didn't apply first time... there was a mismatch somewhere, but this should have your fixes now

joelthompson:
Thanks @andynelson but unfortunately there's a merge conflict. It looks like @sivel made one of the same documentation fixes you did, can you rebase?

andynelson:
The rebase is now done @joelthompson

joelthompson:
Thanks! :)

shipit

gregdek:
Thanks again to @andynelson for this PR, and thanks @joelthompson for reviewing. Marking for inclusion.

PR #1644, opened 2016-02-09
Add option to send puppet apply logs to syslog

emonty:
While returning puppet logs as ansible stdout is useful in some cases,
there are also cases where it's more destructive than helpful. For
those, local logging to syslog so that the ansible logging makes sense
is very useful.

This defaults to stdout so that behavior does not change for people.

### DIFF
--- a/system/puppet.py
+++ b/system/puppet.py
@@ -67,6 +67,12 @@ options:
       - Puppet environment to be used.
     required: false
     default: None
+  logdest:
+    description:
+      - Where the puppet logs should go, if puppet apply is being used
+    required: false
+    default: stdout
+    choices: [ 'stdout', 'syslog' ]
 requirements: [ puppet ]
 author: "Monty Taylor (@emonty)"
 '''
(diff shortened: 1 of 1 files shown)


bcoca:
most modules do this by default, we have no_log and no_target_syslog options to avoid this in many cases, this option would bypass these general settings.

emonty:
So - this is the output from puppet feeding into the ansible module stdout return value, not controlling ansible's logging. The thing we want to express here is "ansible please log things as usual - puppet, on normal operation please do not emit logs to the stdout param of the return dict from ansible" 

Should we be hooking in to the no_log/no_target_syslog options in some way to be able to do that?

emonty:
no_log looks like it would solve half of the issue - but we still want to be able to tell puppet to log its stuff to syslog locally on the box in question. (with only no_log we'd lose the log output altogether)

bcoca:
no_log is there to prevent it, it should be 'obeyed' when True as it lets the user create a 'non disclosure' policy, no_syslog_target does  this partially but only applies to the syslog ON the target (it is always implied if no_log is true).

so i might have misunderstood what this does, will give it another closer look.

gregdek:
Thanks @emonty. Since you are a maintainer of this module, we are marking this PR for inclusion.

jimi-c:
Merged, thanks!

jimi-c:
Also took care of adding the `version_added: "2.1"` info for that new param.

emonty:
On 02/09/2016 01:38 PM, James Cammarata wrote:

> Also took care of adding the |version_added: "2.1"| info for that new param.
> 
> —
> https://github.com/ansible/ansible-modules-extras/pull/1644#issuecomment-182024831.

Oops - forgot that ... thanks!

PR #1649, opened 2016-02-11
win_unzip: overwrite any existing file

ilpianista:
If the destination files exist, the win_unzip module hangs forever because the server asked what to do with the existing files (overwrite or not).

It seems to me that the unarchive module overwrite any existing file so I'm applying the same logic here. 

See https://msdn.microsoft.com/en-us/library/ms723207%28v=vs.85%29.aspx for any documentation about `CopyHere` flags.

### DIFF
--- a/windows/win_unzip.ps1
+++ b/windows/win_unzip.ps1
@@ -61,7 +61,8 @@ If ($ext -eq ".zip" -And $recurse -eq $false) {
         $shell = New-Object -ComObject Shell.Application
         $zipPkg = $shell.NameSpace($src)
         $destPath = $shell.NameSpace($dest)
-        $destPath.CopyHere($zipPkg.Items())
+        # 20 means do not display any dialog (4) and overwrite any file (16)
+        $destPath.CopyHere($zipPkg.Items(), 20)
         $result.changed = $true
     }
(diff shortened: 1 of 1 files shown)


jhawkesworth:
Make sense to make it the same.
I tend to just use 
   win_file:
      state: absent

For anything I don't want before unzipping.

I did wonder for a minute whether this change needed documenting but the behavior you describe sounds like a bug (hanging) so I think no need to modify the documentation.

Again not module maintainer but 
meets_guidelines

robynbergeron:
Thanks @ilpianista. @schwartzmx please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

schwartzmx:
nice catch @ilpianista.   I agree, this is a bug no need for doc changes.

shipit

gregdek:
Thanks again to @ilpianista for this PR, and thanks @schwartzmx for reviewing. Marking for inclusion.

resmo:
f8e1600444b8952ac7a78e79524c027e952341c3 seems to be the source of this bug, which is only in devel.

Issue #766, opened 2015-07-25
pacman: package groups' state always considered absent


hannahdiels:
**Issue Type:** Bug Report

**Ansible Version:**

```
ansible 1.9.2
  configured module search path = None
```

**Ansible Configuration:** Not relevant

**Environment:** Arch Linux: Linux 4.1.2-2-ARCH #1 SMP PREEMPT x86_64 GNU/Linux

**Summary:**

No distinction is made between package groups and packages. Package group state=present is not detected, e.g. pacman -Qi base-devel will always report base-devel missing, as it checks for a package.

Relevant commands:

```
pacman -Sg base-devel  # print all packages in group
pacman -Qg base-devel  # print all packages in group that are installed
(pacman -Qg base-devel & pacman -Sg base-devel) | sort | uniq -u  # print packages in group that are not installed
```

My **suggested fix** is to start the module by building a list of packages to install/remove/whatever:
1. Determine whether {{ name }} is a group or package: `pacman -Sg base-devel` has `return code == 0` if it's a group, `1` otherwise.
2. A. If it's a group: pkgs = second column of the output of step 1.
   B. Else, pkgs is singleton of {{ name }}

You'll need to expand on the above to allow `name=a,b` and such.

**Steps To Reproduce:**

Put this task in a playbook, then run it twice:

```
- name: Install base-devel
  pacman: name=base-devel state=present
```

**Expected Results:**

Assuming base-devel wasn't installed, the first run would install all packages of base-devel. The second run would do nothing. (Ideally, only packages in base-devel that are absent will be installed; no reinstalls)

**Actual Results:**

All packages of base-devel are (re)installed, regardless of whether they were already present.

rightaway:
+1

dcbishop:
+1 from me to. It's kind of painful trying to install xorg for example. I basically have to specify 72 individual packages and that's not maintainable since it might change in the future.

suizman:
Same here install xfce4 always show changes:

```
PLAY [localhost] ************************************************************** 
GATHERING FACTS *************************************************************** 
ok: [localhost]

TASK: [roles/kickass-xfce | Install base package Arch Linux] ****************** 
changed: [localhost]

TASK: [roles/kickass-xfce | Install extra package Arch Linux] ***************** 
ok: [localhost] => (item=xfwm4-themes)
ok: [localhost] => (item=numix-themes)

TASK: [roles/kickass-xfce | Install base package Debian/Ubuntu] *************** 
[...]

PR #1646, opened 2016-02-11
pacman: fix recurse delete. Add force update_cache feature.

k0ste:
Recurse delete was broken on #762.
Now module can force fetch repo databases. It was disabled on #762, now is a 'force' option to 'update_cache'.

PR #1651, opened 2016-02-11
Fixes package group handling in pacman module

### LINKED ISSUE #650: pacman module always reports meta package installation as 'changed'
Issue Type: Bug Report

Ansible Version: 1.9.1

Ansible Configuration: default from FreeBSD port

Environment: Controller FreeBSD 9x. Target Arch x64

Summary: pacman module tasks that install either the 'base-devel' or 'gnome' meta packages always report the status as "changed" and "installed 1 package(s)". This occurs even though the meta package is already installed.

Steps To Reproduce: create and run playbook with a pacman module task that installs an Arch 'meta' package, such as 'base-devel' or 'gnome'

Expected Results: Status message for already installed meta packages will show "changed": false ... "msg": "package(s) already installed"}

Actual Results: status message is "changed" and "installed 1 package(s)".

```
TASK: [workstations | install Arch software pkgs] ************************* 
ok: [host.name] => (item=kexec-tools) => {"changed": false, "item": "kexec-tools", "msg": "package(s) already installed"}
ok: [host.name] => (item=nfs-utils) => {"changed": false, "item": "nfs-utils", "msg": "package(s) already installed"}
changed: [host.name] => (item=base-devel) => {"changed": true, "item": "base-devel", "msg": "installed 1 package(s)"}
TASK: [workstations | install Gnome Desktop environment ] ********* 
changed: [host.name] => {"changed": true, "msg": "installed 1 package(s)"}
```

wkeeling:
This pull request fixes #650 and #766 which both exist on current devel. The pacman module is currently unable to properly handle package groups (also referred to as "meta packages"), because the query command it uses to discover whether a package is installed does not work with groups. As a result, groups such as "gnome" or "base-devel" are always reported as uninstalled even if they are installed, causing the module to needlessly re-install them.

As suggested in #766, this fix modifies the pacman module so it first checks whether a name refers to a group or a package. Where a name refers to a group, that group is then expanded into a list of its indvidual packages before the list is passed to the relevant install/remove logic.

The pull request also removes some unused imports and a redundant remove_packages() function declaration.

### DIFF
--- a/packaging/os/pacman.py
+++ b/packaging/os/pacman.py
@@ -267,6 +267,25 @@ def check_packages(module, pacman_path, packages, state):
         module.exit_json(changed=False, msg="package(s) already %s" % state)
 
 
+def expand_package_groups(module, pacman_path, pkgs):
+    expanded = []
+
+    for pkg in pkgs:
+        cmd = "%s -Sgq %s" % (pacman_path, pkg)
+        rc, stdout, stderr = module.run_command(cmd, check_rc=False)
+
+        if rc == 0:
+            # A group was found matching the name, so expand it
+            for name in stdout.split('\n'):
+                name = name.strip()
+                if name:
+                    expanded.append(name)
+        else:
+            expanded.append(pkg)
+
+    return expanded
+
+
 def main():
     module = AnsibleModule(
         argument_spec    = dict(
@@ -305,7 +324,7 @@ def main():
         upgrade(module, pacman_path)
 
     if p['name']:
-        pkgs = p['name']
+        pkgs = expand_package_groups(module, pacman_path, p['name'])
 
         pkg_files = []
         for i, pkg in enumerate(pkgs):


robynbergeron:
Thanks @wkeeling. @indrajitr please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

packaging/os/pacman.py
```diff
@@ -109,11 +109,6 @@
 - pacman: name=baz state=absent force=yes
 '''
 
-import shlex
-import os
-import re
```
indrajitr, line 116: Hmm, we have calls like `os.path.exists()`, `re.sub()` etc.

packaging/os/pacman.py
```diff
@@ -175,12 +170,6 @@ def upgrade(module, pacman_path):
         module.exit_json(changed=False, msg='Nothing to upgrade')
 
 def remove_packages(module, pacman_path, packages):
-    if module.params["recurse"]:
```
indrajitr: Good call, but `recuse=yes` support needs to be fixed. It was in my todo, but feel free to take it up :)
FWIW, `force=yes` isn't really same as `recurse=yes`.

packaging/os/pacman.py
```diff
@@ -261,6 +250,25 @@ def check_packages(module, pacman_path, packages, state):
         module.exit_json(changed=False, msg="package(s) already %s" % state)
 
 
+def expand_package_groups(module, pacman_path, pkgs):
+    expanded = []
+
+    for pkg in pkgs:
+        cmd = '%s -Sgq %s' % (pacman_path, pkg)
```
indrajitr: Nitpicking here :)
The module uses `"` instead of `'` in creating cmd string in all places. We'd probably want to be consistent.

indrajitr:
needs_revision
Nice PR, just needs a bit of rework and we are all set!

indrajitr:
Let #1646 be applied first. Once that's done, I'd request @wkeeling to rebase this branch and apply group handling changes cleanly.

packaging/os/pacman.py
wkeeling, line 116: The import `from ansible.module_utils.basic import *` on line 331 brings in these dependencies, but I guess it's probably better to be explicit. I'll re-add them.

wkeeling:
Thanks @indrajitr 

I'll wait for #1646 to be applied, then rebase and make the changes.

wkeeling:
Hi @indrajitr - #1646 has now been applied, so I have now rebased this branch and made the amendments based on your initial feedback.

Thanks!

indrajitr:
shipit
Thanks, @wkeeling!

bcoca:
@wkeeling can you squash to single commit?

wkeeling:
@bcoca, commits now squashed.

Many thanks!

PR #1351, opened 2015-12-08
ec2_asg_find.py will allow you to search for auto scale groups.

linuxdynasty:
Search AutoScaleGroups based on the name or prefix of the name.

PR #1659, opened 2016-02-12
ec2_asg_find.py will allow you to search for auto scale groups

linuxdynasty:
This PR will replace the previous PR #1351 @tombamford, this is the new PR for this module.

### DIFF
--- a/cloud/amazon/ec2_asg_facts.py
+++ b/cloud/amazon/ec2_asg_facts.py
@@ -16,7 +16,7 @@
 
 DOCUMENTATION = '''
 ---
-module: ec2_asg_find
+module: ec2_asg_facts
 short_description: Find auto scaling groups by name and/or tags
 description:
   - Searches for and returns list of matching auto scaling groups with comprehensive details
@@ -37,7 +37,7 @@ options:
     description:
       - If no results are returned, you can return either success (with empty results) or failure.
     required: false
-    default: fail
+    default: success
     choices: [ 'success', 'fail' ]
   limit_results:
     description:
@@ -53,36 +53,36 @@ EXAMPLES = '''
 
 # Find all groups
 - local_acton:
-    module: ec2_asg_find
+    module: ec2_asg_facts
   register: asgs
 
 # Find a group with matching name/prefix
-- ec2_asg_find:
+- ec2_asg_facts:
     name: public-webserver-asg
   register: asgs
 
 # Find a group with matching tags
-- ec2_asg_find:
+- ec2_asg_facts:
     tags:
       project: webapp
       env: production
   register: asgs
 
 # Find a group with matching name/prefix and tags
-- ec2_asg_find:
+- ec2_asg_facts:
     name: myproject
     tags:
       env: production
   register: asgs
 
 # Fail if no groups are found
-- ec2_asg_find:
+- ec2_asg_facts:
     name: public-webserver-asg
     no_result_action: fail
   register: asgs
 
 # Fail if more than the limit of 3 groups is found.
-- ec2_asg_find:
+- ec2_asg_facts:
     name: public-webserver-asg
     no_result_action: fail
     limit_results: 1
@@ -353,7 +353,7 @@ def main():
             tags=dict(type='dict'),
             limit_results=dict(required=False, type='int', default=0),
             no_result_action = dict(
-                required=False, type='str', default='fail',
+                required=False, type='str', default='success',
                 choices = ['success', 'fail']
             ),
         )
--- a/cloud/amazon/ec2_asg_find.py
+++ b/cloud/amazon/ec2_asg_find.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/python
 # This file is part of Ansible
 #
 # Ansible is free software: you can redistribute it and/or modify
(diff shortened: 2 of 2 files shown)


linuxdynasty:
@tombamford I linked this PR to the closed one. Now anyone who needs to view the history can.

linuxdynasty:
@gregdek and @sivel, If you could apply the labels to this PR as you did to PR #1351. I would greatly appreciate it.

cloud/amazon/ec2_asg_find.py
```diff
@@ -0,0 +1,410 @@
...
+def main():
+    argument_spec = ec2_argument_spec()
+    argument_spec.update(
+        dict(
+            name=dict(type='str'),
+            tags=dict(type='dict'),
+            limit_results=dict(required=False, type='int', default=0),
+            no_result_action = dict(
+                required=False, type='str', default='fail',
+                choices = ['success', 'fail']
```
wimnat: I don't think this is very 'Ansibley'.  Using other facts modules as an example, if there are no results then it just exits with an empty list. I think it should be on the user to handle that.

wimnat:
I would rename this to ec2_asg_facts

cloud/amazon/ec2_asg_find.py
manicminer: I prefer built-in failure handling over additional tasks to check for common failure states. Perhaps "success" would be a better default, then it's up to the user to choose?

manicminer:
@wimnat Good call, ec2_asg_facts is a better choice.

cloud/amazon/ec2_asg_find.py
linuxdynasty: I agree with you @tombamford. Change the default choice from fail to success. The original idea behind this module, was to search for an ASG and if the results are empty or if the number of results is more than limit_results, than this should fail.

cloud/amazon/ec2_asg_facts.py
```diff
@@ -0,0 +1,410 @@
...
+                [
+                    "subnet-a1b1c1d1",
+                    "subnet-a2b2c2d2",
+                    "subnet-a3b3c3d3"
+                ]
+            }
+        ]
+    """
+    asgs = conn.describe_auto_scaling_groups()['AutoScalingGroups']
+    matched_asgs = []
```
wimnat, line 308: Would it be better to query the API directly for the names passed rather than looping through all asgs further down?

cloud/amazon/ec2_asg_facts.py
```diff
@@ -0,0 +1,410 @@
...
+            matched_asgs.append(asg_details)
+    return matched_asgs
+
+def main():
+    argument_spec = ec2_argument_spec()
+    argument_spec.update(
+        dict(
+            name=dict(type='str'),
+            tags=dict(type='dict'),
+            limit_results=dict(required=False, type='int', default=0),
```
wimnat, line 354: Is this necessary? The user is going to have to check anyway that the result they get is the one they want.

Again, nothing against it really except that from a consistency stand point, other aws facts modules don't do this

cloud/amazon/ec2_asg_find.py
wimnat: I would argue that this is what 'failed_when' is for

cloud/amazon/ec2_asg_facts.py
manicminer, line 308: In this case, the user can specify either/both/neither the `name` and `tags` arguments. The name can also be a partial match and the API call doesn't seem to support wildcards. There are also no other filter options for the API call beyond specifying a list of _exact_ names.

cloud/amazon/ec2_asg_facts.py
manicminer, line 354: Agreed, the user could arguably check this themselves after each ec2_asg_facts task, but I'd contend that is unnecessary boilerplate for the simple case when you want to identify a single auto scaling group (which in the cases I've used this module, is every time). There is a similar precedent in the ec2_ami_find module for sort/limit behaviour (disclaimer: I submitted that module).

linuxdynasty, line 354: I argue that the other modules should have a fail/success option. So that the user does not have to check. If you set the module to fail if no results are found, than it should break in my humble opinion. Since not everyone will agree with that, is the reason why we added the fail/success option. It is left to the user to decide.

manicminer, line 354: @linuxdynasty On further thought, I completely understand @wimnat's point, since these two are equivalent:

``` yaml
- ec2_asg_facts:
    tags:
      foo: bar
    limit_results: 1
  register: asg
```

``` yaml
- ec2_asg_facts:
    tags:
      foo: bar
  register: asg
  failed_when: "{{ asg.results | length > 1 }}"
```

The module is effectively repeating Ansible built-in functionality, so I feel it would make more sense to remove that option. Less code == better maintainability.

linuxdynasty:
@tombamford I completely forgot about the failed_when option using Jinja filters. I just finished rereading the developer documentation http://docs.ansible.com/ansible/developing_modules.html. I do not see an area where it explicitly says, that modules should not have an option to fail. If this option is truly against the community/core standards, than I feel it should be explicitly stated in that document.

Since we can indeed use the failed_when option, I am no longer opposed in removing the fail/success option. @tombamford would you like to handle that or should I?

cloud/amazon/ec2_asg_facts.py
manicminer, line 354: Doh, I meant `failed_when` rather than `when_failed`.

manicminer:
@linuxdynasty Happy to make the changes. Will see what I can do in the next couple of hours.

cloud/amazon/ec2_asg_facts.py
```diff
@@ -0,0 +1,410 @@
...
+                else:
+                    asg_details[new_key] = None
+
+            for key in ['EnabledMetrics', 'Instances', 'SuspendedProcesses', 'Tags']:
+                if key in asg:
+                    new_key = camel_prog.sub(r'_\1', key).lower()
+                    asg_details[new_key] = []
+                    for item in asg[key]:
+                        asg_details[new_key].append(dict((camel_prog.sub(r'_\1', k).lower(), v) for k, v in item.iteritems()))
+
```
wimnat, line 339: Just in case some one wants to beat me to this...

I think we should add a shared function in the aws boilerplate that takes a dict and replaces all the CamelCase keys to non_camel_case.  Every new module is going to need it.

manicminer:
@linuxdynasty @wimnat I've made the suggested changes and resubmitted as https://github.com/ansible/ansible-modules-extras/pull/1710

PR #1660, opened 2016-02-12
dnf module: package not installed with state=latest

ingvagabund:
dnf: name=PACKAGE state=latest is reponsible for two use cases:
- to install a package if not already installed.
- to update the package to the latest if already installed.

The latter use cases is not handled properly as base.upgrade does not
throw dnf.exceptions.MarkingError if a package is not installed.

Setting base.conf.best = True ensures a package is installed or
updated to the latest when calling base.install.

Sign-off: jsilhan@redhat.com (dnf guy)
Sign-off: jchaloup@redhat.com

Thanks Jan Silhan for the 1 minute fix.

### DIFF
--- a/packaging/os/dnf.py
+++ b/packaging/os/dnf.py
@@ -274,11 +274,10 @@ def ensure(module, base, state, names):
                     # If not already installed, try to install.
                     base.group_install(group, const.GROUP_PACKAGE_TYPES)
             for pkg_spec in pkg_specs:
-                try:
-                    base.upgrade(pkg_spec)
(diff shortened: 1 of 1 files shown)


ingvagabund:
@bcoca, @nitzmahone, @abadger  can you review?

I have tested the patch on f22, worked for me.

abadger:
@ignatenkobrain, @DJMuggs  letting you guys know that I'm going to apply this.  Since Jan Silhan looked at this and helped with the fix it seem like it's probably correct.  If you find problems with it  let me know!

PR #1661, opened 2016-02-12
Feature Request Examples

puchrojo:
Feature Request: It would be nice two example with a key as a string, one with absent and other with present. thanks


sivel:
A feature request should not be submitted as a pull request.  If you would like to add these examples, please open a pull request with the added examples.  Otherwise, please resubmit this as an issue that is not placing the feature request in the actual documentation.

Issue #1375, opened 2015-12-15
#446 broke npm state=latest for missing packages


TamsynUlthara:
##### Issue Type:
- Bug Report (`npm` module)
##### Ansible Version:

Running against devel:

``` console
$ ansible --version
ansible 2.1.0 (devel be5488cb60) last updated 2015/12/15 09:36:59 (GMT -400)
  lib/ansible/modules/core: (devel 6b13da738b) last updated 2015/12/15 09:38:18 (GMT -400)
  lib/ansible/modules/extras: (devel f3251de29c) last updated 2015/12/15 09:38:42 (GMT -400)
  config file = /home/tomxtobin/.ansible.cfg
  configured module search path = Default w/o overrides
```
##### Ansible Configuration:

``` ini
[defaults]
hostfile = ~/ansible/hosts
nocows = 1
```
##### Environment:

N/A (but it's Arch Linux)
##### Summary:

It looks like PR #446 broke `npm: name=foo state=latest` for a missing package `foo` (i.e., `foo` isn't present on the system yet).

Suggested fix: for `state == 'latest'`, actually differentiate between the result of checking `len(missing)` and `len(outdated)` to see whether the package is installed or not, and run either `npm.install()` or `npm.update()` as appropriate.
##### Steps To Reproduce:

Let's use the `gulp` package as an example.

On a system that doesn't already have `gulp` installed globally:

``` console
$ ansible example-host -m command -a 'npm ls -g -depth 0 gulp' 
example-host | FAILED | rc=1 >>
/usr/lib
└── (empty)npm ERR! code 1
```

``` console
$ ansible example-host -m command -a 'gulp --version' 
example-host | FAILED | rc=2 >>
[Errno 2] No such file or directory
```

Run a task against such system to install `gulp` globally:

``` console
$ ansible example-host -m npm -a 'name=gulp state=latest global=yes'
example-host | SUCCESS => {
    "changed": true
}
```
##### Expected Results:

The module (`gulp`, above) actually gets installed on the system(s) I'm running that task against.

Against such a system, I can run something like:

``` console
$ ansible example-host -m command -a 'npm ls -g -depth 0 gulp' 
example-host | SUCCESS | rc=0 >>
/usr/lib
└── gulp@3.9.0 
```

``` console
$ ansible example-host -m command -a 'gulp --version'
example-host | SUCCESS | rc=0 >>
[15:24:28] CLI version 3.9.0
```

[...]

Issue #1098, opened 2015-10-14
npm: install doesn't pick up missing devDependencies


deployable:
The `npm` module will only install devDependencies for a fresh install. Anything using the `list()` logic will bypass devDependencies completely, for example adding a new package to an existing install. 

The [module relies](https://github.com/ansible/ansible-modules-extras/blob/15480e3d04f586645516b42863e2c05f55239488/packaging/language/npm.py#L169-L181) on `npm list --json` which will only reports missing `production` dependencies. Even when using `npm list --dev --json` npm doesn't report what's missing. 

The module code also only looks at the `dependencies` hash from `npm list --json` for things to install, which I guess is fine as nothing is reported as missing in devDependencies.

deployable:
Demo setup

```
○ /tmp/ansible → cat package.json 
{
  "name": "ansible-npm-test",
  "version": "1.0.0",
  "description": "ansible-npm-test",
  "main": "index.js",
  "dependencies": {
    "ms": "^0.7.1"
  },
  "repository": "over there",
  "scripts": {
    "test": "node index.js"
  },
  "keywords": [
    "test"
  ],
  "author": "Matt",
  "license": "ISC",
  "devDependencies": {
    "humanize": "0.0.9"
  }
}

○ /tmp/ansible → cat hosts 
[local]
localhost

[local:vars]
ansible_connection=local

○ /tmp/ansible → cat test.yml 
---
- hosts: local
  tasks:
  - npm: |
      path=/tmp/ansible
      state=present
```

deployable:
npm module failing to install missing devDependency of `humanize`

```
○ /tmp/ansible → rm -rf node_modules/

○ /tmp/ansible → ansible-playbook -i hosts test.yml 

PLAY [local] ****************************************************************** 

GATHERING FACTS *************************************************************** 
ok: [localhost]

TASK: [npm path=/tmp/ansible
state=present
] ********************************** 
changed: [localhost]

PLAY RECAP ******************************************************************** 
localhost                  : ok=2    changed=1    unreachable=0    failed=0 

○ /tmp/ansible → find node_modules
node_modules
node_modules/humanize
node_modules/humanize/.npmignore
node_modules/humanize/bower.json
node_modules/humanize/humanize.js
node_modules/humanize/LICENSE
node_modules/humanize/package.json
node_modules/humanize/README.md
node_modules/humanize/specs
node_modules/humanize/specs/dateData.js
[...]

Issue #957, opened 2015-09-11
npm module doesn't install new grunt plugin according to package.json update


visay:
- **Issue Type**: Bug Report
- **Ansible Version**: ansible 1.9.2
- **Environment**: Ubuntu Server 14.04 LTS AMD64
- **Summary**: npm module doesn't install new grunt plugin according to package.json update.

**Steps To Reproduce**:
- Create a `package.json` and store it in a directory:

```
{
    "name": "Dummy",
    "version": "0.0.1",
    "devDependencies": {
        "grunt": "latest",
        "grunt-contrib-clean": "~0.6.0",
        "grunt-contrib-copy": "latest",
        "grunt-contrib-watch": "latest",
        "grunt-contrib-uglify": "latest",
        "grunt-contrib-compass": "latest",
        "grunt-contrib-cssmin": "~0.10.0",
        "grunt-contrib-concat": "~0.5.0"
    }
}
```
- Run ansible command to install them

```
ansible localhost -m npm -a "path=/path/to/my/dir/containing/package-json state=latest"
```

First run, everything is fine and I get `node_modules` directory with all plugins defined in `package.json`.
- Update `package.json` file adding `"grunt-contrib-imagemin": "~0.4.0"` and change version to `0.0.2`
- Call ansible command again

```
ansible localhost -m npm -a "path=/path/to/my/dir/containing/package-json state=latest"
```

**Result**: Nothing updated. imagemin plugin doesn't get installed
**Output**: Always get

```
localhost | success >> {
    "changed": false
}
```
- If runing `npm install` shell command manually inside the directory, it works fine and the new plugin for imagemin get downloaded.

juhau1:
I'm having the same problem. This seems like a critical bug affecting many npm packages.

My main package.json says:

``` javascript
  "dependencies": {
    "my-sub-module": "latest"
  }
```

When updating my-sub-module package ansible npm module is not picking up the change, but continues with the old my-sub-module version.

Current workaround is to use shell module to do npm install as suggested by visay.

Ansible version: 1.9.3-1ppa~trusty

robynbergeron:
[module: packaging/language/npm.py]

robynbergeron:
@visay -- thanks for submitting this bug report, and thanks to @juhau1 as well for providing info around your ability to duplicate.

@chrishoffman -- pinging you to let you know about this bug report for the npm module.

frinux:
[...]

Issue #578, opened 2015-06-03
npm module does not detect that a package if already installed if using the 'name@version' construct


JeanFred:
Issue Type: Bug report / Feature request
Ansible Version: 1.9.1
Environment: Ubuntu 14.04
Summary: the NPM module does not detect that a package is already installed when using the 'name@version' construct

(I do realise that the `name@version` construct is not documented for this module, but I think it’s a fairly popular NPM construct and it does work. Also, it makes it much easier when using `with_items`)

The `list()` method loops on the installed package JSON, comparing names. Since `package@version` does not match `package`, the package gets reinstalled.

gregdek:
attn: @chrishoffman

JeanFred:
If there is interest, I’d be happy to submit a pull request for it :)

ava-miket:
I would like to see this fixed -- it's been a minor irritant to me

ansibot:
@chrishoffman, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@shane-walker, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@shane-walker, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
This repository has been locked. All new issues and pull requests should be filed in https://github.com/ansible/ansible

Please read through the <a href="http://docs.ansible.com/ansible/dev_guide/repomerge.html#move-issues-and-prs-to-new-repo">repomerge</a> page in the dev guide. The guide contains links to tools which automatically move your issue or pull request to the ansible/ansible repo.

ansibot:
This issue was migrated to https://github.com/ansible/ansible/issues/29210

PR #1663, opened 2016-02-14
Update npm module to work better with newer npm versions

moises-silva:
The 'outdated' npm module gives us better information now days
and it can acept a --json option so there is no need to parse
the output with a regex.

Also, it's better to attempt an install operation even if
the state specified is 'latest' because you may need to bootstrap
an empty project and old npm versions wouldn't do anything
when 'npm update' is done and there is only devDependencies.

I believe these fixes have a very good chance of solving:

issue #1375
issue #1098
issue #578
issue #383

And slightly related to:

issue #957

### DIFF
--- a/packaging/language/npm.py
+++ b/packaging/language/npm.py
@@ -103,7 +103,6 @@ description: Install packages based on package.json using the npm installed with
 '''
 
 import os
-
 try:
     import json
 except ImportError:
@@ -127,13 +126,14 @@ class Npm(object):
 
         if kwargs['version']:
             self.name_version = self.name + '@' + self.version
+            self.reqversion = kwargs['version']
         else:
             self.name_version = self.name
+            self.reqversion = None
 
     def _exec(self, args, run_in_check_mode=False, check_rc=True):
         if not self.module.check_mode or (self.module.check_mode and run_in_check_mode):
             cmd = self.executable + args
-
             if self.glbl:
                 cmd.append('--global')
             if self.production:
@@ -161,24 +161,57 @@ class Npm(object):
         return ''
 
     def list(self):
-        cmd = ['list', '--json']
+        """Gather a list of installed and missing
+        node modules. Try first to use 'npm list' and then
+        fallback to 'npm outdated' because 'list' does not
+        list devDependencies (not even with a --dev) when the
+        project consists of a single package.json and nothing
+        has been installed at all yet (that seems like a bug),
+        but npm outdated does list them ...
+        """
+        installed = set()
+        missing = set()
 
-        installed = list()
-        missing = list()
+        # Use list first
+        cmd = ['list', '--json']
         data = json.loads(self._exec(cmd, True, False))
         if 'dependencies' in data:
             for dep in data['dependencies']:
-                if 'missing' in data['dependencies'][dep] and data['dependencies'][dep]['missing']:
-                    missing.append(dep)
-                elif 'invalid' in data['dependencies'][dep] and data['dependencies'][dep]['invalid']:
-                    missing.append(dep)
+                depinfo = data['dependencies'][dep]
+                if ('missing' in depinfo
+                   and depinfo['missing']):
+                    missing.add(dep)
+                elif ('invalid' in depinfo
+                      and depinfo['invalid']):
+                    missing.add(dep)
+                elif (self.reqversion
+                      and 'version' in depinfo
+                      and depinfo['version'] != self.reqversion):
+                    missing.add(dep)
                 else:
-                    installed.append(dep)
-            if self.name and self.name not in installed:
-                missing.append(self.name)
-        #Named dependency not installed
-        else:
-            missing.append(self.name)
+                    installed.add(dep)
+
+        # Now try using 'outdated' because list does not seem to list missing
+        # dev dependencies at all (npm version 3.7.2) even if --dev is
+        # specified (though should not be needed unless you only want
+        # dev dependencies listed). If this bug is ever fixed on npm, this
+        # can be removed.
+        try:
+            cmd = ['outdated', '--json']
+            data = json.loads(self._exec(cmd, True, False))
+            installed.update([pkg for pkg, info in data.items()
+                              if pkg not in missing
+                              and 'location' in info
+                              and info['location']])
+            missing.update([pkg for pkg, info
+                            in data.items()
+                            if pkg not in installed
+                            and 'current' not in info])
+        except ValueError:
+            pass
+
+        if self.name and self.name not in installed:
+            missing.add(self.name)
 
         return installed, missing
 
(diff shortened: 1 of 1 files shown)


deployable:
Great! thanks. 

I've had a quick try and the new `outdated` command is failing to produce any output so the module falls back to the old `list` method.  Although now the old method is also not installing any missing `development` dependencies, which I'm sure it was before, I only had issues with `devDependencies` previously. 

I'm installing a specific package to a specific path

```
  npm: |
    name={{ item.npm }}
    state=present
    production={{ node.env == "production" }}
    path={{ path_apps }}/{{ item.npm }}-{{ item.version }}
    version={{ item.version }}
```

With some added debug, results in;

```
DEBUG:root:Running: ['/bin/npm', 'outdated', '--json', 'node-package@3.0.0'] /apps/node-package-3.0.0
DEBUG:root:rc: 0
DEBUG:root:stdout: 
DEBUG:root:stderr: 
DEBUG:root:outdated failed with:No JSON object could be decoded
DEBUG:root:outdated failed, using list
DEBUG:root:Running: ['/bin/npm', 'list', '--json', 'node-package@3.0.0'] /apps/node-package-3.0.0
DEBUG:root:rc: 1
DEBUG:root:stdout: {
  "name": "node-package",
  "version": "3.0.0",
<snip>
```

`/bin/npm outdated --json` will work when the package name isn't appended to the end, otherwise it only queries if that named package is outdated.

moises-silva:
@deployable That's odd, the old logic is pretty much unchanged. Let me try to run your use case.

moises-silva:
@deployable What's your npm version btw? npm --version

moises-silva:
@deployable I've made significant changes again, this time I tested individual packages as well and included some TESTING documentation at the top. When using npm version 3.7.2, all seems to work for me now (individual packages, updating, packaging.json projects, etc). I did not test some more advanced options though like production=yes, but AFAICS that only adds a --production flag so as long as npm does what is expected with that flag things should be fine.

deployable:
I'll give the updates a try. 

FYI

```
$ npm --version
3.4.1
$ node --version
v4.3.0
```

deployable:
btw do you have a good process for debugging modules? I find it one of the most frustrating parts of ansible, at least on the operational side.  When parts of an ansible run fail with a short error and there's no facility to debug what the module's are doing underneath on the host.

bcoca:
from the repo hacking/test-module can be used to run them and attach a python debugger
also you can configure ansible to keep the remote temp files, which includes the full module code + it's arguments.

deployable:
@bcoca Thanks for that, got it working although it took a while to get running. For some reason the `pdb` bin doesn't start and then `import ipdb` somehow gets written over so I can't run `ipdb.set_trace()`... anyway 

I'm going off on a bit of a tangent here... 

I'm talking more on the operational side though. After the module is locally tested it goes into use to run on X type of external hosts with X type of playbook definitions. A lot of the time's there's no feedback to the front end for issues in the module.  For playbooks that run commands, a standard interface to collect the commands, rc, stderr and stdout would be awesome. Maybe this is purely a module standards thing, but it shouldn't require Python knowledge and a debugger for users to figure out what a playbook is doing or trying to run. 

It's along the lines of the fix on first failure ethos of having enough information available on failure rather than having to try and reproduce complex problems in complex environments (or requiring developers to use debuggers).

The other thing that npm highlights is the lack of streaming output, which I understand there are some design issues to supporting, but when things take a while (usually network related) there is no way to interrogate what is currently happening. Docker builds are another area I run into this. 

Even a local debug log for modules that get's cleaned up with the script temp directory (but saved or dumped back to the master on failure) would help quickly resolve so many of the issues I run into and could be done out of band of the ansible session. 

...but I digress from the issue at hand.

bcoca:
The modules were designed to be simple scripts and presumably did not need that much debugging. 

Still, we do return output on catastrophic errors (better in >=2.0) and have facilities for modules to return more data on error, the issue is that most modules do not pass a lot of that info back (msg and traceback keys in return). 

Also, not everything modules do translates into a shell command, but you CAN log those putting Ansible in debug mode (via ANSIBLE_DEBUG env var or debug in ansible.cfg), you can also keep the remote files to examine using ANSIBLE_REMOTE_TEMP/remote_tmp settings.

The 'live updates' feature is something i'm working on https://github.com/bcoca/ansible/tree/update_json but is still a way off as modules would also have to be updated to take advantage of it.

deployable:
@moises-silva the new update looks pretty solid.  

Are you intentionally falling back to the old list method when `npm outdated` reports ok? What I though was skipping before was outdated exiting with rc=0 but writing nothing to stdout, so the JSON won't parse.

This actually works well for the local install of a named package case I've been testing, because `npm outdated` doesn't work very well when there's not a `package.json` in place.  

```
○ → mkdir npm-debug
○ → cd npm-debug/
○ → npm install debug
/private/tmp/npm-debug
└─┬ debug@2.2.0 
  └── ms@0.7.1 

npm WARN enoent ENOENT: no such file or directory, open '/private/tmp/npm-debug/package.json'
npm WARN npm-debug No description
npm WARN npm-debug No repository field.
npm WARN npm-debug No README data
npm WARN npm-debug No license field.

○ → rm -rf node_modules/ms/
○ → npm outdated --json
○ → npm ls --json
{
  "problems": [
    "missing: ms@0.7.1, required by debug@2.2.0"
  ],
  "dependencies": {
    "debug": {
      "version": "2.2.0",
      "from": "debug@latest",
      "resolved": "http://hlcit003.openans.co.uk:8081/nexus/content/groups/npm-all/debug/-/debug-2.2.0.tgz",
      "problems": [
        "missing: ms@0.7.1, required by debug@2.2.0"
      ],
      "dependencies": {
        "ms": {
          "required": "0.7.1",
          "missing": true
        }
      }
    }
  }
}
npm ERR! missing: ms@0.7.1, required by debug@2.2.0

○ → npm install ms@0.6.0
/private/tmp/npm-debug
├─┬ debug@2.2.0 
│ └── ms@0.7.1 
└── ms@0.6.0 

npm WARN enoent ENOENT: no such file or directory, open '/private/tmp/npm-debug/package.json'
npm WARN npm-debug No description
npm WARN npm-debug No repository field.
npm WARN npm-debug No README data
npm WARN npm-debug No license field.

○ → npm outdated --json
{
  "ms": {
    "current": "0.6.0",
    "wanted": "0.6.0",
    "latest": "0.7.1",
    "location": "node_modules/ms"
  }
}
```

So there's a couple of quirks, `npm outdated --json` doesn't work when modules are missing in a locally installed package (which is fine at the moment because the JSON doesn't parse and it falls back to `list`).

The reporting `ms` is outdated when an `ms` module exists is quirky but all survivable. 

I'll do some more testing with a `package.json` over the next few days.

sivel:
I would prefer the `TESTING` docs be removed as that is not a standard thing.  Instead a separate PR should be submitted to `ansible/ansible` to include integration tests with those `TESTING` steps.

moises-silva:
@sivel yeah not a problem, I was kinda expecting to get some feedback on that and figure out what test strategy (if any) was already in place. It was just a quick way to share with any potential testers how to verify this works.

moises-silva:
@deployable If you are talking about the npm.list() method, yeah, I intentionally start with the old method now and then complement that information with what outdated --json finds, if available. That seemed to have the best results across node versions and different use cases.

moises-silva:
@deployable On another note, I hear your pain with the debugging of modules. Those were my _exact_ thoughts as soon as I started debugging Ansible modules. How the hell do I make the modules log more so I know what's going on. I am used to work on systems (OSS telecom systems) with tons of debugging available that you can turn on/off very granularly, so I am a bit spoiled on that end.

@bcoca ^ .... I know what you mean, modules are meant to be simple, however, right now, every time I tried to fix this npm module (which is written by another person) I had to open() a file and add logging in useful places, then remove that before doing the pull request, then add it back when something went wrong, etc etc. Would you be open to a separate PR to add a logging mechanism to the AnsibleModule() class? so then we can leave log statements in place like:

``` python
module = AnsibleModule()
module.log('Here goes some logging!')
```

Then, we can mask those logs most of the time except when debug is enabled in which case those logs can stay in the host system or even get transferred back to the execution machine. Heck, they could even be turned on individually when calling a module with a special __debug=true parameter or something? so then debugging is triggered on a specific module known to be causing issues?

If that's something that you think Ansible devs could be open to should I send this to ansible dev google group? create a new github issue with a proposal? or perhaps this has been discussed before and is something the devs don't think they need? (I hope not, because certainly would make myself a lot more open to try to debug other modules if I had that facility available).

moises-silva:
@sivel Also, I forgot, any recommended reading (either docs or existing test cases) that would point me in the right direction to create test for this? this kind of thing needs to be tested in a VM or container (I used a VM myself), so not sure exactly how to write those tests that you suggested must be submitted to ansible as new integration test. I peeked briefly to ansible/test/integration, but still unsure about how to proceed exactly.

Is the idea I just code the tests as playbooks? but then how do I verify the side-effects and assert on them?

sivel:
> Would you be open to a separate PR to add a logging mechanism to the AnsibleModule() class? so then we can leave log statements in place like

Ansible 2.0 already has this functionality, specifically there is a `module.debug` method, that only if `ANSIBLE_DEBUG` is enabled, will it log items.

moises-silva:
@sivel sweet! I'm using ansible 2, I'll look into updating the npm module for 2.x to use it. Thanks.

sivel:
> Is the idea I just code the tests as playbooks? but then how do I verify the side-effects and assert on them?

Effectively you create a role.  That role should include tasks, that register results, and potentially use other modules to verify functionality and use the `assert` module, to assert conditions based on registered vars.  All of the tasks I see outlined in `TESTING` can be turned into modules that assert all these things.

Your role should then be called from an existing playbook, where applicable, or a new playbook.

Something like:

```
- npm: name=browserify global=true version=12.0.1 state=present
  register: result

- assert:
    that:
      - result|changed

- npm: name=browserify global=true version=12.0.1 state=present
  register: result

- assert:
    that:
      - not result|changed
```

And so on...

moises-silva:
Alright, I'll see what I can come up with for testing this.

gregdek:
Thanks @moises-silva. All backport requests must be reviewed by the core team, and this can take time. We appreciate your patience.

moises-silva:
@gregdek That's cool, I'd just ask you to please keep in mind that the npm module is pretty much broken without these fixes. If you decide not to 'backport' then you might as well better remove the npm module from 1.x releases because it just becomes a waste of time for those of us that try to use it and find that it half-works and end up better resorting to use raw shell commands to invoke npm.

gregdek:
cc: @ansible/core team for their opinion.

PR #1666, opened 2016-02-15
amazon/ec2_vpc_subnet.py: Add `auto_assign_public_ip` support

gechr:
Adds support for assigning public IPs to instances launched within a given subnet. Implementation is largely based on the solution suggested @ https://github.com/boto/boto/issues/2646

I've named the key itself `auto_assign_public_ip` (rather than `map_public_ip_on_launch`) for consistency with how it's displayed in the AWS Console ([link](https://i.imgur.com/lStbDJA.png)), but happy to rename it as you wish.

It should be noted that Boto3 has [first-class support for (un)setting this attribute](https://boto3.readthedocs.org/en/latest/reference/services/ec2.html#EC2.Subnet.map_public_ip_on_launch), but, in order to maintain backwards compatibility with Boto2, I've avoided using it.

### DIFF
--- a/cloud/amazon/ec2_vpc_subnet.py
+++ b/cloud/amazon/ec2_vpc_subnet.py
@@ -32,6 +32,11 @@ options:
       - "The CIDR block for the subnet. E.g. 10.0.0.0/16. Only required when state=present."
     required: false
     default: null
+  auto_assign_public_ip:
+    description:
+      - "Assign a public IP address to instances launched into the specified subnet."
+    required: false
+    default: no
   tags:
     description:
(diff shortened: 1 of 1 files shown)


sivel:
```
./cloud/amazon/ec2_vpc_subnet.py
ERROR: version_added for new option (auto_assign_public_ip) should be 2.1. Currently 0.0
```

gechr:
Sorted. Should the parameter perhaps be renamed `assign_public_ip` for consistency with the [`ec2` module](http://docs.ansible.com/ansible/ec2_module.html)?

PR #1719, opened 2016-02-23
The size option is required for lvol module

bborysenko:
##### Ansible Version: 2.0.0.2

```
ansible 2.0.0.2
  config file = /etc/ansible.cfg
```
##### Summary:

The size option is required for lvol module actually.

### DIFF
--- a/system/lvol.py
+++ b/system/lvol.py
@@ -48,7 +48,8 @@ options:
     choices: [ "present", "absent" ]
     default: present
     description:
-    - Control if the logical volume exists.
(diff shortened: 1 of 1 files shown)


bcoca:
it is not required in the spec

bborysenko:
Yes you are right, actually it requires only with `state=present`.

bborysenko:
Just updated description for `state` instead.

bcoca:
I see no change, it is still incorrect

bborysenko:
sorry, forgot to push

Issue #1680, opened 2016-02-17
yum_repository: remove unmanaged repo files


timss:
##### ISSUE TYPE

Feature Idea
##### COMPONENT NAME

yum_repository module
##### ANSIBLE VERSION

N/A
##### SUMMARY

The predecessor the the yumrepo module seems to be the [yumrepo role](https://github.com/picotrading/ansible-yumrepo), both made by @jtyr.

One feature I'm missing in the module is the ability to delete unmanaged repositories using the managed mode in the yumrepo role.

> The role also supports managed mode which removes any repo which is not managed by this role.

For anyone unfamiliar with the managed mode or the role itself, it creates `/etc/yum.repos.d/.managed` which includes the name of any repo you specified, and if enabled, deletes any repo not mentioned in this file at the end.

Using the module it is possible to remove existing repo files using `state: absent`. However, if you do not know which files are already present, or you plan to use the same playbook on hosts with different existing repo files it seems to me that this can quickly become a challenge to maintain.

Would it be possible to add a managed mode, or something similar to remove any repo files not specified using the yumrepo module? I might be able to help out, if that's desirable.

Thanks.

jtyr:
The `yumrepo` role is actually creating a temp file for every YUM repo file it creates and then there is a final play which creates the `.managed` file composed of the names of the temp files. I could add the same feature into the `yumrepo` module but then it's up to the user to create the final play which will create the `.managed` file and deletes the unmanaged repos. I believe that's not possible to do on the module level. Just thinking, perhaps using Ansible facts to store the list of created repo files would be better than the temp files. Please feel free to elaborate on the possible solution.

timss:
I have no previous experience developing Ansible modules, but looking at the yumrepo module code and official documentation it does indeed seem like a feasible solution to use facts (reference: [module provided facts](https://docs.ansible.com/ansible/developing_modules.html#module-provided-facts)). 

Not really in a position to say much about how it should be implemented at this point, but I'll try to take a look. My initial impression is that one could possibly add an option to `yumrepo` to trigger removal of repo files which are not available in the facts provided. If this sounds silly, please notify me. Thanks!

jtyr:
> add an option to yumrepo to trigger removal of repo files which are not available in the facts provided

This is definitely possible but it must be called at the end of the playbook in order collect all the facts from all the previous `yumrepo` module calls.

timss:
[...]

PR #1595, opened 2016-02-03
Adding more options to yumrepo module

jtyr:
This PR adds some more options which might be useful for some users (should fix issue #1563). It also fixes one critical typo in the name of `Exception`.

PR #1728, opened 2016-02-24
Adding Managed Mode into the yum_repository module

jtyr:
##### Issue Type:

Existing module enhancement.
##### Plugin Name:

`yum_repository`
##### Ansible Version:

For the 2.1 release.
##### Summary:

This patch is adding support for he Managed Mode into the `yumrepo` module. It addresses the feature request from the issue #1680. It builds upon the changes from the PR #1595 (approved to be merged) so if this patch is tagged with `shipit` the PR #1595 can be closed. 
##### Example output:

```
$ ansible-playbook -i hosts site.yaml 

PLAY [Initial play] ************************************************************

ok: [localhost]

TASK [Enable the managed mode] *************************************************
changed: [localhost]

PLAY [My play] *****************************************************************

ok: [localhost]

TASK [Create a repo file] ******************************************************
changed: [localhost]

PLAY [Final play] **************************************************************

ok: [localhost]

TASK [Delete unmanaged repo files] *********************************************
ok: [localhost]

localhost                  : ok=6    changed=2    unreachable=0    failed=0

$ ansible-playbook -i hosts site.yaml 

PLAY [Initial play] ************************************************************

ok: [localhost]

TASK [Enable the managed mode] *************************************************
ok: [localhost]

PLAY [My play] *****************************************************************

ok: [localhost]

TASK [Create a repo file] ******************************************************
ok: [localhost]

PLAY [Final play] **************************************************************

ok: [localhost]

TASK [Delete unmanaged repo files] *********************************************
ok: [localhost]

localhost                  : ok=6    changed=0    unreachable=0    failed=0
```

### DIFF
--- a/packaging/os/yum_repository.py
+++ b/packaging/os/yum_repository.py
@@ -19,6 +19,7 @@
 # along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
 
 
+from glob import glob
 import ConfigParser
 import os
 from ansible.module_utils.pycompat24 import get_exception
@@ -28,7 +29,7 @@ DOCUMENTATION = '''
 ---
 module: yum_repository
 author: Jiri Tyr (@jtyr)
-version_added: '2.1'
+version_added: '2.2'
 short_description: Add and remove YUM repositories
 description:
   - Add or remove YUM repositories in RPM-based Linux distributions.
@@ -151,6 +152,13 @@ options:
       - C(packages) means that only RPM package downloads should be cached (but
          not repository metadata downloads).
       - C(none) means that no HTTP downloads should be cached.
+  ignore_repo_files:
+    version_added: '2.2'
+    required: false
+    default: None
+    description:
+      - List of repo files which should be ignored during deletion in the
+        Managed Mode. Default is empty list (C([])).
   include:
     required: false
     default: null
@@ -240,7 +248,7 @@ options:
         expire.
       - Default value is 6 hours.
   name:
-    required: true
+    required: false
     description:
       - Unique repository ID.
       - This parameter is only required if I(state) is set to C(present) or
@@ -355,10 +363,19 @@ options:
       - Defines whether yum should verify SSL certificates/hosts at all.
   state:
     required: false
-    choices: [absent, present]
+    choices: [absent, present, mm_enabled, mm_disabled, managed]
     default: present
     description:
-      - State of the repo file.
+      - The state of the repo file (C(absent), C(present)) or the Managed Mode
+        (C(mm_enabled), C(mm_disabled), C(managed)).
+      - The C(mm_enabled) will enable Managed Mode. This state should be used
+        somewhere at the beginning of the playbook before any M(yum_repository)
+        module calls are made.
+      - The C(mm_disabled) will disable the Managed Mode.
+      - The C(managed) will initiate the deletion of unmanaged repo files if
+        the I(mm_enabled) was used before. This state should be used only
+        somewhere at the end of the playbook where M(yum_repository) module is
+        called no more.
   throttle:
     required: false
     default: null
@@ -447,6 +464,37 @@ EXAMPLES = '''
     gpgkey: http://server.com/keys/somerepo.pub
     gpgcheck: yes
     params: "{{ my_role_somerepo_params }}"
+
+#
+# Example of a playbook using the Managed Mode
+#
+
+- name: Initial play
+  hosts: all
+  tasks:
+    - name: Enable the Managed Mode
+      yum_repository:
+        state: mm_enabled
+
+- name: My play
+  hosts: testhost1
+  tasks:
+    - name: Create a repo file
+      yum_repository:
+        name: myrepo
+        description: Managed repo
+        baseurl: http://repo.server.com/path/to/the/repo
+
+- name: Final play
+  hosts: all
+  tasks:
+    - name: Delete unmanaged repo files
+      yum_repository:
+        state: managed
+        # These repo files won't be deleted even if they are not managed
+        ignore_repo_files:
+          - CentOS-Base
+          - CentOS-Media
 '''
 
 RETURN = '''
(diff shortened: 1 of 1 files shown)


jtyr:
@timss Please be so kind and review the code, the documentation and test this module for it's functionality in the normal and the Managed Mode.

timss:
Will do. I'll have a better look at it a bit later.

robynbergeron:
Thanks @jtyr. Since you are a maintainer of this module, we are marking this PR for inclusion.

packaging/os/yumrepo.py
```diff
@@ -549,40 +914,81 @@ def main():
...
+    if state in ['present', 'absent']:
+        # Rename "name" and "description" to ensure correct key sorting
+        module.params['repoid'] = module.params['name']
+        module.params['name'] = module.params['description']
+        del module.params['description']
+        del module.params['ignore_repo_files']
+
+        # Define repo file name if it doesn't exist
+        if module.params['file'] is None:
+            module.params['file'] = module.params['repoid']
```
timss: It might be somewhat obvious, but this will fail if an already existing repo file has a section (`repoid`) which name is not the same as the file name. 

Works as described using the explicit `file` parameter, but the documentation could possibly be updated to include a "warning" of sorts for the `name: Remove repository` example. 

For example, the default CentOS-Media.repo in CentOS 7:

```
$ cat CentOS-Media.repo                                                      
# CentOS-Media.repo
#
#  This repo can be used with mounted DVD media, verify the mount point for
#  CentOS-7.  You can use this repo and yum to install items directly off the
#  DVD ISO that we release.
#
# To use this repo, put in your DVD and use it with the other repos too:
#  yum --enablerepo=c7-media [command]
#  
# or for ONLY the media repo, do this:
#
#  yum --disablerepo=\* --enablerepo=c7-media [command]

[c7-media]
name=CentOS-$releasever - Media
baseurl=file:///media/CentOS/
        file:///media/cdrom/
        file:///media/cdrecorder/
gpgcheck=1
enabled=0
gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7
```

Will fail using:

```
- name: Remove CentOS-Media
  yumrepo:
    name: c7-media
    state: absent
```

jtyr: As you said, this is obvious that this module won't search through all repo files to find if such repo is already defined. That's up to the user to deal with that and specify the `file` parameter to target the right repo file if needed.

packaging/os/yumrepo.py
```diff
@@ -134,6 +174,14 @@
         are allowed. Substitution variables (e.g. C($releasever)) are honored
         here.
       - The list can also be a regular YAML array.
+  ip_resolve:
```
timss: While "whatever" is [official](http://lists.baseurl.org/pipermail/yum-devel/2010-April/006900.html), it's a [silly](https://www.mail-archive.com/yum-devel@lists.baseurl.org/msg01500.html) name. Would it make sense to divert from the official man page and specify that "whatever" is indeed a legitimate option and not "foobar"?

jtyr: The `whatever` is a keyword used in YUM to deal with the IP resolving. We can not change this here.

timss:
Code looks very good! 

Both normal and managed mode works great for me (add, remove, ignore) :+1:

packaging/os/yumrepo.py
timss: I'm aware. More so that I was wondering if maybe any explanation was appropriate, but maybe not.

jtyr: I think I will leave it as it is. This module is no a man page for YUM. If the user wants to know more what the option mean, he should check the man page.

jtyr:
Thanks for the review, @timss. I really appreciate it. Hopefully the PR will be merged soon and it will become a part of the 2.1 release.

packaging/os/yumrepo.py
timss: Fair enough, I agree.

timss:
Happy to be of a little help, don't mention it :-)

Sounds good, looking forward to it. I'll be using it in a local library until then.

timss:
@jtyr Now conflicts because of #1762.

jtyr:
@timss I have renamed the module to `yum_repository` and rebased against the `devel` branch. Please feel free to merge it now.

bcoca:
so many things added here, the 'managed mode' though is not something I think should be in the module itself. there are already ways to 'tidy' directories so I'm not in favor of building this into modules.

jtyr:
@bcoca Please could you enlighten me into the techniques available for tiding up directories from unmanaged files? I know that puppet has the `purge` option of the `file` type which can do that but I'm not aware of any equivalent in Ansible. The Managed Mode is important security feature which I believe should be available in this module for those who want to use it (the feature is implemented to be optional).

jtyr:
I have separated the Managed Mode from other changes in the module (mainly new parameters) in the PR #1772. Please feel free to merge it. I will also update this PR to cover only changes required for the Managed Mode.

jjshoe:
As far as I can tell this only useful when taking a machine in an unknown state, to a known state. Managed mode is because the user doesn't want to specify the known state.

I'm not a big fan of CM tools that do 'magic' under the hood. I think that's why most of us wind up on ansible. This seems like one of those magic things that adds little to no value. This work can be done in two tasks, easily. 

$0.02

jtyr:
The 'tidy' directory thing which @bcoca mentioned above is [this](https://github.com/bcoca/ansible-tidy).

cThrice:
I do not see this Tidy Role referenced as being equivalent.  

With a Tidy Role implementation you would need to have a pre-known list of directories you wish to "manage" and their associated files to be excluded... or have a play compile a running list of files modified that you wish to be in "managed" directories. Where as with a "managed" feature in a module not only do you keep any deletions within the task that truly is calling for it, you also keep it dynamic and simple.  No Running lists to be added to, no pre-compiled lists, just a simple flag.  Thus I would rather see more Ansible modules implemented a "managed" feature.

As for the known state objection:
This addition actually brings Ansible closer to defining a "known state".  This is because it is removing all "unknown files" that are not defined by Ansible.  I don't see this as magic any more than Puppet's Purge feature is magic.  It simply removes "unmanaged" files.

I apologize if I am missing some bigger point here, but without this feature, the entire yumrepo module is useless for maintaining a "known state".  And the suggested Tidy Role just seems Hackish.  The entire purpose of the yumrepo module is to manage a host's yum repos, which are controlled by files in a directory.  

Thank you @jtyr for your work on this.

jtyr:
Thanks @cThrice for your comment. I have discussed this feature with the Android devs and they suggested to implement it as a CallBack (CB) plugin. I [tried that](https://github.com/jtyr/ansible-purge_cb_plugin) but it has several issues comparing to the implementation in the module:
1. The CB plugin needs to call the module to perform the purge so there have to be some logic in the module anyway.
2. The call must be done per module and host sequentially which slows down the purging process significantly.

I still believe that the implementation in the module is the best option.

cThrice:
@jtyr I have to agree.  Why is this getting extra scrutiny?  Isn't this a extras module?   

Can someone tell me if I shouldn't plan on this being shipped? That way I can just start maintaining my own fork.

jjshoe:
I would NEVER run this on a box I manage, because it assumes an unknown state, but again, this can easily be done in a single task:

```
- name: Deploy static host reservation configuration
  shell: "/usr/bin/find /etc/yum.repos.d  -not -path /etc/yum.repos.d -not -path /etc/yum.repos.d/{{ yum_repos | join(' -not -path /etc/yum.repos.d/') }} -exec /bin/rm {} \\;"
```

I don't see why anyone would maintain a separate fork, or such a change would happen to a module, for something that should be a simple task.

jtyr:
@jjshoe Your task would work only if you would have previous knowledge which files you manage. This is very difficult to know if you don't collect the information in the module itself (or in a callback plugin). The Managed Mode takes care of this by recording all managed files into a file which is then processed across all machine. Simple.

jjshoe:
@jtyr - No, my task works when you know which files you want to exist, all files you don't know about will be removed. Did you try it? What bug did you hit? Steps to reproduce?

jtyr:
@jjshoe Managed files are those which you want to exist. To get such list is very difficult if you start using roles which you did not create. In your case, in order to compile the list of Managed Files, you have to walk through all roles/playbooks manually and record the names of yumrepo files. This is what the Managed Mode does for you for free.

cThrice:
A shell command is the solution?  For something that is needed to prevent duplicate entries which will bork yum?

Don't get me wrong, I totally see how the original module would work fine if used on a preknown environment.  But is it really the Ansible way to assume Ansible is being run in clean, known environment?  

This is the first time I heard this.  Your input would help me greatly, and give me a good conversation for my area's local meet up.

jjshoe:
@jtyr 

```
grep baseurl
```

Though, I personally consider it a great no no to not read through and thoroughly understand someone elses role before using them. I would never run untrusted code on servers, and a third party role is no different imho. 

My vote would still be a no. It sounds like people want magic because they don't want to do things 'the right way'.

I have no further input to provide. I simply provided yet another way to do this.

jtyr:
@jjshoe I do review third party roles before I use them. I just don't see a reason why should I maintain a list of files I want to manage if the module can do it for me. On top of that, in your case you assume the yumrepo file name is written in the role and is the same on all servers. In reality, the name can be a variable which can be different on every host depending on some other attributes. So you would have to keep list of yumrepo files you want to manage per host which becomes a nightmare to maintain. Again, the module takes care of this for you for free.

bcoca:
not for free, as the module needs the same information provided that you would need to manage the files independently. 

This feature does NOT remove the need to manage the information, also it assumes that each invocation of this modules has 'perfect' information, the same you would need to manage it externally.

jtyr:
@bcoca That's not true. The module doesn't need the same information. It collects all the information by itself so it actually does remove the need to manage the information.

There is only one assumption - the module is used exclusively to create yumrepo files. If some other method is used (e.g. templates or static files), those must be passed to the `ignore_repo_files` parameter.

gregdek:
Thanks @jtyr for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

jtyr:
I have updated the `version_added` and the PR is `ready_for_review` now.

gregdek:
Thanks @jtyr. To the current maintainers, @jtyr please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

gregdek:
Thanks @jtyr for this PR. Unfortunately, it is not mergeable in its current state due to merge conflicts. Please rebase your PR. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

gregdek:
@jtyr A friendly reminder: this pull request has been marked as needing your action. If you still believe that this PR applies, and you intend to address the issues with this PR, just let us know in the PR itself and we will keep it open pending your changes. When you do address the issues, please respond with ready_for_review in your comment, so that we can notify the maintainer.

jtyr:
This PR is `ready_for_review`.

gregdek:
Thanks @jtyr. To the current maintainers, @jtyr please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

gregdek:
Thanks @jtyr for this PR. Unfortunately, it is not mergeable in its current state due to merge conflicts. Please rebase your PR. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

For help on how to do this cleanly please see http://docs.ansible.com/ansible/community.html#contributing-code-features-or-bugfixes

PR #1730, opened 2016-02-24
Adding jenkins_plugin module

jtyr:
##### Issue Type:

New Module Pull Request
##### Plugin Name:

`jenkins_plugin`
##### Ansible Version:

For Ansible 2.1
##### Summary:

This PR is adding a new module which helps to manage Jenkins plugins. The module supports installation/uninstallation and pinning/unpinning of a module. The `latest` or a specific version can be installed.
##### Example output:

```
$ ansible-playbook -i hosts -l jenkins-master site.yaml

PLAY [Jenkins Master] **********************************************************

ok: [jenkins-master]

TASK [Install Jenkins plugins] *************************************************
ok: [jenkins-master] => (item={u'version': u'latest', u'name': u'build-pipeline-plugin'})
ok: [jenkins-master] => (item={u'version': u'latest', u'name': u'ant'})
ok: [jenkins-master] => (item={u'version': u'latest', u'name': u'ldap'})

TASK [Pin plugin] **************************************************************
changed: [jenkins-master]

jenkins-master             : ok=3    changed=1    unreachable=0    failed=0
```


sivel:
```
./web_infrastructure/jenkins_plugin.py
ERROR: requests import found, should use ansible.module_utils.urls instead
```

jtyr:
@sivel Should be fixed now.

web_infrastructure/jenkins_plugin.py
```diff
@@ -0,0 +1,753 @@
...
+        # Check if we got valid data
+        if 'plugins' not in plugins_data:
+            self.module.fail_json(msg="No valid plugin data found.")
+
+        # Create final list of installed/pined plugins
+        self.is_installed = False
+        self.is_pinned = False
+        self.is_enabled = False
+
+        for p in plugins_data['plugins']:
```
jjshoe, line 397: No break out of this loop?

web_infrastructure/jenkins_plugin.py
```diff
@@ -0,0 +1,753 @@
...
+                    changed = True
+
+        return changed
+
+    def __download_updates(self):
+        updates_file = 'update-center.json'
+        updates_tmp_file = "%s/%s" % (tempfile.gettempdir(), updates_file)
+        download_updates = True
+
+        # Check if we need tu download new updates file
```
jjshoe: typo

web_infrastructure/jenkins_plugin.py
```diff
@@ -0,0 +1,753 @@
...
+        try:
+            fd.close()
+        except IOError, e:
+            self.module.fail_json(
+                msg='Cannot close the plugin file %s.' % f,
+                details=str(e))
+
+        # Change owner and permissions of the file
+        try:
+            os.chmod(f, 0664)
```
jjshoe: Shouldn't this be a param?

web_infrastructure/jenkins_plugin.py
```diff
@@ -0,0 +1,753 @@
...
+    choices: [absent, present, pinned, unpinned, enabled, disabled]
+    default: present
+    description:
+      - Desired plugin state.
+  timeout:
+    required: false
+    default: None
+    description:
+      - Server connection timeout in secs.
+  updates_expiration:
```
jjshoe, line 88: Is this the correct name to use? seems like the apt module uses cache_valid_time or something like that. Is there consistency amongst modules out there?

web_infrastructure/jenkins_plugin.py
```diff
@@ -0,0 +1,753 @@
...
+    # Module settings
+    module = AnsibleModule(
+        argument_spec=dict(
+            jenkins_group=dict(default='jenkins'),
+            jenkins_home=dict(default='/var/lib/jenkins'),
+            jenkins_user=dict(default='jenkins'),
+            name=dict(required=True),
+            params=dict(),
+            state=dict(
+                choices=[
```
jjshoe: No latest here?

jtyr: The `latest` can be specified in the `version` param.

jjshoe: take a look at the apk and apt module

jtyr: But those modules do not allow to specify `version`. I think it's more logical to have the `latest` in the `version` parameter than in the state.

web_infrastructure/jenkins_plugin.py
jtyr, line 88: The `apt` module is the only one using `cache_valid_time`.

web_infrastructure/jenkins_plugin.py
jtyr: Fixed

web_infrastructure/jenkins_plugin.py
jtyr: Fixed

web_infrastructure/jenkins_plugin.py
jtyr, line 397: Fixed

web_infrastructure/jenkins_plugin.py
jjshoe: apk does not allow you to, you are correct there. Both apt and yum do.

jtyr: Both `apt` and `yum` do not allow to specify version as an extra parameter. The version is always as part of the package name. As I said above, I believe that it's more logical to have the `latest` specified in the `version` parameter than in the `state`.

jjshoe: https://github.com/ansible/ansible-modules-core/blob/devel/packaging/os/apt.py#L121

jjshoe: https://github.com/ansible/ansible-modules-core/blob/devel/packaging/os/yum.py#L167

jtyr: Yes, this is exactly what I was saying. The version must be as part of the name of the package. There is no extra param for version. This module has an extra `version` param, this is why I believe the `latest` should be specified in that param instead of in the `state`.

jjshoe: I guess my vote is for consistency

jtyr: I really appreciate your opinion but you are comparing two different things. If this module wouldn't have the `version` param, then I would absolutely agree with you that there should be `state: latest`. But as there is the `version` param, it doesn't feel right to have the **`latest` version** as a `state`.

jtyr:
@jjshoe Thank you very much for reviewing the code. If you think the module follows the [module guidelines](http://docs.ansible.com/ansible/developing_modules.html#module-checklist), please could you comment here with "passes_guidelines"? If you also tested the functionality and it worked for you, please could you also comment here with "works_for_me"?

gregdek:
Thanks @jtyr for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

xenexis:
I just finished testing this module and it looks like what we want it, definitely works for me/us.
Previously I was managed the plugins with ansible command/shell, but this is a far better approach.
Also far as I can see, it follows the module guidelines, it passes the guidelines for me.
Thanks Jiri! 

(adding shipit here, thanks!)

jimi-c:
@jtyr I'm in agreement with @jjshoe. The `state` field is typically where we specify latest, while the version is part of the `name`. I think we should stick with the convention of the other modules here.

jtyr:
Thanks for your comment, @jimi-c. As I explained to @jjshoe, the `jenkins_plugin` module doesn't specify the version as part of the `name` like, for example, in the case of the `yum` module. Version of the `jenkins_plugin` is specified in a separate parameter called [`version`](https://github.com/ansible/ansible-modules-extras/pull/1730/files#diff-e24f65156cc9b7c1dfe44242891e885aR121). Therefore I think it's more logical to specify the `latest` version in the `version` parameter than in the `state`.

If you believe that my logic is wrong, please could you point me to another Ansible module having an extra parameter for version and still using `state: latest`?

jjshoe:
http://docs.ansible.com/ansible/pip_module.html
http://docs.ansible.com/ansible/npm_module.html
http://docs.ansible.com/ansible/gem_module.html

I'm sure there's plenty more.

jtyr:
Thanks for the links, @jjshoe. I can see now that it's really quite common to have `state: latest` even if `version` parameter is available. Although I think it's illogical, I'm going to change the module to follow the _standard_.

jtyr:
@jjshoe @jimi-c I have modified the module to support the `state: latest`. Please feel free to review/merge it.

gregdek:
Thanks @jtyr for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

jtyr:
I have updated the `version_added` and it's `ready_for_review` now.

gregdek:
Thanks @jtyr for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

gregdek:
Thanks @jtyr for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

gregdek:
@jtyr A friendly reminder: this pull request has been marked as needing your action. If you still believe that this PR applies, and you intend to address the issues with this PR, just let us know in the PR itself and we will keep it open pending your changes. When you do address the issues, please respond with ready_for_review in your comment, so that we can notify the maintainer.

jtyr:
This PR is `ready_for_review`.

gregdek:
Thanks @jtyr for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

gregdek:
Thanks @jtyr for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

jtyr:
@gregdek I have rebased this PR and it's 'ready_for_review' now.

gregdek:
Thanks @jtyr for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

abadger:
Reviewed at today's new module meeting.  Bockers:
- We saw that the comments here say that latest has been moved to state but that doesn't seem to be reflected in the code.
- Looks like you maybe should be using module.atomic_move() to move the downloaded plugin file into place so that getting interrupted in the middle of writing the file doesn't leave a partially written plugin file.
- The file_mode parameter is buggy.  If the user specifies a file_mode as a number in their playbook, they could get an unexpected value here.  ie: file_mode: 0644 => the decimal integer 420 => arg_spec validation turns it into the string "420" => then the module's code turns that into the decimal 272.  The ansible core modules that take mode arguments specify that the type of the parameter is raw and then use module.set_mode_if_different() to set the mode if required.

Style notes.  These are optional:
- Generally imports from the stdlib should come before imports from other sources.  So the imports should be reordered to do `from ansible.module_utils.pycompat24 import get_exception` after importing from the stdlib.  Additionally, there is now no drawback to putting module_utils imports at the top of the module so you should move the imports of basic and urls to be near the pycompat24 import.
- We don't generally use double underscore to mark things private.  Python convention works on the principle of "consenting adults" so we use a single underscore to mark things as private instead.

abadger:
needs_revision

web_infrastructure/jenkins_plugin.py
```diff
@@ -0,0 +1,781 @@
...
+
+        return changed
+
+    def __download_updates(self):
+        updates_file = 'update-center.json'
+        updates_tmp_file = "%s/%s" % (tempfile.gettempdir(), updates_file)
+        download_updates = True
+
+        # Check if we need to download new updates file
+        if os.path.isfile(updates_tmp_file):
```
mattclay: Can this be made to work without caching the updates file across module runs?

We generally avoid saving state or caching files on the remote host between module runs, as it tends to make the module more prone to unexpected failures or behavior:
- Different remote users running the module may not all have access to the cached file, which would cause module failure. If they do, the permissions are probably too open, allowing unintended access to the file.
- Different playbooks may have different configuration for using the cached file (in this case the expiration).
- If the cached file is corrupted, the module will be unable to recover until the cached file is expired.
- The cache file could be corrupted if there are two concurrent runs of the module on the remote system.

Some of these issues could be solved by:
- Caching the file in the remote user's home directory instead of the system-wide temp dir.
- Writing the file to a temp location and verifying it can be parsed before saving it for later use.

However, it would be better if the caching could be avoided all together.

gregdek:
Thanks @jtyr for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

jtyr:
Thanks for the review, @abadger and @mattclay.

@abadger, I have implemented all your comments.

@mattclay, I believe that the caching is quite important as otherwise it would require to download 1MB+ of data for every `jenkins_plugin` task execution. That might have significant impact on users with slower internet connection and also on the Jenkins' infrastructure. Therefore I implemented at least your suggestions - store the cache file into a temp file, check if the data are valid and then move it into the user's home directory.

The code is `ready_for_review` now.

gregdek:
Thanks @jtyr for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

abadger:
@jtyr Is mode being used anymore?

jtyr:
@abadger Yes, the `mode`, `owner`, `group` and any other file attributes are used by the `module.set_fs_attributes_if_different()` function. I list them in the documentation and in the list of module params only because I want them to have a default value.

abadger:
Ah yeah-- I just grepped for mode to see where it was used but now I see they're being aggregated into file_args via another helper function just before that.  Cool all the things on my list are resolved.  Just have to figure out if the caching issue is resolved today.

web_infrastructure/jenkins_plugin.py
```diff
@@ -0,0 +1,815 @@
...
+                    self.params['version'] not in [None, 'latest'] or
+                    md5sum_old is None):
+
+                # Download the plugin file directly
+                r = self._download_plugin(plugin_url)
+
+                # Write downloaded plugin into file if checksums don't match
+                if md5sum_old is None:
+                    # No previously installed plugin
+                    self._write_file(plugin_file, r)
```
mattclay: It appears the conditional for check mode is missing here.

web_infrastructure/jenkins_plugin.py
```diff
@@ -0,0 +1,815 @@
...
+                    # Get data for the MD5
+                    data = r.read()
+
+                    # Make new checksum
+                    md5sum_new = hashlib.md5(data).hexdigest()
+
+                    # If the checksum is different from the currently installed
+                    # plugin, store the new plugin
+                    if md5sum_old != md5sum_new:
+                        self._write_file(plugin_file, data)
```
mattclay: It appears the conditional for check mode is missing here.

jtyr: Good catch. Fixed.

web_infrastructure/jenkins_plugin.py
jtyr: Fixed.

web_infrastructure/jenkins_plugin.py
```diff
@@ -0,0 +1,819 @@
...
+                        msg="Cannot load JSON data from the tmp updates file.",
+                        details=e.message)
+
+                break
+
+            i += 1
+
+        # Move the updates file to the right place if we could read it
+        if download_updates:
+            self.module.atomic_move(updates_file, updates_file_orig)
```
mattclay, line 624: Check mode should be observed here too.

mattclay:
@jtyr Did you look at using the Jenkins API to work with updates? I realize an implementation using the API would be completely different than what you currently have. I'm just wondering if there was a reason not to use the Jenkins API.

jtyr:
@mattclay I don't remember why I have chosen the way how I deal with the updates. I can have a look if I can trigger the download of the updates file via API and then read plugins properties via API. That could actually help to get rid of the caching in the Ansible module.

web_infrastructure/jenkins_plugin.py
jtyr, line 624: I would leave this as it is now because the it doesn't change anything on the Jenkins side but only refreshes the caching file.

web_infrastructure/jenkins_plugin.py
```diff
@@ -0,0 +1,819 @@
...
+                    file_args, changed)
+            else:
+                # See the comment above
+                changed = True
+
+        return changed
+
+    def _download_updates(self):
+        updates_filename = 'update-center.json'
+        updates_file = "%s/%s" % (os.path.expanduser('~'), updates_filename)
```
mattclay: This comment assumes you don't end up switching to the API for module installs.

This path should probably be configurable. I'd recommend one of the following:
- Make it required, so it's up to the user to indicate where to store it. This way there are no surprises for the user with the cache file created by the module.
- Make it optional with a default like `~/.ansible/tmp/jenkins-plugin-cache.json`. This way it's clear where the file came from, which module it belongs to, and that it's a temporary file so is safe to delete.

jtyr: The second option looks better to me. Less required parameters is better. Let me fix it.

jtyr:
@mattclay I quickly tried to have a look if I can pull all necessary information via Jenkins API but I can not find all the necessary endpoints. The only information I can find is the list of currently installed plugins ;o(

mattclay:
@jtyr When you have additional changes, you can just push additional commits, rather than updating the one you already have. It makes review of the additional changes easier. When the PR is merged, we'll use the squashing feature on GitHub to merge it as a single commit.

web_infrastructure/jenkins_plugin.py
jtyr: Fixed.

mattclay:
@jtyr Merged. Thanks for the new module!

PR #1738, opened 2016-02-25
Elb tags

nand0p:
##### Issue Type:

- New Module Pull Request
##### Plugin Name:

elb_tags
##### Ansible Version:

```
ansible 1.9.4
```
##### Summary:

Module allows for adding and removing tags from ELBs

As ELB's do not have an AWS Resource-ID, the ec2_tag module
cannot be used to tag ELBs.  This new mod allows for tagging ELBs
and returns changed=True only when tags are actually changed.

### DIFF
--- a/cloud/amazon/elb_tags.py
+++ b/cloud/amazon/elb_tags.py
@@ -1,0 +1,119 @@
+#!/usr/bin/python
+
+DOCUMENTATION = '''
+---
+module: elb_tags
+short_description: allows manipulation of elb tags
+options:
+
+  elb_name:
+    description:
+      - Name of the elb of which to modify tags
+    required: true
+    default: null
+    aliases: []
+
+  tags:
+    description:
+      - list of dictionaries of tags
+    required: true
+
+  state:
+    description:
+      - add or remove tags
+    required: true
+
+
+extends_documentation_fragment: aws
+'''
+
+EXAMPLES = '''
+    # Adds elb tag
+    elb_tag:
+        elb_name: foo
+        tags: "{{ tags_dict }}"
+        state: present
+'''
+
+try:
+    import boto3
+    import botocore
+    HAS_BOTO3 = True
+except ImportError:
+    HAS_BOTO3 = False
+
+def check_if_tags_exist(result):
+    tags_new = [ values for tags in result['tags']
+                        for values in tags.values() ]
+    tags_existing = [ values for tags in result['existing']
+                             for values in tags.values() ]
+    return set(tags_new).issubset(set(tags_existing))
+
+def create_or_update_elb_tags(client, module, result):
+    if check_if_tags_exist(result):
+        module.exit_json(**result)
+    try:
+        result['msg'] = client.add_tags(
+            LoadBalancerNames = [ result['elb_name'] ],
+            Tags = result['tags']
+        )
+        result['changed'] = True
+    except botocore.exceptions.ClientError:
+        result['msg'] = 'Failed to create/update elb tags due to error: ' + \
+                        traceback.format_exc()
+        module.fail_json(**result)
+    else:
+        module.exit_json(**result)
+
+
+def delete_elb_tags(client, module, result):
+    try:
+        result['msg'] = client.remove_tags(
+            LoadBalancerNames = [ result['elb_name'] ],
+            Tags = result['tags']
+        )
+        result['changed'] = True
+    except botocore.exceptions.ClientError:
+        result['msg'] = 'Failed to create/update elb tags due to error: ' + \
+                        traceback.format_exc()
+        module.fail_json(**result)
+    else:
+        module.exit_json(**result)
+
+def main():
+    argument_spec = ec2_argument_spec()
+    argument_spec.update(dict(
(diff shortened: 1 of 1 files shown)


robynbergeron:
Thanks @nand0p for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

smiller171:
shipit!

manicminer:
Would it be feasible to add a `tags` argument to the `ec2_elb_lb` module instead? I feel that would be a cleaner approach to avoid unnecessary module fragmentation. I understand that tagging ELBs requires boto3, but you could conditionally import boto3 based on the arguments passed by the user (with corresponding conditional error handling if boto3 is not present).

smiller171:
@tombamford I think there was a decision to try to keep things a bit more separated in the AWS modules, as they have gotten large and monolithic. @defionscode I think you may know more about this.

juliedavila:
I would prefer to see what @tombamford suggested, but on ec2_tag module. I don't want to set a precedence for tagging modules for each type of taggable aws resource.

nand0p:
@defionscode the issue is that the main required option of ec2_tag is "resource-id", which elb's do not have. 

do we want to restructure the ec2_tag mod to support this? or have separate tagging mod for this one aws resource that has no resource-id?

juliedavila:
That's a good point, thinking out loud, what do you think of simply adding support to ec2_tag so that it can also accept the ELB's DNS record as an argument to the resource-id param? The logic could then simply be reduced to detecting `.elb.amazonaws.com` (and boto3) in the string, at which point the module would know to do operations against an ELB.

nand0p:
Im thinking would we want to pollute the existing ec2_tag module with conditional boto3?  and also the misleading ResourceID-as-FQDN option?  these are two things that make elb tagging unique, and so may make a case for this mod.

However, i can certainly close this PR, and open one up to update ec2_tag for boto3 and elb fqdn as resource-id.... does anyone else have an opinion on this?

Am i correct in thinking the logic you are suggesting @defionscode is if (resource: stringmatch .elb.amazonaws.com && not HAS_BOTO3) then fail? this is certainly doable.  

The only issue would be if amazon ever changes the top level domain(s) for elb, which is highly unlikely. Also, i can make a feature request to amazon to add resource-id's to ELBs.

juliedavila:
Of course I want the opinion of others on this as well before an official decision is made. The 'pollution' should be minimal. For resource id, it would just be, perhaps, a matter of adding an alias to that parameter (for readability) and updating the documentation string.

And a bit of tweak to the logic, it would be more of a 

``` python
if resource.endswith('.elb.amazonaws.com`):
    if not HAS BOTO3:
        module.fail_json(msg="...")
```

this way it only fails on missing boto3 if the user is attempting to tag an elb.

I'm highly skeptical AWS would change the top level domain in any sort of sudden way as it would mess with DNS all over the place, and also skeptical they'll add resource_ids in any sort of quick manner.

Anyways, definitely looking for more opinions.

smiller171:
any solution is good with me. I currently have to use a command module to tag ELBs with the aws cli.

Personally I think tagging for everything should live in the module for that resource and getting rid of the ec2-tag module, but I could understand keeping the ec2-tag module for messing with tags after resources have been created.

juliedavila:
@wimnat if you have time, could you give you thoughts on the comments here? Need a sanity check on my idea.

manicminer:
I would be wary of having logic that introduces a dependency on outside factors that are not part of an API contract, i.e. relying on `*.elb.amazonaws.com` always being the same. You can be sure that something you expect to stay the same, will at some point change. Also my personal opinion is that mangling the `resource` argument to also accept this edge case is something that could be designed better (and it opens the door to a future list of such edge cases).

I still think that handling this with the `ec2_elb_lb` module handle is cleaner. But in lieu of that, how about adding another argument to the `ec2_tag` module? Something like `elb_name`, and make it mutually exclusive to the `resource` argument. Other parameters could feasibly be added in the same way without depending on brittle parsing logic to automagically figure out what needs to happen.

wimnat:
@defionscode I'm personally in favor of @tombamford's approach - modifying the ec2_elb_lb module.

I think we should follow AWS API approach whenever we write Ansible modules.  When ec2_elb_lb was initially written, AWS API didn't allow lb tagging via the elb API.

Amazon have now settled on their tagging approach and allow it pretty much with every resource including ELBs.  I suggest we do the same and add any tagging functionality to the respective modules with the hope of one day no longer needed ec2_tag.

So, as I said, my preferred approach would be to add the boto3 include and tagging functioanlity to ec2_elb_lb.

juliedavila:
Alright, after reading the thoughts @tombamford and @wimnat gave, I'm inclined to @tombamford's suggestion of enhancing the current ec2_elb_lb module. @nand0p does that sound good to you? If so, you can close this PR and make a new one...or I suppose you could just update this PR as well. Whatever works for you.

nand0p:
sounds good. i will go ahead with this direction.

nand0p:
cannot reuse this PR, as ec2_elb_lb is in core, not extras.

PR #1753, opened 2016-02-28
Fixes #1616, check whether interface-list exits before querying its length

### LINKED ISSUE #1616: zabbix_host TypeError on host update without interfaces
I get the following exception when updating a host without specifying interfaces:

> An exception occurred during task execution. The full traceback is:
> Traceback (most recent call last):
>   File "/home/plldp/.ansible/tmp/ansible-tmp-1454663921.27-43711125456205/zabbix_host", line 2430, in <module>
>     main()
>   File "/home/plldp/.ansible/tmp/ansible-tmp-1454663921.27-43711125456205/zabbix_host", line 464, in main
>     exist_interfaces, zabbix_host_obj, proxy_id):
>   File "/home/plldp/.ansible/tmp/ansible-tmp-1454663921.27-43711125456205/zabbix_host", line 334, in check_all_properties
>     if self.check_interface_properties(exist_interfaces, interfaces):
>   File "/home/plldp/.ansible/tmp/ansible-tmp-1454663921.27-43711125456205/zabbix_host", line 293, in check_interface_properties
>     if len(interfaces) >= 1:
> TypeError: object of type 'NoneType' has no len()

I would suggest checking if interfaces is set. Perhaps something like this in zabbix-host.py:

``` python
    # check the exist_interfaces whether it equals the interfaces or not
    def check_interface_properties(self, exist_interface_list, interfaces):
        interfaces_port_list = []
        if interfaces is not None:
          if len(interfaces) >= 1:
            for interface in interfaces
```

Line "if interfaces is not None:" was added.

eikef:
##### Issue Type:

- Bugfix Pull Request
##### Plugin Name:

zabbix_host
##### Summary:

Fixes #1616 

An empty interface list would cause a stacktrace. This commit fixes that.

However, the semantics of passing an empty interface list are unclear since Zabbix does not allow a host to have no interfaces. Right now the ansible module will simply ignore an empty interface list and leave it on the Zabbix server as is.

### DIFF
--- a/monitoring/zabbix_host.py
+++ b/monitoring/zabbix_host.py
@@ -289,9 +289,11 @@ class Host(object):
     # check the exist_interfaces whether it equals the interfaces or not
     def check_interface_properties(self, exist_interface_list, interfaces):
         interfaces_port_list = []
-        if len(interfaces) >= 1:
-            for interface in interfaces:
-                interfaces_port_list.append(int(interface['port']))
+
+        if interfaces is not None:
+            if len(interfaces) >= 1:
+                for interface in interfaces:
+                    interfaces_port_list.append(int(interface['port']))
 
         exist_interface_ports = []
         if len(exist_interface_list) >= 1:


gregdek:
Thanks @eikef. Since you are a maintainer of this module, we are marking this PR for inclusion.

Issue #1759, opened 2016-02-29
zabbix modules use deprecated "exists" api methods that are removed in zabbix 3.0


kepstin:
##### Issue Type:
- Bug Report
##### Plugin Name:

I hit the error in:
zabbix_maintenance

These modules also appear to be affected:
zabbix_group
zabbix_host
zabbix_hostmacro
##### Ansible Version:

```
ansible 1.9.4 (stable-1.9 309996504f) last updated 2015/11/26 11:27:04 (GMT -400)
  lib/ansible/modules/core: (detached HEAD eda33a45e0) last updated 2015/06/17 11:13:07 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD 8c01513e1e) last updated 2015/06/17 11:13:07 (GMT -400)
  configured module search path = None
```

(I've confirmed that the issue still exists in the current ansible-modules-extra via source inspection)
##### Ansible Configuration:

N/A
##### Environment:

N/A
##### Summary:

When using the zabbix_\* modules listed above against a zabbix 3.0 server, you get errors like the following:

```
TASK: [zabbix_maintenance ] *************************************************** 
failed: [host.example.com -> localhost] => {"failed": true}
msg: Failed to check maintenance Ansible host.example.com existance: (u'Error -32602: Invalid params., Incorrect method "maintenance.exists". while sending {"params": {"name": "Ansible host.example.com"}, "jsonrpc": "2.0", "method": "maintenance.exists", "auth": "REDACTED", "id": 1}', -32602)
```

This is because the module is calling the "exists" method that was deprecated in zabbix 2.4, and removed in zabbix 3.0:
https://www.zabbix.com/documentation/3.0/manual/api/changes_2.4_-_3.0

There isn't really a direct replacement; it might make sense to do a 'get', then depending on the status either take no action, or do an 'update' or 'create'. (This shouldn't reduce compatibility with older zabbix versions, I think.)

eikef:
When those 3 PRs are committed the issue should be solved. Thanks :-)

gregdek:
Thanks @eikef!

wellumies:
.exists still in use in ansible (2.0.2.0)

[...]

PR #1763, opened 2016-02-29
Replace deprecated exists API for Zabbix 3.0 compatibility

eikef:
##### Issue Type:
- Bugfix Pull Request
##### Plugin Name:

zabbix_group
zabbix_hostmacro
##### Summary:

Final partial fix for #1759 

This replaces a call to the deprecated exists-api in zabbix_group, making it compatible with Zabbix 3.0 (and keeping compatibility with previous versions). Also removed an unused method in zabbix_hostmacro referencing the same API.

### DIFF
--- a/monitoring/zabbix_group.py
+++ b/monitoring/zabbix_group.py
@@ -114,7 +114,7 @@ class HostGroup(object):
         try:
             group_add_list = []
             for group_name in group_names:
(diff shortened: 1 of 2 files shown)


gregdek:
Thanks @eikef. Since you are a maintainer of this module, we are marking this PR for inclusion.

PR #1768, opened 2016-03-01
Module to manage Cloudflare DNS records

mgruener:
##### Issue Type:
- New Module Pull Request
##### Plugin Name:

cloudflare_dns
##### Ansible Version:

```
ansible 2.1.0 (devel 71402abf21) last updated 2016/03/01 22:18:47 (GMT +200)
  lib/ansible/modules/core: (detached HEAD cd9a7667aa) last updated 2015/12/02 20:42:01 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD 3c4f954f0f) last updated 2015/12/02 20:42:04 (GMT +200)
  config file = 
```
##### Summary:

A module to manage DNS records hosted at Cloudflare.
##### Example output:

```
{
    "changed": false, 
    "result": {
        "records": [
            {
                "content": "111.222.222.111", 
                "created_on": "2016-02-16T19:11:51.905832Z", 
                "id": "9759964ac78087ef8746bc2e2df53b88", 
                "locked": false, 
                "meta": {
                    "auto_added": false
                }, 
                "modified_on": "2016-02-16T19:11:51.905832Z", 
                "name": "www.foo.bar", 
                "proxiable": false, 
                "proxied": false, 
                "ttl": 1, 
                "type": "A", 
                "zone_id": "e87ede0b0583229f94029d2e6b73d85a", 
                "zone_name": "foo.bar"
            }
        ]
    }
}
```

### DIFF
--- a/network/cloudflare_dns.py
+++ b/network/cloudflare_dns.py
@@ -1,0 +1,536 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2016 Michael Gruener <michael.gruener@chaosmoon.net>
+#
+# This file is (intends to be) part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+import json
+import urllib
+
+DOCUMENTATION = '''
+---
+module: cloudflare_dns
+author: "Michael Gruener (@mgruener)"
+version_added: "2.1"
+short_description: manage Cloudflare DNS records
+description:
+   - "Manages dns records via the Cloudflare API, see the docs: U(https://api.cloudflare.com/)"
+options:
+  account_api_token:
+    description:
+      - "Account API token. You can obtain your API key from the bottom of the Cloudflare 'My Account' page, found here: U(https://www.cloudflare.com/a/account)"
+    required: true
+  account_email:
+    description:
+      - "Account email."
+    required: true
+  port:
+    description: Service port. Required for C(type=SRV)
+    required: false
+    default: null
+  priority:
+    description: Record priority. Required for C(type=MX) and C(type=SRV)
+    required: false
+    default: "1"
+  proto:
+    description: Service protocol. Required for C(type=SRV)
+    required: false
+    choices: [ 'tcp', 'udp' ]
+    default: null
+  record:
+    description:
+      - Record to add. Required if C(state=present). Default is C(@) (e.g. the zone name)
+    required: false
+    default: "@"
+    aliases: [ "name" ]
+  service:
+    description: Record service. Required for C(type=SRV)
+    required: false
+    default: null
+  solo:
+    description:
+      - Whether the record should be the only one for that record type and record name. Only use with C(state=present)
+      - This will delete all other records with the same record name and type.
+    required: false
+    default: null
+  state:
+    description:
+      - Whether the record(s) should exist or not
+    required: false
+    choices: [ 'present', 'absent' ]
+    default: present
+  timeout:
+    description:
+      - Timeout for Cloudflare API calls
+    required: false
+    default: 30
+  ttl:
+    description:
+      - The TTL to give the new record. Min 1 (automatic), max 2147483647
+    required: false
+    default: 1 (automatic)
+  type:
+    description:
+      - The type of DNS record to create. Required if C(state=present)
+    required: false
+    choices: [ 'A', 'AAAA', 'CNAME', 'TXT', 'SRV', 'MX', 'NS', 'SPF' ]
+    default: null
+  value:
+    description:
+      - The record value. Required for C(state=present)
+    required: false
+    default: null
+    aliases: [ "content" ]
+  weight:
+    description: Service weight. Required for C(type=SRV)
+    required: false
+    default: "1"
+  zone:
+    description:
+      - The name of the Zone to work with (e.g. "example.com"). The Zone must already exist.
+    required: true
+    aliases: ["domain"]
+'''
+
+EXAMPLES = '''
+# create a test.my.com A record to point to 127.0.0.01
+- cloudflare_dns:
+    zone: my.com
+    record: test
+    type: A
+    value: 127.0.0.1
+    account_email: test@example.com
+    account_api_token: dummyapitoken
+  register: record
+
+# create a my.com CNAME record to example.com
+- cloudflare_dns:
+    zone: my.com
+    type: CNAME
+    value: example.com
+    state: present
+    account_email: test@example.com
+    account_api_token: dummyapitoken
+
+# change it's ttl
+- cloudflare_dns:
+    zone: my.com
+    type: CNAME
+    value: example.com
+    ttl: 600
+    state: present
+    account_email: test@example.com
+    account_api_token: dummyapitoken
+
+# and delete the record
+- cloudflare_dns:
+    zone: my.com
+    type: CNAME
+    value: example.com
+    state: absent
+    account_email: test@example.com
+    account_api_token: dummyapitoken
+
+# create TXT record "test.my.com" with value "unique value"
+# delete all other TXT records named "test.my.com"
+- cloudflare_dns:
+    domain: my.com
+    record: test
+    type: TXT
+    value: unique value
+    state: present
+    solo: true
+    account_email: test@example.com
+    account_api_token: dummyapitoken
+
+# create a SRV record _foo._tcp.my.com
+- cloudflare_dns:
+    domain: my.com
+    service: foo
+    proto: tcp
+    port: 3500
+    priority: 10
+    weight: 20
+    type: SRV
+    value: fooserver.my.com
+'''
+
+RETURN = '''
+records:
+    description: >
+      List containing the records for a zone or the data for a newly created record.
+      For details see https://api.cloudflare.com/#dns-records-for-a-zone-properties.
+    returned: success/changed after record creation
+    type: list
+'''
+
+class CloudflareAPI(object):
+
+    cf_api_endpoint = 'https://api.cloudflare.com/client/v4'
+    changed = False
+
+    def __init__(self, module):
+        self.module            = module
+        self.account_api_token = module.params['account_api_token']
+        self.account_email     = module.params['account_email']
+        self.port              = module.params['port']
+        self.priority          = module.params['priority']
+        self.proto             = module.params['proto']
+        self.record            = module.params['record']
+        self.service           = module.params['service']
+        self.is_solo           = module.params['solo']
+        self.state             = module.params['state']
+        self.timeout           = module.params['timeout']
+        self.ttl               = module.params['ttl']
+        self.type              = module.params['type']
+        self.value             = module.params['value']
+        self.weight            = module.params['weight']
+        self.zone              = module.params['zone']
+
+        if self.record == '@':
+            self.record = self.zone
+
+        if (self.type in ['CNAME','NS','MX','SRV']) and (self.value is not None):
+            self.value = self.value.rstrip('.')
+
+        if (self.type == 'SRV'):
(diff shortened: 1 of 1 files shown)


network/cloudflare_dns.py
```diff
@@ -0,0 +1,448 @@
...
+version_added: "2.1"
+short_description: manage Cloudflare DNS records
+description:
+   - "Manages dns records via the Cloudflare API, see the docs: U(https://api.cloudflare.com/)"
+options:
+  account_api_token:
+    description:
+      - "Account API token. You can obtain your API key from the bottom of the Cloudflare 'My Account' page, found here: U(https://www.cloudflare.com/a/account)"
+    required: true
+    default: null
```
resmo: remove the default: null when the param is required, see http://docs.ansible.com/ansible/developing_modules.html#module-checklist for more infos

resmo:
This module is a good candidate to split the "list" functionality into a separate facts module (preferred). The shared code of these modules could be excluded in to https://github.com/ansible/ansible/tree/devel/lib/ansible/module_utils

network/cloudflare_dns.py
```diff
@@ -0,0 +1,448 @@
...
+  record:
+    description:
+      - Record to add, if blank and value= or type= is specified, record= defaults to the zone name.
+      - "* refers to all records (only use with state=absent)"
+    required: false
+    default: null
+    aliases: [ "name" ]
+  solo:
+    description:
+      - Whether the record should be the only one for that record type and record name. Only use with state=present
```
resmo: for more readabilty use C() for state=present e.g. C(state=present)

network/cloudflare_dns.py
```diff
@@ -0,0 +1,448 @@
...
+        new_record = { "type": type, "name": record, "content": value, "ttl": ttl }
+        if not self.module.check_mode:
+            result, info = self._cf_api_call('/zones/{0}/dns_records'.format(zone_id),'POST',new_record)
+        self.changed = True
+        return result,self.changed
+
+def main():
+    module = AnsibleModule(
+        argument_spec = dict(
+            account_api_token = dict(required=True, default=None, no_log=True, type='str'),
```
resmo: no need for a default here if required, :+1:  for using no_log on secrets :)

network/cloudflare_dns.py
```diff
@@ -0,0 +1,448 @@
...
+
+    # sanity checks
+    if cf_api.record == '*' and cf_api.state == 'present':
+        module.fail_json(msg="record=* can only be used with state=absent")
+    if cf_api.is_solo and cf_api.state == 'absent':
+        module.fail_json(msg="solo=true can only be used with state=present")
+
+    # No zone specified
+    # -> return zone information for available zones
+    if (not cf_api.zone):
```
resmo: no need for () here

network/cloudflare_dns.py
```diff
@@ -0,0 +1,448 @@
...
+            result,changed = cf_api.create_dns_record()
+            module.exit_json(changed=changed,result={'records': result})
+        elif cf_api.state == 'absent':
+            # force solo to False, just to be sure
+            changed = cf_api.delete_dns_records(solo=False)
+            module.exit_json(changed=changed)
+        else:
+            module.fail_json(msg="{0} is an unknown state".format(cf_api.state))
+
+    module.fail_json(msg="Unexpected combination of zone, record, value and type.")
```
resmo: Not sure if it helps but wanted to let you know there are some helpers in utils e.g. mutually_exclusive and required_together for defining the wanted combinations: see https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/basic.py#L512

resmo:
Hi @mgruener, good work. I did a quick review and commented the parts. 

Unfortunately I can not test it. It is always a problem for other users to test modules consuming a "closed" API. 

It would be nice if you can show that all cases works as expected by adding the output in the PR description. Optionally if you could add tests to the ansible/ansible repo in https://github.com/ansible/ansible/tree/devel/test/integration that would be a +. You may get an idea if you look at my cloudstack integration tests.

network/cloudflare_dns.py
mgruener: I was initially considering these but did not use them as the parameters are only required together or mutually exclusive under certain conditions. The mutually_exclusive and required_together parameters have to be provided during the instantiation of the AnsibleModule class. I would need something "require_oneof_ if" or "conflicts_if". Please give me  a pointer to where I have to look if I have overlooked something.

mgruener:
Thanks for the feedback @resmo. I will implement your suggestions as soon as my job lets me ;)

Regarding the creation of a facts module and moving the common parts to https://github.com/ansible/ansible/tree/devel/lib/ansible/module_utils, would that be a hard requirement to get this module into modules-extras? I really like the idea, but for the moment (this being my first contribution) I would like to touch as few parts of ansible as possible. Facts wise, this module functions similar to the dnssimple module, which also only returns data if no domain or just the domain is provided. I was using this module as orientation on what functions I need to implement in my module.

If you think the "list" functionality really does not belong in the module, I would tend to removing it for the moment (it is not needed for the module to function).

Wouldn't the Rackspace tests be a better reference, as it is also a "closed" API?. How are the Rackspace tests intended to work? I see there is something about a credentials file, and the test playbooks use rackspace_username and rackspace_api_key. Is it up to the test designer to provide a test account or up to the tester? If it is up to the tester, I will gladly design the integration tests.

resmo:
it is not a requirement to move common part to module_utils yet, just a suggestion. 

Ansible had no policy at that time about facts and list info. For existing modules, removing the list functionality would mean breaking the backward compatibility.
For new modules, we try to avoid it. Facts are much cleaner as it will not require to register the output in a variable. I would suggest to remove the list functionality.

Regarding the API, sure yes. at the end the "tester" only need to have the credentials and can reproduce the test playbooks, which is up to the tester of course.

gregdek:
Thanks @mgruener for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

mgruener:
As suggested by @resmo I removed the "list" functionality from the module. This allowed me to utilize required_if and required_one_of which led to some code simplifications :smile: 

Next up: creating tests for https://github.com/ansible/ansible/tree/devel/test/integration

sgzijl:
Nice work! 

Record types A and CNAME seems to work, but when testing this with TXT records I run into an issue:

```
# this will fail:
    - name: Create DNS TXT record
      cloudflare_dns:
        state: present
        domain: foo.bar
        type: TXT
        record: '@'
        value: "fear cuts deeper than swords"
        account_email: "{{ email }}"
        account_api_token: "{{ token }}"

# this will work at first run:
    - name: Create DNS TXT record
      cloudflare_dns:
        state: present
        domain: foo.bar
        type: TXT
        record: '@'
        value: "fear%20cuts%20deeper%20than%20swords"
        account_email: "{{ email }}"
        account_api_token: "{{ token }}"

# but second run will fail with:

msg: API bad request; Status: 400; Method: POST: Call: /zones/$zoneid/dns_records

```

I will test the other records type too and let you know if I run into any problems.

Edit: same for SRV records.

sgzijl:
Found another issue with NS records.

Eg:

```
    - name: Delegate subdomain to AWS
      cloudflare_dns:
        state: present
        record: sudomain
        domain: foo.bar
        type: NS
        value: "{{ item }}"
        account_email: "{{ email }}"
        account_api_token: "{{ token }}"
      with_items:
        - "ns-363.awsdns-45.com."
```

First (initial) run works:

```
changed: [localhost] => (item=ns-363.awsdns-45.com.) => {"changed": true, "item": "ns-363.awsdns-45.com.", "result": {"records": {"content": "ns-363.awsdns-45.com", "created_on": "2016-03-08T20:46:34.541512Z", "id": "$id", "locked": false, "meta": {"auto_added": false}, "modified_on": "2016-03-08T20:46:34.541512Z", "name": "sudomain.foo.bar", "proxiable": false, "proxied": false, "ttl": 1, "type": "NS", "zone_id": "$zoneid", "zone_name": "foo.bar"}}}
```

But from the second run it fails:

```
failed: [localhost] => (item=ns-363.awsdns-45.com.) => {"failed": true, "item": "ns-363.awsdns-45.com."}
msg: API bad request; Status: 400; Method: POST: Call: /zones/$zoneid/dns_records
```

It's also unclear to me how to define MX records. Tried values like '10 mailserver.foo.bar', but think that we need an extra priority argument for managing MX records.

mgruener:
Thanks for the testing @sgzijl . Working on a proper test playbook right now so your tests give me good pointers on what to consider there in addition to what I already have.

The HTTP 400 Bad Request errors normally happen when you try to do something that is actually not supported by the Cloudflare API. I have to see if it is possible to get better errors from the API. In theory a failed API call should return a JSON structure with details about the error but in my initial tests this was not the case (or I missed something). I will have to go over this part of the error handling again.

For your TXT record test, what would be your expected result? E.g. what happens if you try to add this record with this value using the CloudFlare web gui?

The NS record test should work, I will have to see what is the problem there.

Why would you need a separate priority argument for MX records? According to the docs, the API does not use a specific priority field, it just expects a value (aside from the record name and type). Is this a case of "it does not work" or would updating the documentation to be more precise in this regard be enough?

sgzijl:
@mgruener thanks for the fast reply.

When I add the TXT record using the CloudFlare API, I add the name `@` and the value `fear cuts deeper then swords`.  Then the DNS TXT record, in this example, should return:

```
~ dig +short -t txt foo.bar
"fear cuts deeper than swords"
```

The module does not like spaces, where they are allowed in TXT records.

NS records do work, but make sure the values do not end with a dot, which is common in DNS...

About the MX records. In the UI, when I add a MX record a pop-up will ask me for the server and priority. Older versions of the Cloudflare API had a priority field. Not sure how they do it now, I find nothing about it in the current API documentation. I assumed it was something like `10 mail.server.tld.`, but that doesn't  work. If you wish, I can turn this question to them, tomorrow.

MX pop-up:
image

Actual content:
image

sgzijl:
FYI.

The payload while debugging the UI via the browser contains:

```
content: "mail.foo.bar"
data: {}
name: foo.bar"
priority: 20
proxiable: false
proxied: false
ttl: 3600
type: "MX"
zone_id: "$zone_id"
zone_name: "foo.bar"
```

Where I set the MX priority to 20, ttl to one hour and value to `mail.foo.bar`.

mgruener:
Ah, I misread your initial comment @sgzijl . I thought the second TXT record was the one failing. This should be relatively easy to fix, I probably just have to use proper url encoding und decoding on the record value.

I will also introduce a priority argument for MX records. And probably for SRV records, as an error code in the API documentation suggest these might need it too. Never trust API documentation to be complete :smile:

mgruener:
The TXT record issue should be fixed. I also did a quick test with other record types and there is still quite some work left to do to handle all the special cases (trailing "." are stripped for NS and CNAME, MX requires a priority parameter, SRV and LOC require even more...) the API documentation does not cover. I will work on that tomorrow.

mgruener:
- Fixed the issues with SRV, MX, NS and so on mentioned by @sgzijl. There are now several new parameters for MX and SRV record creation
- Removed support for LOC records. This would have added a lot (12 to be exact) of module parameters for minimal gain. If someone needs this record type, let me know
- Improved the error handling. The module will now report error details IF ansibles fetch_url returns the response for non-success HTTP status codes. Which it does not right now.
-  deleting multiple records is no longer possible when specifying the type parameter. I removed the respective example

sgzijl:
I get an error about the `follow_redirects` keyword on line 236:

```
TypeError: fetch_url() got an unexpected keyword argument 'follow_redirects'
```

If we really need follow_redirects, shouldn't this be handled at module parameter level? Without the `follow_redirects` keyword all tests passed and seem to work like a charm. Awesome!

I tested: A, AAAA, NS, CNAME, TXT, MX and SRV records.

mgruener:
Seems like this parameter to fetch_url was introduced in the current ansible devel branch. Right now I do not think it is really needed, I just added it as a safety measure. I will remove it to make the module easier to test.

mgruener:
The `follow_redirects` error with ansible < 2.1 has been fixed.

sgzijl:
@mgruener great job! :+1: 

@gregdek are there any community members that can review this module? I'd say shipit ...

network/cloudflare_dns.py
```diff
@@ -0,0 +1,536 @@
...
+    type: TXT
+    value: unique value
+    state: present
+    solo: true
+    account_email: test@example.com
+    account_api_token: dummyapitoken
+
+# create a SRV record _foo._tcp.my.com
+- cloudflare_dns:
+  domain: my.com
```
resmo: non working example: missing indentation

mgruener: Thanks, and fixed.

PR #1772, opened 2016-03-02
Adding more options to the yum_repository module

jtyr:
This PR is a replacement of PR #1595. It helps to separate the shipment of the new parameters from adding of the Managed Mode which is under discussion in the PR #1728.

### DIFF
--- a/packaging/os/yum_repository.py
+++ b/packaging/os/yum_repository.py
@@ -1,7 +1,7 @@
 #!/usr/bin/python
 # encoding: utf-8
 
-# (c) 2015, Jiri Tyr <jiri.tyr@gmail.com>
+# (c) 2015-2016, Jiri Tyr <jiri.tyr@gmail.com>
 #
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @jtyr. Since you are a maintainer of this module, we are marking this PR for inclusion.

packaging/os/yum_repository.py
```diff
@@ -44,23 +51,41 @@
         (bytes/sec) then this option is ignored. Default is C(0) (no bandwidth
         throttling).
   baseurl:
-    required: false
+    required: true
```
resmo: you can keep it as required=false if you have conditional requirements as in the description, required: true should be identical to the params in argument_spec=dict()

jtyr: Thanks for pointing this out. It's fixed now.

abadger:
Thanks!  merged to devel.

PR #1788, opened 2016-03-04
Removing parameter from yum_repository module

jtyr:
##### Issue Type:

Bugfix Pull Request
##### Plugin Name:

`yum_repository`
##### Ansible Version:

```
ansible 2.0.1.0
```
##### Summary:

This PR removes one parameter which I left there when separating the Managed Mode (#1728) from adding additional parameters. The removed parameter appears in the final repo file if present (see bellow).
##### Example output:

```
[test1]
baseurl = http://example.com
ignore_repo_files = []
name = Test repo 1.
priority = 2
```


abadger:
Thanks for catching that and fixing it.

PR #1796, opened 2016-03-05
fixes bug where puppet fails if logdest is not specified

Jmainguy:
##### Issue Type:

- Bugfix Pull Request
  ##### Plugin Name:

system/puppet.py
##### Ansible Version:

```
latest devel
```
##### Summary:

If logdest is not specified, the module fails, because the default is misinterpreted.
##### Example output:

```
[root@phy01 test]# ansible -i hosts all -m puppet
centos7.soh.re | FAILED! => {
    "changed": false, 
    "failed": true, 
    "msg": "value of logdest must be one of: stdout,syslog, got: ['stdout']"
}
```

### DIFF
--- a/system/puppet.py
+++ b/system/puppet.py
@@ -119,7 +119,7 @@ def main():
             puppetmaster=dict(required=False, default=None),
             manifest=dict(required=False, default=None),
             logdest=dict(
-                required=False, default=['stdout'],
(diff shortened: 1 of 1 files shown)


Jmainguy:
@emonty @nibalizer @bcoca ^^

emonty:
shipit

PR #1807, opened 2016-03-08
Hiera new module added, to parse and use hiera variables as node facts

jparrill:
##### Issue Type:
- New Module Pull Request
##### Plugin Name:

Hiera-Ansible
##### Ansible Version:

```
ansible 1.9.4
```
##### Summary:

Ansible-Hiera Parser could retrieve all Hiera data that you want to map as node facts, just copy all hieradata and hiera.yaml from Puppetmaster into destination node (the good way to do it is executing from Puppetmaster the ansible-playbook) and select the variable to store your hiera variable.

**Why?**

We are thinking about create one repository with only Hiera data and hierarchy file, also 2 repositories One with Puppet code and other one with Ansible code that will have the Hiera repository as submodule, this way will centralize all relevant data in one place avoiding replication errors.

I think this is a elegant way to organize your repositories and separate the logic from key/value data
##### Example output:

```
TASK: [Retrieving Hiera Data] *************************************************
ok: [vagrantServer] => (item={'key': 'var_array_line', 'value': 'proxy::array_line'})
ok: [vagrantServer] => (item={'key': 'line', 'value': 'line'})
ok: [vagrantServer] => (item={'key': 'var_array_multi', 'value': 'proxy::array_multi'})
```

### DIFF
--- a/system/hiera.py
+++ b/system/hiera.py
@@ -109,8 +109,36 @@ EXAMPLES = '''
 '''
 
 RETURN = '''
-This module will return the datatype that you got in Hieradata backend.
-
+path:
+    description: bin for hiera execution
+    returned: success
+    type: string
+    sample: "/path/to/file"
+environment:
+    description: scope to be interpreted by hiera
+    returned: success even the scope does not exists
+    type: string
+    sample: "production"
+fqdn:
+    description: other scope for hiera parse
+    returned: success even the scope does not exists
+    type: string
+    sample: ""
+key:
+    description: hiera's value name
+    returned: success
+    type: string
+    sample: "proxy::array_multi"
+fact:
+    description: variable where store the Hiera's value
+    returned: success
+    type: string
+    sample: "var_array_multi"
+source:
+    description: hiera config file
+    returned: success
+    type: string
+    sample: "/path/to/hiera.yaml"
 '''
 
 
@@ -132,7 +160,7 @@ def main():
     """
     module = AnsibleModule(
         argument_spec=dict(
-            name=dict(aliases=['key']),
+            name=dict(required=True, aliases=['key']),
             fact=dict(required=False),
             path=dict(required=False, default="hiera"),
             context=dict(required=False, default={}, type='dict'),
--- a/system/hiera.py
+++ b/system/hiera.py
@@ -36,23 +36,19 @@ options:
         description:
             - This is a fact name where you want to save your variable from Hiera.
         required: false
-        default: null
     key:
         description:
             - This is the Hiera variable name that you want to get.
         required: true
-        default: null
         aliases: ['name']
     context:
         description:
             - This key value will set the scope of hiera key/value. Also will follow the hiera's hierarchi, then if you dont have any variable in the first scope will go down to the next one.
         required: false
-        default: null
     source:
         description:
             - The hiera config file path, if you want to custom every query with multiple hierarchies
         required: false
-        default: null
 '''
 
 EXAMPLES = '''
--- a/system/hiera.py
+++ b/system/hiera.py
@@ -21,7 +21,7 @@ Hiera-Ansible Parser.
 DOCUMENTATION = '''
 ---
 module: hiera
-version_added: "2.1"
+version_added: "2.2"
 short_description: Use Hiera key/values into Ansible as fact.
 author: "Juan Manuel Parrilla @kerbeross jparrill@redhat.com"
 description:
--- a/system/hiera.py
+++ b/system/hiera.py
@@ -21,7 +21,7 @@ Hiera-Ansible Parser.
 DOCUMENTATION = '''
 ---
 module: hiera
-version_added: "1.0.0"
+version_added: "2.1"
 short_description: Use Hiera key/values into Ansible as fact.
 author: "Juan Manuel Parrilla @kerbeross jparrill@redhat.com"
 description:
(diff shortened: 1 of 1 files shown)


sivel:
```
./system/hiera.py
TRACE:
    while scanning a simple key
      in "<string>", line 9, column 5:
            destination node, this module wi ... 
            ^
    could not found expected ':'
      in "hiera.DOCUMENTATION", line 30, column 5:
            and create node facts in executi ... 
            ^
ERROR: DOCUMENTATION is not valid YAML. Line 30 column 5
ERROR: Interpreter line is not "#!/usr/bin/python"
```

system/hiera.py
```diff
@@ -0,0 +1,176 @@
...
+    fact:
+        description:
+            - This is a fact name where you want to save your variable from Hiera.
+        required: false
+        default: null
+    key:
+        description:
+            - This is the Hiera variable name that you want to get.
+        required: true
+        default: null
```
juliedavila: `default: null` here and in the other spots is not needed

system/hiera.py
```diff
@@ -0,0 +1,176 @@
...
+    - source: hiera config file
+        WARNING: This module try to solve a disparity between arch of
+        Puppet/Hiera and Ansible, because of how they works (Puppet compile
+        the values into puppet master node) (Ansible are executed into
+        destination node and have not access to Hiera backend directly)
+
+    """
+    module = AnsibleModule(
+        argument_spec=dict(
+            name=dict(aliases=['key']),
```
juliedavila: here you have `name` as the primary and `key` as the alias, in `DOCUMENTATION` it's the other way around. Also, you have `required: True` in `DOCUMENTATION` and here it's `False` (as that's the default value), should have `required=True`

system/hiera.py
```diff
@@ -0,0 +1,176 @@
...
+        line: "line"
+
+    - debug: msg="{{ item }}"
+      with_items: var_array_multi
+    - debug: msg="{{ item }}"
+      with_items: var_array_line
+    - debug: msg="{{ line }}"
+'''
+
+RETURN = '''
```
juliedavila, line 107: This is not the proper format for `RETURN`, see https://github.com/ansible/ansible-modules-core/blob/devel/files/copy.py#L123-L184 as an example.

I get that it could be any arbitrary key/value, but would be nice to see something here even if it's a mock key/val.

tima:
Have you considered working this module to work more like ec2_facts and the like? From what I'm seeing in the examples in your repo this module fetches one key/fact at a time. It would be better to just call the module once, let it grab everything and bring it back to access as needed. No need for complicated with loops and extra param passing.

bcoca:
inventory script or lookup make more sense imho

nibalizer:
At openstack we have some code that does some of this. http://git.openstack.org/cgit/openstack-infra/ansible-puppet/tree/tasks/main.yml#n22 We copy over subsections of our hiera tree so that when puppet runs it can make hiera lookups.

I think this could be better served with a more simple example. The basic task of looking up a value in hiera is valuable I think.

I don't know much about it but after a little looking I agree that the lookup subsystem in ansible is probably a better place to patch in hiera.

jparrill:
I think that as a first approach it's very simple and clean as already are, now with this first version I can add more functionality.

I will work in how expose all hiera key-values with a big hierarchy, because there are the real problem. If you dont know what you want to retrieve from Hiera, how do you know what you will query to it? There is any big hash with all key-values? 

Thanks for the suggestions, all are great :+1:

nibalizer:
Ya hiera doesn't support a query like 'give me my datas!' it has a simple question-answer interface.

system/hiera.py
```diff
@@ -0,0 +1,200 @@
...
+
+    try:
+        pargs = [params['path']]
+
+        if params['source']:
+            pargs.extend(['-c', params['source']])
+
+        pargs.append(params['name'])
+        pargs.extend(
+            [r'%s=%s' % (k, v) for k, v in params['context'].iteritems()]
```
nibalizer, line 181: In order to get hiera to 'hierarchy' you need to provide filtering variables.

environment=production fqdn=host1.example.com osfamily=Debian etc etc

I think you'd have  to unfold all of facts['FACTER_*'] into key=value pairs and append those to the end of the command to get it to work. Otherwise you're just looking things up in a single namespace.

jparrill, line 181: Yeah, you already got this with:

```
args:
   context:
      environment: 'production'
      fqdn: 'puppet01.localdomain'
      scope: 'zone'
```

It depends in how do you need at ansible execution. If you want a follow a complex hierarchy like this:

```
:hierarchy:
  - "%{location}/%{environment}/%{hostgroup}/%{fqdn}"
  - "%{location}/%{environment}/%{hostgroup}"
  - "%{location}/%{environment}"
  - "%{location}"
  - default
```

You must indicate all steps:

```
args:
  context:
    location: "{{ location }}"
    environment: "{{ environment }}"
    hostgroup: "{{ hostgroup }}"
    fqdn: "{{ fqdn }}"
```

With this distribution you can respect the original hierarchy, and if you don't find any variables in any concrete scope, Hiera will search in the next one.

jparrill:
Searching for another simple approach, there is not easy way to load all hierarchy key-values into a hashmap. Ideas? :S

jparrill:
@bcoca Do you think that the module are ready to be merged?, by my side this approximation it's enough for first instance. When I will use the module and find more needs, I will add it ;)

tima:
I think this will be an important an useful module to those using Ansible with Puppet that I'm not sure this is good enough. Does this help? https://github.com/stefanandres/hiera_dump

PR #1815, opened 2016-03-08
Fix vmware_portgroup throwing an error if port group already exists

camradal:
##### Issue Type:

- Bugfix Pull Request
##### Plugin Name:

vmware_portgroup
##### Ansible Version:

```
ansible 2.0.0.2
  config file =
```
##### Summary:

vmware_portgroup is not idempotent - running the play twice throws an error. This fix will prevent the error from happening, but it doesn't add any additional functionality.
##### Example output:

This is the error that this commit fixes

```
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "A specified parameter was not correct: "}
```


gregdek:
Thanks @camradal. @mtnbikenc @chrrrles please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

cloud/vmware/vmware_portgroup.py
```diff
@@ -68,6 +68,14 @@
     HAS_PYVMOMI = False
 
 
+def find_portgroup_by_name(host, portgroup_name):
```
chrrrles, line 71: It would be better to have this in `module_utils/vmware.py` as `find_host_portgroup_by_name`

chrrrles:
Heya @camradal  -  Adding exception handlers to catch `vim.fault.DuplicateName` and also `vim.fault.InvalidName` _might_ be a more efficient way to handle this condition.

camradal:
@chrrrles Good suggestion, let update it.

camradal:
Unfortunately, it doesn't seem like we're getting good information from exceptions.
The only error is vmodl.fault.InvalidArgument with message "A specified parameter was not correct: " (with no parameter provided)

Let's wait for pull request to ansible pull request (https://github.com/ansible/ansible/pull/15002)

camradal:
Going to fetch upstream and rewrite this since the dependent issue is now merged.

PR #1829, opened 2016-03-10
remove legacy action style from examples

liquidat:
##### Issue Type:
- Docs Pull Request
##### Plugin Name:

module ovirt
##### Ansible Version:

Git devel
##### Summary:

remove legacy action style from examples
- "action" style invoking is a legacy way to call modules
- the examples were updated to the typical style of calling complex
    modules:

```
ovirt:
  parameter1: value1
  parameter2: value2
   ...
```

### DIFF
--- a/cloud/misc/ovirt.py
+++ b/cloud/misc/ovirt.py
@@ -159,51 +159,51 @@ requirements:
 EXAMPLES = '''
 # Basic example provisioning from image.
 
-action: ovirt >
-    user=admin@internal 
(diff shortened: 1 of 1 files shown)


bcoca:
it is not legacy, key=val style is not deprecated nor going anywhere, still will accept this as key=val multiline is fragile and yaml syntax is better in this case

liquidat:
The "legacy" referred to the usage of "action" in the example, not the key/value multiline statement.

PR #1877, opened 2016-03-20
Fix type used by the module

mscherer:
##### Issue Type:
- Bugfix Pull Request
##### Plugin Name:

redis
##### Summary:

Set int for the various port (and so avoid to convert them later)
Set no_log=True for the login_password
Verify that db is a int, so avoid a conversion

### DIFF
--- a/database/misc/redis.py
+++ b/database/misc/redis.py
@@ -166,13 +166,13 @@ def main():
     module = AnsibleModule(
         argument_spec = dict(
             command=dict(default=None, choices=['slave', 'flush', 'config']),
-            login_password=dict(default=None),
+            login_password=dict(default=None, no_log=True),
             login_host=dict(default='localhost'),
-            login_port=dict(default='6379'),
+            login_port=dict(default=6379, type='int'),
             master_host=dict(default=None),
-            master_port=dict(default=None),
+            master_port=dict(default=None, type='int'),
             slave_mode=dict(default='slave', choices=['master', 'slave']),
-            db=dict(default=None),
+            db=dict(default=None, type='int'),
             flush_mode=dict(default='all', choices=['all', 'db']),
             name=dict(default=None),
             value=dict(default=None)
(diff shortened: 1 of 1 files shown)


abadger:
master_port can also have its explicit conversion into int removed: https://github.com/ansible/ansible-modules-extras/pull/1877/files#diff-40a12a7129d99aa78f612e011dcbd1afR196

resmo:
Thanks @mscherer. To the current maintainers, @slok please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

database/misc/redis.py
```diff
@@ -257,15 +253,12 @@ def main():
...
-        except Exception:
-            db = 0
+        db = module.params['db']
         mode = module.params['flush_mode']
 
         #Check if we have all the data
         if mode == "db":
-            if type(db) != int:
+            if db is None:
                 module.fail_json(
```
abadger, line 262: These last two sections of the original code confuse me a little.  It looks like the first section is saying "Make the db parameter into an integer.  If, for any reason, we can't turn it into an integer then set db to 0.  Then the second section is unnecessarily making a check whether db  really is an int now.

Probably those can just be translated into a default value of 0 in the argument spec and then we don't have to check whether db is None here.

mscherer, line 262: I do not know redis enough to know what should be done. I think we should wait for the maintainer of the module to tell what he wanted to express with that construct, but yeah, it would be more logical to default to 0.

But "anything not a number as a argument mean the first db" is a bit curious in the first place.

abadger, line 262: Okay..  well, this PR fixes things, the module is in extras, and this note isn't a bug... It's just not making best use of the arg parsing framework.  I'm going to merge this PR and if the module owner is inclined, they look at making this change as well.

abadger:
Thanks for working on this @mscherer!

PR #1890, opened 2016-03-21
AWS lambda modules

ghost:
##### Issue Type:
- New Modules Pull Request
##### Plugin Name:

lambda
lambda_facts
lambda_alias
lambda_event
lambda_invoke
lambda_policy
##### Summary:

These modules help manage AWS Lambda resources including code, configuration, aliases, versions and event source mappings. 
##### Example:

```

---
- hosts: localhost
  gather_facts: no
  vars:
    state: present
    project_folder: /path/to/your/project/folder
    deployment_package: lambda.zip
    account: 123456789012
    function_name: ingestData

  tasks:
  - name: gather lambda facts
    lambda_facts:
      name: "{{ function_name }}"
      query: all

  - name: AWS Lambda Function
    lambda:
      state: "{{ state | default('present') }}"
      name: "{{ function_name }}"
      publish: True
      description: lambda function description
      code_s3_bucket: package-name
      code_s3_key: "lambda/{{ deployment_package }}"
      local_path: "{{ project_folder }}/{{ deployment_package }}"
      runtime: python2.7
      timeout: 5
      handler: lambda.handler
      memory_size: 192
      role: API2LambdaExecRole

  - name: "alias 'Dev' for function {{ function_name }}"
    lambda_alias:
      state: "{{ state | default('present') }}"
      function_name: "{{ function_name }}"
      name: Dev
      description: Development is $LATEST version

  - name: Lambda policy for S3 event notifications
    lambda_policy:
      state: "{{ state | default('present') }}"
      function_name: functionName
      alias: Dev
      statement_id: lambda-s3-myBucket-create-data-log
      action: lambda:InvokeFunction
      principal: s3.amazonaws.com
      source_arn: arn:aws:s3:eu-central-1:123456789012:bucketName
      source_account: 123456789012

  - name: DynamoDB stream event mapping
    lambda_event:
      state: "{{ state | default('present') }}"
      event_source: stream
      function_name: "{{ function_name }}"
      alias: Dev
      source_params:
        source_arn: arn:aws:dynamodb::123456789012:table/tableName/stream/2016-03-19T19:51
        enabled: True
        batch_size: 120
        starting_position: TRIM_HORIZON

  - name: execute lambda function
    lambda_invoke:
      function_name: "{{ function_name }}"
      qualifier: Dev
      invocation_type: RequestResponse

  - debug: var=lambda_invocation_results

```


resmo:
Thanks @pjodouin for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

jsdalton:
HI @pjodouin!

Just wanted to say thanks for this contribution. I'm toying around with a possible use case for AWS Lambda on a project I'm working on that uses Ansible, so it might give me a chance to test drive this module.

I haven't tried it out just yet or reviewed the code thoroughly but I had one or two questions as I'm poking around: (caveat: I'm very new to using AWS Lambda and still working on wrapping my head around the feature)

Re the `lambda_event` module... If i'm reading it correctly it looks like you're implementing the ability to create both pull event model sources in AWS Lambda itself (i.e. Dynamo and Kinesis) as well as push event model sources (looks like you have S3 and SNS implemented).

A few thoughts there:
- It definitely makes sense to me to implement the "pull" event sources that are part of the core functionality of Lambda.
- Do the push event sources make sense though? It seems like those should be handled by their respective cloud modules in the ansible library, since the event is actually being created in those resources and not in lambda.
- Another issue there is it ends up feeling arbitrary that we'd have chosen two services to implement push events for, leaving off the about 6 or 7 other services that can be configured to push events to lambda: http://docs.aws.amazon.com/lambda/latest/dg/intro-core-components.html

Lastly (and relatedly) I think as an approach to implementing cloud modules in ansible it makes sense to map them pretty closely to the structure and functionality of the boto functions underlying them. This prevents some of the structural issues like my comment about only partially implementing the push events. It also becomes easier to maintain the module going forward, since as new capabilties and features are added to the boto library, the easier it is to extend the module to support them.

If I look at your proposed modules through that lens it's actually pretty close. The major "nouns" I identify when I look at the boto module are:
- lambda functions
- lambda permissions
- lambda aliases
- lambda event source mappings

And you've got:
- lambda (functions)
- lambda_event (source mappings)
- lambda_alias
- lambda_invoke (which is also in boto)

Anyhow, point being this is a pretty good module as is and I think with a bit of evolution could be ready for inclusion.

Not sure how helpful these high level comments are but if you have any questions about any of the above let me know and I can clarify.

I'll try exercising some of the functionality in the coming days, which will also give me an opportunity to dig into the code at a more granular level and provide additional feedback.

Please also feel free to reach out directly if you have any questions or thoughts.

Cheers!

ghost:
Hi @jsdalton,
Thanks for the valuable feedback.  The push notifications may not belong there.  It depends on the design philosophy behind the ansible module architecture, something I could not find.  Originally I had a separate module for permissions and a separate one for the s3 push event as well.  I guess I was trying to simplify the process of adding the policy permissions needed for each event type.  And I was planning on eventually adding the other push notification types.  

All good points and I look forward to getting more feedback.

Cheers!

viper233:
This bug hashicorp/terraform/issues/5767 should be tested, I'm assuming an Ansible provisioned VPC may have the same issue, cloudformation does. (not sure if this is the best place to address this issue, just wanted to make lambda users aware of it.)

ghost:
@viper233 

Lambda functions can be created, deleted and re-created within a VPC without any issue.  However, if you so much as detach the automatically created ENI from a subnet, let alone delete it, the lambda function will no longer be able to access resources inside the VPC, even if you delete it and recreate the function.  I'm not sure how long this condition persists.  Best bet is to ignore the automatically created ENIs.

ghost:
Hi @jsdalton, 
I've removed the _push_  event handling from the lambda_event module and added a lambda_policy module to allow permissions to be added for _push_ model events.  An s3_event module is forthcoming.

gregdek:
Thanks @pjodouin for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

ghost:
Hi @gregdek 

Not sure why this is still failing with the same error.  They've all been changed to 2.2 yet Travis still sees  2.1.  Any ideas?

gregdek:
That is... odd. Maybe rebase against latest?

ghost:
Yeah, did that and it worked the second time but now it's complaining about
"boto" (which is now fixed too)

ghost:
ready_for_review

gregdek:
Thanks @pjodouin for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

ghost:
@gundalow  Thanks John!

cloud/amazon/lambda.py
```diff
@@ -0,0 +1,545 @@
...
+    :param aws: AWS client connection
+    :return:
+    """
+
+    function_name = module.params['function_name']
+
+    # validate function name
+    if not re.search('^[\w\-:]+$', function_name):
+        module.fail_json(
+                msg='Function name {0} is invalid. Names must contain only alphanumeric characters and hyphens.'.format(function_name)
```
gundalow: [bad-continuation] Wrong hanging indentation (remove 4 spaces).

(from pep8)

cloud/amazon/lambda.py
```diff
@@ -0,0 +1,545 @@
...
+
+
+def main():
+    """
+    Main entry point.
+
+    :return dict: ansible facts
+    """
+    argument_spec = ec2_argument_spec()
+    argument_spec.update(dict(
```
gundalow, line 500: Formatting of this (according to pep8) looks wrong here
I think it should be

```
argument_spec.update(
    dict(
        state=dict(required=False, default='present', choices=['present', 'absent']),


```

cloud/amazon/lambda.py
```diff
@@ -0,0 +1,545 @@
...
+    module = AnsibleModule(
+        argument_spec=argument_spec,
+        supports_check_mode=True,
+        mutually_exclusive=[],
+        required_together=[['subnet_ids', 'security_group_ids']]
+    )
+
+    # validate dependencies
+    if not HAS_BOTO3:
+        module.fail_json(msg='Both boto3 & boto are required for this module.')
```
gundalow, line 529: Should this be boto3 and botocore?

cloud/amazon/lambda_alias.py
```diff
@@ -0,0 +1,367 @@
...
+
+author: Pierre Jodouin (@pjodouin)
+options:
+  function_name:
+    description:
+      - The name of the function alias.
+    required: true
+  state:
+    description:
+      - Describes the desired state and defaults to "present".
```
gundalow, line 44: No need to say what the default is, it's documented below. Also missing fullstop.
`Describes the desired state.`

cloud/amazon/lambda.py
gundalow, line 276: pep8 reports

lambda.py:276:17: E126 continuation line over-indented for hanging indent

cloud/amazon/lambda.py
```diff
@@ -0,0 +1,545 @@
...
+      - The function within your code that Lambda calls to begin execution.
+    required: true
+  role:
+    description:
+      - The Amazon Resource Name (ARN) of the IAM role that Lambda assumes when it executes your function to access 
+        any other AWS resources.  If the role belongs to the same account, the simple role name can be used.
+    required: true
+  timeout:
+    description:
+      - The function execution time at which Lambda should terminate the function. Because the execution time has cost
```
gundalow, line 96: The function execution time (in seconds)

Also please remove the detail about default, that's captured below

cloud/amazon/lambda.py
```diff
@@ -0,0 +1,545 @@
...
+      - The function execution time at which Lambda should terminate the function. Because the execution time has cost 
+        implications, we recommend you set this value based on your expected execution time. The default is 3 seconds.
+    required: false
+    default: 3
+  memory_size:
+    description:
+      - The amount of memory, in MB, your Lambda function is given. Lambda uses this memory size to infer the amount of 
+        CPU and memory allocated to your function. Your function use-case determines your CPU and memory requirements. 
+        For example, a database operation might need less memory compared to an image processing function. The default 
+        value is 128 MB. The value must be a multiple of 64 MB.
```
gundalow, line 105: Remove default

gundalow, line 105: If you don't set a multiple of 64 how does this fail. This isn't checked in your code. Do we get a meaningful message back from boto3, or a backtrace?

cloud/amazon/lambda.py
```diff
@@ -0,0 +1,545 @@
...
+    default: 128
+  publish:
+    description:
+      - This boolean parameter is used to publish a version of the function from the current snapshot of $LATEST.
+        The code and configuration cannot be modified after publication.
+    required: false
+    default: false
+  description:
+    description:
+      - A short, user-defined function description. Lambda does not use this value. Assign a meaningful description
```
gundalow, line 116: A short, human readable user-defined function description. Lambda does not use this value. Assign a meaningful as you see fit.

Is there a character limit? How well is that handled if you go over it? A meaningful message or backtrace?

cloud/amazon/lambda.py
```diff
@@ -0,0 +1,545 @@
...
+    required: false
+    default: false
+  description:
+    description:
+      - A short, user-defined function description. Lambda does not use this value. Assign a meaningful description 
+        as you see fit.
+    required: false
+  version:
+    description:
+      -  Version number of the Lambda function to be deleted. This parameter cannot be used with state=present.
```
gundalow, line 121: I(state=present)

gundalow, line 121: Do you validate that these can't be used togther?

gundalow:
I've done a partial review of AWS. A number of those comments will apply to the other modules.

Please ensure all files are pep8 clean

`pep8 -r --ignore=E501,E221,W291,W391,E302,E251,E203,W293,E231,E303,E201,E225,E261,E241,E402 lambda.py`

ghost:
@gundalow 

This is straight from the Ansible web pages: 

> We do not take ‘style only’ requests unless the code is nearly unreadable, we are “PEP8ish”, but not strictly compliant.

Moreover, I ran PEP8 on the Ansible project and got 4,411 violations.  I've confirmed this by perusing many other modules (extras and core) and found several examples of those same violations you raised on this PR and much worse.

So here's the question:  is this a new requirement for new contributions?

If the answer to this question is "yes," then where is this documented?

Thanks,
Pierre

gundalow:
Hi, 
Thank for the quick reply, and you new modules to improve Ansible. 

The reason that the statement above is there is "style only" changes break git blame as they make it difficult to spot when a change/bug was introduced. 

For new modules there isn't any history to break. 

In the "New Modules" meeting we have discussed this and we want to raise the bar for new modules to be pep8 clean (with the exceptions listed in my previous command line). 

Once a module is pep8 clean we can ensure it stays clean, which gives us cleaner code to build on in the future. 

We will be looking at updating the travis-ci.org tests that run for new modules Pull Requests that enforce this. 

You are correct that this isn't currently documented, I've been requesting this manually for all the recent modules I've reviewed. Updating documentation is tracked as part of https://github.com/ansible/community/issues/96

Thank you for your understanding

ghost:
Hi,
Sounds logical to me.  And yes, It would be a good idea to formalize this in the documentation for future contributors. 

Also, documentation guidelines would be helpful too.  When comparing what's already done, it goes from very little to too much.  

I'm all for raising the bar and will make the recommended changes as soon as I have some free time.

gundalow:
Thank you for your understanding. 

One of the things I'm doing by reviewing lots of PRs is to spot common trends, and rase the bar. As you stated documenting this is also required. I have a PR in flight in improve the Developing Modules Checklist, which has the the above tracked.

danvaida:
@pjodouin are you working on extending the **lambda_event** module to allow 'CloudWatch Events' as a source or is that something for the future (i.e. post merge). Other than that, I'm currently testing the **lambda** module and will get back to you once I have some feedback. Thank you!

ghost:
@danvaida, I had originally included "push model" events such as the S3 and SNS event notifications, to the **lambda_event** module but it was suggested that it might be more appropriate to include these push events in their respective service modules.  I've since then worked on an [s3_event](https://github.com/pjodouin/ansible-repo/tree/master/library)  module that includes Lambda invocations.  I can certainly look at doing the same for CloudWatch events if the interest is there and someone is willing to test and review.

danvaida:
@pjodouin indeed, that would make more sense. it's a good idea to follow the separation of concerns principle, whenever possible.
I'd definitely be interested in helping out with a CloudWatch events module.

I see that you're importing ec2 helpers in **lambda_facts** but you're not advertising the inherited parameters (i.e. region) anywhere in the description, examples. Dunno if it'd be acceptable to at least specify this availability in the description. Probably the case with the other lambda_ modules.

jsdalton:
@pjodouin and @danvaida - I have an open PR (#2101) which I submitted a few weeks ago with a CloudWatch events module implementation. I wrote it for this precise use case, i.e. sending events into a lambda. Have a look (and please feel free to review) because I think it would pair nicely with the work @pjodouin is doing here.

@pjodouin My apologies for not having the time yet to review this module. This part of the project I'm working on got sidetracked so I have not had a chance to put your module through its paces. It looks like this week I'll be circling around to my lambda work and will be able to review this.

danvaida:
@jsdalton spot on! I'm gonna start testing your module as well and provide feedback in the PR. _quickly runs to replace the shell tasks invoking the aws cli..._ :)

ghost:
No problem @jsdalton and I'll take a look at your CloudWatch PR as well.  Excellent!

Not sure exactly what you mean @danvaida regarding the description.  Are you referring to the generated html?  It's supposed to have included the aws document fragments which documents the EC2 parameters such as region...etc.

danvaida:
@pjodouin totally right about that. forgot about the clever sphinx rules :)

ghost:
Closed this PR as it was just too time consuming.  Perfectly working modules will always be available in my personal repo.

Cheers!

resmo:
someone pick that up? @defionscode maybe?

juliedavila:
Darn this PR wasn't on my radar. Yes I have a couple of lambda modules in
the works.

ghost:
I can re-open this PR as I have some time during the next week or so.  If another Lambda PR is in the works, that's fine too.

danvaida:
@pjodouin I'm not so sure that it's a good idea to have the `lambda` module uploading the artefact to S3. I mean, that's what the `s3` module is for.

I have a genuine use case where the zip file is living in an S3 bucket that's owned by another AWS account than the one in which I want to create the lambda function. Everything's happening within the same region.

The zip file is not publicly readable and it's quite cumbersome (if not impossible) to handle authentication for both accounts: one time for uploading the object and one time to create the lambda function.

As a workaround, I have an interim `s3` task that handles AWS authentication itself to be able to fetch the object. Once the zip file is downloaded, the `lambda` task kicks in and tries to upload the zip file.

Perhaps I got it wrong, please correct me.

ghost:
@danvaida The local_path was used to make the module idempotent.  I would have to remove the requirement of a local_path to support your use case. And as long as the running account had read-access to the bucket, it should work.

danvaida:
idempotence is indeed very important and the other `lambda` module that got merged is also doing this, but has this parameter optional IIRC. Cross-account access is something that doesn't come cheap and read-access is not always OK to have.
I'm not a very big fan of the idea of having the artefact locally and accessible by the lambda module.
Think of the following use case:
1. your build process generates a zip file and uploads it to an S3 bucket
2. you deployment process is running somewhere else and takes this readable bucket object and feeds it to task that runs the lambda module.

So, unless I have another task before the lambda task that takes care of downloading the zip file into a specific dir locally, I cannot use this module. This doesn't seem right to me. What if your module would take care of this behind the scenes?
Maybe I'm going too far with this. Would be interested to see what other say.
Thanks.

PR #1896, opened 2016-03-23
Remove dead code from pacman module

djmattyg007:
The manual check to see if get_bin_path() returned anything is
redundant, because we pass True to the required parameter of
get_bin_path(). This automatically causes the task to fail if the pacman
binary isn't available. Therefore, the code within the if statement
being removed is never called.

### DIFF
--- a/packaging/os/pacman.py
+++ b/packaging/os/pacman.py
@@ -301,9 +301,6 @@ def main():
 
     pacman_path = module.get_bin_path('pacman', True)
 
-    if not os.path.exists(pacman_path):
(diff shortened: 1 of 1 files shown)


resmo:
Thanks @djmattyg007. To the current maintainers, @indrajitr please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

indrajitr:
Nice!
shipit

resmo:
Thanks again to @djmattyg007 for this PR, and thanks @indrajitr for reviewing. Marking for inclusion.

PR #1902, opened 2016-03-24
Added modules to manage Atomic Host Platform (host and image)

krsacme:
##### ISSUE TYPE
- New Module Pull Request
##### COMPONENT NAME

atomic_host
atomic_image
##### ANSIBLE VERSION

```
ansible 2.1.0
  configured module search path = /usr/share/my_modules/
```
##### SUMMARY

Atomic Host Plaform provides atomic command to manage the host and to manage the container images on the host. Details of the command can be found at [1]. Added 2 modules in here:
atomic_host :- Module to mange the host itself, upgrade or rollback
atomic_image :- Module to manage the container images

[1] http://www.projectatomic.io/docs/usr-bin-atomic/, https://github.com/projectatomic/atomic/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -117,7 +117,7 @@ install:
   - pip install git+https://github.com/ansible/ansible.git@devel#egg=ansible
   - pip install git+https://github.com/sivel/ansible-testing.git#egg=ansible_testing
 script:
-  - python2.4 -m compileall -fq -x 'cloud/|monitoring/zabbix.*\.py|/dnf\.py|/layman\.py|/maven_artifact\.py|clustering/(consul.*|znode)\.py|notification/pushbullet\.py|database/influxdb/influxdb.*\.py|database/mssql/mssql_db\.py|/letsencrypt\.py|network/f5/bigip.*\.py|bmc/ipmi/.*\.py' .
+  - python2.4 -m compileall -fq -x 'cloud/|monitoring/zabbix.*\.py|/dnf\.py|/layman\.py|/maven_artifact\.py|clustering/(consul.*|znode)\.py|notification/pushbullet\.py|database/influxdb/influxdb.*\.py|database/mssql/mssql_db\.py|/letsencrypt\.py|network/f5/bigip.*\.py|bmc/ipmi/.*\.py|system/atomic_.*\.py' .
   - python2.6 -m compileall -fq .
   - python2.7 -m compileall -fq .
   - python3.4 -m compileall -fq . -x $(echo "$PY3_EXCLUDE_LIST"| tr ' ' '|')
--- a/system/atomic_host.py
+++ b/system/atomic_host.py
@@ -25,18 +25,24 @@ description:
     - Rebooting of Atomic host platform should be done outside this module
 version_added: "2.2"
 author: "Saravanan KR @krsacme"
+notes:
+    - Host should be an atomic platform (verified by existence of '/run/ostree-booted' file)
 options:
     revision:
         description:
           - The version number of the atomic host to be deployed. Providing ```latest``` will upgrade to the latest available version.
         required: false
         default: latest
+        aliases: ["version"]
 '''
 
 EXAMPLES = '''
 
 # Upgrade the atomic host platform to the latest version (atomic host upgrade)
 - atomic_host: revision=latest
+
+# Deploy a specific revision as the atomic host (atomic host deploy 23.130)
+- atomic_host: revision=23.130
 
 '''
 
@@ -74,14 +80,13 @@ def core(module):
 def main():
     module = AnsibleModule(
                 argument_spec = dict(
-                    revision = dict(default='latest', required=False),
+                    revision = dict(default='latest', required=False, aliases=["version"]),
                 ),
             )
 
-    # Verify that the platform supports atomic command
-    rc, out, err = module.run_command('atomic -v', check_rc=False)
-    if rc != 0:
-        module.fail_json(msg="Error in running atomic command", err=err)
+    # Verify that the platform is atomic host
+    if not os.path.exists("/run/ostree-booted"):
+        module.fail_json(msg="Module atomic_host is applicable for Atomic Host Platforms only")
 
     try:
         core(module)
--- a/system/atomic_host.py
+++ b/system/atomic_host.py
@@ -1,0 +1,95 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public licenses
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION='''
+---
+module: atomic_host
+short_description: Manage the atomic host platform
+description:
+    - Manage the atomic host platform
+    - Rebooting of Atomic host platform should be done outside this module
+version_added: "2.2"
+author: "Saravanan KR @krsacme"
+options:
+    revision:
+        description:
+          - The version number of the atomic host to be deployed. Providing ```latest``` will upgrade to the latest available version.
+        required: false
+        default: latest
+'''
+
+EXAMPLES = '''
+
+# Upgrade the atomic host platform to the latest version (atomic host upgrade)
+- atomic_host: revision=latest
+
+'''
+
+RETURN = '''
+msg:
+    description: The command standard output
+    returned: always
+    type: string
+    sample: 'Already on latest'
+'''
+
+def core(module):
+    revision = module.params['revision']
+    args = []
+
+    if revision == 'latest':
+        args = ['atomic', 'host', 'upgrade']
+    else:
+        args = ['atomic', 'host', 'deploy', revision]
+
+    out = {}
+    err = {}
+    rc = 0
+
+    rc, out, err = module.run_command(args, check_rc=False)
+
+    if rc == 77 and revision == 'latest':
+        module.exit_json(msg="Already on latest")
+    elif rc != 0:
+        module.fail_json(rc=rc, msg=err)
+    else:
+        module.exit_json(msg=out, changed=True)
+
+
+def main():
+    module = AnsibleModule(
+                argument_spec = dict(
+                    revision = dict(default='latest', required=False),
+                ),
+            )
+
+    # Verify that the platform supports atomic command
+    rc, out, err = module.run_command('atomic -v', check_rc=False)
+    if rc != 0:
+        module.fail_json(msg="Error in running atomic command", err=err)
+
+    try:
+        core(module)
+    except Exception as e:
+        module.fail_json(msg=str(e))
+
+
+# import module snippets
+from ansible.module_utils.basic import *
+if __name__ == '__main__':
+    main()
--- a/system/atomic_host.py
+++ b/system/atomic_host.py
@@ -27,6 +27,9 @@ version_added: "2.2"
 author: "Saravanan KR @krsacme"
 notes:
     - Host should be an atomic platform (verified by existence of '/run/ostree-booted' file)
+requirements:
+  - atomic
+  - "python >= 2.6"
 options:
     revision:
         description:
@@ -58,6 +61,8 @@ def core(module):
     revision = module.params['revision']
     args = []
 
+    module.run_command_environ_update = dict(LANG='C', LC_ALL='C', LC_MESSAGES='C')
+
     if revision == 'latest':
         args = ['atomic', 'host', 'upgrade']
     else:
@@ -70,7 +75,7 @@ def core(module):
     rc, out, err = module.run_command(args, check_rc=False)
 
     if rc == 77 and revision == 'latest':
-        module.exit_json(msg="Already on latest")
+        module.exit_json(msg="Already on latest", changed=False)
     elif rc != 0:
         module.fail_json(rc=rc, msg=err)
     else:
--- a/system/atomic_image.py
+++ b/system/atomic_image.py
@@ -29,6 +29,7 @@ notes:
     - Host should be support ```atomic``` command
 requirements:
   - atomic
+  - "python >= 2.6"
 options:
     name:
         description:
@@ -81,6 +82,8 @@ def core(module):
     state = module.params['state']
     started = module.params['started']
     is_upgraded = False
+
+    module.run_command_environ_update = dict(LANG='C', LC_ALL='C', LC_MESSAGES='C')
 
     if state == 'present' or state == 'latest':
         if state == 'latest':
--- a/system/atomic_image.py
+++ b/system/atomic_image.py
@@ -1,0 +1,128 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION='''
+---
+module: atomic_image
+short_description: Manage the container images on the atomic host platform
+description:
+    - Manage the container images on the atomic host platform
+    - Allows to execute the commands on the container images
+    - Commands ```install``` and ```run``` will download the container image if not present already
+version_added: "2.2"
+author: "Saravanan KR @krsacme"
+notes:
+    - Host should be an atomic platform (verified by existence of '/run/ostree-booted' file)
+requirements:
+  - atomic
+options:
+    name:
+        description:
+          - Name of the container image
+        required: True
+        default: null
+    upgrade:
+        description:
+          - Forcefully upgrade if the container, even if the container is already running
+        required: False
+        choices: ["yes", "no"]
+        default: no
+    state:
+        description:
+          - The state of the container image
+        required: False
+        choices: ["started", "stopped"]
+        default: started
+'''
+
+EXAMPLES = '''
+
+# Execute the run command on rsyslog container image (atomic run rhel7/rsyslog)
+- atomic_image: name=rhel7/rsyslog state=started
+
+'''
+
+RETURN = '''
+msg:
+    description: The command standard output
+    returned: always
+    type: string
+    sample: [u'Using default tag: latest ...']
+'''
+
+def do_upgrade(module, image):
+    args = ['atomic', 'update', '--force', image]
+    rc, out, err = module.run_command(args, check_rc=False)
+    if rc != 0: # something went wrong emit the msg
+        module.fail_json(rc=rc, msg=err)
+    elif 'Image is up to date' in out:
+        return False
+
+    return True
+
+
+def core(module):
+    image = module.params['name']
+    upgrade = module.params['upgrade']
+    state = module.params['state']
+    is_upgraded = False
+
+    if state == 'started':
+        if upgrade:
+            is_upgraded = do_upgrade(module, image)
+        args = ['atomic', 'run', image]
+    elif state == 'stopped':
+        args = ['atomic', 'stop', image]
+
+    out = {}
+    err = {}
+    rc = 0
+    rc, out, err = module.run_command(args, check_rc=False)
+
+    if rc != 0: # something went wrong emit the msg
+        module.fail_json(rc=rc, msg=err)
+    elif state == 'started' and 'Container is running' in out:
+        module.exit_json(result=out, changed=is_upgraded)
+    else:
+        module.exit_json(msg=out, changed=True)
+
+
+def main():
+    module = AnsibleModule(
+                argument_spec = dict(
+                    name    = dict(default=None, required=True),
+                    upgrade = dict(default='no', type='bool'),
+                    state   = dict(default='started', choices=['started', 'stopped']),
+                ),
+            )
+
+    # Verify that the platform supports atomic command
+    rc, out, err = module.run_command('atomic -v', check_rc=False)
+    if rc != 0:
+        module.fail_json(msg="Error in running atomic command", err=err)
+
+    try:
+        core(module)
+    except Exception as e:
+        module.fail_json(msg=str(e))
+
+
+# import module snippets
+from ansible.module_utils.basic import *
+if __name__ == '__main__':
+    main()
--- a/system/atomic_image.py
+++ b/system/atomic_image.py
@@ -23,11 +23,10 @@ short_description: Manage the container images on the atomic host platform
 description:
     - Manage the container images on the atomic host platform
     - Allows to execute the commands on the container images
-    - Commands ```install``` and ```run``` will download the container image if not present already
 version_added: "2.2"
 author: "Saravanan KR @krsacme"
 notes:
-    - Host should be an atomic platform (verified by existence of '/run/ostree-booted' file)
+    - Host should be support ```atomic``` command
 requirements:
   - atomic
 options:
@@ -36,24 +35,25 @@ options:
           - Name of the container image
         required: True
         default: null
-    upgrade:
-        description:
-          - Forcefully upgrade if the container, even if the container is already running
-        required: False
-        choices: ["yes", "no"]
-        default: no
     state:
         description:
-          - The state of the container image
+          - The state of the container image.
+          - The state ```latest``` will ensure container image is upgraded to the latest version and forcefully restart container, if running.
         required: False
-        choices: ["started", "stopped"]
-        default: started
+        choices: ["present", "absent", "latest"]
+        default: latest
+    started:
+        description:
+          - Start or Stop the continer
+        required: False
+        choices: ["yes", "no"]
+        default: yes
 '''
 
 EXAMPLES = '''
 
 # Execute the run command on rsyslog container image (atomic run rhel7/rsyslog)
-- atomic_image: name=rhel7/rsyslog state=started
+- atomic_image: name=rhel7/rsyslog state=latest
 
 '''
 
@@ -78,25 +78,31 @@ def do_upgrade(module, image):
 
 def core(module):
     image = module.params['name']
-    upgrade = module.params['upgrade']
     state = module.params['state']
+    started = module.params['started']
     is_upgraded = False
 
-    if state == 'started':
-        if upgrade:
+    if state == 'present' or state == 'latest':
+        if state == 'latest':
             is_upgraded = do_upgrade(module, image)
-        args = ['atomic', 'run', image]
-    elif state == 'stopped':
-        args = ['atomic', 'stop', image]
+
+        if started:
+            args = ['atomic', 'run', image]
+        else:
+            args = ['atomic', 'install', image]
+    elif state == 'absent':
+        args = ['atomic', 'uninstall', image]
 
     out = {}
     err = {}
     rc = 0
     rc, out, err = module.run_command(args, check_rc=False)
 
-    if rc != 0: # something went wrong emit the msg
+    if rc < 0:
         module.fail_json(rc=rc, msg=err)
-    elif state == 'started' and 'Container is running' in out:
+    elif rc == 1 and 'already present' in err:
+        module.exit_json(restult=err, changed=is_upgraded)
+    elif started and 'Container is running' in out:
         module.exit_json(result=out, changed=is_upgraded)
     else:
         module.exit_json(msg=out, changed=True)
@@ -106,8 +112,8 @@ def main():
     module = AnsibleModule(
                 argument_spec = dict(
                     name    = dict(default=None, required=True),
-                    upgrade = dict(default='no', type='bool'),
-                    state   = dict(default='started', choices=['started', 'stopped']),
+                    state   = dict(default='latest', choices=['present', 'absent', 'latest']),
+                    started = dict(default='yes', type='bool'),
                 ),
             )
 


gregdek:
Thanks @krsacme for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

system/atomic_host.py
```diff
@@ -0,0 +1,96 @@
...
+    elif rc != 0:
+        module.fail_json(rc=rc, msg=err)
+    else:
+        module.exit_json(msg=out)
+
+
+def main():
+    module = AnsibleModule(
+                argument_spec = dict(
+                    state = dict(default=None, choices=['rollback', 'status', 'upgrade', 'rebase'], required=True),
```
miabbott: What about support for the `deploy` verb?

https://github.com/projectatomic/atomic/blob/master/docs/atomic-host.1.md

system/atomic_host.py
```diff
@@ -0,0 +1,96 @@
...
+    args = ['atomic', 'host']
+    args.append(state)
+
+    out = {}
+    err = {}
+    rc = 0
+
+    rc, out, err = module.run_command(args, check_rc=False)
+
+    if rc == 77:
```
miabbott: Return code 77 only means unchanged if the user explicitly passes the `--check` option.  

See the following commit - https://github.com/projectatomic/rpm-ostree/pull/226/commits/c3f1e7c85a18182669899f4065d23fec6ac0962c

system/atomic_host.py
miabbott: Also, to properly use the `rebase` verb, you'll also need to provide a `refspec` argument.

```
# atomic host rebase --help
usage: atomic host rebase [-h] [--os OS] refspec ...

positional arguments:
  refspec     Origin refspec for new deployment
  args        Additional arguments appended to the rebase method. Use `--
              --OPTION=VAL` if you want to pass additional unsupported
              arguments to rpm-ostree.

optional arguments:
  -h, --help  show this help message and exit
  --os OS     Operate on provided OSNAME
```

miabbott:
I'd like to see the `atomic host rollback/upgrade` commands also support the automatic reboot option that they have.  This may not be possible to implement because a reboot would sever the connection Ansible has made to the server, but still worth investigating.

See the reboot options to the commands below:

```
# atomic host rollback --help
usage: atomic host rollback [-h] [-r] ...

positional arguments:
  args          Additional arguments appended to the rollback method. Use `--
                --OPTION=VAL` if you want to pass additional unsupported
                arguments to rpm-ostree.

optional arguments:
  -h, --help    show this help message and exit
  -r, --reboot  initiate a reboot after rollback is prepared

# atomic host upgrade --help
usage: atomic host upgrade [-h] [-r] [--allow-downgrade] [--os OS]
                           [--check-diff]
                           ...

positional arguments:
  args               Additional arguments appended to the upgrade method. Use
                     `-- --OPTION=VAL` if you want to pass additional
                     unsupported arguments to rpm-ostree.

optional arguments:
  -h, --help         show this help message and exit
  -r, --reboot       if an upgrade is available, reboot after deployment is
                     complete
  --allow-downgrade  Permit deployment of chronologically older trees
  --os OS            Operate on provided OSNAME
  --check-diff       Check for upgrades and print package diff only
```

system/atomic_image.py
```diff
@@ -0,0 +1,102 @@
...
+    name:
+        description:
+          - Name of the container image
+        required: True
+        default: null
+    state:
+        description:
+          - Whether to install (C(install) or C(run) or C(update)), or remove (C(stop) or C(uninstall))
+        required: True
+        choices: ["install", "run", "stop", "update", "uninstall"]
```
tima: It is considered best practice that Ansible module are declarative. Refrain from verb _especially_ in a param called state like this. Look at other similar modules (don't look at the Docker module -- that one is an anti-pattern we are going to address soon enough) and you'll see state use values like "present", "latest", "absent" if its for deployment or package management. Use  "started", "restarted" and "stopped" for process management. I'd hesitate about having a module that does both deployment and service management.

system/atomic_host.py
```diff
@@ -0,0 +1,96 @@
...
+short_description: Manage the atomic host platform
+description:
+    - Manage the atomic host platform
+version_added: "2.1"
+author: "Saravanan KR @krsacme"
+notes:
+    - Host should be an atomic platform (verified by existence of '/run/ostree-booted' file)
+requirements:
+  - atomic
+options:
```
tima, line 33: The purpose of this module isn't very clear to me. I admit to not being very knowledgable about how Atomic works. Curious why only one parameter. What value does this module provide over calling atomic directly with the command or shell module?

system/atomic_host.py
```diff
@@ -0,0 +1,96 @@
...
+    state:
+        description:
+          - The state of the atomic host to be in
+        required: true
+        choices: ["rollback", "status", "upgrade", "rebase"]
```
tima: See my comments on the state param of the atomic_image module in this PR.

system/atomic_host.py
```diff
@@ -0,0 +1,96 @@
...
+DOCUMENTATION='''
+---
+module: atomic_host
+short_description: Manage the atomic host platform
+description:
+    - Manage the atomic host platform
+version_added: "2.1"
+author: "Saravanan KR @krsacme"
+notes:
+    - Host should be an atomic platform (verified by existence of '/run/ostree-booted' file)
```
cdrage, line 29: This has been fixed upstream for Fedora Rawhide / next release to use /etc/os-release to parse "atomichost" from RELEASE_ID just thought to let you know :)

system/atomic_host.py
cdrage, line 33: Main project atomic OS' use github.com/projectatomic/atomic as it's main entry-point for everything (launching containers, restarting, upgreading the host, etc)

system/atomic_host.py
cdrage: See comment above on: github.com/projectatomic/atomic :)

system/atomic_image.py
krsacme: @tima, the atomic image management is similar to the docker, it uses pull (install), start (run). This particular utility is more of both deployment and service management. I will analyze further if it can be limited to only a particular type. 
Having said, I will modify the verbs and will re-submit it.

system/atomic_host.py
krsacme, line 29: @cdrage thanks for pointing out. I will keep a track to update.

krsacme:
@miabbott As you mentioned, reboot cannot be done as part of this execution, if it required, probably the playbook should be written in such a way that on successful upgrade, the next step should be reboot and that will be the end of playbook. It is appropriate to add a caution as part of the description to remind the user about this. I am just thinking loud.

gregdek:
Thanks @krsacme for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

system/atomic_host.py
krsacme: @miabbott, by default atomic host upgrade returns 77 when the upgrade is not available. I have even checked it on rpm-ostree command, which also returns 77 when the upgrade is not available, without --check option itself. I am looking in to the code to understand the difference in the expectation. Sample output below.

```
[vagrant@atomicnode ~]$ sudo rpm-ostree upgrade 
Updating from: centos-atomic-host:centos-atomic-host/7/x86_64/standard

1 metadata, 0 content objects fetched; 313 B transferred in 2 seconds
No upgrade available.
[vagrant@atomicnode ~]$ echo $?
77
```

The changes that you have mentioned is part of the later release, and the current version of rpm-ostree is 

```
rpm-ostree 2015.11
```

Probably on the next releaes, we need to upgrade this check.

krsacme:
ready_for_review

gregdek:
Thanks @krsacme for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

system/atomic_host.py
miabbott: @krsacme You are correct about the behavior of the current released version of `rpm-ostree` in CentOS, Fedora, and Red Hat Atomic Host.  

It will get tricky if one of those OSes releases a new version of `rpm-ostree` but the others are still on 2015.11

system/atomic_host.py
```diff
@@ -0,0 +1,128 @@
...
+    - Manage the atomic host platform
+    - Reboot option for states ```rollback```, ```upgrade``` and ```deploy``` are not supported, the playbook should do explicit reboot
+version_added: "2.2"
+author: "Saravanan KR @krsacme"
+notes:
+    - Host should be an atomic platform (verified by existence of '/run/ostree-booted' file)
+requirements:
+  - atomic
+options:
+    state:
```
miabbott: Would you consider renaming this from `state` to `action` or `command`?

I feel like `state` does not accurately describe what is happening.  Using any of deploy/rebase/rollback/upgarde is really just an action and not really a state of the system.  These are more like transitions from one ostree deployment to another.

system/atomic_image.py
```diff
@@ -0,0 +1,110 @@
...
+
+import os
+import sys
+
+def core(module):
+    state_map = {'present'  : 'install',
+                 'start'    : 'run',
+                 'stop'     : 'stop',
+                 'update'   : 'update',
+                 'remove'   : 'uninstall'}
```
miabbott: Mapping these seems unnecessary and I think would be confusing to anyone used to using `atomic` on the command line.  Users will have to check the docs to verify that the state they provide to the Ansible module matches the command they are familiar with using.

Additionally, I think the use of `state` here is a misnomer and should be renamed to `command`

system/atomic_host.py
tima: Agreed @miabbott. A param called `state` has specific connotations in good Ansible module design which this does not conform. You shouldn't use `action` either since that is a play directive and could get confusing. I recommend `command` which is commonplace for most modules like this one.

system/atomic_image.py
tima: Agreed. With changing `state` to `command` these mappings do seem unnecessary and confusing and should be removed.

bcoca: bad use of state, it should be preset/absent, with present including 'update'

started=yes|no should be a different property that is a companion to 'present'.

system/atomic_image.py
```diff
@@ -0,0 +1,110 @@
...
+    err = {}
+    rc = 0
+
+    rc, out, err = module.run_command(args, check_rc=False)
+
+    if rc != 0: # something went wrong emit the msg
+        module.fail_json(rc=rc, msg=err)
+    else:
+        module.exit_json(msg=out, changed=True)
+
```
bcoca, line 112: if i'm reading this right it is 'always changed on success', this is not what users will expect from the module.

we need to check if it is present/absent and/or up to date, also if started or not and compare to the options to give users real value

system/atomic_host.py
```diff
@@ -0,0 +1,128 @@
...
+        description:
+          - The state of the atomic host to be in
+        required: true
+        choices: ["rollback", "upgrade", "rebase", "deploy"]
```
bcoca: We try to keep a declarative language in our modules, specially for state. This is mostly imperative, which i would expect of using the CLI tool directly.

Most of these choices look like state=present, but with different implications depending on refspec or revision.

system/atomic_host.py
```diff
@@ -0,0 +1,128 @@
...
+import sys
+
+def core(module):
+    state = module.params['state']
+
+    args = ['atomic', 'host']
+    args.append(state)
+
+    refspec = module.params['refspec']
+    if state == 'rebase':
```
bcoca: most of these checks can be done using the 'required_together' function from basic.py.

system/atomic_host.py
```diff
@@ -0,0 +1,128 @@
...
+    rc = 0
+
+    rc, out, err = module.run_command(args, check_rc=False)
+
+    if rc == 77:
+        module.exit_json(msg="No upgrades available")
+    elif rc != 0:
+        module.fail_json(rc=rc, msg=err)
+    else:
+        module.exit_json(msg=out, changed=True)
```
bcoca, line 82: this always returns changed  unless it fails, as explained in the other plugin, this is not what users expect

system/atomic_host.py
krsacme: As per [comments](https://github.com/ansible/ansible-modules-extras/pull/1902#discussion-diff-62712075) by @miabbott and @tima, i agree that it is appropriate to change the name from `state` to `command`, in which case, the choices can be as it is.

system/atomic_image.py
krsacme: @miabbott and @tima, I agree that the atomic_host operations are inline with ```command``` and cannot be ```state```. But the atomic_image operations are purely ```state``` operations (i will change```remove``` to ```absent``` and ```update``` to ```latest``` ). And along with ```atomic run <image>```, it is also possible to give the command to be executed on the image like ```atomic run rhel7/rhel-tools hostname```, in which case, naming it as ```command``` is not appropriate.
@bcoca 's suggestion of using```started=yes|no``` with ```present``` is explicit and easy to understand. I will follow it.

system/atomic_host.py
krsacme: `required_together` and `required_if` are really useful. I am modifying according to it. One drawback is that if i use `required_if`, it is checking for presence of the parameter along with the value, but if it not checking for not present with other parameters, like

```
                required_if = ([
                    ('command', 'rebase', ['refspec']),
                    ('command', 'deploy', ['revision']),
                ]),
```

in above example, when `command` has value of `rebase`, then `refspec` is mandatory, but having `revision` along with `rebase` is not allowed.

krsacme:
@bcoca atomic host and image modules are just commands that will be executed. For example, referring to `atomic install rhel7/rsyslog` command, the expectation of this command is to execute the command assigned to [LABEL INSTALL](https://github.com/projectatomic/atomic#atomic-install). And `atomic run rhel7/rsyslog` command does 2 things, checks if the container is running, if not  executes the command under LABEL RUN to run the container. I have added the check for already running use case. Rest of the commands are based on the command result only.

krsacme:
ready_for_review

gregdek:
Thanks @krsacme for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

tima:
I checked the code of these modules again and they look already pretty solid to me. Thanks @krsacme. I have had no chance to test them and would appreciate if someone with more atomic host familiarity like @miabbott  could do verify this work before I'd says personally ship it.

miabbott:
I was able to test out these modules using Fedora Atomic Host and RHEL Atomic Host and they worked well.  ~~I think they are ready to be included.~~

Edit: Other reviewers have pointed out some changes that would benefit the module.  See my most [recent comment](https://github.com/ansible/ansible-modules-extras/pull/1902#issuecomment-223111803).

system/atomic_host.py
```diff
@@ -0,0 +1,119 @@
...
+    - Manage the atomic host platform
+    - Reboot option for commands ```rollback```, ```upgrade``` and ```deploy``` are not supported, the playbook should do explicit reboot
+version_added: "2.2"
+author: "Saravanan KR @krsacme"
+notes:
+    - Host should be an atomic platform (verified by existence of '/run/ostree-booted' file)
+requirements:
+  - atomic
+options:
+    command:
```
detiber: So, This playbook looks like it would be great for orchestration, but doesn't really have a good ongoing configuration management story. I would expect there to be a way for me to specify a specific version to be installed (without having to handle the associated logic outside of the module).

For example, I would expect to be able to accomplish the following tasks:
- Ensure a particular version:

```
- atomic_host:
    revision: <some version>
  register: result
- <task to reboot>
   when: result | changed
```
- Or ensuring the latest (though, arguably upgrade does this):

```
- atomic_host:
    revision: latest
  register: result
- <task to reboot>
   when: result | changed
```

system/atomic_image.py
```diff
@@ -0,0 +1,109 @@
...
+    - Host should be an atomic platform (verified by existence of '/run/ostree-booted' file)
+requirements:
+  - atomic
+options:
+    name:
+        description:
+          - Name of the container image
+        required: True
+        default: null
+    command:
```
detiber: I don't think command is the appropriate argument here. Instead there should be a state argument that drives the behavior.

For example, I would expect the module to then do all the necessary things to make sure that I have the latest tag for rhel7/rhel-tools image installed and running:

```
- atomic_image:
    name: rhel7/rhel-tools
    version: latest
    state: running
```

In general, I expect Ansible modules to present an interface for describing the desired state of the system rather than providing me a way to just drive the underlying commands, since I can already easily wrap the underlying command line tooling using the command module.

system/atomic_image.py
```diff
@@ -0,0 +1,109 @@
...
+    module = AnsibleModule(
+                argument_spec = dict(
+                    name = dict(default=None, required=True),
+                    command = dict(default=None, choices=['install', 'run', 'stop', 'uninstall'], required=True),
+                ),
+            )
+
+    # Verify that the platform is atomic host
+    if not os.path.exists("/run/ostree-booted"):
+      module.fail_json(msg="Module atomic_image is applicable for Atomic Host Platforms only")
```
jlebon: We could make this module more generic by dropping this requirement. We only actually require the `atomic` package to be installed.

detiber: I agree here, We (The OpenShift Productization team) are looking to leverage the atomic tooling for running non-docker containers on both atomic and non-atomic hosts in the future.

miabbott:
@detiber said:

> In general, I expect Ansible modules to present an interface for describing the desired state of the system rather than providing me a way to just drive the underlying commands, since I can already easily wrap the underlying command line tooling using the command module.

As a relative novice to Ansible, I missed this larger view of the modules and the Ansible way of doing things.  So my review/testing was mostly focused on the functional aspect of the code.

I think these modules could be re-factored to be more declarative, as @tima pointed out in [his comment](https://github.com/ansible/ansible-modules-extras/pull/1902#discussion_r59795588).

system/atomic_host.py
krsacme: Agreed, It is not important to expose all arguments of a command (which i was doing earlier), instead focus on the usecase.

system/atomic_image.py
krsacme: @detiber atomic_run command provides install/run/stop/uninstall options which are nothing but executing the corresponding LABEL commands. It does not relate to state directly, that's the reason behind naming it as command. I understand the proposed way, i will incorporate it.

system/atomic_image.py
krsacme: OK. I will add a check to verify the atomic command existence as the only pre-condition for this module.

gregdek:
Thanks @krsacme for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

krsacme:
ready_for_review

gregdek:
Thanks @krsacme for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

system/atomic_image.py
```diff
@@ -0,0 +1,128 @@
...
+    - Host should be an atomic platform (verified by existence of '/run/ostree-booted' file)
+requirements:
+  - atomic
+options:
+    name:
+        description:
+          - Name of the container image
+        required: True
+        default: null
+    upgrade:
```
detiber: Instead of an upgrade option here, would it make sense to be able to specify a specific tag and have a 'force' option that would allow for changing the tag of a running container?

system/atomic_image.py
```diff
@@ -0,0 +1,128 @@
...
+        description:
+          - Forcefully upgrade if the container, even if the container is already running
+        required: False
+        choices: ["yes", "no"]
+        default: no
+    state:
+        description:
+          - The state of the container image
+        required: False
+        choices: ["started", "stopped"]
```
detiber: I would suggest that there should be a `present` and an `absent` state for consistency with other modules. I would also suggest `running` instead of `started`.

system/atomic_image.py
```diff
@@ -0,0 +1,128 @@
...
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION='''
+---
+module: atomic_image
+short_description: Manage the container images on the atomic host platform
+description:
+    - Manage the container images on the atomic host platform
+    - Allows to execute the commands on the container images
+    - Commands ```install``` and ```run``` will download the container image if not present already
```
detiber: The above comments are now out of date.

detiber:
@krsacme these are looking a lot better, can't wait until we can start using these modules in openshift-ansible.

system/atomic_image.py
krsacme: Removed

system/atomic_image.py
krsacme: @detiber  AFAIU atomic command does not take image version, thats why i have modeled like it. Let me know if there is a way to specify tag.

system/atomic_image.py
krsacme: @detiber I have named according to the previous comments from @tima which is 
_Look at other similar modules (don't look at the Docker module -- that one is an anti-pattern we are going to address soon enough) and you'll see state use values like "present", "latest", "absent" if its for deployment or package management. Use "started", "restarted" and "stopped" for process management. I'd hesitate about having a module that does both deployment and service management._

detiber: @krsacme thanks for the context.

@tima I think I would have to disagree with not wanting a module to do both deployment and service management, since those things are so intertwined at the container level currently, it would seem odd to me to have to specify that I want a particular version of a container image installed and then have to separately say that I want that specific version of that container image to also be running (since there could be multiple versions of a given container present).

@chouseknecht I'm curious to your take on the situation, since ansible-container faces similar issues.

system/atomic_image.py
detiber: @krsacme ah, it appears you are correct, I only see options for specifying a image to use and to update the image to the latest. I wish I had a better suggestion for usability, but maybe s/upgrade/update/ to match the underlying tooling.

system/atomic_host.py
```diff
@@ -0,0 +1,95 @@
...
+---
+module: atomic_host
+short_description: Manage the atomic host platform
+description:
+    - Manage the atomic host platform
+    - Rebooting of Atomic host platform should be done outside this module
+version_added: "2.2"
+author: "Saravanan KR @krsacme"
+options:
+    revision:
```
miabbott, line 34: Can `revision` be renamed to `version`?

It seems to make more sense since the description actually describes it as a 'version number'.

system/atomic_host.py
```diff
@@ -0,0 +1,95 @@
...
+          - The version number of the atomic host to be deployed. Providing ```latest``` will upgrade to the latest available version.
+        required: false
+        default: latest
+'''
+
+EXAMPLES = '''
+
+# Upgrade the atomic host platform to the latest version (atomic host upgrade)
+- atomic_host: revision=latest
+
```
miabbott, line 46: I think it would be helpful to show another example where the `revision` argument is not `latest`

The `atomic host deploy` command accepts version numbers and commit IDs as arguments, so I could see benefit in showcasing both.

system/atomic_image.py
```diff
@@ -0,0 +1,128 @@
...
+version_added: "2.2"
+author: "Saravanan KR @krsacme"
+notes:
+    - Host should be an atomic platform (verified by existence of '/run/ostree-booted' file)
```
miabbott: In the case of the `atomic update` command that is used below, there is no requirement that the host be an Atomic Host.

However, for the commands in the `atomic_host.py` module, the host **does** need to be an Atomic Host.

I would suggest moving this note to the `atomic_host.py` module.

system/atomic_host.py
miabbott, line 34: Ohh...I see the the help for `atomic host deploy` actually specifies `revision`.

Hmm...I guess this is fine the way it is, then.

system/atomic_image.py
tima: @detiber I see your point. I was looking to remain consistent amongst modules but should have narrowed my consideration to container tech modules rather than general purpose guidance I gave @krsacme. I will defer to you and @chouseknecht on the matter.

system/atomic_host.py
bcoca, line 34: this is what aliases are for, it can be both.

system/atomic_image.py
bcoca: actually, what makes sense here is having `started=yes|no` and leave `state=present|absent|latest` which would also take over upgrade

system/atomic_host.py
```diff
@@ -0,0 +1,95 @@
...
+    - Manage the atomic host platform
+    - Rebooting of Atomic host platform should be done outside this module
+version_added: "2.2"
+author: "Saravanan KR @krsacme"
+options:
+    revision:
+        description:
+          - The version number of the atomic host to be deployed. Providing ```latest``` will upgrade to the latest available version.
+        required: false
+        default: latest
```
bcoca, line 38: does it make sense to have `state=present|absent`?

system/atomic_image.py
detiber: @bcoca :+1:

krsacme: @tima and @chouseknecht - if no objection, then i will proceed with @bcoca suggestion.

system/atomic_host.py
krsacme, line 34: I will add alias as version.

system/atomic_image.py
tima: no objection from me.

system/atomic_host.py
krsacme, line 38: @bcoca atomic host command does providing a mechanism to remove the images as of now.

system/atomic_image.py
krsacme: Moved

system/atomic_host.py
krsacme, line 46: Added

detiber:
shipit

miabbott:
shipit

system/atomic_host.py
abadger, line 38: @krsacme Did you leave off a "not" in what you said?  Or are you agreeing that it should be added?

system/atomic_host.py
```diff
@@ -0,0 +1,100 @@
...
+        args = ['atomic', 'host', 'deploy', revision]
+
+    out = {}
+    err = {}
+    rc = 0
+
+    rc, out, err = module.run_command(args, check_rc=False)
+
+    if rc == 77 and revision == 'latest':
+        module.exit_json(msg="Already on latest")
```
abadger: Should specify changed=False explicitly here.

.travis.yml
```diff
@@ -117,7 +117,7 @@ install:
   - pip install git+https://github.com/ansible/ansible.git@devel#egg=ansible
   - pip install git+https://github.com/sivel/ansible-testing.git#egg=ansible_testing
 script:
-  - python2.4 -m compileall -fq -x 'cloud/|monitoring/zabbix.*\.py|/dnf\.py|/layman\.py|/maven_artifact\.py|clustering/(consul.*|znode)\.py|notification/pushbullet\.py|database/influxdb/influxdb.*\.py|database/mssql/mssql_db\.py|/letsencrypt\.py|network/f5/bigip.*\.py|bmc/ipmi/.*\.py' .
+  - python2.4 -m compileall -fq -x 'cloud/|monitoring/zabbix.*\.py|/dnf\.py|/layman\.py|/maven_artifact\.py|clustering/(consul.*|znode)\.py|notification/pushbullet\.py|database/influxdb/influxdb.*\.py|database/mssql/mssql_db\.py|/letsencrypt\.py|network/f5/bigip.*\.py|bmc/ipmi/.*\.py|system/atomic_.*\.py' .
```
abadger, line 120: Why is this not built for python2.4?  If there's a good reason then there also needs to be a requirements: field in the documentation that lists the newer python version.

system/atomic_host.py
```diff
@@ -0,0 +1,100 @@
...
+---
+module: atomic_host
+short_description: Manage the atomic host platform
+description:
+    - Manage the atomic host platform
+    - Rebooting of Atomic host platform should be done outside this module
+version_added: "2.2"
+author: "Saravanan KR @krsacme"
+notes:
+    - Host should be an atomic platform (verified by existence of '/run/ostree-booted' file)
```
abadger, line 29: You can put the requirements field here if python 2.6 or greater turns out to be a real requirement.

system/atomic_image.py
```diff
@@ -0,0 +1,134 @@
...
+            args = ['atomic', 'run', image]
+        else:
+            args = ['atomic', 'install', image]
+    elif state == 'absent':
+        args = ['atomic', 'uninstall', image]
+
+    out = {}
+    err = {}
+    rc = 0
+    rc, out, err = module.run_command(args, check_rc=False)
```
abadger, line 102: Since we're screenscraping stdout later, do we need to make sure that the output is in the C locale?  If so, you can add locale setting to either  the `environ_update` parameter of run_command or add to the `module.run_command_environ_update` dict.

Example:

``` python
lang_env = dict(LANG='C', LC_ALL='C', LC_MESSAGES='C')
 rc, out, err = module.run_command(args, check_rc=False, environ_update=lang_env)
```

abadger:
I found a couple bugs and had a few questions.  Made inline comments for those.

needs_revision

.travis.yml
krsacme, line 120: @abadger Handling exception for python 2.4 and other version is tricky. If i need to support 2.4, i have to follow like "e = get_exception()". Since, this is a fairly new feature (has lot of other dependencies) and python2.4 on RHEL5 is till Apr'17, I felt it makes more sense to ignore support for python2.4. I will add it in the requirements.

system/atomic_host.py
krsacme: Added

system/atomic_host.py
krsacme, line 29: Added

system/atomic_image.py
krsacme, line 102: Added

krsacme:
ready_for_review

gregdek:
Thanks @krsacme for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

.travis.yml
abadger, line 120: We don't get to make an arbitrary decision like that, unfortunately :-(  Ansible modules are supporting python2.4 until the Ansible release after April 2017 unless there's a dependency that prevents them from working with earlier python.  We might be able to stretch that a little bit... Does atomic run on RHEL5?  If it doesn't maybe we can say it only works on python-2.6 and above because atomic is needed and atomic isn't available on the platforms where python-2.4 is the minimum version.

abadger:
Further probing about the requirement on python-2.6.

needs_info

.travis.yml
miabbott, line 120: > Does atomic run on RHEL5? If it doesn't maybe we can say it only works on python-2.6 and above because atomic is needed and atomic isn't available on the platforms where python-2.4 is the minimum version.

Since `atomic` interacts with Docker mostly (and `rpm-ostree` on Atomic Host), the support for `atomic` is basically in line with Docker.  In the case of RHEL releases, Docker is currently only supported for RHEL 7 [1].  And the same goes for `rpm-ostree`.

Additionally, RHEL7 shipped with python 2.7, so I believe `atomic` would need that version as a minimum.  Not sure what that means for this module, but just wanted to report this info.

[1] https://access.redhat.com/solutions/1378023

abadger, line 120: misc and maxamillion confirmed on IRC that atomic does not work on RHEL5:

```
<maxamillion> abadger1999: yeah, atomic is only RHEL7+
```

So the python dependency is okay.  Yay!

abadger:
All issues addressed.  Merged to devel.  Thanks!

maxamillion:
shipit

maxamillion:
oh, ha ... my page hadn't refreshed yet :)

system/atomic_image.py
```diff
@@ -0,0 +1,137 @@
...
+    module.run_command_environ_update = dict(LANG='C', LC_ALL='C', LC_MESSAGES='C')
+
+    if state == 'present' or state == 'latest':
+        if state == 'latest':
+            is_upgraded = do_upgrade(module, image)
+
+        if started:
+            args = ['atomic', 'run', image]
+        else:
+            args = ['atomic', 'install', image]
```
cevich, line 95: Arguments-wise, IMHO, this doesn't feel very intuitive.  It also assumes you can run 'install' w/o first 'uninstall' - that's entirely container-image dependent.  How about making `latest` into a boolean, and reflect the desired state with `state`?

e.g. atomic_image: name=foo latest=true state=installed
(or uninstalled or running or something like that)

tima:
I just caught that this got merged. 

Great work @krsacme! Thanks for all of your time and persistence on this contribution.

abadger:
@krsacme See @cevich's note -- this will be in the 2.2 release so we have until early/mid september before we can't change the parameters anymore.

Also -- @bcoca mentioned today that this should be categorized in cloud similar to docker and lxd.  So I'll move it from system/atomic into cloud/atomic today.

abadger:
@krascme in case it wasn't apparent, if you want to make changes to address @cevich's notes, it would be a new PR.  Feel free to ping me in IRC so that we can be sure to look at any backwards incompatible parameter changes before 2.2 goes out.

cevich:
@abadger thanks for consideration.  Sorry I didn't get to make it before merge.

system/atomic_image.py
krsacme: @cevich Thanks for your comments. Sorry for delayed response, stuck with other issues. 
There was a discussion [1] with @detiber @tima @bcoca regarding how we should name the arguments. The suggestion was to not to mix package and process management together and finally, came up with above. Please let us know if it make sense.

[1] https://github.com/ansible/ansible-modules-extras/pull/1902/files/50eb61f2070dd19cc4975d2b6f3f705ff2949aa2#r74254464

system/atomic_image.py
```diff
@@ -0,0 +1,128 @@
...
+  - atomic
+options:
+    name:
+        description:
+          - Name of the container image
+        required: True
+        default: null
+    upgrade:
+        description:
+          - Forcefully upgrade if the container, even if the container is already running
```
cevich: `s/upgrade if the container/upgrade the container/`

system/atomic_image.py
```diff
@@ -0,0 +1,128 @@
...
+    upgrade = module.params['upgrade']
+    state = module.params['state']
+    is_upgraded = False
+
+    if state == 'started':
+        if upgrade:
+            is_upgraded = do_upgrade(module, image)
+        args = ['atomic', 'run', image]
+    elif state == 'stopped':
+        args = ['atomic', 'stop', image]
```
cevich: LGTM.  Nice and simple, does not hard-code any install/uninstall.

system/atomic_image.py
krsacme: @cevich You are looking at an older version, please take a look at the later revision.

system/atomic_image.py
krsacme: Thanks. Yes, the idea is not to expose all options but expose the required ones with which the most of the common uses cases can be achieved.

PR #1914, opened 2016-03-26
The current module supporting F5 BIGIP pool creation does not support…

jwitko:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

Networking/F5/bigip_pool.py
##### ANSIBLE VERSION

```
ansible 2.0.0 (v2_final 816b20af0b) last updated 2016/03/25 16:14:47 (GMT -400)
```
##### SUMMARY

The current module supporting F5 BIGIP pool creation does not support a setup where the port number must be zero to signify the pool will listen on multiple ports. This change implements that functionality and fixes an illogical conditional.

### DIFF
--- a/packaging/os/pkgin.py
+++ b/packaging/os/pkgin.py
@@ -63,10 +63,6 @@ EXAMPLES = '''
 '''
 
 
-import shlex
-import os
-import sys
-import pipes
 import re
 
 def query_package(module, pkgin_path, name):
@@ -214,14 +210,14 @@ def main():
     module = AnsibleModule(
             argument_spec    = dict(
                 state        = dict(default="present", choices=["present","absent"]),
-                name         = dict(aliases=["pkg"], required=True)),
+                name         = dict(aliases=["pkg"], required=True, type='list')),
             supports_check_mode = True)
 
     pkgin_path = module.get_bin_path('pkgin', True, ['/opt/local/bin'])
 
     p = module.params
 
-    pkgs = p["name"].split(",")
+    pkgs = p["name"]
 
     if p["state"] == "present":
         install_packages(module, pkgin_path, pkgs)


resmo:
Thanks @jwitko. To the current maintainers, @mhite please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

mhite:
I'm afraid there is some other logic that needs work.

If you supply host=1.1.1.1 and port=0, you'll end up with the following error:

`127.0.0.1 | FAILED! => {
    "changed": false,
    "failed": true,
    "msg": "both host and port must be supplied"
}`

Did I invoke it wrong?

Also, can you explain a bit more about how you would use this? I'm a little confused why you can't just use the bigip_pool module to make a pool, and then just add your pool members (same IP, multiple ports) with multiple calls to bigip_pool_member. 

If you can explain what this allows you to accomplish specifically -- maybe with a full, verbose example -- that would be extremely helpful.

Sidenote: The same logic you are dealing with is also in bigip_pool_member. The pool membership feature of bigip_pool is deprecated in favor of using bigip_pool_member.

needs_revision

jwitko:
Hey Matt,
I'm a bit confused by the extra scrutiny as nothing has changed in the code submission over all the attempts I've made at these pull requests (Which I apologize for).  You had already signed off on this exact set of changes previously? 

Setting a pool member or VIP to port zero is for hosts that are listening across a range of ports and indicates "Any port" when set on a pool member.  You could almost certainly add the same IP over and over again with many ports but if you have a host listening across 1000 ports that gets a little crazy.  Having a node with port zero is a documented and standard F5 implementation.  

An example implementation is as follows (Which works and has been tested by me with the modifications attempting to be made in this repo):

Example Use Case:  
a 'Redirector' server which is responsible for the redirection of SIP traffic to appropriate channels within my environment.  Standard SIP protocol port is 5060 however when dealing with multiple telephony carriers and multiple SIP streams it is considered best practice to spread and segregate these traffic streams across multiple ports.  When in a business situation dealing with many different carriers this can lead to well over 1000 ports being used for SIP traffic.  This traffic is then carrier to a 'redirector' server listening on all ports and performing translations as necessary.  

Here is an example of the above implementation via code:

1)  My group_var file for all redirectors within a datacenter:

```
virtual_servers:
    - name: "redirect_udp_all"
      dstaddr: "10.200.5.16"
   ** s_port: '0'**
      protocol: "udp"
      description: "SIP traffic to redirector"
      pool: "redirect_all"
      snat: { type: "snat", pool: "/Common/redirector_snat_pool" }
      rules:
        - "/Common/SIP_Fix_F5_broken_Via_processing"
        - "/Common/SIP_Logging"
        - "/Common/SIP_URI_rewrite"
      profiles:
        - { name: "udp", context: "all" }
        - { name: "redirect_sip_profile", context: "all" }

ltm_pools:
    - { name: "redirect_5060", monitors: "tcp", port: 5060 } 
    - { name: "redirect_all",  monitors: "tcp_5060", port: 0 }
```

2)  Within my playbook, inheriting the variables above, an example of the execution of the task

```
    - name: Add node to appropriate pools
      local_action:
        module: bigip_pool
        server: "{{ item.0 }}"
        user: "{{f5_admin}}"
        password: "{{f5_password}}"
        state: present
        name: "{{ item.1.name }}"
        port: "{{ item.1.port|int }}"
        host: "{{ ansible_fqdn }}"
      with_nested:
        - groups['ltm']
        - ltm_pools
      tags:
        - f5
        - add_pool
```

As to why your setup of host=1.1.1.1 and port=0 is returning that error there are only two possibilities as to why that might happen:
1. Host is defined and port is not
2. Port is defined and host is not

The error check for that can be seen here:

```
    if (host and not port) or (port and not host):
        module.fail_json(msg="both host and port must be supplied")
```

So I would have to see your implementation for possible reasons as to why that might be happening but this code has not changed and is from your original implementation.  

Please let me know if more is needed here.

mhite:
Thanks for the example -- it really helps clarify the intent of your change a lot better. I've never worked with the scenario you describe (wildcard services?), so you'll have to excuse my ignorance. I definitely have no problem supporting the port 0 scenario and am excited to incorporate your contribution.

Sorry if my testing has introduced extra scrutiny that you weren't expecting at this point. I noticed the module failure error and I want to make sure we resolve that problem before we move forward. I am only trying to point out that it does not appear that you can actually add a host with port 0 still, despite your PR. The following code still prevents the module from proceeding:

```
if (host and not port) or (port and not host):
    module.fail_json(msg="both host and port must be supplied")
```

It indeed fails because port == 0, and 0 evaluates to False. Hence the first expression in the if statement evaluates to True, causing the module to fail with the "both host and port must be supplied" error message. 

Or am I just totally misunderstanding something?

Thanks.

jwitko:
> Sorry if my testing has introduced extra scrutiny that you weren't expecting at this point.

No, my apologies for even bringing it up.  I don't know why I reacted that way, extra scrutiny is always welcome and fixing more than one thing in this change is something I'm happy to do  (May the commit squashing gods help me along the way).  

> It indeed fails because port == 0, and 0 evaluates to False.

So this was my initial thought as well and was only rebuked by the fact that I use this exact code in my setups today and it works fine.  I'm thinking the difference here might be how I'm applying it in ansible using the `port: "{{ item.1.port|int }}"`.  As I'm forcing it to be interpreted as an integer (or maybe a string containing an integer character?  not sure how the quotes affects it). 

I think your issue is still valid and your analysis of why is probably correct.  What are your thoughts on the following change?

```
    if (host and not port) or (port and not host):
        if port != 0:
          module.fail_json(msg="both host and port must be supplied")
```

gregdek:
Thanks @jwitko for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

mhite:
@jwitko -- No worries. Text is a terrible communications medium, I know. Sometimes I am too terse.

How about this:

```
if (host and port is None) or (port is not None and not host):
    module.fail_json(msg="both host and port must be supplied")
```

jwitko:
Yes that sounds better.  I'll go ahead and implement and attempt to squash, although I still don't understand how to do that.

mhite:
Awesome. I think something like this will do the trick:

http://stackoverflow.com/questions/33578468/is-it-possible-to-squash-git-commits-after-a-pr-has-been-submitted-and-changes-t

jwitko:
I do not understand how this got closed.  I am struggling to rebase this correctly and it just ended up closed?

mhite:
Github is a fickle beast. Feel free to @ me on the new PR if you go that route.

gormeroth:
We should re-open this.  Once updating, I get "msg": "valid ports must be in range 0 - 65535" when trying to delete the pool when changing the state to absent.  The funny thing is, I am not providing a port at all.  Can someone confirm that deleting pools work now?

mhite:
The final merged version is PR #1927. Let's open an issue if this is indeed a problem.

mhite:
I've opened issue #1992.

mhite:
Should be fixed in PR #1994. Please confirm fix works for you in the PR.

PR #1916, opened 2016-03-26
system/puppet: add --tags parameter

s10:
--tags [1] is used to apply a part of the node’s catalog.

In puppet:
puppet agent --tags update,monitoring

In ansible:
puppet: tags=update,monitoring

[1] https://docs.puppetlabs.com/puppet/latest/reference/lang_tags.html#restricting-catalog-runs

### DIFF
--- a/system/puppet.py
+++ b/system/puppet.py
@@ -112,6 +112,9 @@ EXAMPLES = '''
 # Run puppet using a specific piece of Puppet code. Has no effect with a
 # puppetmaster.
 - puppet: execute='include ::mymodule'
+
+# Run puppet using a specific tags
+- puppet: tags=update,nginx
 '''
 
 
--- a/system/puppet.py
+++ b/system/puppet.py
@@ -80,6 +80,12 @@ options:
     required: false
     default: None
     version_added: "2.1"
+  tags:
+    description:
+      - Comma-separated list of puppet tags without spaces to be used.
+    required: false
+    default: None
+    version_added: "2.1"
   execute:
     description:
       - Execute a specific piece of Puppet code. It has no effect with
@@ -147,6 +153,7 @@ def main():
             facter_basename=dict(default='ansible'),
             environment=dict(required=False, default=None),
             certname=dict(required=False, default=None),
+            tags=dict(required=False, default=None),
             execute=dict(required=False, default=None),
         ),
         supports_check_mode=True,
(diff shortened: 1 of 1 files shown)


system/puppet.py
```diff
@@ -147,6 +156,7 @@ def main():
             facter_basename=dict(default='ansible'),
             environment=dict(required=False, default=None),
             certname=dict(required=False, default=None),
+            tags=dict(required=False, default=None),
```
resmo: I recommend to use `type='list'`here. This would allow the use list style syntax:

```
puppet:
  tags:
    - update
    - nginx
```

but this still works as well

```
puppet: tags=update,nginx
```

I usually also add an alias `aliases=['tag']`, so this would also work:

```
puppet:
  tag: update
```

system/puppet.py
```diff
@@ -211,6 +221,8 @@ def main():
             cmd += " --show_diff"
         if p['environment']:
             cmd += " --environment '%s'" % p['environment']
+        if p['tags']:
+            cmd += " --tags '%s'" % p['tags']
```
resmo: of course this would than become `cmd += " --tags '%s'" % ','.join(p['tags'])`

system/puppet.py
```diff
@@ -227,6 +239,8 @@ def main():
             cmd += " --certname='%s'" % p['certname']
         if p['execute']:
             cmd += " --execute '%s'" % p['execute']
+        if p['tags']:
+            cmd += " --tags '%s'" % p['tags']
```
resmo: of course this would than become `cmd += " --tags '%s'" % ','.join(p['tags'])`

resmo:
Thanks @s10. To the current maintainers, @emonty, @nibalizer please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

system/puppet.py
s10: Looks pretty, but I'm concerned about collision with ansible tags [2] in case of using list type.

[2] http://docs.ansible.com/ansible/playbooks_tags.html

resmo: No worries, tags for tasks and tags for puppet can be mixed as they are not on the same level, and btw there are other modules already have tags args.

``` yaml
- puppet:
    tags:
      - update
      - nginx
  tags:
    - puppet
```

Without list, there are many things you can't do, e.g.:

``` yaml
vars:
  puppet_tags: 
    - update
    - nginx
tasks:
- puppet:
    tags: "{{ puppet_tags }}"
```

nibalizer:
:shipit:

gregdek:
Thanks again to @s10 for this PR, and thanks @emonty, @nibalizer for reviewing. Marking for inclusion.

PR #1933, opened 2016-03-30
do not set a default config for lxc containers

evgeni:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

lxc_container
##### ANSIBLE VERSION

2.0.0-0.5.beta3
##### SUMMARY

do not set a default config for lxc containers otherwise deploying user-containers fail as these require information from ~/.config/lxc/default.conf. that file is loaded by default by the the LXC if no --config was supplied

Signed-off-by: Evgeni Golov evgeni@golov.de

### DIFF
--- a/cloud/lxc/lxc_container.py
+++ b/cloud/lxc/lxc_container.py
@@ -57,7 +57,6 @@ options:
         description:
           - Path to the LXC configuration file.
         required: false
-        default: /etc/lxc/default.conf
     lv_name:
         description:
           - Name of the logical volume, defaults to the container name.
@@ -1687,7 +1686,6 @@ def main():
             ),
             config=dict(
                 type='str',
-                default='/etc/lxc/default.conf'
             ),
             vg_name=dict(
                 type='str',
--- a/cloud/lxc/lxc_container.py
+++ b/cloud/lxc/lxc_container.py
@@ -57,6 +57,7 @@ options:
         description:
           - Path to the LXC configuration file.
         required: false
+        default: null
     lv_name:
         description:
           - Name of the logical volume, defaults to the container name.


gregdek:
Thanks @evgeni. To the current maintainers, @cloudnull please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

cloudnull:
needs_revision -- because this is something that many deployers are using today and have likely developed playbooks with the config in line already I'd like to see if we can make this an optional setting in the module going forward which would support both old users of the module and new whom may want to deploy using user containers.

cloudnull:
I'd be happy to work on that if @bcoca doesn't mind me doing so.

evgeni:
Hey, so my idea was that the behavior change is minimal as the user still can define an alternative config file:
- if the user already had a custom config and wanted to use it, they put it as param and this will be working after my change
- if the user already had a custom config and for some reason did not want to use it, they maybe did not specify the config param, this would change now
- if the user had no custom config, lxc tools will use the one in /etc/lxc regardless if we pass --config or not

so the only change is in the case the user did not want to use the user-specific lxc config, which I would expect is pretty seldom.

Or did I miss something obvious?

gregdek:
Thanks @evgeni for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

cloudnull:
@bcoca sorry i missed tagged in the previous comment.

@evgeni You are correct and your change is perfect. When I read the diff I read it wrong and I assumed the config file option was being removed.

cloudnull:
shipit

gregdek:
Thanks again to @evgeni for this PR, and thanks @cloudnull for reviewing. Marking for inclusion.

cloud/lxc/lxc_container.py
```diff
@@ -57,7 +57,6 @@
         description:
           - Path to the LXC configuration file.
         required: false
-        default: /etc/lxc/default.conf
```
resmo, line 60: instead of removing the default, change its value to null or none, e.g. :`default: null`

evgeni:
@resmo thanks for the hint, done.

cloudnull:
thanks @evgeni  for the PR and @resmo for reviewing.

cloudnull:
shipit

PR #1936, opened 2016-03-31
fix security vulnerability in lxc module

retropc:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

lxc
##### SUMMARY

octal/decimal confusion makes script world-writable before executing it.

```
before:
  $ ls testit
  --wx-ws-wt 1 chris chris 0 Mar 31 22:42 testit

after:
  $ ls testit
  -rwx------ 1 chris chris 0 Mar 31 22:42 testit
```

PR #1941, opened 2016-04-01
do not use a predictable filenames in the LXC plugin

evgeni:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

lxc_container
##### ANSIBLE VERSION

Git
##### SUMMARY

The attach script of the LXC module currently uses predictable names, allowing symlink attacks.
Using mkstemp and not NamedTempFile to keep Py 2.4 compat :(

CVE-2016-3096

PoC:

```
root@container:/# cd /opt/
root@container:/opt# echo foo > foo
root@container:/opt# ln -s foo .lxc-attach-script
root@container:/opt# ls -alh
total 12K
drwxr-xr-x  2 root root 4.0K Apr  1 13:46 .
drwxr-xr-x 22 root root 4.0K Apr  1 13:39 ..
lrwxrwxrwx  1 root root    3 Apr  1 13:46 .lxc-attach-script -> foo
-rw-r--r--  1 root root    4 Apr  1 13:46 foo
root@container:/opt# cat foo 
foo

# run ansible with "container_command: echo bar"

root@container:/opt# cat foo 
#!/usr/bin/env bash
pushd "$(grep $(whoami) /etc/passwd | awk -F':' '{print $6}')"
    if [[ -f ".bashrc" ]];then
        source .bashrc
    fi
popd

# User defined command
echo bar
```

### DIFF
--- a/cloud/lxc/lxc_container.py
+++ b/cloud/lxc/lxc_container.py
@@ -144,7 +144,7 @@ options:
         description:
           - Path the save the archived container. If the path does not exist
             the archive method will attempt to create it.
-        default: /tmp
+        default: null
     archive_compression:
         choices:
           - gzip
@@ -557,13 +557,8 @@ def create_script(command):
     import subprocess
     import tempfile
 
-    # Ensure that the directory /opt exists.
-    if not path.isdir('/opt'):
-        os.mkdir('/opt')
-
-    # Create the script.
-    script_file = path.join('/opt', '.lxc-attach-script')
-    f = open(script_file, 'wb')
+    (fd, script_file) = tempfile.mkstemp(prefix='lxc-attach-script')
+    f = os.fdopen(fd, 'wb')
     try:
         f.write(ATTACH_TEMPLATE % {'container_command': command})
         f.flush()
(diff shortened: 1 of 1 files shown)


evgeni:
@retropc could you also please have a look? you mentioned you had seen more issues in #1936

gregdek:
Thanks @evgeni. To the current maintainers, @cloudnull please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

abadger:
@evgeni another path that could be fixed is archive_path.  It currently defaults to /tmp/ so an attacker could make symlinks there.  https://github.com/ansible/ansible-modules-extras/pull/1941/files#diff-cf760a9c318e06abfcf03558a58b2bc4R143  https://github.com/ansible/ansible-modules-extras/blob/devel/cloud/lxc/lxc_container.py#L1750

Since the purpose of archive is to create a backup of the container that the user can find later we probably can't make it into an unpredictable tempfile.  I think the best thing for that is to remove the default for archive_path.  then add a required_if to arg spec so that if archive is True, archive_path must be set by the user.

abadger:
@evgeni Code here looks good.  If you're around go ahead and squash your commits for easier cherry-picking and let me know.

I'm going to start working on the archive_path issue unless you tell me that you're already working on it (want to get these cherry-picked to stable-2.0 so that if we decide to do a 2.0.2rc3, it gets included)  (Not sure if we'll do an rc3 yet... but I want to make sure the code is ready if we do.)

evgeni:
@abadger I have the code ready on my laptop. Just needs a round if testing. Will push later tonight.

abadger:
@evgeni Excellent thanks.  I'll merge and cherry-pick once you're ready.

evgeni:
@abadger there, pushed.

sorry it took longer, but I just fell asleep yesterday after arriving home from the customer

abadger:
@evgeni No worries.  Thanks for taking the time to create this PR and test!

abadger:
Merged to devel, stable-2.0 and stable-1.9 branches.  CHANGELOG updated on stable-2.0 and stable-1.9.  Thanks for the fix!  it will be in our next release on all three branches.

PR #1945, opened 2016-04-02
Add git_config module

djmattyg007:
This module can be used for reading and writing git configuration at all
three scopes (local, global and system). It supports --diff and --check
out of the box.

This module is based off of the following gist:
https://gist.github.com/mgedmin/b38c74e2d25cb4f47908
I tidied it up and added support for the following:
- Reading values on top of writing them
- Reading and writing values at any scope
  The original author is credited in the documentation for the module.


gregdek:
Thanks @djmattyg007 for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

system/git_config.py
```diff
@@ -0,0 +1,207 @@
...
+        new_value = None
+
+    args = [git_path, "config"]
+    if params['list_all']:
+        args.append('-l')
+    if scope:
+        args.append("--" + scope)
+    if name:
+        args.append(name)
+
```
gundalow: Minor thing, but there is some trailing whitespace here

system/git_config.py
```diff
@@ -0,0 +1,207 @@
...
+    default: null
+  repo:
+    description:
+      - Path to a git repository for reading and writing values from a
+        specific repo.
+    required: false
+    default: null
+  scope:
+    description:
+      - ...
```
gundalow: missing description, maybe something like:
`The values are read/set from the system, global and repository local configuration files`

system/git_config.py
```diff
@@ -0,0 +1,207 @@
...
+    required: false
+    default: null
```
gundalow, line 64: As this is a fixed list you can update the help to say which options are valid by adding:
`choices: [ 'local', 'global', 'system'] ]`

system/git_config.py
```diff
@@ -0,0 +1,207 @@
...
+    This is needed if you don't want to use M(template) for the entire git
+    config file (e.g. because you need to change just C(user.email) in
+    /etc/.git/config).  Solutions involving M(command) are cumbersone or
+    don't work correctly in check mode.
+options:
+  list_all:
+    description:
+      - List all settings (optionally limited to a given scope)
+    required: false
+    default: no
```
gundalow, line 43: As this is a fixed list you can update the help to say which options are valid by adding:
`choices: [ "true", "false" ]`

system/git_config.py
```diff
@@ -0,0 +1,207 @@
...
+description:
+  - The M(git_config) module changes git configuration by invoking 'git config'.
+    This is needed if you don't want to use M(template) for the entire git
+    config file (e.g. because you need to change just C(user.email) in
+    /etc/.git/config).  Solutions involving M(command) are cumbersone or
+    don't work correctly in check mode.
+options:
+  list_all:
+    description:
+      - List all settings (optionally limited to a given scope)
```
gundalow: Minor suggestion to increase readability:
`- List all settings (optionally limited to a given I(scope))`

system/git_config.py
```diff
@@ -0,0 +1,207 @@
...
+            repo=dict(type='str'),
+            scope=dict(required=False, type='str', choices=['local', 'global', 'system']),
+            value=dict(required=False)
+        ),
+        mutually_exclusive=[['list_all', 'name'], ['list_all', 'value']],
+        required_if=[('scope', 'local', ['repo'])],
+        required_one_of=[['list_all', 'name']],
+        supports_check_mode=True,
+    )
+    git_path = module.get_bin_path('git')
```
gundalow, line 135: Need to deal with the case if git isn't found, currently this fails with

```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TypeError: sequence item 0: expected string, NoneType found
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "parsed": false}
```

Have at https://github.com/ansible/ansible-modules-extras/blob/devel/system/puppet.py#L160 for how to gracefully fail

gundalow:
Good module, for the following reasons:
- Clearly stated why this modules and why the existing modules (template/command) don't make sense. Which I agree with.
- Solid set of real world examples (+1 for etckeeper)
- Clean pep8, pyflakes and validate_branches
- Supports check mode

1) I've left a few comments on the diff, mainly to improve the HTML docs that will be generated.

2) Documentation needs a:
`requirements: [ git ]`

3) No graceful fail if git isn't available
See https://github.com/ansible/ansible-modules-extras/blob/devel/system/puppet.py#L160 for an example of how to achieve this

I've done a quick bit of functional testing, which has all look good.

Note for other testers:

```
mkdir library
cd library/
wget https://raw.githubusercontent.com/djmattyg007/ansible-modules-extras/add_git_config_module/system/git_config.py
```

djmattyg007:
Thank you a lot for the feedback! I'll do my best to update this soon.

djmattyg007:
I've updated the module and rebased my branch against the latest commits on `devel`. Please let me know if there's anything else that needs updating.

gundalow:
Thank you for the prompt turn around on those fixes :) 

I confirm that those fixes address all of my comments. I need to test the module a but more first.

system/git_config.py
```diff
@@ -0,0 +1,216 @@
...
+    alias.remotev: "remote -v"
+'''
+
+
+def main():
+    module = AnsibleModule(
+        argument_spec=dict(
+            list_all=dict(required=False, type='bool', default=False),
+            name=dict(type='str'),
+            repo=dict(type='str'),
```
mscherer: Shoud be type='path', so the filename is expansed automatically.

system/git_config.py
```diff
@@ -0,0 +1,216 @@
...
+        dir = params['repo']
+    elif params['list_all'] and params['repo']:
+        # Include local settings from a specific repo when listing all available settings
+        dir = params['repo']
+    else:
+        # Run from root filesystem to avoid accidentally picking up any local config settings
+        dir = "/"
+
+    (rc, out, err) = module.run_command(' '.join(args), cwd=dir)
+    if params['list_all'] and scope and rc == 128 and 'unable to read config file' in err:
```
mscherer, line 179: Since you look at the error message, you need to make sure the locales are properly set ( https://github.com/ansible/ansible-modules-core/pull/3257 , https://github.com/ansible/ansible-modules-core/pull/3286 for example ). Since ssh also transmit the locales from the client, this can impact remote systems.

RichVRed, line 179: I would suggest including the ability to unset a config setting, similar to other modules that use 'state' of absent or present.

Example:

```
# Unset setting in ~/.gitconfig
git_config: name=alias.ci scope=global set=no
```

would equate to:

```
git config --unset alias.ci
```

quinot:
Not clear what the added value is compared to using ini_file:

  ini_file: dest=/etc/.git/config section=user option=email value="root@{{ ansible_fqdn }}"
  ini_file: dest=~/.gitconfig section=alias option=ci value=commit
  ini_file: dest=~/.gitconfig section=alias option=ci state=absent

djmattyg007:
Thanks for the additional feedback. I should be able to push an update on the weekend.

@quinot By performing the updates with git, you can be sure that everything is set according to how git does things. You don't have to have platform-specific code to handle differences in system-wide gitconfig files, you don't have to worry about formatting issues in the file, you don't have to worry about remembering that a specific repository is bare, etc.

djmattyg007:
It also makes it remarkably easy to obtain a list of all config settings as git would see them, by using `list_all=yes` with this module.

djmattyg007:
I have updated the `repo` parameter to be of type `path`, and set the locale appropriately before running any git commands. I also rebased the branch against the latest commits on `devel` again.

Regarding adding the ability to unset configuration values: I would like to do this as I believe it's important functionality, but i'm not sure how it would work when just trying to read config values. If you set a `state` parameter and provide a default, what happens when you just want to read the value? Can you have a list of choices and the type `str`, but the default be `None`? Would `None` have to go into the list of choices?

gundalow:
@djmattyg007  Not sure, I've asked on ansible-devel on freenode for someone to have a look. If you don't get a response it might be worth you joining the Ansible New Module meeting https://github.com/ansible/community/blob/master/MEETINGS.md

mscherer:
I think that the way it is done is to have 1 module to read the value (like git_config_facts), and one to write.

system/git_config.py
```diff
@@ -0,0 +1,219 @@
...
+        required_if=[('scope', 'local', ['repo'])],
+        required_one_of=[['list_all', 'name']],
+        supports_check_mode=True,
+    )
+    git_path = module.get_bin_path('git')
+    if not git_path:
+        module.fail_json(msg="Could not find git. Please ensure it is installed.")
+
+    params = module.params
+    # We check error message for a pattern, so we need to make sure the messages appear in the form we're expecting.
```
mscherer: That comment is false, since there is no checking of the error message as far as I see.

mscherer: Ok so I found where we do check, ignore that comment.

abadger:
> If you set a state parameter and provide a default, what happens when you just want to read the value? Can you have a list of choices and the type str, but the default be None? Would None have to go into the list of choices?

IIRC:
- yes, you can have a list of choices, type str and the default is None.
- no, None does not need to be in the list of choices.

PR #1995, opened 2016-04-10
Client_secret is supposed to be kept secret, so mark it as no_log

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

typetalk


abadger:
Thanks @mscherer!

PR #2047, opened 2016-04-17
Handle empty IPS publishers

### LINKED ISSUE #975: pkg5_publisher throws KeyError with unexpected output from `pkg publisher -Ftsv`
##### Issue Type:

Bug Report
##### Ansible Version:

```
$ ansible --version
ansible 2.0.0
```
##### Summary:

I have a system with the output of publishers listed like so:

```
$ pkg publisher
PUBLISHER                             TYPE     STATUS   URI
delphix.com                           origin   online   file:///data/ips_repo/
ms.omniti.com                         origin   online   file:///data/ips_repo/
omnios                                origin   online   file:///data/ips_repo/
on-nightly                            origin   online   file:///net/pharos/export/pkg/dlpx-os-gate/nightly/repo.redist/
openindiana.org             
```

When I try to use the `pkg5_publisher` with the system configured like this, I get the following Python exception:

```
$ sudo python /tmp/ansible/ansible-tmp-1442291882.82-184265772797279/pkg5_publisher
  File "/tmp/ansible/ansible-tmp-1442291882.82-184265772797279/pkg5_publisher", line 1902, in <module>
    main()
  File "/tmp/ansible/ansible-tmp-1442291882.82-184265772797279/pkg5_publisher", line 95, in main
    modify_publisher(module, module.params)
  File "/tmp/ansible/ansible-tmp-1442291882.82-184265772797279/pkg5_publisher", line 102, in modify_publisher
    existing = get_publishers(module)
  File "/tmp/ansible/ansible-tmp-1442291882.82-184265772797279/pkg5_publisher", line 184, in get_publishers
    publishers[name][values['type']].append(values['uri'])
KeyError: ''
```

Looking at the source, `pkg5_publisher` tries to parse each line from the output of `pkg publisher -Ftsv` and assumes there will be a value for each column. Obviously this is not a valid assumption to make, as my system is proof this isn't guaranteed to be true, so it'd be nice of the module was able to handle this situation gracefully.

If I manually remove this publisher from the system with:

```
$ sudo pkg unset-publisher openindiana.org
$ pkg publisher -Ftsv
PUBLISHER   STICKY  PREFERRED   ENABLED TYPE    STATUS  URI
delphix.com true    true    true    origin  online  file:///data/ips_repo/
ms.omniti.com   true    false   true    origin  online  file:///data/ips_repo/
omnios  true    false   true    origin  online  file:///data/ips_repo/
on-nightly  true    false   true    origin  online  file:///net/pharos/export/pkg/dlpx-os-gate/nightly/repo.redist/
```

then the `pkg5_publisher` module will work, but the whole point of using this module is to remove the need to manually configure the system.

To make matters worse, I can't simply use:

```
- name: Remove openindiana.org Publisher
  command: pkg unset-publisher openindiana.org
```

prior to using `pkg5_publisher` because that command will fail if the publisher doesn't exist (e.g. on the second invocation of the playbook or role, after the publisher was removed the first time).

mavit:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

`packaging/os/pkg5_publisher.py`
##### SUMMARY

It turns out it's possible to set up a publisher with no URIs.  Handle this gracefully.

Fixes #975.

### DIFF
--- a/packaging/os/pkg5_publisher.py
+++ b/packaging/os/pkg5_publisher.py
@@ -180,13 +180,14 @@ def get_publishers(module):
             publishers[name]['origin'] = []
             publishers[name]['mirror'] = []
 
-        publishers[name][values['type']].append(values['uri'])
+        if values['type'] is not None:
+            publishers[name][values['type']].append(values['uri'])
 
     return publishers
 
 
 def unstringify(val):
-    if val == "-":
+    if val == "-" or val == '':
         return None
     elif val == "true":
         return True


gregdek:
Thanks again to @mavit. This module is going into community review.

We notice that you are one of the maintainers (@mavit) of this module, so when you think it's ready, please comment "shipit" and we will consider it for merging.

mavit:
Shipit.

resmo:
Thanks @mavit, turns out the bot missed your shipit because of case sensitiveness.

PR #2059, opened 2016-04-18
unknown type fix

swizzley:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

yum_repository
##### ANSIBLE VERSION

```
2.4.1
```
##### SUMMARY

this fixes an issue when running from tower, ERROR implementing reposdir unknown type path

### DIFF
--- a/packaging/os/yum_repository.py
+++ b/packaging/os/yum_repository.py
@@ -663,7 +663,7 @@ def main():
             proxy_password=dict(no_log=True),
             proxy_username=dict(),
             repo_gpgcheck=dict(type='bool'),
-            reposdir=dict(default='/etc/yum.repos.d', type='path'),
+            reposdir=dict(default='/etc/yum.repos.d'),
             retries=dict(),
             s3_enabled=dict(type='bool'),
             skip_if_unavailable=dict(type='bool'),


gregdek:
Thanks @swizzley. To the current maintainers, @jtyr please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

robinro:
Shouldn't tower be fixed to understand `type=path`?
It's used in quite a few other places as well.

jtyr:
If it's Tower-specific error, then I would prefer to fix it in Tower than removing the type from all modules. I'm voting to close this issue ('close_me').

gregdek:
Thanks @swizzley. One of the current maintainers, @jtyr requested to close this PR.

swizzley:
would you point to the fix in tower then please?

resmo:
@swizzley did you manually copy this module into your ansible setup or did it came with tower? This module is new in 2.1 and will probably not work with 2.0.

swizzley:
@resmo I copied it directly from the github devel release because it was not in the stable 2.0 release, figuring that it was fine since it's already in the docs here: http://docs.ansible.com/ansible/yum_repository_module.html I simply loaded it into my Library and ran the project from Tower on an agent-less node.

swizzley:
I also tried type='string' with the same result, before removing it completely

resmo:
it would be type='str'

bcoca:
closing this as it is an issue of mismatching feature with ansible version

swizzley:
@bcoca lame sauce

bcoca:
@swizzley you cannot expect us to remove a new feature to make it work with an old version, it is fine you do this locally, but there should be no reason to do this in the upstream.

swizzley:
@bcoca  I'd like like to see some sort of action, I agree that my PR should be closed, but the modules in your documentation should be supported for your premier tower clients, so patching somewhere seems like something that should be done. If this is just going to be ignored, it would make sense to remove it from your docs or at least note in the docs that tower 2.x will not support it.

bcoca:
@swizzley its not a question of tower support or not, its the underlying ansible version must match, they are not in lockstep.

The ansible docs are not the tower docs, if you use `ansible-doc` on your machine you will get the correct docs for your installed version of ansible. You cannot expect to pull things from a newer version into an older one and expect it to always work (tower is not a factor here).

PR #2069, opened 2016-04-20
Added portenforce module

ndavison:
##### ISSUE TYPE
- New Module Pull Request
##### COMPONENT NAME

portenforce
##### ANSIBLE VERSION

```
ansible 2.1.0 (devel 3c3061378b) last updated 2016/04/18 19:52:38 (GMT +1100)
  lib/ansible/modules/core: (detached HEAD 5409ed1b28) last updated 2016/04/18 19:52:51 (GMT +1100)
  lib/ansible/modules/extras: (detached HEAD 3afe117730) last updated 2016/04/18 19:53:03 (GMT +1100)
```
##### SUMMARY

This module allows the user to define whitelists for TCP and/or UDP ports that a host is allowed to have processes listening on, killing processes that are listening on a port not in the whitelist. This is designed to be a small component in a larger offensive security Ansible setup.

### DIFF
--- a/system/portenforce.py
+++ b/system/portenforce.py
@@ -78,10 +78,11 @@ def netStatParse(raw):
         if listening_search:
             splitted = line.split()
             conns = re.search('[^ ]+:([^ ]+)', splitted[3])
+            pidstr = ''
             if 'tcp' in splitted[0]:
                 proto = 'tcp'
                 pidstr = splitted[6]
-            else:
+            elif 'udp' in splitted[0]:
                 proto = 'udp'
                 pidstr = splitted[5]
             pids = re.search('([0-9]+)/', pidstr)
--- a/system/portenforce.py
+++ b/system/portenforce.py
@@ -110,15 +110,6 @@ def applyWhitelist(portspids, whitelist=list()):
                 kill_pids.append(dict(pid=p['pid'], port=p['port'], proto=p['proto'], name=p['name'], stime=p['stime']))
     return kill_pids
 
-def getPidSTime(pid):
-    p1 = Popen(['ps', '-o', 'lstart', '-p', str(pid)], stdout=PIPE, stderr=PIPE)
-    ps_output = p1.communicate()[0]
-    stime = ''
-    for line in iter(ps_output.splitlines()):
-        if 'started' not in line:
-            stime = line
-    return stime
-
 def main():
 
     module = AnsibleModule(
@@ -129,13 +120,25 @@ def main():
         supports_check_mode=True
     )
 
+    def getPidSTime(pid):
+        ps_cmd = module.get_bin_path('ps', True)
+        p1 = Popen([ps_cmd, '-o', 'lstart', '-p', str(pid)], stdout=PIPE, stderr=PIPE)
+        ps_output = p1.communicate()[0]
+        stime = ''
+        for line in iter(ps_output.splitlines()):
+            if 'started' not in line:
+                stime = line
+        return stime
+
     result = {}
     result['changed'] = False
     result['killed'] = list()
 
     try:
+        netstat_cmd = module.get_bin_path('netstat', True)
+
         # which TCP ports are listening for connections?
-        p1 = Popen(['netstat', '-plnt'], stdout=PIPE, stderr=PIPE)
+        p1 = Popen([netstat_cmd, '-plnt'], stdout=PIPE, stderr=PIPE)
         output_tcp = p1.communicate()[0]
         kill_tcp = netStatParse(output_tcp)
         for i, p in enumerate(kill_tcp):
@@ -143,7 +146,7 @@ def main():
             kill_tcp[i] = p
 
         # which UDP ports are listening for connections?
-        p1 = Popen(['netstat', '-plnu'], stdout=PIPE, stderr=PIPE)
+        p1 = Popen([netstat_cmd, '-plnu'], stdout=PIPE, stderr=PIPE)
         output_udp = p1.communicate()[0]
         kill_udp = netStatParse(output_udp)
         for i, p in enumerate(kill_udp):
--- a/system/portenforce.py
+++ b/system/portenforce.py
@@ -72,80 +72,80 @@ import re
 from subprocess import Popen, PIPE
 
 def netStatParse(raw):
-	results = list()
-	for line in iter(raw.splitlines()):
-		listening_search = re.search('[^ ]+:[^ ]+', line)
-		if listening_search:
-			splitted = line.split()
-			conns = re.search('[^ ]+:([^ ]+)', splitted[3])
-			if 'tcp' in splitted[0]:
-				proto = 'tcp'
-				pidstr = splitted[6]
-			else:
-				proto = 'udp'
-				pidstr = splitted[5]
-			pids = re.search('([0-9]+)/', pidstr)
-			if conns and pids:
-				port = conns.group(1)
-				pid = pids.group(1)
-				result = dict(pid=int(pid), port=int(port), proto=proto)
-				if result not in results:
-					results.append(result)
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @ndavison for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

system/portenforce.py
```diff
@@ -0,0 +1,153 @@
...
+        supports_check_mode=True
+    )
+
+    result = {}
+    result['changed'] = False
+    result['killed'] = list()
+
+    try:
+        # which TCP ports are listening for connections?
+        p1 = Popen(['netstat', '-plnt'], stdout=PIPE, stderr=PIPE)
```
mscherer: Not portable on freebsd:

```
[root@freebsd ~]# netstat -plnt
netstat: lnt: unknown or uninstrumented protocol
[root@freebsd ~]# uname -a
FreeBSD freebsd.example.org 10.0-RELEASE-p7 FreeBSD 10.0-RELEASE-p7 #0: Tue Jul  8 06:37:44 UTC 2014     root@amd64-builder.daemonology.net:/usr/obj/usr/src/sys/GENERIC  amd64
```

system/portenforce.py
```diff
@@ -0,0 +1,153 @@
...
+    results = list()
+    for line in iter(raw.splitlines()):
+        listening_search = re.search('[^ ]+:[^ ]+', line)
+        if listening_search:
+            splitted = line.split()
+            conns = re.search('[^ ]+:([^ ]+)', splitted[3])
+            if 'tcp' in splitted[0]:
+                proto = 'tcp'
+                pidstr = splitted[6]
+            else:
```
mscherer: If something is neither tcp or udp, this would be classified as udp, that's likely not what you want. It would be ok because there is previous filtering since we only look on a specific option of netstat, but I would suggest to be more defensive in this part of the code.

system/portenforce.py
```diff
@@ -0,0 +1,153 @@
...
+                proto = 'udp'
+                pidstr = splitted[5]
+            pids = re.search('([0-9]+)/', pidstr)
+            if conns and pids:
+                port = conns.group(1)
+                pid = pids.group(1)
+                result = dict(pid=int(pid), port=int(port), proto=proto)
```
mscherer: If a process is listening on 2 ports (or udp + tcp), it will be listed twice there. So if the pid is reused in the mean time, it will be killed again a 2nd time, while it would be a different process.

system/portenforce.py
```diff
@@ -0,0 +1,153 @@
...
+    # gather all the pids to kill
+    kill_pids_tcp = applyWhitelist(kill_tcp, module.params['whitelist_tcp'])
+    kill_pids_udp = applyWhitelist(kill_udp, module.params['whitelist_udp'])
+    killed = list(kill_pids_tcp)
+    killed.extend(x for x in kill_pids_udp if x not in kill_pids_tcp)
+
+    # kill! kill!
+    if not module.check_mode:
+        for p in killed:
+            p1 = Popen(['kill', str(p['pid'])], stdout=PIPE)
```
mscherer: What if the pid of the process got changed between the netstat exec and the kill exec, wouldn't it kill another unrelated pid ?

ndavison: This is a good point. It may be mitigated a bit by capturing and comparing process name, but the possibility would remain.

system/portenforce.py
ndavison: Good pick up - [ab76f86](https://github.com/ndavison/ansible-modules-extras/commit/ab76f868ef0ea7e5d3f512927b6017a96f579b14) addresses this by creating a new unique list just for the offending pids.

system/portenforce.py
ndavison: I've added a check on process stime before killing.

system/portenforce.py
ndavison: Would it be acceptable to add a note about not being compatible with freebsd?

system/portenforce.py
```diff
@@ -0,0 +1,182 @@
...
+def applyWhitelist(portspids, whitelist=list()):
+    kill_pids = list()
+    for p in portspids:
+        if int(p['port']) not in whitelist and str(p['port']) not in whitelist:
+            if dict(pid=p['pid'], port=p['port'], proto=p['proto'], name=p['name'], stime=p['stime']) not in kill_pids:
+                kill_pids.append(dict(pid=p['pid'], port=p['port'], proto=p['proto'], name=p['name'], stime=p['stime']))
+    return kill_pids
+
+def getPidSTime(pid):
+    p1 = Popen(['ps', '-o', 'lstart', '-p', str(pid)], stdout=PIPE, stderr=PIPE)
```
mscherer: That is not portable on solaris, but it work on freebsd and netbsd, albeit with a different output format (on both) that wouldn't be filtered by the grep ( ie, STARTED without leading spaces)

system/portenforce.py
```diff
@@ -0,0 +1,182 @@
...
+    p2 = Popen(['grep', '^[^ ]'], stdin=p1.stdout, stdout=PIPE, stderr=PIPE)
```
mscherer: I suspect this would take less ressources and be cleaner to do that in python directly.

system/portenforce.py
mscherer: I would rather see that being checked in the code (ie, verify the platform). 

In fact, I also check on netbsd and solaris and that's also blocked. And obviously, this is not working on windows (and likely os x, but I do not have one to test)

system/portenforce.py
```diff
@@ -0,0 +1,182 @@
...
+    kill_unique = list()
+    for p in kill_all:
+        if p['pid'] not in kill_unique:
+            kill_unique.append(p)
+
+    # kill! kill!
+    if not module.check_mode:
+        for p in kill_unique:
+            if p['stime'] == getPidSTime(p['pid']):
+                p1 = Popen(['kill', str(p['pid'])], stdout=PIPE)
```
mscherer: What about using os.kill rather than a Popen ?

system/portenforce.py
```diff
@@ -0,0 +1,182 @@
...
+        # which TCP ports are listening for connections?
+        p1 = Popen(['netstat', '-plnt'], stdout=PIPE, stderr=PIPE)
+        output_tcp = p1.communicate()[0]
+        kill_tcp = netStatParse(output_tcp)
+        for i, p in enumerate(kill_tcp):
+            p['stime'] = getPidSTime(p['pid']) 
+            kill_tcp[i] = p
+
+        # which UDP ports are listening for connections?
+        p1 = Popen(['netstat', '-plnu'], stdout=PIPE, stderr=PIPE)
```
mscherer: If you run a command, you have to use module.get_bin_path, since it will verify the binary is here ( check the others modules for example ). Some distribution no longer install netstat by default (for example, RHEL 7, because net-tools is replaced by iputils)

system/portenforce.py
ndavison: I have added a platform check, failing on anything but Linux. To extend beyond Linux I would likely need other contributions. Alternatively, if I introduce psutil as a dependency I could refactor away from system commands directly, but would this be a worse overall posture for an Ansible module?

gregdek:
Thanks @ndavison for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

ndavison:
ready_for_review

gregdek:
Thanks @ndavison for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

PR #2075, opened 2016-04-21
new modules aws_tags for aws cloud

ozbillwang:
##### ISSUE TYPE
- New Module Pull Request
##### COMPONENT NAME

aws_tags
##### ANSIBLE VERSION

```
Ansible 2.0.1+
```
##### SUMMARY

Generate facts from aws ec2 instance tags, and register for other modules.

Best practice

Suppose when you created the instances, assign IAM role with AmazonEC2ReadOnlyAccess policy.

aws_tags module works perfect with ansible ec2_facts module

```
- name: get ec2 facts
  action: ec2_facts

- name: get instance tags
  action: aws_tags instance_id={{ ansible_ec2_instance_id }} region={{ ansible_ec2_placement_region }}

- name: export aws_tags
  command: echo {{ ansible_ec2_tag_Name }}
```


gregdek:
Thanks @SydOps for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

ozbillwang:
some weird error messages, let me check. 

```
ERROR: version_added should be 2.1. Currently 1.0
ERROR: GPLv3 license header not found
```

ozbillwang:
Please review, thanks.

@jsmartin @scicoin-project @tombamford @garethr @jarv @jsdalton @silviud @adq @zbal @zeekin @willthames @lwade @carsongee @defionscode @tastychutney @bpennypacker @loia

juliedavila:
Why not make this a lookup plugin instead?

ozbillwang:
@defionscode 

Will it have to be lookup plugin? There are a lot of aws cloud modules, but I don't see any cloud related  codes in current lookup plugin.

We need modules or plugin to be consistent.

willthames:
Aside from the fact that this definitely shouldn't be called `aws_tags` (as it only works against EC2 objects), the need for this module suggests something is lacking elsewhere. 

I'd rather augment `ec2_facts` to return the tags as well as the instance metadata, if that's possible (the way that `ec2_facts` works, by logging onto the instance to get the facts, rather than querying the EC2 service, might mean it's not). 

The other thing worth noting is that this information is already available using the existing ec2 dynamic inventory script, if you're able to use that instead.

ozbillwang:
@defionscode 

I did research about aws ec2 facts, compare with below ways, in fact the output are similar

```
ansible all -m setup -a 'filter=facter_ec2_*'
```

and

```
 name: get ec2 facts
  action: ec2_facts
```

Module `setup` and  `ec2_facts` both get the `metadata` facts only, which missed ec2 tags. 

Regarding inventory `ec2.py` which is not in both core or extras modules, export is controlled by separated config file `ec2.ini`, and the code is not efficient enough (run 4 times connection for every query) to make it very slow. And the code is not suitable to fetch the individual ec2 instance information. 

Core module `ec2_facts` has missed informations from beginning. So what I can do?
1. raise pull request to core module ec2_facts dirctly
2. raise pull request to add extra module, as I did here. Which I think it will be easier to get merge than Core module.

Regarding the name of the module `aws_tags`, in fact I'd like to call it `ec2_tags`, but there is another modules named it already, which is only used to manage ec2 tags, such as create and remove tags. 

The idea I use `aws_tags` as module name is to have the space to extend the usage for this module, which will not only collect ec2 instances tags, it will collect volume, vpc, or other tags.

ozbillwang:
wait, module `ec2_tags` has the list function, let me check if it can list ec2 instance tags directly. 

```
Creates, removes and lists tags from any EC2 resource
```

ozbillwang:
Ok, this module is not required. I can use module `ec2_tags` to get the same information. 

```
- name: get ec2 facts
  action: ec2_facts

- name: list the tags for the instance
  ec2_tag:
    region: "{{ ansible_ec2_placement_region }}"
    resource: "{{ ansible_ec2_instance_id }}"
    state: list
  register: ec2_tags

- name: check tags
  shell: echo {{ ec2_tags.tags.Name }}
```

PR #2081, opened 2016-04-21
Streamlined ec2_elb_facts module return values with naming in ec2_elb…

stefanhorning:
##### ISSUE TYPE
- Bugfix/Change Pull Request
##### COMPONENT NAME

Change for `cloud/amazon/ec2_elb_facts.py` module
##### ANSIBLE VERSION

```
ansible 2.0.0.2
```
##### SUMMARY

Streamlined `ec2_elb_facts` module return values with naming in `ec2_elb_lb`. As ec2_elb_lb uses hosted_zone_name / hosted_zone_id this module should do the same so can use return values interchangably. See https://github.com/ansible/ansible-modules-core/blob/devel/cloud/amazon/ec2_elb_lb.py#L502-L503

Although this change would certainly break backwards compatibility for this module.


gregdek:
Thanks @stefanhorning. To the current maintainers, @mjschultz please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

mjschultz:
@gregdek: I'm not sure what the ansible policy is on changing return names because it'll end up break someones script somewhere which they generally seem to want to avoid. Is there a way to mark a field as deprecated and suggest the new name?

@stefanhorning: We can certainly just dupe the field names for now so that we don't break existing scripts and can use the output as input to the ec2_elb module (which is what I intended the output to be capable of). If you make that change I'll mark this for inclusion for sure.

stefanhorning:
@mjschultz Yes, that seems like reasonable solution. I brought back the old return keys now.

mjschultz:
Thanks!

shipit

gregdek:
Thanks again to @stefanhorning for this PR, and thanks @mjschultz for reviewing. Marking for inclusion.

PR #2103, opened 2016-04-26
netconf module with edit-config operation

lpenz:
##### ISSUE TYPE
- New Module Pull Request
##### COMPONENT NAME

netconf module
##### ANSIBLE VERSION

```
ansible 2.0.2.0
```
##### SUMMARY

This PR adds a module that allows the user to configure a network device using netconf.
The module is named netconf_config, and supports only the "edit-config" netconf operation with a user-provided XML text. The operation itself is very flexible, though, and the user can add, replace or remove configuration entities with it (examples are provided). The user should observe the idempotency of the provided XML.

A netconf_facts submodule might be useful in the feature, that's why the module is not named simply "netconf"


gregdek:
Thanks @lpenz for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

network/netconf/netconf_config.py
```diff
@@ -0,0 +1,229 @@
...
+def main():
+
+    module = AnsibleModule(
+        argument_spec=dict(
+            host=dict(type='str', required=True),
+            port=dict(type='int', default=830),
+            hostkey_verify=dict(type='bool', default=True),
+            username=dict(type='str', required=True, no_log=True),
+            password=dict(type='str', required=True, no_log=True),
+            operation=dict(choices=['edit_config'], default='edit_config'),
```
privateip: Is this argument needed?   If there is only one choice, I don't think it needs to be exposed to the module playbook designer

network/netconf/netconf_config.py
```diff
@@ -0,0 +1,229 @@
...
+        return
+    retkwargs['server_capabilities'] = list(m.server_capabilities)
+    try:
+        changed = netconf_edit_config(
+            m=m,
+            xml=module.params['xml'],
+            commit=True,
+            retkwargs=retkwargs,
+        )
+    finally:
```
privateip, line 212: Not compatible with py24... consider revising

network/netconf/netconf_config.py
lpenz: You are right, it's not even used. I intend to add other operations, but I'll remove it until it is needed.

network/netconf/netconf_config.py
lpenz, line 212: I think using "try", "except" and "finally" together is incompatible (I'm guessing from what is written at https://docs.python.org/2.4/tut/node10.html and https://github.com/schacon/hg-git/issues/56). That is not the case here, as there is no "except"... Am I missing something?

resmo:
LGTM

gregdek:
Thanks again to @lpenz for this PR, and thanks @ansible/core for reviewing. Marking for inclusion.

PR #2106, opened 2016-04-26
Rename azure_deploy to azure_rm_deployment. Refactor to use azure_com…

chouseknecht:
##### ISSUE TYPE
- New Module Pull Request
##### COMPONENT NAME

azure_rm_deployment
##### ANSIBLE VERSION

```
ansible 2.1.0 (docker_common 8b50e93151) last updated 2016/04/25 15:04:12 (GMT -400)
  lib/ansible/modules/core: (devel 804a8e6378) last updated 2016/04/21 18:24:50 (GMT -400)
  lib/ansible/modules/extras: (devel 3031105e78) last updated 2016/04/21 18:24:27 (GMT -400)
  config file =
```
##### SUMMARY

Changed name to match naming convention of Azure Resource Manager modules, and refactored to use azure_rm_common utility module.


gregdek:
Thanks @chouseknecht for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

cloud/azure/azure_rm_deployment.py
```diff
@@ -0,0 +1,646 @@
...
+  resource_group_name:
+    description:
+      - The resource group name to use or create to host the deployed template
+    required: true
+  location:
+    description:
+      - The geo-locations in which the resource group will be located.
+    required: false
+    default: westus
+  state:
```
bcoca, line 41: missing requried

cloud/azure/azure_rm_deployment.py
```diff
@@ -0,0 +1,646 @@
...
+      adminUsername:
+        value: devopscle
+      adminPassword:
+        value: Password1!
+      dnsNameForPublicIP:
+        value: devopscleazure
+'''
+
+RETURN = '''
+msg:
```
bcoca: do not add the default keys to return

cloud/azure/azure_rm_deployment.py
```diff
@@ -0,0 +1,646 @@
...
+  description: String indicating if the deployment was created or deleted
+  returned: always
+  type: string
+  sample: "deployment created"
+deployment:
```
bcoca, line 319: inadequate return docs, just passing a key and a sample of a complex structure is not very useful documentation, the subkeys should be documented also, look at stat module for examples.

cloud/azure/azure_rm_deployment.py
```diff
@@ -0,0 +1,646 @@
...
+            "sshCommand": {
+                "type": "String",
+                "value": "ssh chouseknecht@testvm9910001.westus.cloudapp.azure.com"
+            }
+        }
+    }
+'''
+
+import time
+import yaml
```
bcoca: yaml is not part of core python and cannot be assumed present, use try/except to capture this fact and return useful message to user.

cloud/azure/azure_rm_deployment.py
```diff
@@ -0,0 +1,646 @@
...
+            deployment = self.deploy_template()
+            self.results['deployment'] = dict(
+                name=deployment.name,
+                group_name=self.resource_group_name,
+                id=deployment.id,
+                outputs=deployment.properties.outputs,
+                instances=self._get_instances(deployment)
+            )
+            self.results['changed'] = True
+            self.results['msg'] = 'deployment created'
```
bcoca: should we not check if it exists already? like it is done for deletion, otherwise this 'always updates' and is always changed.

cloud/azure/azure_rm_deployment.py
chouseknecht, line 41: Actually, it has a default, so required is not required. But adding anyways.

cloud/azure/azure_rm_deployment.py
chouseknecht: Removing defaults.

cloud/azure/azure_rm_deployment.py
chouseknecht: pyyaml is a prerequisite of ansible, as you know. so one cannot run ansible without it installed. here is what happens when you run ansible without pyyaml installed:

```
ansible-playbook -i inventory test_deployment.yml
ERROR! Unexpected Exception: No module named yaml
the full traceback was:

  File "/Users/chouseknecht/projects/azure/ansible/bin/ansible-playbook", line 79, in <module>
    mycli = getattr(__import__("ansible.cli.%s" % sub, fromlist=[myclass]), myclass)
  File "/Users/chouseknecht/projects/azure/ansible/lib/ansible/cli/__init__.py", line 27, in <module>
    import yaml
```

Not sure how you expect the module to trap that, but just satisfy you I'm adding the unnecessary try/except block.

cloud/azure/azure_rm_deployment.py
chouseknecht: Not sure if there is a way to know if the deployment was created or updated. That's a question for the original authors. For now I've changed the messages to something more generic - 'deployment succeeded'

cloud/azure/azure_rm_deployment.py
bcoca, line 41: even if there is a default, required field is ALWAYS requried.

cloud/azure/azure_rm_deployment.py
bcoca: that is only for the controller machine, for targets only json is required. As much as cloud modules normally run on the same machine as Ansible, this is not a guarantee nor can it be a requirement, there are plenty of cases in which this module will run on a different machine.

cloud/azure/azure_rm_deployment.py
bcoca: for delete (just a few lines down) you detect if it exists. 

A different question is if the deployment has changed or not and if it should be updated or not.

cloud/azure/azure_rm_deployment.py
chouseknecht, line 319: I removed the actually useful JSON, and replaced it with minimal YAML. This is not a better UX experience. However, I am not going to translate nested JSON into YAML and provide description, returned, type, etc. down to the nth level.

cloud/azure/azure_rm_deployment.py
chouseknecht: Again, take it up with the original authors. I have spent all the time I am going to spend on this module. It's good enough.

chouseknecht:
@nitzmahone @mckerrj 

I have made all the changes I'm going to make on this module. I have spent all the time I'm going to spend on it. It's in an infinitely better place than where it started. It has been tested, and it works. I get that it may not be just perfect enough for a core module, but imho it is more than sufficient for an extras module. Can we please get this merged.

bcoca:
I think you are confusing what 'extras'  is for, its not for low quality, its for community maintained.

The barrier to entry is lower in that we don't question 'is this useful?' or 'do we want to maintain this?' but we still examine it for consistency, quality and user friendliness. 

Putting a module in 'extras' is not an excuse to ignore these things.

mckerrj:
Er...other than build stuff, why isn't this going in?  If there aren't further specifics or critical issues, @chouseknecht  please check with Jimi-c and merge.

PR #2119, opened 2016-04-27
message template variable fix

hypergig:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

datadog_monitor.py
##### ANSIBLE VERSION

```
ansible 1.9.2
```
##### SUMMARY

Due to ansible/jinja2 templating, it is difficult to use the monitor message template variables as they need to be surrounded by `{{` and `}}`, this change addresses that issue by allowing the user to use `[[` and `]]` instead. @skornehl, I am not sure if this is something you would approve, as it does change some base functionality, but I will say that I have tried everything to escape those curly braces on the ansible side. Ansible variables are resolved recursively so this is a real tricky problem.  Please let me know if there is anything you would like for me to change. Thanks!

### DIFF
--- a/monitoring/datadog_monitor.py
+++ b/monitoring/datadog_monitor.py
@@ -107,7 +107,7 @@ datadog_monitor:
   name: "Test monitor"
   state: "present"
   query: "datadog.agent.up".over("host:host1").last(2).count_by_status()"
-  message: "Some message."
+  message: "Host [[host.name]] with IP [[host.ip]] is failing to report to datadog."
   api_key: "9775a026f1ca7d1c6c5af9d94d9595a4"
   app_key: "87ce4a24b5553d2e482ea8a8500e71b8ad4554ff"
 
--- a/monitoring/datadog_monitor.py
+++ b/monitoring/datadog_monitor.py
@@ -63,7 +63,7 @@ options:
         description: ["The name of the alert."]
         required: true
     message:
-        description: ["A message to include with notifications for this monitor. Email notifications can be sent to specific users by using the same '@username' notation as events."]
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @hypergig. To the current maintainers, @skornehl please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

skornehl:
@hypergig thanks for this PR. Give me please a few days to thing this through and test it. 
I'm not sure, but I believe I already had some project where I did something similar. Could you provide an example in the docs?

hypergig:
@skornehl yes certainly

hypergig:
@skornehl c0013af14cb75d4ff189ee5c445c32a14d471bb0, used the existing documented example to demonstrate how you would use builtin monitoring message template variables with ansible.

skornehl:
@hypergig I believe this is an improvement. This is indeed much better than escaping/using raw and much more readable.

Thank you!
shipit

gregdek:
Thanks again to @hypergig for this PR, and thanks @skornehl for reviewing. Marking for inclusion.

hypergig:
Thank you @skornehl, it is my pleasure.

Just a note about your previous comment, escaping and using raw still doesn't address the issue. For instance:

```
08:11 $ ansible all -i localhost, --connection=local -m shell -a "echo 'This should be fully escaped but is not {%raw%}{{Hello}}{%endraw%}'"
localhost | success | rc=0 >>
This should be fully escaped but is not {#Hello#}
```

Obviously `{# #}` won't work.  I tried everything, even html encoding for `{ }` which did sort of work, but datadog didn't decode it before setting the monitor message.

PR #2132, opened 2016-04-29
Zypper repo autoimport keys

### LINKED ISSUE #65: zypper_repository module should have a fingerprint option
When adding a repository using zypper_repository, its public key may not automatically be stored as trusted. In case of the "zypper" command, it would ask interactively whether a certain GPG key fingerprint would be accepted by the user or  not. However, when scripting this using the ansible zypper_repository module, this is not possible.

Some users use the "disable_gpg_check" option, but this disables the GPG check completely, thus opening a security vulnerability.

Thus, the user of the ansible zypper_repository module should be able to explicitly specify an acceptable GPG key (or multiple acceptable GPG keys) by verbatimly quoting the fingerprint of that key. This way, it is prevented that untrusted software is installed using ansible.
## Example

Instead of 

```
zypper_repository: repo=http://download.opensuse.org/repositories/Application:/Geo/openSUSE_13.1/ name=/Application:/Geo/openSUSE_13.1/ state=present disable_gpg_check=yes
```

use

```
zypper_repository: repo=http://download.opensuse.org/repositories/Application:/Geo/openSUSE_13.1/ name=/Application:/Geo/openSUSE_13.1/ state=present acceptable_gpg_key_fingerprint=195E211106BC205D2A9C2222CC7F07489591C39B
```

robinro:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

packaging/os/zypper_repository.py
##### ANSIBLE VERSION

2.2, devel
##### SUMMARY

Adds the option `auto_import_keys` to the zypper_repository module, which imports the repo key automatically when a repo is added. Fixes #65

Also there is some cleanup of unused code and the return values are more standardized, returning stdout, stderr and the return code on failures.


gregdek:
Thanks @robinro. To the current maintainers, @matze please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

alxgu:
"Has an effect only if state is `present'."

Maybe you should add this hint into the doc. So it doesn't remove the gpg key if the repo will be removed, but on the other hand the option name should be is clearly enough.

shipit

robinro:
@alxgu thanks for the feedback. I reworded the doc.

robinro:
I added an option `runrefresh` which forces a run of `zypper refresh`. This also implements `zypper_repository: name=* runrefresh=yes`, which can refresh all repos (useful as a handler maybe). See https://github.com/ansible/ansible-modules-extras/pull/2411 for the discussion on adding refresh to the zypper module.

robinro:
The build failure complains about my rename of `refresh` to `autorefresh`. Since the old name is still there as an alias and the new name better distincts between `autorefresh` and `runrefresh` I think this should be ok.

gregdek:
Thanks @robinro for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

gregdek:
@robinro A friendly reminder: this pull request has been marked as needing your action. If you still believe that this PR applies, and you intend to address the issues with this PR, just let us know in the PR itself and we will keep it open pending your changes. When you do address the issues, please respond with ready_for_review in your comment, so that we can notify the maintainer.

robinro:
ready_for_review

gregdek:
Thanks @robinro. To the current maintainers, @matze please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

alxgu:
@matze Could you have a look on that PR? Thanks

packaging/os/zypper_repository.py
```diff
@@ -238,14 +264,16 @@ def get_zypper_version(module):
...
 
-def fail_if_rc_is_null(module, rc, stdout, stderr):
-    if rc != 0:
-        #module.fail_json(msg=stderr if stderr else stdout)
-        if stderr:
-            module.fail_json(msg=stderr)
-        else:
-            module.fail_json(msg=stdout)
+    rc, stdout, stderr = module.run_command(cmd, check_rc=True)
+    return rc, stdout, stderr
```
matze, line 275: You could return the result directly from `module.run_command` and avoid the superfluous assignments here and similar places above. But from my point of view this is not a blocker for the PR.

matze:
shipit

gregdek:
Thanks again to @robinro for this PR, and thanks @matze for reviewing. Marking for inclusion.

packaging/os/zypper_repository.py
robinro: The idea was to make each of the functions self-documenting, by naming the return values.
For the next refactor it's probably a good idea to move the run_command to the main function and only build the command strings in the individual ones. Then one could also better deal with multiple return codes.

Thanks for the review!

packaging/os/zypper_repository.py
```diff
@@ -59,12 +59,14 @@
...
         choices: [ "yes", "no" ]
-    refresh:
+    autorefresh:
         description:
             - Enable autorefresh of the repository.
         required: false
         default: "yes"
         choices: [ "yes", "no" ]
+        aliases: [ "refresh" ]
+        version_added: "2.2"
```
resmo: this was already there before 2.2, no need to add version_added, build will fail, but this is "ok" in this case. Please remove it.

robinro:
@resmo: Thanks for the review. I removed the version_added line.

PR #2133, opened 2016-04-29
Add ec2_elb_tag module (see issue #439)

ericcitaire:
##### ISSUE TYPE

Feature Pull Request
##### COMPONENT NAME

ec2_elb_tag
##### ANSIBLE VERSION

```
ansible 2.0.2.0
  config file = 
```
##### SUMMARY

Add ec2_elb_tag module

Fixes #439

This module creates, removes or lists ELB tags.


ericcitaire:
Related to https://github.com/ansible/ansible-modules-core/pull/3163

gregdek:
Thanks @ericcitaire for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

admscale:
Great stuff ! No need to wrap cli shell anymore ...

gedl:
Great stuff! I'm with @admscale, no more shell commands or extra dependencies.

wimnat:
This shouldn't be a separate module. There's already a merged PR to add tag functionality to ec2_elb_lb

ericcitaire:
@wimnat : Adding tag functionality to ec2_elb_lb is great. But having a separate module is really useful if you need to tag an existing resource which is not created by Ansible.
It's very similar to the ec2_tag module which adds tags on existing resource, even if the ec2 module is able to tag an ec2 at creation.
Here the desired state is not to have a tagged ELB present or absent, the desired state is to have a bunch of tags present or absent on an existing ELB.

wimnat:
@ericcitaire i see your point but in that case, would it not be better to use the ec2_tag module?   Add a parameter elb_name that is mutually exclusive of the resource id parameter and then if you get that passed in you can just do the functionality you've written here.

ericcitaire:
@wimnat Why not. I personnaly prefer simple contracts and a clear separation of concerns. Either way, you'll need two separate actions to tag both resources with ids and ELBs.

PR #2136, opened 2016-04-30
Fix deploy

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

deploy_helper

### DIFF
--- a/web_infrastructure/deploy_helper.py
+++ b/web_infrastructure/deploy_helper.py
@@ -230,8 +230,6 @@ procedure to remove it during cleanup.
 class DeployHelper(object):
 
     def __init__(self, module):
-        module.params['path'] = os.path.expanduser(module.params['path'])
-
         self.module    = module
         self.file_args = module.load_file_common_arguments(module.params)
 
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @mscherer. To the current maintainers, @ramondelafuente please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

ramondelafuente:
@gregdek I'm afraid I'm going to need a little help on this one. Is it policy to use `path` instead of `str` for module parameters? I'm not familiar with it, so I'm not sure about the implications of this change.

mscherer:
'path' is just 'str' with proper path expansion, see https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/basic.py#L1394

The change is here so we can be sure to have a consistent behavior for path arguments. 

And maybe in the future, if we need to do some conversion of path or anything, we would be able to do it in one single place (even if that's just a theorical possibility, I do see how it would be needed)

abadger:
Took a look.  I was a little hesitant about expanding shared_path and current_path as those can be relative and the path type was created with absolute paths in mind but looking at the code I can't think of any way that this will break.

Should releases_path also be made into type path?

PR #2207, opened 2016-05-10
iptables module - add position argument (append/prepend/line-number)

vansickle:
##### ISSUE TYPE

Feature Pull Request
##### COMPONENT NAME

iptables
##### ANSIBLE VERSION

```
ansible 2.0.2.0
```
##### SUMMARY

Adds argument which describes how iptables rule must be setup - appending (iptables -A) or prepend (insert, -I). Very usable for simple chaining of rules.

### DIFF
--- a/system/iptables.py
+++ b/system/iptables.py
@@ -266,10 +266,10 @@ options:
     description:
       - "Specifies the error packet type to return while rejecting."
     required: false
-  setup:
+  position:
     version_added: "2.2"
     description:
-      - "Specifies setup method: can be 'append' (default) or 'prepend'"
+      - "Specifies the position where rule will be added: can be 'append' (default), 'prepend' or line number"
     required: false
     default: append
 '''
@@ -355,6 +355,8 @@ def push_arguments(iptables_path, action, params):
     cmd = [iptables_path]
     cmd.extend(['-t', params['table']])
     cmd.extend([action, params['chain']])
+    if 'rulenum' in params:
+        cmd.append(params['rulenum'])
     cmd.extend(construct_rule(params))
     return cmd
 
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @vansickle for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

system/iptables.py
```diff
@@ -266,6 +266,12 @@
     description:
       - "Specifies the error packet type to return while rejecting."
     required: false
+  setup:
```
bcoca: setup does not sound right to me (yep, i'm bikeshedding), it might make more sense to have `postition=apend|prepend|<line number>`

vansickle:
ready_for_review

gregdek:
Thanks @vansickle. To the current maintainers, @LinusU please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

LinusU:
I'm very much against this approach for a number of reasons, and this has actually been discussed before a few times.

This will _only_ work if the rules are not already present, it will not do any attempt to reorder rules and thus goes against one of the most basic expectation of all Ansible rules; that it doesn't matter in which order you run them, or how the current state looks, you just describe the end state.

If this is going to get in, I believe that we need to make a system where you can specify the position relative to other rules. That order also needs to be checked when running in check mode, and also updated if both rules are present but in the wrong order.

This could possibly be achieved by using `--comment` to name the different rules, and then specifying positions in a way similar to this:

``` yml
- iptables: chain=INPUT source=8.8.8.8 jump=DROP name=first
  become: yes

- iptables: chain=INPUT source=8.8.4.4 jump=DROP name=second after=first
  become: yes

- iptables: chain=INPUT source=4.4.4.4 jump=DROP after=second
  become: yes
```

vansickle:
@LinusU I agree that "position" isn't "idiomatic" way, it rather describes what to do than state. But it's simple and very useful in the following basic scenario:  when you, for example, have "ACCEPT" rules for specific IP, but need to "DROP" another. 

iptables: chain=INPUT source=8.8.8.8 jump=ACCEPT position=prepend
iptables: chain=INPUT source=8.8.4.4 jump=ACCEPT position=prepend
iptables: chain=INPUT jump=DROP position=append

In these case ACCEPT tasks are prepended to chain and DROP - appended. And it doesn't matter in which order you run tasks or the current state of the system - you'll always receive proper state.

vansickle:
OK, found that #1180 does the same and even marked as 'shipit' but never merged.

LinusU:
Hmm, I wonder why it never shipped, I would be okay with shipping that implementation...

vansickle:
Close this pull request, hope you merge #1180

PR #2135, opened 2016-04-29
Feature Pull Request: New shrink lvol option

Yannig:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

lvol
##### ANSIBLE VERSION

```
ansible 2.2.0 (devel 1d6608e84f) last updated 2016/04/29 23:18:16 (GMT +200)
  lib/ansible/modules/core: (detached HEAD e78ee3b128) last updated 2016/04/29 23:19:49 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD ca310f3d15) last updated 2016/04/29 23:19:49 (GMT +200)
  config file = /etc/ansible/ansible.cfg
  configured module search path = Default w/o overrides
```
##### SUMMARY

Allow lvol to not shrink logical volume if size is lower than the defined size.

This allow to ignore error of oversized logical volume when customizing your FS size outside of Ansible.

Before :

``` yaml
    - lvol: lv=test-lv size=128m vg=test
    - lvol: lv=test-lv size=32m  vg=test
```

```
TASK [lvol] ********************************************************************
changed: [debian]

TASK [lvol] ********************************************************************
fatal: [debian]: FAILED! => {"changed": false, "failed": true, "msg": "Sorry, no shrinking of test-lv without force=yes."}
```

After:

``` yaml
    - lvol: lv=test-lv size=128m vg=test
    - lvol: lv=test-lv size=32m  vg=test shrink=no
```

```
TASK [lvol] ********************************************************************
changed: [debian]

TASK [lvol] ********************************************************************
ok: [debian]
```

Playbook example:

``` yaml

---

- name: "Test PV/LV"
  hosts: debian
  gather_facts: no
  tasks:
    - lvol: lv=test-lv vg=test state=absent force=yes
    - lvg: vg=test pvs=/dev/sdb
    - lvol: lv=test-lv size=128m vg=test
    - lvol: lv=test-lv size=32m  vg=test
      ignore_errors: yes
    - lvol: lv=test-lv size=32m  vg=test shrink=no
```

PR #2220, opened 2016-05-12
lvol: Fix check-mode incorrectly returning changed

dagwieers:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

lvol
##### SUMMARY

The lvol module has a different logic in check-mode for knowing when a change is induced. And this logic is _only_ based on a size check. However during a normal run, it is the `lvreduce` or `lvextend` tool that decides when a change is performed (or when the requested and existing sizes differ using different criteria). So while in check-mode the module reports a change, in real run-mode it does not in fact changes anything and instead reports ok.

One solution would be to implement the exact size-comparison logic that is implemented in `lvextend` and `lvreduce`, but we opted to use the `--test` option to each command to verify if a change is induced or not. In effect both check-mode and run-mode use the exact same logic and conclusion.

### DIFF
--- a/system/lvol.py
+++ b/system/lvol.py
@@ -214,6 +214,12 @@ def main():
     if opts is None:
         opts = ""
 
+    # Add --test option when running in check-mode
+    if module.check_mode:
+        test_opt = ' --test'
+    else:
+        test_opt = ''
+
     if size:
         # LVCREATE(8) -l --extents option with percentage
         if '%' in size:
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @dagwieers. To the current maintainers, @jhoekx, @abulimov please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

jhoekx:
Is this also applicable to the percentage based resizing code? If yes, could you also make the change there to keep things consistent?

dagwieers:
@jhoekx Obviously, I first wanted to make sure this was the right approach and acceptable for inclusion.

dagwieers:
@jhoekx Adapted to include all `lvcreate`, `lvreduce`, `lvextend` and `lvremove` calls. Also rebased with change from #2135.

jhoekx:
Seems to work fine for me :-)

jhoekx:
shipit

gregdek:
Thanks again to @dagwieers for this PR, and thanks @jhoekx, @abulimov for reviewing. Marking for inclusion.

PR #2235, opened 2016-05-14
Add support for checking module on python3, like on core

mscherer:


gregdek:

abadger:
My instinct on these is if it presently compiles with python3 and there's
no reason it shouldn't use python3  (a library dep that doesn't run with
python3) then we should apply any requests like these to make sure the code
continues to compile. In the future.

PR #2253, opened 2016-05-18
Fix ValueErrors if extra output picked up from rabbitmqctl

dwcramer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

rabbitmq_vhost and rabbitmq_user
##### ANSIBLE VERSION

```
ansible 2.0.1.0
  config file = /Users/davcrame/.ansible.cfg
```
##### SUMMARY

Is the shell is configured to output extra information about the last login etc like so, these modules parse these lines as if they were part of the output of rabbitmqctl. This results in ValueErrors as the modules attempt to parts this output:

```
Last login: Wed May 18 14:54:11 EDT 2016 from 10.152.245.63 on pts/0
Welcome to Ubuntu 14.04.4 LTS (GNU/Linux 3.16.0-70-generic x86_64)
```

This fix protects against the error and is identical to how this check is done in other parts of these modules. See https://github.com/ansible/ansible-modules-extras/blob/devel/messaging/rabbitmq_user.py#L157 for example.

### DIFF
--- a/messaging/rabbitmq_user.py
+++ b/messaging/rabbitmq_user.py
@@ -178,6 +178,8 @@ class RabbitMqUser(object):
 
         perms_list = list()
         for perm in perms_out:
+            if '\t' not in perm:
+                continue
             vhost, configure_priv, write_priv, read_priv = perm.split('\t')
             if not self.bulk_permissions:
                 if vhost == self.permissions[0]['vhost']:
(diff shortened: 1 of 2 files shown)


gregdek:
Thanks @dwcramer. To the current maintainers, @manuel-sousa, @chrishoffman, @romanek-adam please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

romanek-adam:
I don't know the internals of Ansible, but shouldn't modules parse output of commands they issue themselves?

If all modules were about to skip output like last login info then it would complicate their implementation unnecessarily.

The question is: is this the right place to fix this issue?

PR #2254, opened 2016-05-18
Move running py3 compile test from whitelist to blacklist

abadger:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

Testing of modules
##### ANSIBLE VERSION

```
2.2
```
##### SUMMARY

At today's new module meeting we decided that new modules should be syntax compatible with python3 (not necessarily run correctly under python3 but able to be parsed correctly by the interpreter). It was decided that a prerequisite of that was to switch testing of the syntax of modules from a whitelist to a blacklist so that all new modules would be automatically tested in travis.  This PR implements the blacklist approach and blacklists the modules which are currently failing.


gregdek:

PR #2287, opened 2016-05-24
Fix pkgng.py documentation. Remove useless imports. Remove some whitespaces.

jody-frankowski:


gregdek:
Thanks @jody-frankowski. To the current maintainers, @JoergFiedler please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

gregdek:
Thanks @jody-frankowski for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

packaging/os/pkgng.py
```diff
@@ -24,10 +24,10 @@
 DOCUMENTATION = '''
 ---
 module: pkgng
-short_description: Package manager for FreeBSD >= 9.0
+short_description: Package manager for FreeBSD >= 8.3 and >= 9.1
```
bcoca, line 27: doesn't >=8.3 also cover 9.1?

bcoca:
most changes seem to be just style changes, we don't normally accept these as they obscure attribution

packaging/os/pkgng.py
JoergFiedler, line 27: The new package manager is part of the official release since version 9.1. It might be possible that there was an option to use pkgng with the 8.3 release as well.

I would prefer to stick to the offical [man page](https://www.freebsd.org/cgi/man.cgi?query=pkg&apropos=0&sektion=0&manpath=FreeBSD+10.3-RELEASE+and+Ports&arch=default&format=html) which states:

> The pkg command first appeared in FreeBSD 9.1.

JoergFiedler:
I always welcome changes that make the code more consitent, which also include small improvements like removing unnecessary imports and white spaces. So, if there is no code style/best practice that this PR violates, I don't mind to ship those style changes with the updated documentation.

jody-frankowski:
@bcoca @JoergFiedler Indeed, as you can see [here](https://www.freebsd.org/releases/), FreeBSD maintains a "production release" and a "legacy release" at the same time. On the same page you can see that 8.3 was released after 9.0.
You can also check [here](http://ftp-archive.freebsd.org/pub/FreeBSD-Archive/old-releases/amd64/8.3-RELEASE/packages/Latest/) that pkg was in 8.3 (search for pkg.tbz).
I'd prefer to point this out in order to avoid user confusion.

The removal of useless imports aren't cosmetic changes and I could do a separated PR for them if you wish.

JoergFiedler:
@jody-frankowski … thanks for looking it up … I do know that 8.3 was released after 9.0 and I assumed there might have been a package for that, but didn't check the package repo to verify the assumtion …

IMHO you don't need to create another PR for the "style" changes … 

@bcoca: do you agree?

bcoca:
no, we don't take style changes, you CAN create a separate PR but it will also not be accepted.

As for the text, i would leave it as `>=8.3` (where it became optional) as that also covers 9 (where it became default).

The only change that I would take from this PR is removing the unused imports.

jody-frankowski:
@bcoca 9.0 doesn't support pkg. The 8.x branch got support for 'pkg' on the 8.3 release. The 9.x branch got support for 'pkg' only on the 9.1 release. I know it sounds confusing.
I guess I should close this PR then?

JoergFiedler:
It seems the PR rules for this projects are rather strict. So, @jody-frankowski please create another PR to remove the unused imports, if you want to contribute. This is a more detailed explanation from @bcoca  regarding code style changes:

> whitespace changes or PEPization changes we don't normally take as it obscures attribution, we really only take style changes on lines you are modifying anyways or when the original code has real maintainability problems

bcoca:
fixed via 8fac01a3cca168ced59152c04ee8bca16e67586f

jody-frankowski:
Thanks.

PR #2300, opened 2016-05-25
Add sat_content_views module

liquidat:
##### ISSUE TYPE
- New Module Pull Request
##### COMPONENT NAME

sat_content_views (in cloud/satellite menu)
##### SUMMARY

This is a module to manage Satellite, specifically the so called content views. It uses the same technical principals as the already existing module (currently PR) to manage activation keys.

From the commit changelog:

DONE:
- creation, update, deletion of content views
- promotion and removal of content view versions
- manages Red Hat Satellite and thus Katello REST API calls
- uses Ansible internal URL calls
- check_mode supported
- catches errors when parameters cannot be found
- no pep8 errors

TODO:
- all major parameters for create and update are covered


gregdek:
Thanks @liquidat for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

cloud/satellite/sat_content_views.py
```diff
@@ -0,0 +1,349 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+DOCUMENTATION = '''
+---
+module: satellite_content_views
```
tima: The docs say "satellite_content_views" but the examples use "sat_content_views". I suggest using the former in the spirit of human-meaningful naming.

liquidat:
@tima I was just not sure if it would be better to have short-named modules (gce-..., aws-...) or better readable ones. But if it fits better in, I can rename this module (and all corresponding ones) to "satellite_", would that be better?

tima:
I think it would be better. gce- and aws- are acronyms while sat- is not nor is is a (common) nickname for satellite. "Sat" is also a common word. So yeah... for all those reasons I think the module would be better served being named "satellite".

liquidat:
@tima I haven't thought about it that way, it makes totally sense. I will change the naming for both submitted modules and push a new version. Thanks for the feedback.

missnebun:
+1

liquidat:
@missnebun There are now modules included for generic Foreman and Katello management. For that reason I did not try to submit these modules upstream again. Would you see a large benefit in this module compared to the other modules?

PR #2324, opened 2016-05-27
Freebsd pkgng autoremove support

AMDmi3:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

pkgng module
##### ANSIBLE VERSION

```
ansible 2.0.2.0
  config file = /home/amdmi3/.ansible.cfg
```
##### SUMMARY

Add autoremove capability to pkgng module, e.g. an ability to automatically remove no longer used dependencies (most useful after package deinstallation).

Before:

```
% ansible myhost -m pkgng --sudo -v -a "name=p5-Geo-IP state=present
myhost | SUCCESS => {
    "changed": true, 
    "msg": "added 1 package(s)"
}
```

this actally installs two packages, p5-Geo-IP and it's dependency GeoIP.

```
% ansible ihor-1.amdmi3.ru -m pkgng --sudo -v -a "name=p5-Geo-IP state=absent"                
myhost | SUCCESS => {
    "changed": true, 
    "msg": "removed 1 package(s)"
}
```

GeoIP package is still installed, though it's no longer needed.

After:

```
% ansible myhost -m pkgng --sudo -v -a "name=p5-Geo-IP state=present autoremove=yes"
myhost | SUCCESS => {
    "changed": true, 
    "msg": "added 1 package(s), no package(s) to autoremove"
}
% ansible myhost -m pkgng --sudo -v -a "name=p5-Geo-IP state=absent autoremove=yes" 
myhost | SUCCESS => {
    "changed": true, 
    "msg": "removed 1 package(s), autoremoved 1 package(s)"
}
```

The leftover dependency is now removed.


packaging/os/pkgng.py
```diff
@@ -75,6 +75,11 @@
             - pkg will chroot in the specified environment
             - can not be used together with 'rootdir' option
         required: false
+    autoremove:
+        version_added: "2.2"
+        description:
+            - remove automatically installed packages which are no longer needed
```
bcoca, line 81: default?

bcoca:
looks good, needs minor doc update, but I still should note:
- several style changes, we don't normally take those but these are minor and I'm going to pass on that
- you are returning 'changed/autoremoved' packages in msg=, it might be better to shift this to use the diff facility in the future, but not required for this PR.

gregdek:
Thanks @AMDmi3. To the current maintainers, @JoergFiedler please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

packaging/os/pkgng.py
AMDmi3, line 81: Added

AMDmi3:
@bcoca, I plan more changes to this module, so could you please elaborate on both notes? I might try to improve it in other ways.

bcoca:
- whitespace changes or PEPization changes we don't normally take as it obscures attribution, we really only take style changes on lines you are modifying anyways or when the original code has real maintainability problems.
- modules now get a flag indicating if --diff was passed on the command line or not, this gives them the option to return a diff with what has changed, mostly this was used only by those that modified files (copy, template, etc) but we have expanded it to other modules, packaging seems a good candidate for this, the apt module has already been updated and the yum module will be soon.

AMDmi3:
Got it. Gotta think this over, though this would be pretty complex with pkgng. Even now the module reports incorrect number of installed packages, as it doesn't count dependencies. Would it be OK to add some more features without --diff support for now? (I plan to implement `clean` (cleaning up downloaded packages), `upgrade` (upgrading all packages to the latest versions, similar to what upgrade in apt does), and make specifying package name optional so only e.g `clean=yes` could be specified to do its work).

bcoca:
it was suggestion, not a mandate.

AMDmi3:
Ok, thank you!

PR #2347, opened 2016-05-31
Pass through YAML parsed object instead of string.

davidedmonds:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

kubernetes
##### ANSIBLE VERSION

```
ansible 2.1.0.0 (stable-2.1 2452bd2135) last updated 2016/05/31 10:33:57 (GMT +100)
  lib/ansible/modules/core: (detached HEAD 21ef499e04) last updated 2016/05/31 10:34:05 (GMT +100)
  lib/ansible/modules/extras: (fix-k8s-module c6fdce8764) last updated 2016/05/31 12:13:06 (GMT +100)
  config file = /Users/david.edmonds/code/unicorn-ops/ansible.cfg
```
##### SUMMARY

Before this change, running the following play:

```
- name: Create app namespace
  kubernetes:
    api_endpoint: localhost:8080
    insecure: true
    inline_data:
      apiVersion: v1
      kind: Namespace
      metadata:
        name: app
        labels:
          name: app
    state: present
```

failed with the following:

```
fatal: [***]: FAILED! => {"changed": false, "failed": true, "module_stderr": "", "module_stdout": "Traceback (most recent call last):\r\n  File \"/tmp/ansible_VW8Mqv/ansible_module_kubernetes.py\", line 392, in <module>\r\n    main()\r\n  File \"/tmp/ansible_VW8Mqv/ansible_module_kubernetes.py\", line 361, in main\r\n    namespace = item.get('metadata', {}).get('namespace', \"default\")\r\nAttributeError: 'str' object has no attribute 'get'\r\n", "msg": "MODULE FAILURE", "parsed": false}
```

By parsing the item string as YAML, and passing it through to the subsequent functions, we are able to successfully create the namespace:

```
TASK [Create app namespace] ****************************************************
changed: [***]
```

First contribution, be gentle :D


gregdek:

metal3d:
It seems this is important... I've got this error and cannot make use of the module.
It worked with ansible < 2, now it's really an important problem

Can you please merge ?

metal3d:
Note: I just tested with ansible 2.1.0.0 and it woks like a charm

clustering/kubernetes.py
```diff
@@ -358,24 +358,25 @@ def main():
...
             except KeyError:
-                module.fail_json("invalid resource kind specified in the data: '%s'" % kind)
+                module.fail_json(msg="invalid resource kind specified in the data: '%s'" % kind)
             url = url.replace("{namespace}", namespace)
         else:
             url = target_endpoint
 
         if state == 'present':
-            item_changed, item_body = k8s_create_resource(module, url, item)
+            item_changed, item_body = k8s_create_resource(module, url, parsed_item)
```
metal3d: reference before assignement here, and in other following lines -> parsed_item

gregdek:
Thanks @davidedmonds for this PR. Unfortunately, it is not mergeable in its current state due to merge conflicts. Please rebase your PR. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

davidedmonds:
ready_for_review

gregdek:
Thanks @davidedmonds for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

davidedmonds:
Wonder if @erjohnso would be interested, as the original author (according to the docs anyway)?

supertom:
LGTM.

clustering/kubernetes.py
```diff
@@ -357,25 +357,27 @@ def main():
 
     for item in data:
         namespace = "default"
+        parsed_item = {}
         if item and 'metadata' in item:
-            namespace = item.get('metadata', {}).get('namespace', "default")
-            kind = item.get('kind', '').lower()
+            parsed_item = yaml.load(item)
```
detiber: Since above there is an explicit yaml.load_all when file_reference is used, I think it would be better to only attempt to yaml.load the info when inline_data is specified..., so I would push this up under `if inline_data:` rather than applying this here.

The other thing that I would suggest is to only attempt to do a yaml.load of the data `if not isinstance(dict, inline_data) and not isinstance(list, inline_data):`, so I would do suggest something like this around line 332:

```
if inline_data:
    if not isinstance(dict, inline_data) and not isinstance(list, inline_data):
        data = yaml.load(inline_data)
    else:
        data = inline_data
```

detiber:
@erjohnso @supertom @ryansb: one semi-unrelated question I have. Should file_reference also accept json formatted documents (or does yaml.load_all just do the right thing in that case)? It seems a bit odd to exclude json formatted files since that is the most common format I've seen used for upstream examples.

davidedmonds:
@detiber Updated as requested.

detiber:
:+1: from me

ryansb:
@detiber JSON is a subset of YAML, so all valid JSON will be parsed just by the YAML loader.

ryansb:
Also, :+1: from me

gregdek:
Thanks again to @davidedmonds for this PR, and thanks @ansible/core for reviewing. Marking for inclusion.

clustering/kubernetes.py
```diff
@@ -330,7 +330,10 @@ def main():
     file_reference = module.params.get('file_reference')
 
     if inline_data:
-        data = inline_data
+        if not isinstance(inline_data, dict) and not isinstance(inline_data, list):
+            data = yaml.load(inline_data)
+        else:
+            data = inline_data
```
bcoca, line 336: this is done automatically if you set inline_date to type='json'

detiber, line 336: @bcoca I see a jsonarg type, but that appears to expect to keep something as a json'ified string rather than convert from json to a python data structure, and it also doesn't appear to handle yaml. Am I missing something?

davidedmonds:
Wondering on the state of this - its labeled as both "pending_action" and "shipit" which I'd assumed would be mutually exclusive. Happy to make further changes if they're needed, just would like to know what needs to be changed :)

detiber:
@davidedmonds it looks like the pending_action is related the question of maintanership.

@gregdek any updates in that area?

resmo:
Thanks! Cherry-picking back for 2.1.1

davidedmonds:
Awesome, thanks all!

PR #2404, opened 2016-06-10
Fix win_environment: strict-mode fix for 'value'

ilpianista:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

win_environment
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  config file =
```
##### SUMMARY

After updating to 2.1.0.0 from 2.0.0.2 I got the following error while executing the task `win_environment: name=NOPAUSE level=user state=absent`:

```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: + ~~~~~~~~~~~~~~~~~~~~~~
fatal: [myhost]: FAILED! => {"changed": false, "failed": true, "msg": "The property 'value' cannot be found on this object. Verify that the property exists."
```

I fixed that with the attached patch.


gregdek:
Thanks @ilpianista. To the current maintainers, @jhawkesworth please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

jhawkesworth:
@nitzmahone - I know this is my module but could you take a look at this?  Something strange seems to be going on.  In 2.0.0.2 (installed from pip) this module had Set-StrictMode -Off 
However, that's been removed in latest devel.  running gitk on the file doesn't seem to show the Set-StrictMode in any of the changesets for this file.

If I recall Set-StrictMode off was set on a number of modules which hadn't yet been fixed up so my fear is that change has been lost against all the windows modules in 2.1 which would not be good.

jhawkesworth:
Just to update on this, it seems this and a few other modules have lost the (temporary) fix for strict-mode compliance.  So thanks for bringing this to my attention.

Also Get-Attr has been improved since this module was written, so there are a couple of other things that could be improved which I will note in line.

jhawkesworth:
Ok I'd forgotten you can only comment on the diffs.
L 23 would be best replaced with 

```
$state = Get-Attr -obj $params -name "state" -validateSet "present","absent" -default "present"
```

That would allow L 27 to 34 inclusive to be deleted.

You could also remove  L 36 to 41 inclusive by inserting

```
$name = Get-Attr -obj $params -name "name" -failifempty $true
```

I would probably leave L45 to 50 just because it handles upper or lower case names for the 'level' parameter.

Also, I'd prefer it if all the Get-Attr calls were grouped together before the rest of the code that validates the parameters and actually does the environment get / set logic.

@ilpianista - would you be ok making the above changes?

Many thanks,

Jon

needs_revision

gregdek:
Thanks @ilpianista for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

ilpianista:
@jhawkesworth thank you for the hints, done!

ready_for_review

gregdek:
Thanks @ilpianista. To the current maintainers, @jhawkesworth please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

jhawkesworth:
Thank you @ilpianista for making the changes.
There is a little thing that it would be good to get right - I should have explained more clearly.
I think we shouldn't use the following syntax any more:
$level = $params.level
$value = $params.value

but instead always use the 
$level = Get-Attr -obj $params -name "level" # further options here
syntax.

By using Get-Attr, we avoid the possibility of attempting to use a variable before it has been initialized (which is the biggest change that Strict-Mode enforces).

Please can you have a look at the options available for Get-Attr (it is documented in the
lib/ansible/module_utils/powershell.ps1 file) and switch over to the Get-Attr syntax for both 'level' and 'value' parameters.  I know I said above to leave 'level' as it is but I just mean to leave the current code that checks that it has a valid value - it still better to call Get-Attr to set it.

Once that is done I think the module is ready.

Many thanks once again for picking this up.

Jon

ilpianista:
Done.

ready_for_review

ilpianista:
Done.

ready_for_review

jhawkesworth:
Great, thank you.
I hadn't realized that -contains is not case sensitive so I've learned something from this too.
The code looks great now.  I will test it as soon as possible.

nitzmahone:
Tests ok for me- merging and cherry-picking to 2.1. Thanks!

jhawkesworth:
Thanks for the test and merge @nitzmahone

PR #2411, opened 2016-06-10
Add zypper refresh support in zypper module

AnderEnder:
##### ISSUE TYPE

- Feature Pull Request
##### COMPONENT NAME

packaging/os/zypper
##### ANSIBLE VERSION

```
ansible 2.2.0
```
##### SUMMARY

This PR adds the manual refresh repositories before operation. Can be run as part of the package installation or as a separate step.
Also there is a clean-up of  trialling symbols.

```
ok: [host] => {"changed": false, "cmd": ["/usr/bin/zypper", "--quiet", "--non-interactive", "--xmlout", "install", "--type", "package", "--auto-agree-with-licenses", "--no-recommends", "--", "openssh"], "invocation": {"module_args": {"disable_gpg_check": false, "disable_recommends": true, "force": false, "name": ["openssh"], "refresh": true, "state": "latest", "type": "package"}, "module_name": "zypper"}, "name": ["openssh"], "rc": 0, "refresh": true, "state": "latest"}
```

### DIFF
--- a/packaging/os/zypper.py
+++ b/packaging/os/zypper.py
@@ -36,6 +36,7 @@ author:
     - "Alexander Gubin (@alxgu)"
     - "Thomas O'Donnell (@andytom)"
     - "Robin Roth (@robinro)"
+    - "Andrii Radyk (@AnderEnder)"
 version_added: "1.2"
 short_description: Manage packages on SUSE and openSUSE
 description:
@@ -72,7 +73,7 @@ options:
     disable_recommends:
         version_added: "1.8"
         description:
-          - Corresponds to the C(--no-recommends) option for I(zypper). Default behavior (C(yes)) modifies zypper's default behavior; C(no) does install recommended packages. 
+          - Corresponds to the C(--no-recommends) option for I(zypper). Default behavior (C(yes)) modifies zypper's default behavior; C(no) does install recommended packages.
         required: false
         default: "yes"
         choices: [ "yes", "no" ]
@@ -83,9 +84,17 @@ options:
         required: false
         default: "no"
         choices: [ "yes", "no" ]
+    refresh:
+        version_added: "2.2"
+        description:
+          - Run the equivalent of C(zypper refresh) before the operation. Can be run as part of the package installation or as a separate step.
+        required: false
+        default: "no"
+        choices: [ "yes", "no" ]
+
 
 # informational: requirements for nodes
-requirements: 
+requirements:
     - "zypper >= 1.0  # included in openSuSE >= 11.1 or SuSE Linux Enterprise Server/Desktop >= 11.0"
     - rpm
 '''
@@ -114,6 +123,12 @@ EXAMPLES = '''
 
 # Apply all available patches
 - zypper: name=* state=latest type=patch
+
+# Refresh repositories and update package "openssl"
+- zypper: name=openssl state=present refresh=yes
+
+# Refresh repositories only
+- zypper: refresh=yes
 '''
 
 
@@ -160,7 +175,7 @@ def parse_zypper_xml(m, cmd, fail_not_found=True, packages=None):
         # zypper exit codes
         # 0: success
         # 106: signature verification failed
-        # 103: zypper was upgraded, run same command again 
+        # 103: zypper was upgraded, run same command again
         if packages is None:
             firstrun = True
             packages = {}
@@ -185,14 +200,15 @@ def parse_zypper_xml(m, cmd, fail_not_found=True, packages=None):
 def get_cmd(m, subcommand):
     "puts together the basic zypper command arguments with those passed to the module"
     is_install = subcommand in ['install', 'update', 'patch']
+    is_refresh = subcommand == 'refresh'
     cmd = ['/usr/bin/zypper', '--quiet', '--non-interactive', '--xmlout']
 
     # add global options before zypper command
-    if is_install and m.params['disable_gpg_check']:
+    if (is_install or is_refresh) and m.params['disable_gpg_check']:
         cmd.append('--no-gpg-checks')
 
     cmd.append(subcommand)
-    if subcommand != 'patch':
+    if subcommand != 'patch' and not is_refresh:
         cmd.extend(['--type', m.params['type']])
     if m.check_mode and subcommand != 'search':
         cmd.append('--dry-run')
@@ -325,24 +341,56 @@ def package_absent(m, name):
 
     return retvals
 
+
+def repo_refresh(m):
+    "update the repositories"
+    retvals = {'rc': 0, 'stdout': '', 'stderr': '', 'changed': False, 'failed': False}
+
+    cmd = get_cmd(m, 'refresh')
+
+    retvals['cmd'] = cmd
+    result, retvals['rc'], retvals['stdout'], retvals['stderr'] = parse_zypper_xml(m, cmd)
+    if retvals['rc'] == 0:
+        # refresh was successed
+        retvals['changed'] = True
+    else:
+        retvals['failed'] = True
+
+    return retvals
+
 # ===========================================
 # Main control flow
 
 def main():
     module = AnsibleModule(
         argument_spec = dict(
-            name = dict(required=True, aliases=['pkg'], type='list'),
+            name = dict(required=False, aliases=['pkg'], type='list'),
             state = dict(required=False, default='present', choices=['absent', 'installed', 'latest', 'present', 'removed']),
             type = dict(required=False, default='package', choices=['package', 'patch', 'pattern', 'product', 'srcpackage', 'application']),
             disable_gpg_check = dict(required=False, default='no', type='bool'),
             disable_recommends = dict(required=False, default='yes', type='bool'),
             force = dict(required=False, default='no', type='bool'),
+            refresh = dict(required=False, default='no', type='bool'),
         ),
+        required_one_of = [['name', 'refresh']],
         supports_check_mode = True
     )
 
     name = module.params['name']
     state = module.params['state']
+    refresh  = module.params['refresh']
+
+    # Refresh repositories
+    if refresh:
+        retvals = repo_refresh(module)
+
+        failed = retvals['failed']
+        del retvals['failed']
+        if failed:
+            module.fail_json(msg="Zypper refresh run failed.", **retvals)
+
+        if not name:
+            module.exit_json(name=name, state=state, refresh=refresh, **retvals)
 
     # Perform requested action
     if name == ['*'] and state == 'latest':
@@ -366,7 +414,7 @@ def main():
         del retvals['stdout']
         del retvals['stderr']
 
-    module.exit_json(name=name, state=state, **retvals)
+    module.exit_json(name=name, state=state, refresh=refresh, **retvals)
 
 # import module snippets
 from ansible.module_utils.basic import *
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @AnderEnder. To the current maintainers, @robinro please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

robinro:
Thanks @AnderEnder for your contribution to the zypper module!

I'm not (yet) convinced this is the best way to introduce manual refresh.
In my view `refresh` is an action on repositories and not packages and should thus happen in `zypper_repository`. There is currently an open PR (https://github.com/ansible/ansible-modules-extras/pull/2132) to autoimport keys for repos when they are added which does refresh. Maybe one can add an explicit option there? 

One of the aspects that show that refresh doesn't quite fit here is that you make `name` optional, which does not make sense if the core tasks is to **install/remove** `name`. If one were to add refresh in zypper I would prefer `name=*` over it being optional.

On a practical note: I basically never do a manual refresh and rely on the automatic ones. What is your use case for manual refreshs? Do we maybe need the force option then?

robinro:
needs_revision

AnderEnder:
> In my view refresh is an action on repositories and not packages and should thus happen in zypper_repository. 

Actually repository can be setup without auto-refresh and this is a normal behaviour, so if you need to update some packages to latest version, all repositories should be updated manually, but this is still package related logic.

> There is currently an open PR (#2132) to autoimport keys for repos when they are added which does refresh.
> It looks like a side-effect, but not a good self-described action.
> 
> On a practical note: I basically never do a manual refresh and rely on the automatic ones. What is your use case for manual refreshs? Do we maybe need the force option then?
- As I mentioned before auto-resfresh for repositories can be disabled, but should be run manually.
- In CI environments remote repositories are updated very often, so we can not rely on auto-refresh, so we always force zypper refresh using a command module, which is not very good solution. It is a very good solution to have everything in one module and one action. 
- Also there is a similar approach in other ansible modules like apt.

gregdek:
Thanks @AnderEnder for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

packaging/os/zypper.py
```diff
@@ -325,24 +341,56 @@ def package_absent(m, name):
...
+    "update the repositories"
+    retvals = {'rc': 0, 'stdout': '', 'stderr': '', 'changed': False, 'failed': False}
+
+    cmd = get_cmd(m, 'refresh')
+
+    retvals['cmd'] = cmd
+    result, retvals['rc'], retvals['stdout'], retvals['stderr'] = parse_zypper_xml(m, cmd)
+    if retvals['rc'] == 0:
+        # refresh was successed
+        retvals['changed'] = True
```
robinro: changed means "package was installed/removed", so it should not be set here

Actually the `retvals['failed']` construct is overkill and you can just check `retvals['rc']` in the main function.

packaging/os/zypper.py
```diff
@@ -325,24 +341,56 @@ def package_absent(m, name):
...
     state = module.params['state']
+    refresh  = module.params['refresh']
+
+    # Refresh repositories
+    if refresh:
+        retvals = repo_refresh(module)
+
+        failed = retvals['failed']
+        del retvals['failed']
+        if failed:
```
robinro: do `if retvals['rc'] != 0:` and leave out the whole `failed` business

robinro:
apt does not have auto-refresh for repos, so the necessity is different. Also if you argue with apt (or yum for that matter), the implementation should be the same, especially:
- still require name
- call the option `update_cache`
  With this I can probably agree on adding it.

Also it would be nice to have the option added to the integration tests. OpenSuSE Leap has recently been added to the list of OS to run the tests.

AnderEnder:
> apt does not have auto-refresh for repos, so the necessity is different. Also if you argue with apt (or yum for that matter), the implementation should be the same, especially:

The implementation is quite similar. It was completely inspired by apt module approach.

>  still require name

Apt module does not require name:
https://github.com/ansible/ansible-modules-core/blob/devel/packaging/os/apt.py#L689
https://github.com/ansible/ansible-modules-core/blob/devel/packaging/os/apt.py#L44

>   call the option update_cache With this I can probably agree on adding it.

IMHO, "refresh" or "update_cache" are OS specific terms and used differently in the documentation, e.g. "refresh" is self-described for zypper users, but cache for zypper usually used for "caching rpm files for specific repository". I propose to add update_cache as alias to have similar naming with other modules, but leave "refresh" as a main parameter name.

> Also it would be nice to have the option added to the integration tests. OpenSuSE Leap has recently been added to the list of OS to run the tests.

Do you mean this test: https://github.com/ansible/ansible/blob/devel/test/integration/roles/test_zypper/tasks/zypper.yml ? I will try to update it.

robinro:
yum requires name. You CI scenario also is fine with requiring name. Having the requirement makes the docs clearer about the purpose of the module. I prefer implementing actual usecases over the most general case. Many modules (also apt) are quite hard to maintain due to all the options that were added. Zypper is fairly clean at the moment.

Having update_cache as an alias for refresh sounds good, then people coming both from zypper and apt/yum will understand it.

@alxgu @agx Do you have an opinion on this PR? Especially what do you think about dropping the requirement for name?

robinro:
I meant the integration test which you linked.

packaging/os/zypper.py
```diff
@@ -325,24 +341,56 @@ def package_absent(m, name):
...
+def repo_refresh(m):
+    "update the repositories"
+    retvals = {'rc': 0, 'stdout': '', 'stderr': '', 'changed': False, 'failed': False}
+
+    cmd = get_cmd(m, 'refresh')
+
+    retvals['cmd'] = cmd
+    result, retvals['rc'], retvals['stdout'], retvals['stderr'] = parse_zypper_xml(m, cmd)
+    if retvals['rc'] == 0:
+        # refresh was successed
```
agx: s/successed/successful/ ?

packaging/os/zypper.py
agx: Can we somehow detect if zypper fetched new information? And only set changed = True in this case? This would prevent changed entries on every ansible run (otherwise making it less useful to catch config drift).

agx:
We should require name on package installations still, not having "always required" is fine IMHO.

alxgu:
I would vote for the current behaviour. `name` is requried

robinro:
Thanks @alxgu and @agx for your feedback. I would thus suggest to keep name required.

For your usecase `zypper: refresh=yes`, I would recommend to move this to `zypper_repository`. I'd be happy to support the addition there. One would then have to discussion the details there. I would prefer something along `zypper_repository: name=* refresh=yes`.

Regarding the naming `refresh/update_cache`: If you follow [ansibles guidelines](http://docs.ansible.com/ansible/developing_modules.html#deprecating-and-making-module-aliases) unified naming between modules is preferred over underlying API names (those should be aliased), so please make `refresh` the alias and `update_cache` the attribute name.

Thanks @AnderEnder for including the other comments. With those details this should be ready to get merged.

AnderEnder:
> Regarding the naming refresh/update_cache: If you follow ansibles guidelines unified naming between modules is preferred over underlying API names (those should be aliased), so please make refresh the alias and update_cache the attribute name.

@robinro, what about internal api naming(function names)? Should it be also renamed?

robinro:
I wouldn't call functions within the module "API". You can't import/call them without ansible core, so the only official way to access them is via them ansible core library. Thus I don't care much about function names.

gregdek:
@AnderEnder A friendly reminder: this pull request has been marked as needing your action. If you still believe that this PR applies, and you intend to address the issues with this PR, just let us know in the PR itself and we will keep it open pending your changes. When you do address the issues, please respond with ready_for_review in your comment, so that we can notify the maintainer.

robinro:
I implemented `zypper_repository: name=* runrefresh=yes` and `zypper_repository: repo=http://myurl runrefresh=yes` in https://github.com/ansible/ansible-modules-extras/issues/2132. This should nicely extend the implementation here.

gregdek:
@AnderEnder Another friendly reminder: this pull request has been marked as needing your action. If you still believe that this PR applies, and you intend to address the issues with this PR, just let us know in the PR itself and we will keep it open. If you have addressed the issues and believe it's ready for review, please comment with the text "ready_for_review". If we don't hear from you within another 14 days, we will close this pull request.

robinro:
@AnderEnder 
Please
- exchange name and alias for `refresh`, `update_cache`
- leave name required
  Then this can be merged.

I'd be happy to get your feedback on #2132 as well, which implements the part without name in `zypper_repository`.

AnderEnder:
@robinro, sorry for the delay.
I have updated the pull requests

robinro:
Thanks @AnderEnder
shipit

mattclay:
@AnderEnder Tests on shippable failed because this PR branch was created before shippable was enabled. Please rebase your branch so tests can run.

gregdek:
Thanks again to @AnderEnder for this PR, and thanks @robinro for reviewing. Marking for inclusion.

packaging/os/zypper.py
```diff
@@ -114,6 +124,9 @@
 
 # Apply all available patches
 - zypper: name=* state=latest type=patch
+
+# Refresh repositories and update package "openssl"
+- zypper: name=openssl state=present update_cache=yes
```
robinro, line 129: If the comment says "update package" it should be `state=latest`

robinro:
While testing I once got 

```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: xml.parsers.expat.ExpatError: not well-formed (invalid token): line 4, column 56
myhost | FAILED! => {
    "changed": false, 
    "failed": true, 
    "module_stderr": "Traceback (most recent call last):\n  File \"/tmp/ansible_OsDqHA/ansible_module_zypper.py\", line 408, in <module>\n    main()\n  File \"/tmp/ansible_OsDqHA/ansible_module_zypper.py\", line 377, in main\n    retvals = repo_refresh(module)\n  File \"/tmp/ansible_OsDqHA/ansible_module_zypper.py\", line 350, in repo_refresh\n    result, retvals['rc'], retvals['stdout'], retvals['stderr'] = parse_zypper_xml(m, cmd)\n  File \"/tmp/ansible_OsDqHA/ansible_module_zypper.py\", line 164, in parse_zypper_xml\n    dom = parseXML(stdout)\n  File \"/usr/lib64/python2.7/xml/dom/minidom.py\", line 1928, in parseString\n    return expatbuilder.parseString(string)\n  File \"/usr/lib64/python2.7/xml/dom/expatbuilder.py\", line 940, in parseString\n    return builder.parseString(string)\n  File \"/usr/lib64/python2.7/xml/dom/expatbuilder.py\", line 223, in parseString\n    parser.Parse(string, True)\nxml.parsers.expat.ExpatError: not well-formed (invalid token): line 4, column 56\n", 
    "module_stdout": "", 
    "msg": "MODULE FAILURE", 
    "parsed": false
}
```

I couldn't reproduce it so I assume it was some temporary zypper issue, but we should keep an eye on it.

If the refresh fails we fail quite badly in parse_xml, but we don't really care about the refresh xml output, only the rc being 0. So one might change that in the future.

resmo:
Thanks!

PR #2438, opened 2016-06-16
Fixes check mode error on Python 2.4 and wrong changed state

kustodian:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

postgresql_ext
##### ANSIBLE VERSION

```
ansible 2.1.0.0
```
##### SUMMARY

`postgresql_ext` in check mode always returned an error when it was executed on Python 2.4. Also in check mode the `changed` return value was wrong. This simple pull request fixes both issues.

### DIFF
--- a/database/postgresql/postgresql_ext.py
+++ b/database/postgresql/postgresql_ext.py
@@ -164,9 +164,9 @@ def main():
 
     try:
         if module.check_mode:
-            if state == "absent":
+            if state == "present":
                 changed = not ext_exists(cursor, ext)
-            elif state == "present":
+            elif state == "absent":
                 changed = ext_exists(cursor, ext)
             module.exit_json(changed=changed,ext=ext)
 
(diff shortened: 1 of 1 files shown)


database/postgresql/postgresql_ext.py
```diff
@@ -175,6 +175,8 @@ def main():
 
         elif state == "present":
             changed = ext_create(cursor, ext)
+    except SystemExit, e:
+        raise
```
bcoca: I'm not sure what you are trying to do here, it seems like a noop and also we try to avoid returning exceptions from modules

gregdek:
Thanks @kustodian. To the current maintainers, @dschep please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

PR #2521, opened 2016-07-05
adding jenkins_job module

sermilrod:
##### ISSUE TYPE

New Module Pull Request
##### COMPONENT NAME

```
jenkins_job
```
##### ANSIBLE VERSION

For Ansible 2.2
##### SUMMARY

This PR is adding a new module to manage Jenkins jobs configuration. This module allows you to create, delete, disable and reconfigure jenkins jobs.

### DIFF
--- a/web_infrastructure/jenkins_job.py
+++ b/web_infrastructure/jenkins_job.py
@@ -328,12 +328,12 @@ def jenkins_builder(module):
 def main():
     module = AnsibleModule(
         argument_spec = dict(
-            config_file = dict(required=True),
+            config_file = dict(required=True, type='path'),
             name        = dict(required=True),
-            password    = dict(required=False),
-            port        = dict(required=False, default=80),
+            password    = dict(required=False, no_log=True),
+            port        = dict(required=False, default=80, type='int'),
             state       = dict(required=False, default='present', choices=['present', 'absent', 'disabled']),
-            token       = dict(required=False),
+            token       = dict(required=False, no_log=True),
             url         = dict(required=True),
             user        = dict(required=False)
         ),
--- a/web_infrastructure/jenkins_job.py
+++ b/web_infrastructure/jenkins_job.py
@@ -224,20 +224,22 @@ class Jenkins:
     def reconfig_job(self, module):
         changed = False
         if self.configuration_changed():
-            try:
-                self.server.reconfig_job(self.name, xml_to_string(self.config_file))
-                changed = True
-            except Exception:
-                e = get_exception()
-                module.fail_json(msg='Unable to reconfigure job, %s for %s' % (str(e), self.jenkins_url))
+            changed = True
+            if not module.check_mode:
+                try:
+                    self.server.reconfig_job(self.name, xml_to_string(self.config_file))
+                except Exception:
+                    e = get_exception()
+                    module.fail_json(msg='Unable to reconfigure job, %s for %s' % (str(e), self.jenkins_url))
 
         module.exit_json(changed=changed, name=self.name, state=self.state, url=self.jenkins_url)
 
     def create_job(self, module):
         changed = False
         try:
-            self.server.create_job(self.name, xml_to_string(self.config_file))
             changed = True
+            if not module.check_mode:
+                self.server.create_job(self.name, xml_to_string(self.config_file))
         except Exception:
             e = get_exception()
             module.fail_json(msg='Unable to create job, %s for %s' % (str(e), self.jenkins_url))
@@ -247,12 +249,13 @@ class Jenkins:
     def delete_job(self, module):
         changed = False
         if self.job_exists(module):
-            try:
-                self.server.delete_job(self.name)
-                changed = True
-            except Exception:
-                e = get_exception()
-                module.fail_json(msg='Unable to delete job, %s for %s' % (str(e), self.jenkins_url))
+            changed = True
+            if not module.check_mode:
+                try:
+                    self.server.delete_job(self.name)
+                except Exception:
+                    e = get_exception()
+                    module.fail_json(msg='Unable to delete job, %s for %s' % (str(e), self.jenkins_url))
 
         module.exit_json(changed=changed, name=self.name, state=self.state, url=self.jenkins_url)
 
@@ -262,8 +265,9 @@ class Jenkins:
             status = self.get_job_status(module)
             try:
                 if status != "disabled":
-                    self.server.disable_job(self.name)
                     changed = True
+                    if not module.check_mode:
+                        self.server.disable_job(self.name)
             except Exception:
                 e = get_exception()
                 module.fail_json(msg='Unable to disable job, %s for %s' % (str(e), self.jenkins_url))
@@ -316,7 +320,7 @@ def main():
         ),
         required_one_of = [['password', 'token']],
         mutually_exclusive = [['password', 'token']],
-        supports_check_mode=False,
+        supports_check_mode=True,
     )
 
     test_dependencies(module)
--- a/web_infrastructure/jenkins_job.py
+++ b/web_infrastructure/jenkins_job.py
@@ -28,7 +28,11 @@ options:
   config_file:
     description:
       - Absolute path to the config.xml for an specific job.
-    required: true
+    required: false
+  config:
+    description:
+      - config.xml file to use as job config within your Ansible repo.
+    required: false
   name:
     description:
       - Name of the Jenkins job.
@@ -70,7 +74,7 @@ EXAMPLES = '''
 
 # Create a jenkins job using the token
 - jenkins_job:
-    config_file: /path/to/config.xml
+    config: "{{ lookup('template', 'templates/test.xml.j2') }}"
     name: test
     token: asdfasfasfasdfasdfadfasfasdfasdfc
     state: present
@@ -79,7 +83,7 @@ EXAMPLES = '''
 
 # Delete a jenkins job using basic authentication
 - jenkins_job:
-    config_file: /path/to/config.xml
+    config: "{{ lookup('template', 'templates/test.xml.j2') }}"
     name: test
     password: admin
     state: absent
@@ -97,7 +101,7 @@ EXAMPLES = '''
 
 # Disable a jenkins job using basic authentication
 - jenkins_job:
-    config_file: /path/to/config.xml
+    config: "{{ lookup('template', 'templates/test.xml.j2') }}"
     name: test
     password: admin
     state: disabled
@@ -115,14 +119,14 @@ EXAMPLES = '''
 
 # To re-enable a job just make it be present
 - jenkins_job:
-    config_file: /path/to/config.xml
+    config: "{{ lookup('template', 'templates/test.xml.j2') }}"
     name: test
     password: admin
     state: present
     url: "http://localhost:8080"
     user: admin
 
-# A production ready example
+# A production ready example using the config_file option
 - template:
     src: test-job.xml.j2
     dest: /tmp/test-job-config.xml
@@ -132,6 +136,15 @@ EXAMPLES = '''
 
 - jenkins_job:
     config_file: /tmp/test-job-config.xml
+    name: test-job
+    token: abcdefghijklmnopqrstuvwxyz
+    state: present
+    url: "https://jenkins.mydomain.com"
+    user: my_user
+
+# A production ready example using the config option
+- jenkins_job:
+    config: "{{ lookup('template', 'templates/test-job.xml.j2') }}"
     name: test-job
     token: abcdefghijklmnopqrstuvwxyz
     state: present
@@ -171,8 +184,9 @@ except ImportError:
     python_lxml_installed = False
 
 class Jenkins:
-    def __init__(self, config_file, name, password, state, token, url, user):
+    def __init__(self, config_file, config, name, password, state, token, url, user):
         self.config_file = config_file
+        self.config = config
         self.name = name
         self.password = password
         self.state = state
@@ -217,9 +231,15 @@ class Jenkins:
         else:
             self.disable_job(module)
 
+    def get_config(self):
+        if self.config_file:
+            return xml_to_string(self.config_file)
+        else:
+            return job_config_to_string(self.config)
+
     def configuration_changed(self):
         changed = False
-        config_file = xml_to_string(self.config_file)
+        config_file = self.get_config()
         machine_file = job_config_to_string(self.server.get_job_config(self.name).encode('utf-8'))
         if not machine_file == config_file:
             changed = True
@@ -238,7 +258,7 @@ class Jenkins:
             changed = True
             if not module.check_mode:
                 try:
-                    self.server.reconfig_job(self.name, xml_to_string(self.config_file))
+                    self.server.reconfig_job(self.name, self.get_config())
                 except Exception:
                     e = get_exception()
                     module.fail_json(msg='Unable to reconfigure job, %s for %s' % (str(e), self.jenkins_url))
@@ -250,7 +270,7 @@ class Jenkins:
         try:
             changed = True
             if not module.check_mode:
-                self.server.create_job(self.name, xml_to_string(self.config_file))
+                self.server.create_job(self.name, self.get_config())
         except Exception:
             e = get_exception()
             module.fail_json(msg='Unable to create job, %s for %s' % (str(e), self.jenkins_url))
@@ -303,6 +323,7 @@ def xml_to_string(source):
 def jenkins_builder(module):
     return Jenkins(
         module.params.get('config_file'),
+        module.params.get('config'),
         module.params.get('name'),
         module.params.get('password'),
         module.params.get('state'),
@@ -314,7 +335,8 @@ def jenkins_builder(module):
 def main():
     module = AnsibleModule(
         argument_spec = dict(
-            config_file = dict(required=True, type='path'),
+            config_file = dict(required=False, type='path'),
+            config      = dict(required=False),
             name        = dict(required=True),
             password    = dict(required=False, no_log=True),
             state       = dict(required=False, default='present', choices=['present', 'absent', 'disabled']),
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @sermilrod for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

web_infrastructure/jenkins_job.py
```diff
@@ -0,0 +1,349 @@
...
+                            module.params.get('url'),
+                            str(e)))
+
+def main():
+    module = AnsibleModule(
+        argument_spec = dict(
+            config_file = dict(required=True),
+            name        = dict(required=True),
+            password    = dict(required=False),
+            port        = dict(required=False, default=80),
```
resmo: please add `type='int'`

web_infrastructure/jenkins_job.py
```diff
@@ -0,0 +1,349 @@
...
+                            module.params.get('state'),
+                            module.params.get('url'),
+                            str(e)))
+
+def main():
+    module = AnsibleModule(
+        argument_spec = dict(
+            config_file = dict(required=True),
+            name        = dict(required=True),
+            password    = dict(required=False),
```
resmo: please add `no_log=True` for secrets.

web_infrastructure/jenkins_job.py
```diff
@@ -0,0 +1,349 @@
...
+        module.fail_json(msg='name, state, url are required: name=%s, state=%s, url=%s' % (
+                            module.params.get('name'),
+                            module.params.get('state'),
+                            module.params.get('url'),
+                            str(e)))
+
+def main():
+    module = AnsibleModule(
+        argument_spec = dict(
+            config_file = dict(required=True),
```
resmo: please add `type='path'`

sermilrod:
Hi, I fixed the things suggested but I get an strange error in shippable tests:

python2.4 -m compileall -fq -x 'cloud/|monitoring/zabbix._.py|/dnf.py|/layman.py|/maven_artifact.py|clustering/(consul._|znode).py|notification/pushbullet.py|database/influxdb/influxdb._.py|database/mssql/mssql_db.py|/letsencrypt.py|network/f5/bigip._.py' . 0s  
/root/634078eb-c9c4-48ae-9bde-fc41ad8c28f6.sh: line 58: python2.4: command not found

Whereas the same validation in travis is passing:

The command "python2.4 -m compileall -fq -x 'cloud/|monitoring/zabbix._.py|/dnf.py|/layman.py|/maven_artifact.py|clustering/(consul._|znode).py|notification/pushbullet.py|database/influxdb/influxdb._.py|database/mssql/mssql_db.py|/letsencrypt.py|network/f5/bigip._.py' ." exited with 0.

It seems like in the container python 2.4 is missing but I am not sure if that is a result due to an issue of mine or it is something wrong with the testing. Can somebody assist me? 

Thanks in advance.

gregdek:
Thanks @sermilrod for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

sermilrod:
ready_for_review I think the PR failed due to an issue with the new shippable testing:

/root/634078eb-c9c4-48ae-9bde-fc41ad8c28f6.sh: line 58: python2.4: command not found

mattclay:
Tests on shippable failed because this PR branch was created before shippable was enabled. If you rebase the branch the tests will pass, although in this case the tests will just no-op since there are no tests for this module.

resmo:
I looked at the url handling and IMHO it makes things unnecessary complicated, I don't see any benefit to split port out into a separate arg and also "guessing" the protocol by looking at the port. The current implementation does not allow to use https on a non standard port e.g. `8443`.

I would suggest to use one arg `url` and pass the full url, e.g. `http://localhost:8080` or `https://jenkins.example.com:8443` or just `https://jenkins.example.com`

Further I would suggest to default that url to `http://localhost:8080` as this would be the default port used by jenkins on a local machine.

example for a http://localhost:8080

``` yaml
- jenkins_job:
    config_file: /path/to/config.xml
    name: test
    password: admin
    user: admin
```

example for using a different url:

``` yaml
- jenkins_job:
    config_file: /path/to/config.xml
    name: test
    url: "https://jenkins.example.com:8443"
    password: admin
    user: admin
```

What do you think? Did I miss anything?

web_infrastructure/jenkins_job.py
```diff
@@ -0,0 +1,349 @@
...
+        argument_spec = dict(
+            config_file = dict(required=True, type='path'),
+            name        = dict(required=True),
+            password    = dict(required=False, no_log=True),
+            port        = dict(required=False, default=80, type='int'),
+            state       = dict(required=False, default='present', choices=['present', 'absent', 'disabled']),
+            token       = dict(required=False, no_log=True),
+            url         = dict(required=True),
+            user        = dict(required=False)
+        ),
```
resmo, line 311: Seems password and token are mutually exclusive, but one would be required right? You should exentend the arg_spec here with:

``` python
required_one_of = (
    ['password', 'token']
),
mutually_exclusive = (
    ['password', 'token']
),
```

sermilrod:
@resmo good catch, it makes perfectly sense to me. I will make both changes you suggested.
h
@mattclay It is strange, I tried to rebase but my branch is up to date with devel, excluding the new module of course. I will make the changes suggested and I hope this time it passes the tests.

sermilrod:
Hi, I performed the suggested changes suggested by @resmo but the CI seems to be stuck. And I still get the issue in Shippable /root/0a68573a-c8e4-4001-91a3-0798ba3cc97e.sh: line 58: python2.4: command not found

resmo:
@sermilrod 

Did you try to rebase?

```
git checkout devel
git pull git@github.com:ansible/ansible-modules-extras.git devel
git checkout Adding_jenkins_job_module
git rebase devel
git push -f origin Adding_jenkins_job_module
```

resmo:
Regarding check mode support, based on your work, it would be super easy to implement check mode support:
e.g. for state=absent the delete_job() would something like:

``` python
def delete_job(self, module):
    changed = False
    if self.job_exists(module):
        changed = True
        if not module.check_mode:
            try:
                self.server.delete_job(self.name)
            except Exception:
                e = get_exception()
                module.fail_json(msg='Unable to delete job, %s for %s' % (str(e), self.jenkins_url))

    module.exit_json(changed=changed, name=self.name, state=self.state, url=self.jenkins_url)
```

sermilrod:
@resmo I had a problem with my git client, but I can pull and rebase with no problems. And now that I am happy that both checks are passing I could easily implement the check mode in reconfig_job, create_job, delete_job and disable_job :)

sermilrod:
ready_for_review Now the check mode is in place and the CI looks good.

web_infrastructure/jenkins_job.py
```diff
@@ -0,0 +1,332 @@
...
+RETURN = '''
+'''
+
+import os
+
+try:
+    import jenkins
+    python_jenkins_installed = True
+except ImportError:
+    e = get_exception()
```
resmo: please remove this line, unused and since the import is later, it causes an error anyway.

web_infrastructure/jenkins_job.py
```diff
@@ -0,0 +1,332 @@
...
+    python_jenkins_installed = False
+
+try:
+    from lxml import etree as ET
+    python_lxml_installed = True
+except ImportError:
+    e = get_exception()
```
resmo: please remove this line, unused and since the import is later, it causes an error anyway.

web_infrastructure/jenkins_job.py
```diff
@@ -0,0 +1,332 @@
...
+    token: abcdefghijklmnopqrstuvwxyz
+    state: present
+    url: "https://jenkins.mydomain.com"
+    user: my_user
+'''
+
+RETURN = '''
+'''
+
+import os
```
resmo: this is unused, please remove it.

web_infrastructure/jenkins_job.py
```diff
@@ -0,0 +1,332 @@
...
+def main():
+    module = AnsibleModule(
+        argument_spec = dict(
+            config_file = dict(required=True, type='path'),
+            name        = dict(required=True),
+            password    = dict(required=False, no_log=True),
+            state       = dict(required=False, default='present', choices=['present', 'absent', 'disabled']),
+            token       = dict(required=False, no_log=True),
+            url         = dict(required=False, default="http://localhost:8080"),
+            user        = dict(required=False)
```
resmo, line 310: While testing it seems user is required, or we default it to `admin` alternatively

resmo:
I found a few tweaks but overall it worked fine. 

Concerning the `config_file` argument: I would prefer this not to be a path to a file. Instead I would prefer just the xml config itself (arg would be renamed to `config` of course).

This would allow to make use of lookup and template, e.g.

``` yaml
- jenkins_job:
    config: "{{ lookup('template', 'path/to/config.xml.j2') }}"
    name: test
    password: admin
    user: admin
```

or just a xml file as is:

``` yaml
- jenkins_job:
    config: "{{ lookup('file', 'path/to/config.xml') }}"
    name: test
    password: admin
    user: admin
```

I don't know if this is out of your scope, just wanted to let you know.

web_infrastructure/jenkins_job.py
sermilrod, line 310: You can have a Jenkins server with no authentication and because of that reason I don't think it should have default value either. Regarding the import conditions mentioned in previous comments, let me remove the unusued lines.

resmo, line 310: regarding auth, good point, agreed

sermilrod:
It is out of the scope but it looks to me like a very good idea. Let me fix the two things mentioned before and test the use case you propose to me.

resmo:
I would give my ok without the "config" feature, it could be implemented later btw.

web_infrastructure/jenkins_job.py
```diff
@@ -0,0 +1,332 @@
...
+        ),
+        required_one_of = [['password', 'token']],
```
resmo: Regarding auth again, if auth is "optional" then password or token must also not be given and  `required_one_of` can be removed. Sorry, my bad.

sermilrod:
Yeah it is not something I can change quickly, I would have to look into it. If it is ok with you I can work on that feature later.

web_infrastructure/jenkins_job.py
sermilrod: Oh so just letting them be mutually exclusive if they are provided right?

web_infrastructure/jenkins_job.py
```diff
@@ -0,0 +1,326 @@
...
+- jenkins_job:
+    config_file: /tmp/test-job-config.xml
+    name: test-job
+    token: abcdefghijklmnopqrstuvwxyz
+    state: present
+    url: "https://jenkins.mydomain.com"
+    user: my_user
+'''
+
+RETURN = '''
```
resmo, line 120: As the code returns results, these must be documented here (but do not document the common return `changed`:

```
module.exit_json(changed=changed, name=self.name, state=self.state, url=self.jenkins_url)
```

web_infrastructure/jenkins_job.py
```diff
@@ -0,0 +1,326 @@
...
+              "see http://lxml.de/installation.html")
+
+def job_config_to_string(xml_str):
+    return ET.tostring(ET.fromstring(xml_str))
+
+def xml_to_string(source):
+    return ET.tostring(ET.parse(source).getroot())
+
+def jenkins_builder(module):
+    if module.params.get('name') and module.params.get('config_file') and module.params.get('url'):
```
resmo: this condition is redundant, it is already defined in the arg_spec.

sermilrod:
Hi @resmo I fixed the two things you mentioned. I was not sure about how to put the information in the return block so I checked other modules. Let me know if you think it needs to be some other way.

resmo:
@sermilrod looks good. Thanks!

sermilrod:
@resmo I managed to make it work with lookups :) For generic purposes I would like to give the user the option to import a file in the remote machine if that is the case so I would like to maintain config and config_file options and let the user select the one that is more convenient. What do you think?

resmo:
@sermilrod makes totally sense. 

I tested it and found some more minor improvements:

self.configuration_changed() must be in the try block:

``` python
    def reconfig_job(self, module):
        changed = False
        try:
            if self.configuration_changed():
                changed = True
                if not module.check_mode:
                    self.server.reconfig_job(self.name, self.get_config())
        except Exception:
            e = get_exception()
            module.fail_json(msg='Unable to reconfigure job, %s for %s' % (str(e), self.jenkins_url))

        module.exit_json(changed=changed, name=self.name, state=self.state, url=self.jenkins_url)
```

Otherwise you get an ugly traceback if auth fails.

sermilrod:
Hi @resmo I fixed it and included the configuration_changed method in the try block

sermilrod:
I am going to be on holidays 10 days, please do not close this due to inactivity. I will be more than willing to fix any other issue that you can point when I am back or happy to hear you decided to ship it :). Thanks for the understanding.

resmo:
@sermilrod sure this won't be closed! Thanks so far for your effort. I'll see some improvements, I'll send you a PR against your branch for review and discussion.

sermilrod:
Sure @resmo no problem

sermilrod:
Hi, I am back. Any updates on this?

sermilrod:
Hi @resmo do you have any update on this? Is it ready or do I still need to work on it?

resmo:
Thanks for your patience @sermilrod, I didn't have time yet to sent your a PR but let me describe the thinks I would like to see. Basicially I don't like the "user interface" yet. E.g. if you want to delete a job, you have to provide the config, even the config is not necessary to delete the job, so I like it to be like:

``` yaml
# delete a job
- jenkins_job:
    name: test-job
    token: abcdefghijklmnopqrstuvwxyz
    state: absent
```

this can achieved using build in required_if of ansible module_utils.

Next is related to state=disabled. I would like to see both states, disabled and enabled. State=present should not automatically enable it should keep in whatever state the job was.

Optionally if the job is not present on state=disabled, as a user I would expect the module creates a disabled job. E.g.

``` yaml
# disable an exising job, if job is not present, this should fail with "missing required param config/config_file" 
# The same counts for state=enabled.
- jenkins_job:
    name: test-job
    token: abcdefghijklmnopqrstuvwxyz
    state: disabled
```

example job is present and enabled.

``` yaml
# ensure a job is present and enabled
- jenkins_job:
    name: test-job
    config_file: /tmp/test-job-config.xml  
    token: abcdefghijklmnopqrstuvwxyz
    state: enabled
```

example job is present, but do not force enabled.

``` yaml
# ensure a job is present whether it be disabled or enabled
- jenkins_job:
    name: test-job
    config_file: /tmp/test-job-config.xml  
    token: abcdefghijklmnopqrstuvwxyz
    state: present
```

Any thoughts?

sermilrod:
Hi @resmo it makes sense to me and it should be easy to implement. I can do it myself so if you are busy don't worry about the extra PR. Regarding the states logic just to be clear let's summarize:

```
STATE     CURRENT_JOB_STATUS   ACTION

present    does not exists                     create it enabled
disabled   does not exists                     create it disabled 
absent     does not exists                     do nothing
enabled   does not exists                     create it enabled

present    exists                                    do nothing but update config (preserves the disabled/enabled state)
disabled   exists                                    force state to disabled
absent     exists                                    delete job
enabled   exists                                    force state to enabled if not already there
```

Am I missing something else, or is this they way you would like the module to behave?

I will start working on this and any other suggestion you make :)

resmo:
The matrix looks good to me. Please let me know in case of any questions or tests.

sermilrod:
@resmo I am almost done but I have a question. How do you specify if a parameter matches a value, then require only one parameter (2 possible). Right now the parameter validation is like this:

```
required_one_of = [['config_file', 'config']],
mutually_exclusive = [['config_file', 'config'], ['password', 'token']],
```

I can remove the 1st validation because as we agreed config should be optional depending on the state. So the question is how to specify ansible to require either 'config_file' or 'config' if state matches the value we want. I have checked other modules and you can specify 'required_if' to ensure that more than 1 parameter is passed but not 1 of a list.

```
required_if = [
            ('state', 'present', ['config_file', 'config']),
            ('state', 'enabled', ['config_file', 'config'])
        ],
```

Something like above will require config_file and config and what I need is an OR condition.

bcoca:
instead of overloading `state` i would create an `enabled=yes|no` to allow for combining states

sermilrod:
Thanks, it would actually simplify the code, let me try that.

sermilrod:
Hi there! 

I think with the new changes the module will be flexible with the required parameters to perform enable/disable/create/delete. This way we do not have to pass the config if we are disabling a job, or we do not have to pass the enable attribute if we want to delete a job as an example.

I would like to point that to make this change possible and easy to maintain I chose to get rid of the old "config_file" option as we can use lookups to place files or templates by using the "config" option instead. There is no point on having 2 attributes to do the same. I hope this makes the module easier to use.

I also updated the module documentation accordingly with the new changes so it should be pretty straightforward to test it.

@resmo, @bcoca If you find something else to enhance please let me know :)

web_infrastructure/jenkins_job.py
```diff
@@ -0,0 +1,327 @@
...
+    )
+
+def main():
+    module = AnsibleModule(
+        argument_spec = dict(
+            config      = dict(required=False),
+            name        = dict(required=True),
+            password    = dict(required=False, no_log=True),
+            state       = dict(required=True,  choices=['present', 'absent']),
+            enable      = dict(required=False, type='bool', choices=[True, False, 'yes', 'no']),
```
resmo: no need to add choices list, just `type='bool'`

sermilrod: Fixed

resmo:
@sermilrod cool, getting rid of config_file makes sense to me. I will take another effort to thest the module the next days. Thanks for your works!

sermilrod:
@resmo Anytime :+1: :). Thanks also to you for the effort testing and thinking of improvements.

resmo:
LGTM

The only thing I would like to see changed is `state` should be defaulted to `present`. And probably `enable` defaulted to `True`. 

However this is not a reason to not merge it. Good job.

PR #2531, opened 2016-07-07
Add iptables_raw module

kustodian:
##### ISSUE TYPE
- New Module Pull Request
##### COMPONENT NAME

`iptables_raw`
##### ANSIBLE VERSION

```
ansible 2.1.0.0
```

But the module also works from 1.9+.
##### SUMMARY

This module makes it easy to manage `iptables`. The module has the following features:
- use iptables syntax to define rules
- keep iptables state
- order rules by weight
- support different iptables tables (filter, nat, raw,...)
- management of unmanaged rules
- diff mode (2.1+)
- ip6tables
- safe flushing of table rules

Full documentation of the module can be read in the module code, or [here](http://blog.nordeus.com/files/libraryblog/articles/managing-iptables-with-ansible/iptables_raw.html). I also wrote a [blog post](http://blog.nordeus.com/dev-ops/managing-iptables-with-ansible-the-easy-way.htm) explaining the usage of the module and why we decided to build it.

Here are a few examples how to use it:

``` yml
# Allow all IPv4 traffic coming in on port 80 (http)
- iptables_raw:
    name=allow_tcp_80
    rules='-A INPUT -p tcp -m tcp --dport 80 -j ACCEPT'

# Allow all IPv6 traffic coming in on port 443 (https) with weight 50
- iptables_raw:
    ipversion=6
    weight=50
    name=allow_tcp_443
    rules='-A INPUT -p tcp -m tcp --dport 443 -j ACCEPT'

# Remove the above rule
- iptables_raw:
    state=absent
    ipversion=6
    name=allow_tcp_443

# Reset all IPv4 iptables rules in all tables and allow all traffic
- iptables_raw:
    name=*
    table=*
    state=absent
```

On the implementation side this module keeps the state in a json file in `/etc/ansible-iptables` directory, since the file generated by `iptables-save` doesn't hold enough information to keep everything that is needed (weight of the rules, information which chains are made by this module, etc). The module only uses `iptables-save` and `iptables-restore` commands to save and restore rules.

Module has been tested on CentOS 5, 6 and 7 and the module saves the generated rules into `/etc/sysconfig/iptables` from which RHEL and it's derivatives read the rules on boot. The module should work without a problem on other Linux distributions as long as they are set to read the iptables rules on boot from `/etc/sysconfig/iptables`. Out of the box support for other distributions can be added into the module, by detecting the path of this file and saving to it.

When the module adds rules it automatically adds `ansible[name]` as a comment to a rule (if a rule has a comment `ansible[name]:` is prepended to that comment), so that the module can distinguish what rules were created with the module, or outside of it. It also allows the user to easily see which task created a specific rule. For example this task:

``` yml
- iptables_raw:
    name=allow_tcp_80
    rules='-A INPUT -p tcp -m tcp --dport 80 -j ACCEPT'
```

generates this rule:

```
-A INPUT -p tcp -m tcp --dport 80 -m comment --comment ansible[allow_tcp_80] -j ACCEPT
```

The module also implements a locking mechanism, so if you run the module on the same host at the same time, one instance will wait for the other to finish execution. This was done just in case so that state files don't get overwritten in a wrong order.

We also wrote test playbooks for this module, but I guess those need to be pushed to the ansible core repo?

### DIFF
--- a/system/iptables_raw.py
+++ b/system/iptables_raw.py
@@ -1,0 +1,871 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+"""
+(c) 2016, Strahinja Kustudic <strahinjak@nordeus.com>
+(c) 2016, Damir Markovic <damir@damirda.com>
+
+This file is part of Ansible
+
+Ansible is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+Ansible is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+"""
+
+DOCUMENTATION = '''
+---
+module: iptables_raw
+short_description: Manage iptables rules
+version_added: "2.1"
+description:
+  - Adds and removes iptables rules
+options:
+  backup:
+    description:
+      - Create a backup of /etc/sysconfig/iptables file before overwriting it.
+    required: false
+    choices: ["yes", "no"]
+    default: "no"
+  ipversion:
+    description:
+      - Target the IP version this rule is for.
+    required: false
+    default: "4"
+    choices: ["4", "6"]
+  keep_unmanaged:
+    description:
+      - If set to C(yes) keeps active iptables (unmanaged) rules for the target C(table) and gives them C(weight=90). This means these rules will be ordered after most of the rules,
+      - since default priority is 40, so they shouldn't be able to block any allow rules. If set to C(no) deletes all rules which are not set by this module.
+      - "WARNING: Be very careful when running C(keep_unmanaged=no) for the first time, since if you don't specify correct rules, you can block yourself out of the managed host."
+    required: false
+    choices: ["yes", "no"]
+    default: "yes"
+  name:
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @kustodian for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

selivan:
Good job. Unlike already included in 2.0 module `iptables`, it has persistence. And best thing - rule priorities. So you can add rules from different roles, without concerning of roles apply order - rules will always get where they are needed. Looking forward to including this module in ansible.

iskrant:
It would be helpful!

linuxdynasty:
@kustodian, great module. I sent you a PR that includes some minor changes, so that this module works on Debian based systems. The only requirement is iptables-persistent package, so that iptables rules can persist.

gregdek:
Thanks @kustodian for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

kustodian:
ready_for_review

gregdek:
Thanks @kustodian for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

matamouros:
Dear @kustodian what's the status on this? Did you guys follow up on this so that it gets merged into Ansible core?

kustodian:
@matamouros we were just super busy when they locked down other the repositories, so we didn't get a PR. I want to fix a few small bugs first, then I will make a PR.

kustodian:
For those interested to upvote the new PR for this module, you can find it here https://github.com/ansible/ansible/pull/21054

PR #1114, opened 2015-10-15
Made win_iis_webapppool Strict Compliant

blakfeld:
This module was failing on me. It appears to be due to Strict being toggled. This change gets the module working as expected.

Issue #1093, opened 2015-10-13
win_iis_webapppool: The property 'Value' cannot be found on this object


rriggs:
Please see https://github.com/ansible/ansible/issues/12494

``` text
TASK [foo : Create fooservice app pool] **************************************
fatal: [foohost]: FAILED! => {"changed": false, "failed": true, "msg": "The property 'Value' cannot be found on this object. Verify that the property exists."}
```

``` yml
- name: Create fooservice app pool
  win_iis_webapppool:
    name: "fooservice"
    state: started
    attributes: "enable32BitAppOnWin64:true|managedRuntimeVersion:v4.0|managedPipelineMode:0"
```

> it sounds like a result of #12384, which turned on strict mode by default for modules.
> Updating the module to include Set-StrictMode -Off just below the # POWERSHELL_COMMON line
> would be a quick workaround to get it working again.

rriggs:
The issue appears to be triggered by the "managedPipelineMode:0" portion.

``` PowerShell
$a = Get-ItemProperty ("IIS:\AppPools\" + $name) -name managedPipelineMode
$a
Integrated
```

jimi-c:
Per the comment on #1114, can you please confirm whether this has been resolved for you?

Thanks!

gregdek:
Closed due to inactivity.

dverg:
This should be fixed and not Closed. The properties on the apppool could be strings or complex objects and checking against (a non-existing) Value property won't work. 
See the ref above: https://github.com/ansible/ansible-modules-extras/issues/2212

Issue #2072, opened 2016-04-20
win_iis_webapppool failing with 'Specified cast is not valid'


michelleperz:
##### ISSUE TYPE

- Bug Report
##### COMPONENT NAME

win_iis_webapppool
##### ANSIBLE VERSION

```
ansible --version
ansible 2.0.0.2
  config file = 
  configured module search path = Default w/o overrides
```
##### CONFIGURATION

No changes
##### OS / ENVIRONMENT

N/A
##### SUMMARY

The win_iis_webappool documentation states that you can separate attributes by a pipe and name values by colon like  "foo:1|bar:2"  However, doing this in the task results in a 'Specified cast is not valid error.'
##### STEPS TO REPRODUCE

Create a basic playbook with attributes that are separated by a pipe:

```
- name: Setting up IIS application pool 
    win_iis_webapppool: 
      name: PoolParty
      state: started 
      attributes: "managedRuntimeVersion:v4.0|managedPipelineMode:0|queueLength:30000"
```

##### EXPECTED RESULTS

App pool would get created successfully.
##### ACTUAL RESULTS

```
TASK [Setting up IIS application pool] *****************************************
fatal: [52.39.68.141]: FAILED! => {"changed": false, "failed": true, "msg": "Specified cast is not valid."}
```

ansibot:
@whenrik ping, this issue is waiting for your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@whenrik, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@whenrik, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@whenrik, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@whenrik, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@whenrik, ping. This issue is still waiting on your response.
[...]

PR #2578, opened 2016-07-15
Set-StrictMode -Off to support attributes with no Value object.

deekayen:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

windows/win_iis_webapppool.ps1
##### ANSIBLE VERSION

```
ansible 2.1.0.0
```
##### SUMMARY

Some attempts to get the win_iis_webapppool module to be strict compliant have still not fixed an issue with setting attributes on the application pool.

After some lengthy debugging on my Windows machine, I found that on line ~82, the `Get-ItemProperty ("IIS:\AppPools\" + $name) $newParameter.Key` doesn't always return a property with a .Value object. For most properties, the value of the returned `$newParameter` is `Microsoft.IIs.PowerShell.Framework.ConfigurationAttribute`, which has a `.Value` object.

One specific case where `Get-ItemProperty` returns merely a value with no .Value object is the managedPipelineMode attribute noted on https://github.com/ansible/ansible-modules-extras/issues/1093#issuecomment-147817712. It returns a simple string like `Integrated`.

Another I found was trying to set `processModel.identityType`, which returns `SpecificUser`.

The absence of .Value is a problem on the following line because it needs to test for the existence of a null .Value object. Merely doing the check in the if statement to `$currentParameter.Value -ne $null` itself causes a warning in strict mode. Catching the warning also seems to cause an exception.

This introduces another exception to the strict rule as have been done with other iis modules.

Long list of references to link:

https://github.com/ansible/ansible/issues/12497
https://github.com/ansible/ansible/issues/12494#issuecomment-142679608
https://github.com/ansible/ansible/issues/12554
https://github.com/ansible/ansible-modules-extras/pull/1114
https://github.com/ansible/ansible-modules-extras/issues/1093#issuecomment-147817712


gregdek:
Thanks @deekayen. To the current maintainers, @whenrik please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

deekayen:
Issue #2072 may also be related to this pull request because they also reference the managedPipelineMode that I wrestled with.

I also noticed that @bahaula created a win_iis role on Galaxy and included a win_iis_webapppool.ps1 library file that adds the `Set-StrictMode -Off` that is suggested by this pull request. https://github.com/bahaula/win_iis/blob/master/library/win_iis_webapppool.ps1

jhawkesworth:
I hit this today, and I've found that if you change the script to turn of strict mode, you can set managedPipelineMode|Classic
or 
managedPipelineMode|Integrated

I sort of feel that this is half a fix though, as the attributes returned by the module still have 0 or 1 for the managedPipelineMode.  This seems to be because the options are based off an xml schema and this particular attribute is an enum.  I have searched around but haven't been able to find a way to look up the enum name using the methods available in Microsoft.IIs.PowerShell.Framework.ConfigurationAttribute or Microsoft.IIs.PowerShell.Framework.ConfigurationAttributeSchema but I'm probably missing something.

I think it is worth getting it to return the enum names so that users can know what correct values to set are.

deekayen:
Yes, my pull request is only half the fix. It was the easiest way I could get a solution to set managedPipelineMode and share a solution with others.

jhawkesworth:
Having read this I think it might be possible to convert the enums into names.  
http://stackoverflow.com/questions/37571351/how-can-powershell-tell-me-the-enum-type-of-an-iis-property
The missing piece of the puzzle for me is discovering the enum names.  I guess it would mean looping through all of the returned attributes and finding their type.  I suppose they could be hard coded into the module but it ought to be possible to determine them at runtime, which would protect against schema changing in future IIS versions.

jhawkesworth:
I have proof of concept code that looks up the enum names. I just need to handle the fact that there isn't always a Value property returned when iterating over the existing pool attributes.

jhawkesworth:
@deekayen - sorry it has taken me a while but I have created #3257 which I think fixes this issue without turning off strict mode and lets you set the controlled values by name.  It also returns the names of the possible values.  If you get the chance please could you give it a try.

PR #1939, opened 2016-04-01
Adding new ec2 security group facts module.

shousper:
##### ISSUE TYPE
- New Module Pull Request
##### COMPONENT NAME

ec2_group_facts
##### ANSIBLE VERSION

```
ansible 2.0.1.0
  config file = /etc/ansible/ansible.cfg
  configured module search path = Default w/o overrides
```
##### SUMMARY

New module provides EC2 security group facts via boto3.

Implementation essentially exposes the same features as the boto3 API for `describe_security_groups`, including group names, ids and the filters.

PR #1908, opened 2016-03-24
New security group facts module

dfjaimes:
##### ISSUE TYPE
- New Module Pull Request
##### COMPONENT NAME

ec2_group_facts
##### ANSIBLE VERSION

```
ansible 2.1.0 (devel 92f29746d9) last updated 2016/03/24 16:10:14 (GMT -500)
  lib/ansible/modules/core: (detached HEAD 7efc09ef08) last updated 2016/03/23 15:15:03 (GMT -500)
  lib/ansible/modules/extras: (detached HEAD 7f9cdc0350) last updated 2016/03/23 15:15:04 (GMT -500)
  config file = /Users/djaimes/.ansible.cfg
  configured module search path = /usr/share/ansible:~/working/git/ansible-modules-extras
```
##### SUMMARY

I needed a way to gather security group information for use in various roles and playbooks.
Following other AWS *facts modules as examples, this module simply gathers facts about security groups in AWS. Results can be optionally filtered to narrow results.

PR #1352, opened 2015-12-09
ec2_group_facts.py - new module for gathering EC2 security group facts (list associated instances, loadbalancers, ENIs)

Constantin07:
##### Issue Type:

New Module Pull Request
##### Summary:

This PR add a new module to query EC2 security group facts.
It includes details about associated resources to the security group like:
1. EC2 Instances
2. Elastic Load Balancers
3. Elastic Network Interfaces
4. RDS Instances

The 'lookup' option may take any combination of 'ec2', 'eni', 'elb', 'rds' to query a specific set of resources or just 'all' for all resources. By default it doesn't query any resource types.

Reason: From AWS console there is no easy way to identify to what resources a security group is associated.

Example:

```
  - name: Get security group details
    ec2_group_facts:
      region: "eu-west-1"
      filters:
        group-id: "sg-25c0a341"
      lookup:
        - ec2
        - eni
        - elb
        - rds
    register: group_facts

  - debug: var=group_facts
```

Here is a sample of output:

```
TASK: [debug var=group_facts] *************************************************
ok: [127.0.0.1] => {
    "var": {
        "group_facts": {
            "changed": false,
            "groups": [
                {
                    "resources": {
                        "ec2": [ "i-f9fc5440" ],
                        "elb": [ "test-dev-elb01" ],
                        "eni": [ "eni-5c0e5a7f" ],
                        "rds": [ "test-rds-mysql" ]
                    },
                    "description": "ELB security group",
                    "id": "sg-05250563",
                    "name": "default",
                    "rules": [
                        {
                            "from_port": null,
                            "grants": [
                                "sg-05250563"
                            ],
                            "ip_protocol": "-1",
                            "to_port": null
                        }
                    ],
                    "rules_egress": [
                        {
                            "from_port": null,
                            "grants": [
                                "0.0.0.0/0"
                            ],
                            "ip_protocol": "-1",
                            "to_port": null
                        }
                    ],
                    "vpc_id": "vpc-8a2ca6ee"
                }
            ],
            "invocation": {
                "module_args": "",
                "module_complex_args": {
                    "region": "us-east-1"
                },
                "module_name": "ec2_group_facts"
            }
        }
    }
}
```

PR #2591, opened 2016-07-20
New ec2_group_facts module to be able to get facts from EC2 security groups

Sodki:
##### ISSUE TYPE
- New Module Pull Request
##### COMPONENT NAME

ec2_group_facts
##### ANSIBLE VERSION

```
ansible 2.1.0.0
```
##### SUMMARY

Ansible doesn't have a way to obtain information from EC2 security groups, which is very useful. This implementation is **heavily** inspired on ec2_snapshot_facts. 

The only option available is `filter`. I didn't implement boto3's `GroupNames` and `GroupIds` because they're a source of major confusion (default vs non-default VPC) and `filter` has that functionality anyway, so I thought it was better to keep it simple.

I know there are other PRs for the same module, but they were either broken or had a larger scope than what was intended. I'll list them here for reference:
#1939
#1908
#1352

### DIFF
--- a/cloud/amazon/ec2_group_facts.py
+++ b/cloud/amazon/ec2_group_facts.py
@@ -65,6 +65,12 @@ EXAMPLES = '''
     filters:
       group-id: sg-12345678
 
+# Gather facts about a security group with multiple filters, also mixing the use of underscores as filter keys
+- ec2_group_facts:
+    filters:
+      group_id: sg-12345678
+      vpc-id: vpc-12345678
+
 # Gather facts about various security groups
 - ec2_group_facts:
     filters:
--- a/cloud/amazon/ec2_group_facts.py
+++ b/cloud/amazon/ec2_group_facts.py
@@ -93,6 +93,8 @@ try:
 except ImportError:
     HAS_BOTO3 = Falsentry
 
+import traceback
+
 
 def main():
     argument_spec = ec2_argument_spec()
@@ -126,7 +128,7 @@ def main():
             Filters=ansible_dict_to_boto3_filter_list(module.params.get("filters"))
         )
     except ClientError as e:
-        module.fail_json(msg=e.message)
+        module.fail_json(msg=e.message, exception=traceback.format_exc(e))
 
     # Turn the boto3 result in to ansible_friendly_snaked_names
     snaked_security_groups = []
--- a/cloud/amazon/ec2_group_facts.py
+++ b/cloud/amazon/ec2_group_facts.py
@@ -20,7 +20,7 @@ module: ec2_group_facts
 short_description: Gather facts about ec2 security groups in AWS.
 description:
     - Gather facts about ec2 security groups in AWS.
-version_added: "2.2"
+version_added: "2.3"
 author: "Henrique Rodrigues (github.com/Sodki)"
 options:
   filters:
--- a/cloud/amazon/ec2_group_facts.py
+++ b/cloud/amazon/ec2_group_facts.py
@@ -130,9 +130,11 @@ def main():
     else:
         module.fail_json(msg="region must be specified")
 
+    # Replace filter key underscores with dashes, for compatibility, except if we're dealing with tags
     sanitized_filters = module.params.get("filters")
     for key in sanitized_filters:
-      sanitized_filters[key.replace("_", "-")] = sanitized_filters.pop(key)
+        if not key.startswith("tag:"):
+            sanitized_filters[key.replace("_", "-")] = sanitized_filters.pop(key)
 
     try:
         security_groups = connection.describe_security_groups(
--- a/cloud/amazon/ec2_group_facts.py
+++ b/cloud/amazon/ec2_group_facts.py
@@ -27,7 +27,8 @@ options:
     description:
       - A dict of filters to apply. Each dict item consists of a filter key and a filter value. See \
       U(https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeSecurityGroups.html) for \
-      possible filters. Filter names and values are case sensitive.
+      possible filters. Filter names and values are case sensitive. You can also use underscores (_) \
+      instead of dashes (-) in the filter keys, which will take precedence in case of conflict.
     required: false
     default: {}
 notes:
@@ -123,9 +124,13 @@ def main():
     else:
         module.fail_json(msg="region must be specified")
 
+    sanitized_filters = module.params.get("filters")
+    for key in sanitized_filters:
+      sanitized_filters[key.replace("_", "-")] = sanitized_filters.pop(key)
+
     try:
         security_groups = connection.describe_security_groups(
-            Filters=ansible_dict_to_boto3_filter_list(module.params.get("filters"))
+            Filters=ansible_dict_to_boto3_filter_list(sanitized_filters)
         )
     except ClientError as e:
         module.fail_json(msg=e.message, exception=traceback.format_exc(e))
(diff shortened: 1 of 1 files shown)


dfjaimes:
#1939 was closed a while back, so its irrelevant here
#1352 is certainly out of scope
#1908 isn't broken, nor is it out of scope - its just waiting on review again

I wrote it back in the boto2 days then retrofitted it to boto3 for the PR, so if the GroupNames and GroupIds aren't needed, I could simply pull that logic out - but, that hasn't been brought up in past reviews so I've never given it any thought.

gregdek:
Thanks @Sodki for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

Sodki:
@calamityman Cool, at the time of writing this your PR was still in old boto. Instead of spamming your PR with comments, can I just send a bunch of PRs to your own branch, to try and make it better? Then I can close this one down.

dfjaimes:
@Sodki of course. I'd be more than happy to work with you on this module if you've got some improvements.

Sodki:
@calamityman Unfortunately it doesn't seem to be possible to send PRs to your branch, since it's a fork. I'll comment straight to your PR.

Sodki:
#2619 is another PR related to this one.

Sodki:
ready_for_review

aairey:
Any update on this? Seems stalled.

Sodki:
@aairey Well, the module is finished and works, but no one has reviewed it yet. I don't know if there's something else that I need to do to speed things up.

cloud/amazon/ec2_group_facts.py
```diff
@@ -0,0 +1,148 @@
...
+- ec2_group_facts:
+    filters:
+      group-name:
+        - example-1
+        - example-2
+        - example-3
+
+# Gather facts about any security group with a tag key Name and value Example
+- ec2_group_facts:
+    filters:
```
ryansb, line 84: On this example can you add a comment to the effect of "The quotes around `tag:Name` are important because of the colon in the value" because not everyone realizes the significance there.

cloud/amazon/ec2_group_facts.py
```diff
@@ -0,0 +1,148 @@
...
+
+# Gather facts about a security group
+- ec2_group_facts:
+    filters:
+      group-name: example-1
+
+# Gather facts about a security group by id
+- ec2_group_facts:
+    filters:
+      group-id: sg-12345678
```
ryansb, line 66: It would be really nice to be able to do underscore-separated tags as well, since that's typically how folks expect to do params. Do you think that'd be a useful addition? 

``` diff
diff --git a/cloud/amazon/ec2_group_facts.py b/cloud/amazon/ec2_group_facts.py
index 777fc59..bc9a75d 100644
--- a/cloud/amazon/ec2_group_facts.py
+++ b/cloud/amazon/ec2_group_facts.py
@@ -121,9 +121,11 @@ def main():
     else:

+    sanitized_filters = {k.replace('_', '-'): v for k, v in module.params.get("filters", {}).items() if 'tag:' not in k}
+    sanitized_filters.update({k: v for k, v in module.params.get("filters", {}).items() if 'tag:' in k})
     try:
         security_groups = connection.describe_security_groups(
-            Filters=ansible_dict_to_boto3_filter_list(module.params.get("filters"))
+            Filters=ansible_dict_to_boto3_filter_list(sanitized_filters)
         )
     except ClientError as e:
```

cloud/amazon/ec2_group_facts.py
```diff
@@ -0,0 +1,148 @@
...
+        )
+    else:
+        module.fail_json(msg="region must be specified")
+
+    try:
+        security_groups = connection.describe_security_groups(
+            Filters=ansible_dict_to_boto3_filter_list(module.params.get("filters"))
+        )
+    except ClientError as e:
+        module.fail_json(msg=e.message)
```
ryansb: Here can you add the exception to the failure, something like:

```
import traceback # do this with the other imports, not in the exception
module.fail_json(msg=e.message, exception=traceback.format_exc(e))
```

cloud/amazon/ec2_group_facts.py
```diff
@@ -0,0 +1,148 @@
...
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: ec2_group_facts
+short_description: Gather facts about ec2 security groups in AWS.
+description:
+    - Gather facts about ec2 security groups in AWS.
+version_added: "2.2"
```
ryansb: Need to update this to 2.3

ryansb (changes requested):
Overall looks great, and works in my environment. Just a couple changes before we can get this merged in.

cloud/amazon/ec2_group_facts.py
Sodki: It's been updated to 2.3.

cloud/amazon/ec2_group_facts.py
Sodki: Your comment was spot on, so I just used it verbatim.

cloud/amazon/ec2_group_facts.py
Sodki: It's done. Let me know if you disagree with the location of the import. I really don't have others like that one and I wanted to keep the `boto` ones separate.

cloud/amazon/ec2_group_facts.py
Sodki: I think it's a good idea. I've implemented it a little bit different from you, but I think it's easier to read. I've also changed the documentation accordingly.

Sodki:
Thank you @ryansb for your review. I've implemented all your change requests as separate commits because I think it's easier to understand the flow. If you're happy I can then squash them all before merging.

cloud/amazon/ec2_group_facts.py
```diff
@@ -65,6 +65,12 @@
     filters:
       group-id: sg-12345678
 
+# Gather facts about a security group with multiple filters, also mixing the use of underscores as filter keys
+- ec2_group_facts:
+    filters:
+      group_id: sg-12345678
```
ryansb, line 71: :+1:

cloud/amazon/ec2_group_facts.py
```diff
@@ -0,0 +1,161 @@
...
+            region=region,
+            endpoint=ec2_url,
+            **aws_connect_params
+        )
+    else:
+        module.fail_json(msg="region must be specified")
+
+    sanitized_filters = module.params.get("filters")
+    for key in sanitized_filters:
+      sanitized_filters[key.replace("_", "-")] = sanitized_filters.pop(key)
```
ryansb: There's one issue with this: if someone uses a filter such as `"tag:my_tag_name": value` this code will replace it and make the filter `tag:my-tag-name`. In the example I gave it skipped replacing `_` for any `tag:` filters.

ryansb:
@Sodki really close, just that last comment on filter underscore-replacement. No worries about squashing this PR, I can do so on merge. Thanks for doing separate commits, makes reviewing much easier. :smile:

cloud/amazon/ec2_group_facts.py
Sodki: Oh, I completely forgot about that edge case. It's fixed now.

Sodki:
Alright, I fixed the last issue. Thanks again for taking the time to review this.

ryansb (approved):

PR #2593, opened 2016-07-20
Allow value to be bool where 'yes'/'no' are in choices

tmshn:
**NOTE** This pull request work on same issue as [#4220 on ansible-modules-core](https://github.com/ansible/ansible-modules-core/pull/4220). Therefore, below comment is almost copy-and-paste.
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAMES
- `ecs_service_facts`
- `virt_net`
- `virt_pool`
- `portage`
- `nmcli`
- `debconf`
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  config file =
```
##### SUMMARY

Many arguments from many modules has limited choices including `'yes'` and/or `'no'` as a option. However, in some case, these are treated as string type; boolean values cause error.

This pull request fix this problem, in the following two way:
##### 1. Change type to bool

If choices are only `'yes'` and `'no'`, this argument can be easily modified to bool type. Even parameter is passed as string, it will be [converted to boolean](https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/basic.py#L1439-L1446), so it is 100% backward compatible.

**Modules:**
- `ecs_service_facts`
- `virt_net`
- `virt_pool`
- `nmcli`

**Example:** `ecs_service_facts` module's case

``` yaml

---
# This still works,
- name: Basic listing example
  ecs_service_facts: details=true # snip...

---
# and this still works too,
- name: Basic listing example
  ecs_service_facts:
    details: "true"
    # snip...

---
# AND! this will work now.
- name: Basic listing example
  ecs_service_facts:
    details: true
    # snip...
```
##### 2. Add options to catch boolean

_UPDATED_ Thanks to ansible/ansible#16961, this case doesn't need to be considered any more.

~~If choices contains other values than `'yes'` or `'no'`, this argument should be kept to be treated as string. In this case, if you pass boolean argument here, it will be [converted to `'True'` or `'False'` respectively](https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/basic.py#L1376-L1381). Therefore, adding them to the choices will solve the problem.~~

~~**Modules:**~~
- ~~`portage`~~

~~**Example**: `portage` module's case~~

~~The choices of `sync` argument of `portage` modules contain `"web"`, so it cannot be treated as boolean.~~

``` yaml

---
# This still works,
- name: sync repositories
  portage: sync=yes

---
# and this still works too,
- name: sync repositories
  portage:
    sync: "yes"

---
# AND! this will work now.
- name: sync repositories
  portage:
    sync: yes

---
# Of course, this
- name: sync repositories via web
  portage: sync=web

---
# or this still works.
- name: sync repositories via web
  portage:
    sync: web
```

**Another example**: `debconf` module's case
The `value` argument of `debconf` modules does not have any explicit choices in the modules source code, but actually its choices are limited when `vtype` is `'boolean'`. So I added conversion process here.

### DIFF
--- a/cloud/amazon/ecs_service_facts.py
+++ b/cloud/amazon/ecs_service_facts.py
@@ -55,7 +55,7 @@ EXAMPLES = '''
 - ecs_service_facts:
     cluster: test-cluster
     service: console-test-service
-    details: "true"
+    details: true
 
 # Basic listing example
 - ecs_service_facts:
@@ -201,7 +201,7 @@ def main():
 
     argument_spec = ec2_argument_spec()
     argument_spec.update(dict(
-        details=dict(required=False, choices=['true', 'false'] ),
+        details=dict(required=False, type='bool', default=False ),
         cluster=dict(required=False, type='str' ),
         service=dict(required=False, type='str' )
     ))
@@ -214,9 +214,7 @@ def main():
     if not HAS_BOTO3:
       module.fail_json(msg='boto3 is required.')
 
-    show_details = False
-    if 'details' in module.params and module.params['details'] == 'true':
-        show_details = True
+    show_details = module.params.get('details', False)
 
     task_mgr = EcsServiceManager(module)
     if show_details:
--- a/cloud/misc/virt_net.py
+++ b/cloud/misc/virt_net.py
@@ -534,16 +534,16 @@ def core(module):
         else:
             module.fail_json(msg="Command %s not recognized" % basecmd)
 
-    if autostart:
+    if autostart is not None:
         if not name:
             module.fail_json(msg = "state change requires a specified name")
 
         res['changed'] = False
-        if autostart == 'yes':
+        if autostart:
             if not v.get_autostart(name):
                 res['changed'] = True
                 res['msg'] = v.set_autostart(name, True)
-        elif autostart == 'no':
+        else:
             if v.get_autostart(name):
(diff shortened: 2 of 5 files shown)


gregdek:
Thanks @tmshn. To the current maintainers, @sayap, @Java1Guy, @alcamie101, @drybjed, @bcoca please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

tmshn:
Same as ansible/ansible-modules-core#4220, this PR is pended until ansible/ansible#16961 will be merged. needs_revision

tmshn:
Thanks to @abadger (ansible/ansible#16961), modification for case 2 is meaningless any more. Thus, I dropped commit on `packaging/os/portage.py`.

Other than that, below modules are welcome for reviews.

**Case 1**
- ecs_service_facts (e5675b3)
- virt_net & virt_pool (cb5a22b)
- nmcli (46651b5)

**Case 3**
- debconf (1ee56fb)
  - In debconf's case, the choices are implicit (invisible from ansible module), so still needs conversion.

abadger:
Thanks @tmshn !  Merged to devel and cherry-picked to stable-2.1.

PR #2690, opened 2016-08-07
Change argument 'no_reload' to 'reload', fix #2321

### LINKED ISSUE #2321: selinux_permissive no_reload option should be renamed/deprecated
##### ISSUE TYPE
- Feature Idea
##### COMPONENT NAME

selinux_permissive module
##### ANSIBLE VERSION

```
ansible 2.0.2.0
```
##### OS / ENVIRONMENT

CentOS 7
##### SUMMARY

The "no_reload" option is badly named.  It starts with a "negative" concept (no_), then you set either true or false to enable/disable the negative concept.  This double-negative is confusing and should be renamed to use a more obvious/positive concept.

Name should be changed to "reload" with default "true".  The no_reload option could still be supported as deprecated to prevent breakage in current playbooks.

mscherer:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

selinux_permissive
##### SUMMARY

Fix #2321 

As I think almost no one use the option, I guess it is
safe to break backward compatibility for that.

### DIFF
--- a/system/selinux_permissive.py
+++ b/system/selinux_permissive.py
@@ -36,14 +36,16 @@ options:
         - "indicate if the domain should or should not be set as permissive"
     required: true
     choices: [ 'True', 'False' ]
-  no_reload:
+  reload:
     description:
         - "automatically reload the policy after a change"
-        - "default is set to 'false' as that's what most people would want after changing one domain"
+        - "default is set to 'true' as that's what most people would want after changing one domain"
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks again to @mscherer. This module is going into community review.

We notice that you are one of the maintainers (@mscherer) of this module, so when you think it's ready, please comment "shipit" and we will consider it for merging.

bcoca:
tempted to keep no_reload as an option in the code and assign to reload if present, just to keep backwards compatibility

PR #2748, opened 2016-08-17
Add facts and storage system modules.

hulquest:
##### ISSUE TYPE
- New Module Pull Request
##### COMPONENT NAME

storage/netapp_e/na_eseries_facts.py
storage/netapp_e/na_eseries_storage_system.py
##### ANSIBLE VERSION

```
ansible 2.2.0 (devel 4cc4dc6793) last updated 2016/07/19 12:02:12 (GMT -600)
  lib/ansible/modules/core: (detached HEAD 7de287237f) last updated 2016/07/19 12:02:18 (GMT -600)
  lib/ansible/modules/extras: (detached HEAD 68ca157f3b) last updated 2016/07/19 12:02:21 (GMT -600)
  config file = /home/khulques/ansible.cfg
  configured module search path = ['/home/khulques/src/projects/ansible-modules-extras/storage/netapp_e', '/home/khulques/src/projects/ansible_eseries/ansible_modules']
```
##### SUMMARY

Adding facts and core system for initial review.  

###### Add/remove systems.

```
10.251.230.29 | SUCCESS => {
    "alreadyExists": false, 
    "alreadyExits": false, 
    "changed": true, 
    "id": "ansible4"
}

10.251.230.29 | SUCCESS => {
    "changed": true, 
    "msg": "Storage system removed."
}
```
###### Facts

```
10.251.230.29 | SUCCESS => {
    "ansible_facts": {
        "netapp_disks": [
            {
                "available": false, 
                "firmware_version": "MS09", 
                "id": "010000005000CCA03C7C9FCC0000000000000000", 
                "media_type": "hdd", 
                "product_id": "HUC106060CSS601 ", 
                "serial_number": "PZJ6JPXD", 
                "status": "optimal", 
                "tray_ref": "0E00000000000000000000000000000000000000", 
                "usable_bytes": "594221686784"
            }, 
            {
                "available": false, 
                "firmware_version": "MS09", 
                "id": "010000005000CCA03C7DC49C0000000000000000", 
                "media_type": "hdd", 
                "product_id": "HUC106060CSS601 ", 
                "serial_number": "PZJ756ED", 
                "status": "optimal", 
                "tray_ref": "0E00000000000000000000000000000000000000", 
                "usable_bytes": "594221686784"
            }, 
            {
                "available": false, 
                "firmware_version": "MS09", 
                "id": "010000005000CCA03C806DD80000000000000000", 
                "media_type": "hdd", 
                "product_id": "HUC106060CSS601 ", 
                "serial_number": "PZJ8MKVD", 
                "status": "optimal", 
                "tray_ref": "0E00000000000000000000000000000000000000", 
                "usable_bytes": "594221686784"
            }, 
            {
                "available": false, 
                "firmware_version": "MS09", 
                "id": "010000005000CCA03C809EB00000000000000000", 
                "media_type": "hdd", 
                "product_id": "HUC106060CSS601 ", 
                "serial_number": "PZJ8RUPD", 
                "status": "optimal", 
                "tray_ref": "0E00000000000000000000000000000000000000", 
                "usable_bytes": "594221686784"
            }, 
            {
                "available": false, 
                "firmware_version": "MS09", 
                "id": "010000005000CCA03C80E4180000000000000000", 
                "media_type": "hdd", 
                "product_id": "HUC106060CSS601 ", 
                "serial_number": "PZJ8WEZD", 
                "status": "optimal", 
                "tray_ref": "0E00000000000000000000000000000000000000", 
                "usable_bytes": "594221686784"
            }, 
            {
                "available": false, 
                "firmware_version": "MS09", 
                "id": "010000005000CCA03C8112A80000000000000000", 
                "media_type": "hdd", 
                "product_id": "HUC106060CSS601 ", 
                "serial_number": "PZJ8ZK3D", 
                "status": "optimal", 
                "tray_ref": "0E00000000000000000000000000000000000000", 
                "usable_bytes": "594221686784"
            }, 
            {
                "available": false, 
                "firmware_version": "MS09", 
                "id": "010000005000CCA03C8117700000000000000000", 
                "media_type": "hdd", 
                "product_id": "HUC106060CSS601 ", 
                "serial_number": "PZJ8ZVZD", 
                "status": "optimal", 
                "tray_ref": "0E00000000000000000000000000000000000000", 
                "usable_bytes": "594221686784"
            }, 
            {
                "available": false, 
                "firmware_version": "MS09", 
                "id": "010000005000CCA03C8119580000000000000000", 
                "media_type": "hdd", 
                "product_id": "HUC106060CSS601 ", 
                "serial_number": "PZJ8ZZXD", 
                "status": "optimal", 
                "tray_ref": "0E00000000000000000000000000000000000000", 
                "usable_bytes": "594221686784"
            }, 
            {
                "available": false, 
                "firmware_version": "MS09", 
                "id": "010000005000CCA03C8122E40000000000000000", 
                "media_type": "hdd", 
                "product_id": "HUC106060CSS601 ", 
                "serial_number": "PZJ90MMD", 
                "status": "optimal", 
                "tray_ref": "0E00000000000000000000000000000000000000", 
                "usable_bytes": "594221686784"
            }, 
            {
                "available": false, 
                "firmware_version": "MS09", 
                "id": "010000005000CCA03C8126600000000000000000", 
                "media_type": "hdd", 
                "product_id": "HUC106060CSS601 ", 
                "serial_number": "PZJ90VUD", 
                "status": "optimal", 
                "tray_ref": "0E00000000000000000000000000000000000000", 
                "usable_bytes": "594221686784"
            }, 
            {
                "available": false, 
                "firmware_version": "MS09", 
                "id": "010000005000CCA03C8128F40000000000000000", 
                "media_type": "hdd", 
                "product_id": "HUC106060CSS601 ", 
                "serial_number": "PZJ9114D", 
                "status": "optimal", 
                "tray_ref": "0E00000000000000000000000000000000000000", 
                "usable_bytes": "594221686784"
            }, 
            {
                "available": false, 
                "firmware_version": "MS09", 
                "id": "010000005000CCA03C8136600000000000000000", 
                "media_type": "hdd", 
                "product_id": "HUC106060CSS601 ", 
                "serial_number": "PZJ91XVD", 
                "status": "optimal", 
                "tray_ref": "0E00000000000000000000000000000000000000", 
                "usable_bytes": "594221686784"
            }, 
            {
                "available": false, 
                "firmware_version": "MS09", 
                "id": "010000005000CCA03C8138BC0000000000000000", 
                "media_type": "hdd", 
                "product_id": "HUC106060CSS601 ", 
                "serial_number": "PZJ922RD", 
                "status": "optimal", 
                "tray_ref": "0E00000000000000000000000000000000000000", 
                "usable_bytes": "594221686784"
            }, 
            {
                "available": false, 
                "firmware_version": "MS09", 
                "id": "010000005000CCA03C8144940000000000000000", 
                "media_type": "hdd", 
                "product_id": "HUC106060CSS601 ", 
                "serial_number": "PZJ92W5D", 
                "status": "optimal", 
                "tray_ref": "0E00000000000000000000000000000000000000", 
                "usable_bytes": "594221686784"
            }, 
            {
                "available": false, 
                "firmware_version": "MS09", 
                "id": "010000005000CCA03C8144C00000000000000000", 
                "media_type": "hdd", 
                "product_id": "HUC106060CSS601 ", 
                "serial_number": "PZJ92WJD", 
                "status": "optimal", 
                "tray_ref": "0E00000000000000000000000000000000000000", 
                "usable_bytes": "594221686784"
            }, 
            {
                "available": false, 
                "firmware_version": "MS09", 
                "id": "010000005000CCA03C814D580000000000000000", 
                "media_type": "hdd", 
                "product_id": "HUC106060CSS601 ", 
                "serial_number": "PZJ93G8D", 
                "status": "optimal", 
                "tray_ref": "0E00000000000000000000000000000000000000", 
                "usable_bytes": "594221686784"
            }, 
            {
                "available": false, 
                "firmware_version": "MS09", 
                "id": "010000005000CCA03C8155B00000000000000000", 
                "media_type": "hdd", 
                "product_id": "HUC106060CSS601 ", 
                "serial_number": "PZJ940HD", 
                "status": "optimal", 
                "tray_ref": "0E00000000000000000000000000000000000000", 
                "usable_bytes": "594221686784"
            }, 
            {
                "available": false, 
                "firmware_version": "MS09", 
                "id": "010000005000CCA03C817FF00000000000000000", 
                "media_type": "hdd", 
                "product_id": "HUC106060CSS601 ", 
                "serial_number": "PZJ96URD", 
                "status": "optimal", 
                "tray_ref": "0E00000000000000000000000000000000000000", 
                "usable_bytes": "594221686784"
            }, 
            {
                "available": false, 
                "firmware_version": "MS09", 
                "id": "010000005000CCA03C8187680000000000000000", 
                "media_type": "hdd", 
                "product_id": "HUC106060CSS601 ", 
                "serial_number": "PZJ97A4D", 
                "status": "optimal", 
                "tray_ref": "0E00000000000000000000000000000000000000", 
                "usable_bytes": "594221686784"
            }, 
            {
                "available": false, 
                "firmware_version": "MS09", 
                "id": "010000005000CCA03C8187A40000000000000000", 
                "media_type": "hdd", 
                "product_id": "HUC106060CSS601 ", 
                "serial_number": "PZJ97AMD", 
                "status": "optimal", 
                "tray_ref": "0E00000000000000000000000000000000000000", 
                "usable_bytes": "594221686784"
            }, 
            {
                "available": false, 
                "firmware_version": "MS09", 
                "id": "010000005000CCA03C8199640000000000000000", 
                "media_type": "hdd", 
                "product_id": "HUC106060CSS601 ", 
                "serial_number": "PZJ98J8D", 
                "status": "optimal", 
                "tray_ref": "0E00000000000000000000000000000000000000", 
                "usable_bytes": "594221686784"
            }, 
            {
                "available": false, 
                "firmware_version": "MS09", 
                "id": "010000005000CCA03C81A0B00000000000000000", 
                "media_type": "hdd", 
                "product_id": "HUC106060CSS601 ", 
                "serial_number": "PZJ990AD", 
                "status": "optimal", 
                "tray_ref": "0E00000000000000000000000000000000000000", 
                "usable_bytes": "594221686784"
            }, 
            {
                "available": false, 
                "firmware_version": "MS09", 
                "id": "010000005000CCA03C8234F40000000000000000", 
                "media_type": "hdd", 
                "product_id": "HUC106060CSS601 ", 
                "serial_number": "PZJ9LWED", 
                "status": "optimal", 
                "tray_ref": "0E00000000000000000000000000000000000000", 
                "usable_bytes": "594221686784"
            }, 
            {
                "available": false, 
                "firmware_version": "MS09", 
                "id": "010000005000CCA03C823D680000000000000000", 
                "media_type": "hdd", 
                "product_id": "HUC106060CSS601 ", 
                "serial_number": "PZJ9MEWD", 
                "status": "optimal", 
                "tray_ref": "0E00000000000000000000000000000000000000", 
                "usable_bytes": "594221686784"
            }
        ], 
        "netapp_enabled_features": [
            "autoCodeSync", 
            "autoLunTransfer", 
            "driveSlotLimit", 
            "flashReadCache", 
            "mixedDriveTypes", 
            "performanceTier", 
            "protectionInformation", 
            "raid6", 
            "ssdSupport", 
            "stagedDownload", 
            "storagePoolsType2", 
            "subLunsAllowed", 
            "totalNumberOfArvmMirrorsPerArray", 
            "totalNumberOfPitsPerArray", 
            "totalNumberOfThinVolumesPerArray"
        ], 
        "netapp_storage_pools": [
            {
                "available_capacity": "9277129359360", 
                "id": "0400000060080E5000299F88000005F25790BD06", 
                "name": "Disk_Pool_1", 
                "total_capacity": "10045928505344", 
                "used_capacity": "768799145984"
            }
        ], 
        "netapp_volumes": [
            {
                "capacity": "536870912000", 
                "id": "0200000060080E5000299F1C0000061C5790C84E", 
                "is_thin_provisioned": false, 
                "name": "1ASG", 
                "parent_storage_pool_id": "0400000060080E5000299F88000005F25790BD06"
            }, 
            {
                "capacity": "214748364800", 
                "id": "0200000060080E5000299F88000005F35790CC51", 
                "is_thin_provisioned": false, 
                "name": "Bravo", 
                "parent_storage_pool_id": "0400000060080E5000299F88000005F25790BD06"
            }, 
            {
                "capacity": "10737418240", 
                "id": "0200000060080E5000299F1C0000061E5790D362", 
                "is_thin_provisioned": false, 
                "name": "BravoMT", 
                "parent_storage_pool_id": "0400000060080E5000299F88000005F25790BD06"
            }, 
            {
                "capacity": "4294967296", 
                "id": "0200000060080E5000299F1C0000061F579107FC", 
                "is_thin_provisioned": false, 
                "name": "repos_0037", 
                "parent_storage_pool_id": "0400000060080E5000299F88000005F25790BD06"
            }
        ], 
        "snapshot_images": [], 
        "ssid": "ansible2"
    }, 
    "changed": false, 
    "msg": "Gathered facts for ansible2."
}
```


gregdek:
Thanks @hulquest for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

storage/netapp_e/na_eseries_storage_system.py
```diff
@@ -0,0 +1,319 @@
...
+        return resp_code, data
+
+
+def main():
+    argument_spec = basic_auth_argument_spec()
+    argument_spec.update(dict(
+        state=dict(required=True, choices=['present', 'absent']),
+        ssid=dict(required=True, type='str'),
+        controller_addresses=dict(type='list'),
+        array_wwn=dict(required=False, type='str'),
```
gundalow: Required true according to the documentation you've added above.

storage/netapp_e/na_eseries_storage_system.py
```diff
@@ -0,0 +1,319 @@
...
+        array_password=dict(required=False, type='str'),
```
gundalow: To avoid password appearing in logs you should set `no_log=True`

See http://docs.ansible.com/ansible/developing_modules.html

storage/netapp_e/na_eseries_storage_system.py
```diff
@@ -0,0 +1,319 @@
...
+
+try:
+    from urllib.error import HTTPError
+except ImportError:
+    from urllib2 import HTTPError
+
+DOCUMENTATION = '''
+module: na_eseries_storage_system
+version_added: "2.2"
+short_description: Add/remove arrays from  the Web Services Proxy
```
gundalow: Looks like you have a double space between "from"  and "the"

gundalow:
needs_revision

storage/netapp_e/na_eseries_storage_system.py
hulquest: Umm, looks false to me.  Am I missing something?

```
+  array_wwn:
 +    required: false
 +    description:
 +    - The WWN of the array to manage. Only necessary if in-band managing multiple arrays on the same agent host.
```

hulquest:
ready_for_review

abadger:
Module should be in storage/netapp/  The e for e-series is contained in the module name already.

storage/netapp_e/na_eseries_facts.py
```diff
@@ -0,0 +1,194 @@
...
+import json
+
+from ansible.module_utils.api import basic_auth_argument_spec
+from ansible.module_utils.basic import AnsibleModule, get_exception
+from ansible.module_utils.urls import open_url
+
+try:
+    from urllib.error import HTTPError
+except ImportError:
+    from urllib2 import HTTPError
```
abadger: You can retrieve this from ansible.module_utils.six.moves.urllib.error

storage/netapp_e/na_eseries_facts.py
```diff
@@ -0,0 +1,194 @@
...
+    if resp_code >= 400 and not ignore_errors:
+        raise Exception(resp_code, data)
+    else:
+        return resp_code, data
+
+
+def main():
+    argument_spec = basic_auth_argument_spec()
+    argument_spec.update(
+        ssid=dict(required=True))
```
abadger: Looks like you need to add documentation of this parameter?

storage/netapp_e/na_eseries_facts.py
```diff
@@ -0,0 +1,194 @@
...
+
+    # fetch the list of storage-pool objects and look for one with a matching name
+    try:
+        (rc, resp) = request(api_url + "/storage-systems/{0}/graph".format(ssid),
+                             headers=dict(Accept="application/json"),
+                             url_username=api_usr, url_password=api_pwd, validate_certs=validate_certs)
+    except:
+        error = get_exception()
+        module.fail_json(
+            msg="Failed to obtain facts from storage array with id [{0}]. Error [{1}]".format(ssid, str(error)))
```
abadger: .format() is only available in python2.6+.  Probably should use percent formatting for python-2.4 compatibility.

```
msg="Failed to obtain facts from storage array with id [%s]. Error [%s]" % ssid, str(error))
```

storage/netapp_e/na_eseries_facts.py
```diff
@@ -0,0 +1,194 @@
...
+
+from ansible.module_utils.api import basic_auth_argument_spec
+from ansible.module_utils.basic import AnsibleModule, get_exception
+from ansible.module_utils.urls import open_url
+
+try:
+    from urllib.error import HTTPError
+except ImportError:
+    from urllib2 import HTTPError
+
```
abadger: style note -- usually we find the three DOCUMENTATION strings before the imports.  Not a blocker but it seems to be the convention.

storage/netapp_e/na_eseries_facts.py
```diff
@@ -0,0 +1,194 @@
...
+        api_pwd = os.environ['API_PASSWORD']
+        api_url = os.environ['API_URL']
+
+    if not api_usr or not api_pwd or not api_url:
+        module.fail_json(msg="You must pass in api_username "
+                             "and api_password and api_url to the module. Alternatively "
+                             "you can specify API_URL, API_USERNAME and API_PASSWORD "
+                             "as environment variables.")
+    facts = dict(ssid=ssid)
+    result = dict(ansible_facts=facts, changed=False)
```
abadger: Style note: Since result isn't used until the very end I wouldn't assign it until just before exit_json().  At first I was confused as to why we were adding all these values to `facts` and then returning `result`.  Once I read all the way back up to the top it became apparent what was going on but it means that you have to have both these initial lines and exit_json lines at the end in your mind at the same time.

storage/netapp_e/na_eseries_facts.py
```diff
@@ -0,0 +1,194 @@
...
+
+    try:
+        api_usr = p['api_username']
+        api_pwd = p['api_password']
+        api_url = p['api_url']
+    except KeyError:
+        import os
+        api_usr = os.environ['API_USERNAME']
+        api_pwd = os.environ['API_PASSWORD']
+        api_url = os.environ['API_URL']
```
abadger: Is this something we want to do?  If so, it should be added to the generic handling of basic auth in module_utils.

abadger:
Not sure if we want this to be a facts module.  The values aren't facts of the machine it runs on -- they're facts of the remote netapp device.  On the other hand, I don't think we have a standard idiom for storing information from modules.  register is there but it's kind of ad hoc.

storage/netapp_e/na_eseries_storage_system.py
```diff
@@ -0,0 +1,319 @@
...
+from ansible.module_utils.api import basic_auth_argument_spec
+from ansible.module_utils.basic import AnsibleModule
+from ansible.module_utils.pycompat24 import get_exception
+from ansible.module_utils.urls import open_url
+
+try:
+    from urllib.error import HTTPError
+except ImportError:
+    from urllib2 import HTTPError
+
```
abadger: Same two comments about imports as with the other module.

storage/netapp_e/na_eseries_storage_system.py
```diff
@@ -0,0 +1,319 @@
...
+    - Whether the specified array should be configured on the Web Services Proxy or not.
+    choices: ['present', 'absent']
+  controller_addresses:
+    required: true
+    description:
+    - The list addresses for the out-of-band management adapter or the agent host.
+  array_wwn:
+    required: false
+    description:
+    - The WWN of the array to manage. Only necessary if in-band managing multiple arrays on the same agent host.
```
abadger: Need to say what the default is for any entry where required is false.

abadger: The default can be None/Null if you want to express that the entry is not necessarily needed by the underlying library/hardware/service.

storage/netapp_e/na_eseries_storage_system.py
```diff
@@ -0,0 +1,319 @@
...
+    if rc == 200:
+        array_exists = True
+        array_detail = resp
+
+        if state == 'absent':
+            changed = True
+            array_exists = False
+        elif state == 'present':
+            current_addresses = set(
+                [i for i in [array_detail['ip1'], array_detail['ip2']] if i is not None and len(i) > 1])
```
abadger: If valid values of array_detail['ip*'] are strings or None than this can be simplified like:

``` python

frozenset(i for i in (array_detail['ip1'], array_detail['ip2']) if i)
```

storage/netapp_e/na_eseries_storage_system.py
```diff
@@ -0,0 +1,319 @@
...
+            if set(controller_addresses) != current_addresses:
+                changed = True
+            if array_detail['wwn'] != array_wwn and array_wwn != None:
```
abadger: Only matters in very very corner cases but `array_wwn is not None` is a better way to compare to None.

storage/netapp_e/na_eseries_storage_system.py
```diff
@@ -0,0 +1,319 @@
...
+        api_url = os.environ['API_URL']
+
+    if not api_usr or not api_pwd or not api_url:
+        module.fail_json(msg="You must pass in api_username "
+                             "and api_password and api_url to the module. Alternatively "
+                             "you can specify API_URL, API_USERNAME and API_PASSWORD "
+                             "as environment variables.")
+
+    changed = False
+    array_exists = False
```
abadger: I'd move definition of these two defaults to just before the `if rc == 200`.  That way people know that they're default values that are about to be overridden.

storage/netapp_e/na_eseries_storage_system.py
```diff
@@ -0,0 +1,319 @@
...
+                    array_add_req['wwn'] = array_wwn
+
+                if array_password:
+                    array_add_req['password'] = array_password
+
+                post_headers = dict(Accept="application/json")
+                post_headers['Content-Type'] = 'application/json'
+
+                try:
+                    request_data = json.dumps(array_add_req)
```
abadger: Probably should move the json.dumps() outside of this exception block.  Smaller exception blocks lead to more targetted exception handling and less surprises.

storage/netapp_e/na_eseries_storage_system.py
```diff
@@ -0,0 +1,319 @@
...
+            else:  # array exists, modify...
+                post_headers = dict(Accept="application/json")
+                post_headers['Content-Type'] = 'application/json'
+                post_body = dict(
+                    controllerAddresses=controller_addresses,
+                    removeAllTags=True,
+                    enableTrace=enable_trace,
+                    metaTags=meta_tags
+                )
+
```
abadger: Seems like the following section should be extracted into a function that's shared between the update-existing and the create-new case.

abadger:
In addition to the inline comments  storage/netapp_e/na_eseries_storage_system.py, almost all of the comments on the facts module also apply to the na_eseries_storage_system module.

storage/netapp_e/na_eseries_facts.py
```diff
@@ -0,0 +1,194 @@
...
+
+    ssid = p['ssid']
+    validate_certs = p['validate_certs']
+
+    try:
+        api_usr = p['api_username']
+        api_pwd = p['api_password']
+        api_url = p['api_url']
+    except KeyError:
+        import os
```
abadger: Should do import of os with the other imports.  Since it comes fro mthe stdlib there's no reason to make it conditional.

abadger:
Putting this ticket on the meeting agenda so we can answer the following two questions:
- basic auth api reading from environment: https://github.com/ansible/ansible-modules-extras/pull/2748/files#diff-d456352230dca855dd4a0606113e1a5aR107
- Should we write facts modules that are targetting external devices?  ie:

```
- hosts: localhost
  tasks:
    - netapp_facts:
        device_ip: 10.0.0.17
```

The theoretical facts module is getting facts about the storage device at 10.0.0.17.  But because it's a facts module, that information is saved on the inventory_host, in this case localhost.  Is that what we want?  If we have no other ways to offer, then perhaps we do want this for now and we should design something better in the future.

storage/netapp_e/na_eseries_facts.py
hulquest: You'll have to help me out here.  

``` python
from ansible.module_utils.six.moves.urllib.error import HTTPError 
```

Doesn't resolve.  I see a whole bunch of classes in six.py but I don't understand what's happening.  Looks like a python equivalent of a custom classloader.

storage/netapp_e/na_eseries_facts.py
hulquest: Oh!  My editor said PEP 8 dictates imports before docstrings.  I checked and it's clear they should be after.

storage/netapp_e/na_eseries_facts.py
hulquest: This design was implemented by Ansible consultants; I thought they'd know the standards.  I'll remove it.

storage/netapp_e/na_eseries_facts.py
hulquest: Agree.  None of the other modules written by the Ansible team do this up-front assignment technique.  Will resolve.

storage/netapp_e/na_eseries_facts.py
hulquest: Fixed.

storage/netapp_e/na_eseries_storage_system.py
hulquest: Fixed.

storage/netapp_e/na_eseries_facts.py
abadger: Which version of the Ansible library are you running against?  This works for me with the current devel tree:

```
$ export |grep PYTHONPATH                                     *[devel]  (15:27:39)
PYTHONPATH=/srv/ansible/abadger/lib:
$ python -c 'from ansible.module_utils.six.moves.urllib.error import HTTPError'
```

abadger: $ ansible --version                                           *[devel](15:28:43)
ansible 2.2.0 (devel 0a39700b36) last updated 2016/08/25 15:00:15 (GMT -700)

storage/netapp_e/na_eseries_facts.py
abadger: Cool.  I can see the utility of falling back to the environment.  We have something similar for AWS and openstack moudles I think.  Those are done in module_utils/ libraries for that type of module though.  As I'm thinking about this more, if you want to have environment variables in the future, they probably would have to use names that reference netapp.  One module might use basic auth for netapp and another for a web based notification service.  The user probably wouldn't have the same username, password, and URL for both.

storage/netapp_e/na_eseries_facts.py
hulquest: It's something in my editor that isn't configured correctly.  From the command line, it looks good.  Running 2.2.0 btw.

I have a test to run through the http exception handler and it's working as expeced.

```
10.251.230.29 | FAILED! => {
    "changed": false, 
    "failed": true, 
    "msg": "Error accessing storage-system with id [ansible4]. Error [<urlopen error [Errno 113] No route to host>]"
}
```

storage/netapp_e/na_eseries_facts.py
hulquest: After we get these modules released, I'll make a change and submit a PR.  I like the idea and utility.

storage/netapp_e/na_eseries_storage_system.py
hulquest: I added some words about array_wwn and controller_addresses being mutually exclusive and enforce it in the AnsibleModule constructor.

storage/netapp_e/na_eseries_storage_system.py
hulquest: Agree.

storage/netapp_e/na_eseries_storage_system.py
hulquest: I refactored the common lines into a function.  Across all the other modules I'll be sending across there is _a lot_ of refactoring that can be done to remove technical debt like this.  I have it on our ansible backlog because I'm under pressure to get all the modules available to the community.

storage/netapp_e/na_eseries_storage_system.py
```diff
@@ -160,48 +184,36 @@ def main():
...
-                [i for i in [array_detail['ip1'], array_detail['ip2']] if i is not None and len(i) > 1])
+            current_addresses = frozenset(i for i in (array_detail['ip1'], array_detail['ip2']) if i)
             if set(controller_addresses) != current_addresses:
                 changed = True
-            if array_detail['wwn'] != array_wwn and array_wwn != None:
+            if array_detail['wwn'] != array_wwn and array_wwn is not None:
                 module.fail_json(
-                    msg='It seems you may have specified a bad WWN. The storage system ID you specified, %s, currently has the WWN of %s' % (
-                        ssid, array_detail['wwn']))
+                    msg='It seems you may have specified a bad WWN. The storage system ID you specified, %s, " + '
```
abadger: Typo here?  Seems like you wanted:

``` python
 msg='It seems you may have specified a bad WWN. The storage system ID you specified, %s, ' \
```

abadger: Actually this is in the call to fail_json so no need for the backslash either:

``` python
 msg='It seems you may have specified a bad WWN. The storage system ID you specified, %s, '
```

storage/netapp_e/na_eseries_facts.py
```diff
@@ -0,0 +1,188 @@
...
+    else:
+        return resp_code, data
+
+
+def main():
+    argument_spec = basic_auth_argument_spec()
+    argument_spec.update(
+        api_username=dict(type='str', required=True),
+        api_password=dict(type='str', required=True, no_log=True),
+        api_url=dict(type='str', required=True),
```
abadger: Since we're making these required=true now, we also need to document them as such.  Talked to bcoca and unfortunately there's no current way to override documentation fragments so you'll have to remove the extends_documentation-fragment from the DOCUMENTATION section, copy the text into the modules documentation and then modify it to show that they are now required parameters.

abadger:
Down to just a few inline comments now, the question about facts modules that we'll resolve at the meeting this week, and git mv the module's directory to storage/netapp (from storage/netapp_e)

storage/netapp_e/na_eseries_storage_system.py
hulquest: Thanks!  I mixed up usage of single and double quotes.  I put the string addition in to keep the line less than 120.

storage/netapp_e/na_eseries_facts.py
hulquest: Ok, makes sense.  I just pushed some changes to reflect the updated api\* comments and to address the silly mistake of mixing single and double quotes.  Unless it's a deal breaker, I'd like to keep netapp_e in the namespace.  Our puppet and chef plugins are known as netapp_e.  ONTAP is simply 'netapp' and I imagine SolidFire will be netapp_s.

storage/netapp_e/na_eseries_storage_system.py
```diff
@@ -212,7 +224,7 @@ def main():
                 changed = True
             if array_detail['wwn'] != array_wwn and array_wwn is not None:
                 module.fail_json(
-                    msg='It seems you may have specified a bad WWN. The storage system ID you specified, %s, " + '
+                    msg='It seems you may have specified a bad WWN. The storage system ID you specified, %s, ' +
                         'currently has the WWN of %s' % (ssid, array_detail['wwn']))
```
abadger: The plus will mess things up.  Like C, python will combine adjacent strings automatically into a single string literal.  Using the plus, otoh, will make python create two separate string literals and then concat them as a third string.  So the plus is unnecessary and makes things slightly less efficient.

In this case, there's a further wrinkle.  We're doing string formatting using %.  because of that, precedence rules come into play.  Python will first combine adjacent literal strings, then do percent formatting, and finally concatenate strings via plus.  That means that with the plus, when we get to the percent formatting operation we don't yet have a string with both %s's in it.  So we'd get a TypeError:

```
TypeError: not all arguments converted during string formatting
```

Removing the plus will resolve that.

storage/netapp_e/na_eseries_facts.py
```diff
@@ -21,9 +21,21 @@
...
+  api_password:
+    required: true
+    description:
+    - The password to authenticate with the SANtricity WebServices Proxy or embedded REST API.
+  api_url:
+    required: true
+    description:
+    - The url to the SANtricity WebServices Proxy or embedded REST API.
+    example:
+    - https://prod-1.wahoo.acme.com/devmgr/v2
```
abadger: There's also a validate_certs parameter that should be added to the documentation (since the basic_auth arg_spec provides it).

storage/netapp_e/na_eseries_storage_system.py
hulquest: Ok, I'm not a python pro; thanks for the detail.  I just pushed a commit with that corrects.

abadger:
Using a facts module was okay'd at the meeting yesterday.  So I think all that's left are:
- documenting validate_certs parameter
- moving the modules from the storage/netapp_e/ directory to storage/netapp/

After those are fixed, this should be good to merge.

storage/netapp_e/na_eseries_facts.py
hulquest: Just pushed a commit that documents validate_certs.

storage/netapp_e/na_eseries_facts.py
```diff
@@ -36,6 +36,10 @@
     - The url to the SANtricity WebServices Proxy or embedded REST API.
     example:
     - https://prod-1.wahoo.acme.com/devmgr/v2
+  validate_certs:
+    required: false
+    description:
+    - Should https certificates be validated?
```
abadger: Add default value here (default: True)

abadger:
@hulquest I noticed you came to the new module meeting today.  Sorry I missed you.  Regarding whether to move the module to a netapp directory, I asked that of the other core committers previously and got a unanimous response in favour of netapp.

Here's the discussion from the meeting:

> <hulquest> i just pushed a commit to doc validate_certs.  i'd like to keep the netapp_e name because our puppet and chef modules are known as netapp_e.  ontap plugins are simply netapp
> <ryansb> what does the _e stand for?
> <hulquest> e-series.  it's the block storage platform in the netapp portfolio.
> <hulquest> the solidfire platform will probably be netapp_s or _sf when they implement modules
> <hulquest> imho, having ontap, eseries, and solidfire modules in the same directory will get kind of messy

And here's the discussion I had previously with the other core committers:

> 7:51 AM tkuratomi Do we want the netapp modules to be in storage/netapp_e?  Or storage/netapp/ ?  ("module for NetApp E/EF-Series" )
> 7:51 AM jctanner storage/netapp
> 7:51 AM jimi-c storage/netapp
> 7:51 AM jctanner name the modules according to the series
> 7:51 AM i don't a module subtopic per company product
> 7:52 AM tkuratomi The code is small and pretty easy to read, btw.
> 7:54 AM bcoca storage/netapp ... leave the _e_ for the modules
> 7:55 AM otherwise we need 40 cisco_<x> dirs

If @jimi-c, @bcoca, or @jctanner if you'd like to expand on that, for hulquest, that would be great.

ryansb:
After further name discussion, here's what we have in mind:
1. Move all these modules into the `storage/netapp` directory
2. Rename the files themselves to `netapp_e_name.py`

We can't have a `storage/netapp_[series]` directory for every series, and our documentation generator can't handle deeper nesting (also ruling out `storage/netapp/e_series/foo.py`). The `na_e` prefix, while unique, makes it hard for users to search for all netapp modules with something like `ansible-doc | grep netapp`. Also when using the modules users won't necessarily see that `na` means `netapp` in that context. 

I can see why you  want to have them separated by series, but that's not a distinction we have for any other modules. For example, `network/` is aligned by operating system, and `cloud/` is separated by vendor, not cloud product.

hulquest:
@ryansb, that's reasonable and makes a lot of sense.  I think the key difference between Ansible and our other IT auto plugins is they are delivered autonomously.  I'll change the name of the directory and module names and commit/push shortly.  We're still good to make another PR for the 17 other modules?  I'll make sure all the comments from this PR are included in that one.  We're taking heat to get the modules available to our customers (internal and external).

ryansb:
Yes, I believe other than the names you're all set.

hulquest:
Thank you @abadger and @ryansb.  I think it's ready to merge.

abadger:
Thanks @hulquest everything looks good now.  Merged to devel!

PR #2800, opened 2016-08-27
Use six for iteration, to make it run on python3

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

timezone

### DIFF
--- a/system/timezone.py
+++ b/system/timezone.py
@@ -20,6 +20,7 @@
 
 import os
 import re
+from ansible.module_utils.six import iteritems, iterkeys
 
 DOCUMENTATION = '''
 ---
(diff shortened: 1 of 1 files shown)


gregdek:

system/timezone.py
```diff
@@ -110,7 +111,7 @@ def __init__(self, module):
         # Initially there's only info of "planned" phase, but the
         # `self.check()` function will fill out it.
         self.value = dict()
-        for key in module.argument_spec.iterkeys():
+        for key in iterkeys(module.argument_spec):
```
abadger, line 114: Stylewise we don't need to do this.  Iterating over the dict is mostly the same as iterkeys.

``` python
for key in module.argument_spec:
```

abadger:
Note -- stylewise, I usually tell people, if the list isn't long (doesn't take up much memory), just use:

```
for item in dict.items()
```

but it's not a problem to use iteritems from six if we don't want to change what the original author did here.

Issue #2384, opened 2016-06-07
yum_repository: allow enable/disable of repository


hhsnow:
##### ISSUE TYPE

- Feature Request
##### COMPONENT NAME

yum_repository
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  config file = /etc/ansible/ansible.cfg
  configured module search path = Default w/o overrides
```
##### SUMMARY

I am unable to use yum_repository to update `enabled = 1` without providing additional configuration. For example, in Amazon Linux, they ship the epel repository with `enabled = 0` by default. This would allow this repository to be enabled permanently. While I currently use the `ini_file` module, this seems like the right place. 
##### STEPS TO REPRODUCE

Proposed item:

```
- name: yum | enable epel repository
  yum_repository:
    name: epel
    enabled: yes
```

Current Item:

```
- name: yum | enable epel repository
  yum_repository:
    name: epel
    description: "Extra Packages for Enterprise Linux 6 - $basearch"
    mirrorlist: "https://mirrors.fedoraproject.org/metalink?repo=epel-6&arch=$basearch"
    failovermethod: priority
    gpgkey: "file:///etc/pki/rpm-gpg/RPM-GPG-KEY-EPEL-6"
    gpgcheck: yes
    enabled: yes
```

##### EXPECTED RESULTS

`/etc/yum.repo.d/epel.repo` should have the following line set:

```
[epel]
...
enabled=1
...
```
##### ACTUAL RESULTS

```
fatal: [example.com]: FAILED! => {"changed": false, "failed": true, "msg": "Parameter 'baseurl' or 'mirrorlist' is required."}
```

jtyr:
Current version of `yum_repository` module was mainly designed to create new repo files although it allows to override repo definition in existing files. For that you have to define complete repo definition including at least the `name` and one of the `baseurl` or `mirrorlist`.

I can imagine to extend the module with `state: update` which would allow to add/modify properties of existing repos:

```
- yum_repository:
    name: epel
    enabled: yes
    state: update
```

To remove an option from the repo definition, one could set the option to `null` value:

```
- yum_repository:
    name: epel
    failovermethod: null
    state: update
```

In both cases, there can be whole list of options to modify/remove.

Please let me know what do you think about it and I will try to implement it.

hhsnow:
That makes sense to me. Makes it easier to distinguish between a full change and a single element update. Do any other modules have an 'update' state?

ansibot:
[...]

PR #2820, opened 2016-08-31
add module with enable/disable yum repo functionality

tcstang:
##### ISSUE TYPE

- New Module Pull Request
##### COMPONENT NAME

yum_repo_enabler
##### ANSIBLE VERSION

```
ansible 2.1.0.0
```
##### SUMMARY

There is currently no functionality in yum_repository and yum modules to enable/disable repositories as necessary, which I find is very useful when working in development environments. This module attempts to remedy that.

### DIFF
--- a/packaging/os/yum_repo_enabler.py
+++ b/packaging/os/yum_repo_enabler.py
@@ -1,0 +1,132 @@
+#!/usr/bin/python
+# encoding: utf-8
+
+# (c) 2016, Timothy Tang <tcstang@gmail.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+import ConfigParser
+from ConfigParser import NoSectionError
+
+DOCUMENTATION = '''
+---
+module: yum_repo_enabler
+extends_documentation_fragment: files
+description:
+    - Enables/disables repository files contained in a yum directory.
+version_added: 1.1
+options:
+    directory:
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @tcstang for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

alxgu:
It is possible to disable/enable repos during the `yum` module run:
- disablerepo
      `Repoid' of repositories to disable for the install/update operation. These repos will not persist beyond the transaction. When specifying multiple repos,
      separate them with a ",". [Default: None]
- enablerepo
      `Repoid' of repositories to enable for the install/update operation. These repos will not persist beyond the transaction. When specifying multiple repos,
      separate them with a ",". [Default: None]

tcstang:
@alxgu:

The problem with the yum module is that the enabled/disabled repos don't persist beyond the transaction as is stated. Often in development environments, we need to enable/disable specific repos accordingly and yum and yum_repository do not provide this functionality.

There is actually an issue filed, #2384, that requests this functionality from yum_repository but interestingly, it didn't fit the module's goal.

robinro:
This functionality should be added to the yum_repository module. There is already an `enabled` flag in that module, so the only thing to change is make it work on existing repositories. This should be a good idea anyhow to ensure idempotency of the module.

Adding a separate module would duplicate some functionality without gaining anything imo.

tcstang:
@robinro 

That module creates and updates yum repo files given a specific file name. Since the yum_repository module's purpose is to either create or remove a yum repo file, it doesn't really make sense (according to the maintainer) to add this functionality. He was considering adding an "update" to the state which would edit it, but decided against it.

parberge:
I agree with @jtyr comment in #2384 that using init_file module should already handle this.

verdurin:
I'll give this a try - I need this functionality, provided in a cleaner way than using init_file, and while I disagree with @jtyr regarding the purpose of `yum_repository`, it's his right to make that decision and this plugin would be a suitable alternative method.

jtyr:
@verdurin I appreciate your determination but unfortunately this module is not meant to tap into an existing YUM repo file if that file wasn't created by this module. For tapping into a YUM repo file (which is actually INI file) is more suitable to use the `init_file` module.

verdurin:
@jtyr - sure, but I think the module in this PR - `yum_repo_enabler` - will do exactly what I'm after, while allowing you to keep `yum_repository` as is.

jtyr:
@verdurin I really don't see a reason to create a new module if we already have the `ini_file` module which does absolutely the same thing.

PR #2842, opened 2016-09-02
Bugfix bigip_facts that was trying to check the length of an iterator

caphrim007:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

network/f5/bigip_facts.py
##### ANSIBLE VERSION

```
ansible 2.2.0
  config file = 
```
##### SUMMARY

Recently, a user reported that the bigip_facts module was failing with the error

```
received exception: object of type 'itertools.imap' has no len()
```

This reported was occurring at line 1657 of the bigip_facts module

bug report is here

https://github.com/F5Networks/f5-ansible/issues/25

Upon further investigation, the map function for returning the specified
includes was returning an iterator, and calling len() on an iterator does
not work.

I believe this problem was caused by part of the Python 3.x effort insofar
as the inclusion of this line

https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/basic.py#L143

seems to affect our usage of map(), probably for the better anyway, and we need
to change our expectations in our module's code to no longer assume a list, but
instead assume an iterator.

After crawling through the module_utils/basic code, I think a list
comprehension is more appropriate here anyway, so I'm changing it to be
that. The affected user reported it works this way, and my own testing
on 2.2.0 supports that.

### DIFF
--- a/network/f5/bigip_facts.py
+++ b/network/f5/bigip_facts.py
@@ -1641,7 +1641,7 @@ def main():
         regex = fnmatch.translate(fact_filter)
     else:
         regex = None
-    include = map(lambda x: x.lower(), module.params['include'])
+    include = [x.lower() for x in module.params['include']]
     valid_includes = ('address_class', 'certificate', 'client_ssl_profile',
                       'device', 'device_group', 'interface', 'key', 'node',
                       'pool', 'provision', 'rule', 'self_ip', 'software',


caphrim007:
@mhite @gundalow for review

gregdek:
Thanks again to @caphrim007. This module is going into community review.

We notice that you are one of the maintainers (@mhite, @caphrim007) of this module, so when you think it's ready, please comment "shipit" and we will consider it for merging.

mhite:
shipit

gundalow:
@abadger FYI as this may be Python 3 related. Is this something we should look for in other bits of the code?

gregdek:
Thanks again to @caphrim007 for this PR, and thanks @mhite, @caphrim007 for reviewing. Marking for inclusion.

abadger:
@gundalow yes, it could turn up in other code ported to python3 as well.

abadger:
In python3 many of the functools return iterators instead of lists.  Importing from six makes the functions on python2 behave in the same way.  Since the effort is new, it shouldn't be hard to look at the places that we import map, reduce, filter (although filter is usually better replaced with a list comprehension or generator expression) from six to see if they need lists.

PR #2848, opened 2016-09-04
update asa_config module

privateip:
- removes get_module() factory method for NetworkModule
- add src argument to provide path to config file
- add new choice to match used to ignore current running config
- add update argument with choices merge or check
- add backup argument to backup current running config to control host
- add defaults argument to control collection of config with or withoutdefaults
- add save argument to save current running config to startup config


gregdek:

PR #2850, opened 2016-09-04
added new functionality to asa_command

privateip:
- commands argument now accepts a dict arguments
- only show commands are allowd when check mode is specified
- config mode is no longer allowed in the command stack
- add argument match with valid values any, all


gregdek:

PR #2257, opened 2016-05-19
Added os_host_aggregate and os_host_aggregate_member  modules

timl:
##### ISSUE TYPE
- New Module Pull Request
##### COMPONENT NAME

os_host_aggregate and os_host_aggregate_member
##### ANSIBLE VERSION

```
ansible 2.2.0 (devel ddd9f92255) last updated 2016/05/10 10:00:59 (GMT -500)
  lib/ansible/modules/core: (detached HEAD 1f5cf669dd) last updated 2016/05/12 09:44:10 (GMT -500)
  lib/ansible/modules/extras: (detached HEAD 816673dd6f) last updated 2016/05/12 09:44:10 (GMT -500)
  config file =
  configured module search path = Default w/o overrides
```
##### SUMMARY

Adds support for management of OpenStack host aggregates.

`os_host_aggregate` handles creating, updating, and deleting aggregates.
`os_host_aggregate_member` handles host aggregate memberships.

These functions are split to match the existing OpenStack tools and APIs.  These modules depend on features available in the next release of shade (>1.8.0).

PR #2873, opened 2016-09-07
added os_aggregate module

kuboj:
##### ISSUE TYPE
- New Module Pull Request
##### COMPONENT NAME

os_aggregate
##### ANSIBLE VERSION

```
ansible 2.1.1.0
```
##### SUMMARY

This PR adds new module for OpenStack - `os_aggregate` - for managing host aggregates.

### DIFF
--- a/cloud/openstack/os_aggregate.py
+++ b/cloud/openstack/os_aggregate.py
@@ -1,0 +1,196 @@
+#!/usr/bin/python
+# Copyright 2016 Jakub Jursa <jakub.jursa1@gmail.com>
+#
+# This module is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This software is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this software.  If not, see <http://www.gnu.org/licenses/>.
+
+try:
+    import shade
+    HAS_SHADE = True
+except ImportError:
+    HAS_SHADE = False
+
+from distutils.version import StrictVersion
+
+DOCUMENTATION = '''
+---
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @kuboj. To the current maintainers, @emonty, @shrews, @juliakreger, @j2sol, @rcarrillocruz please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

cloud/openstack/os_aggregate.py
```diff
@@ -0,0 +1,196 @@
...
+                if metadata:
+                    cloud.set_aggregate_metadata(aggregate.id, metadata)
+                changed = True
+            else:
+                if _needs_update(module, aggregate):
+                    if availability_zone is not None:
+                        aggregate = cloud.update_aggregate(aggregate.id,
+                            name=name, availability_zone=availability_zone)
+                    if metadata is not None:
+                        metas = metadata
```
emonty, line 166: This seems like a leftover from something? metas is only used 3 lines down ...

cloud/openstack/os_aggregate.py
```diff
@@ -0,0 +1,196 @@
...
+                             "the shade library MUST be >=1.6.0")
+
+    name = module.params['name']
+    metadata = module.params['metadata']
+    availability_zone = module.params['availability_zone']
+    hosts = module.params['hosts']
+    state = module.params['state']
+
+    if metadata is not None:
+        metadata.pop('availability_zone', None)
```
emonty, line 134: Does putting availability_zone in the metadata conflict with the ansible module parameter then? If so, should we toss an error if the user tries to supply one? (For that matter, should we update shade's set_aggregate_metadata to validate and throw an error?)

emonty:
This looks generally great - however, could we rename it to os_nova_host_aggregate? I know that's long - but the current convention is that things that are for Operators and not End Users carry the service name in them. Also, people refer to this feature as "host aggregate" when talking, so I think it would be more discoverable by people if it we don't drop the host.

haad:
this seems to duplicate #2257

Shrews:
As @haad pointed out, there is an existing PR for this. Please coordinate with that author to converge on one PR.

PR #2898, opened 2016-09-12
New Module (validate) to provide a set of methods for variable/input validation

exodusftw:
##### ISSUE TYPE
- New Module Pull Request
##### COMPONENT NAME

test/validate.py
##### ANSIBLE VERSION

```
ansible 2.1.1.0
```
##### SUMMARY

A new module to provide a set of methods for validating variables.

This includes the ability to:
- Validate variable types (bool, str, list, dict, int)
- Validate variables against a regex
- Validate Integer variables against an accepted range
- Accept variable values from a whitelist of acceptable values
- Reject variable values from a blacklist of unacceptable values
#### **SPECIAL NOTE**
- **requires `PyYAML` and `re` python modules on validating host - defaults to ansible server**
- **utilizes `delegate_to: localhost(127.0.0.1)` by default for validation methods**

```
# PASSING EXAMPLES
TASK [validate str type] **********************************************
ok: [localhost.localdomain] => {"changed": false, "invocation": {"module_args": {"blacklist": null, "delegate_to": "127.0.0.1", "input_type": "str", "matcher": null, "num_range": null, "value": "default_string", "whitelist": null}, "module_name": "validate"}, "msg": "PASS: Input value 'default_string' of: '<type 'str'>' matches validation requirement for value to be of: '<type 'str'>"}

TASK [validate list type] **********************************************
ok: [localhost.localdomain] => {"changed": false, "invocation": {"module_args": {"blacklist": null, "delegate_to": "127.0.0.1", "input_type": "list", "matcher": null, "num_range": null, "value": "['list_value1', 'list_value2']", "whitelist": null}, "module_name": "validate"}, "msg": "PASS: Input value '['list_value1', 'list_value2']' of: '<type 'list'>' matches validation requirement for value to be of: '<type 'list'>"}

TASK [validate dict type] **********************************************
ok: [localhost.localdomain] => {"changed": false, "invocation": {"module_args": {"blacklist": null, "delegate_to": "127.0.0.1", "input_type": "dict", "matcher": null, "num_range": null, "value": "{'dict_key1': 'dict_value1', 'dict_key2': 'dict_value2'}", "whitelist": null}, "module_name": "validate"}, "msg": "PASS: Input value '{'dict_key1': 'dict_value1', 'dict_key2': 'dict_value2'}' of: '<type 'dict'>' matches validation requirement for value to be of: '<type 'dict'>"}

TASK [validate integer range] **********************************************
ok: [localhost.localdomain] => {"changed": false, "invocation": {"module_args": {"blacklist": null, "input_type": null, "matcher": null, "num_range": "1-20", "value": "6", "whitelist": null}, "module_name": "validate"}, "msg": "PASS: Input value '6' is within bounds of Number Range 1-20 for validation requirement"}

TASK [validate against regex] **********************************************
ok: [localhost.localdomain] => {"changed": false, "invocation": {"module_args": {"blacklist": null, "input_type": null, "matcher": "^.*test string.*$", "num_range": null, "value": "test string", "whitelist": null}, "module_name": "validate"}, "msg": "PASS: Input value 'test string' matches validation requirement against regex: /^.*test string.*$/"}

TASK [validate against whitelist] **********************************************
ok: [localhost.localdomain] => {"changed": false, "invocation": {"module_args": {"blacklist": null, "input_type": null, "matcher": null, "num_range": null, "value": "valid", "whitelist": ["valid", "test", "yes"]}, "module_name": "validate"}, "msg": "PASS: Input value 'valid' is contained within value whitelist '['valid', 'test', 'yes']' validation requirement"}
```

```
# FAILURE EXAMPLES
TASK [validate against regex] **************************************************
fatal: [localhost.localdomain]: FAILED! => {"changed": false, "failed": true, "msg": "FAIL: Input value 'this is a failing test' does not match validation requirement against regex: /^.*example string.*$/"}

TASK [validate user bool] ******************************************************
fatal: [localhost.localdomain]: FAILED! => {"changed": false, "failed": true, "invocation": {"module_args": {"blacklist": null, "input_type": "bool", "matcher": null, "num_range": null, "value": "nope", "whitelist": null}, "module_name": "validate"}, "msg": "FAIL: Input value 'nope' of: '<type 'str'>' does not match validation requirement for value to be of: '<type 'bool'>"}

```

### DIFF
--- a/test/validate.py
+++ b/test/validate.py
@@ -1,0 +1,308 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2016, Jeremy Grant <jeremy.grant@outlook.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: validate
+author: "Jeremy Grant (@exodusftw)"
+version_added: 1.0
+requirements: [ 're', 'yaml']
+short_description: Methods for variable validation
+description:
+  - The M(validate) module provides a set of validation tools
+    for variables. This is handled through a
+    set of options to provide basic type enforcement,
+    argument matching against a regular expression, integer range
+    enforcement, and the ability to validate inputs
+    against whitelist/blacklist entries.
+options:
+  blacklist:
+    description:
+      - List of user input values to reject
+    required: false
+    default: null
+  delegate_to:
+    description:
+      - Host to execute validation tasks - defaults to ansible server
+    required: true
+    default: 127.0.0.1
+  input_type:
+    description:
+      - Required data type for input value - if passed, module will fail
+        if data type provided by user does not match input_type passed
+        to validate
+    required: false
+    choices: [ I(boolean), I(bool), I(str), I(string), I(array), I(list), I(hash), I(dict), I(int) ]
+    default: null
+  matcher:
+    description:
+      - Regular expression to match against user input
+        Expression should be passed as a string rather than I(/regex/)
+    required: false
+    default: null
+  num_range:
+    description:
+      - Number Range to be used for validation
+        against number vars - should be formatted as
+        I(minimum-maximum) i.e. I(1-100)
+    required: false
+    default: null
+  value:
+    description:
+      - User input variable to validate
+    required: true
+    default: null
+  whitelist:
+    description:
+      - List of user input values to accept
+    required: false
+    default: null
+'''
+
+EXAMPLES = '''
+# Basic input validation examples *assumes value field is user input var*
+
+- name: validate boolean example with variable expansion
+  validate:
+  args:
+    value: "{{ example_boolean_var }}"
+    input_type: bool
+
+- name: validate boolean stub variable
+  validate:
+  args:
+    value: True
+    input_type: bool
+
+- name: validate stub integer variable range
+  validate:
+  args:
+    value: 9
+    num_range: '1-100'
+
+- name: validate stub variable against regex
+  validate:
+  args:
+    value: 'http://test.example.com/example/made-up-site'
+    matcher: '^.*/example/.*$'
+
+- name: validate stub variable against value whitelist
+  validate:
+  args:
+    value: 'accepted_value1'
+    whitelist:
+      - 'accepted_value1'
+      - 'accepted_value2'
+
+- name: validate stub variable against value blacklist
+  validate:
+  args:
+    value: 'rejected_value1'
+    blacklist:
+      - 'rejected_value1'
+      - 'rejected_value2'
+'''
+
+RETURN = '''
+---
+pass:
+  description: Returns a string containing the validation result
+  returned: pass
+  type: string
+  sample: "PASS: Input value 'true' of type I(type bool) matches validation requirement for value to be of type: I(type bool)"
+fail:
+  description: Returns a string containing the validation result
+  returned: fail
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @exodusftw for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

exodusftw:
Thought it might be beneficial to add some usage examples:

```
# Basic input validation examples *assumes value field is user input var*
- name: validate boolean example with variable expansion
  validate:
  args:
    value: "{{ example_boolean_var }}"
    input_type: bool

- name: validate boolean stub variable
  validate:
  args:
    value: True
    input_type: bool

- name: validate stub integer variable range
  validate:
  args:
    value: 9
    num_range: '1-100'

- name: validate stub variable against regex
  validate:
  args:
    value: 'http://test.example.com/example/made-up-site'
    matcher: '^.*/example/.*$'

- name: validate stub variable against value whitelist
  validate:
  args:
    value: 'accepted_value1'
    whitelist:
      - 'accepted_value1'
      - 'accepted_value2'

- name: validate stub variable against value blacklist
  validate:
  args:
    value: 'rejected_value1'
    blacklist:
      - 'rejected_value1'
      - 'rejected_value2'
```

jhawkesworth:
Not looked at this in any detail but feels like there might be a bit of overlap with existing 'assert' module - 
http://docs.ansible.com/ansible/assert_module.html 

I'm open to persuasion but I wonder if it might be better to extend 'assert' if necessary.  Perhaps you have a use case that would persuade me that validate needs to be a separate thing though?

exodusftw:
@jhawkesworth - I can see the potential for overlap with assert, I guess my thought is that assert would be limiting in that I don't see the potential for support of regex matching, strong type enforcement, or `if var (in|not in) (whitelist|blacklist)` type operations. In that regard, it made sense to me to separate it out into a completely separate module where the sole purpose is validation of vars/types.

I can see the perspective of potentially extending assert, given that the assert module is basically just an empty module ([this?](https://github.com/ansible/ansible-modules-core/blob/1a0e15094fc940a98f8fdc5acca3506f2ae633c0/utilities/logic/assert.py) and [this?](https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/action/assert.py)), my primary use case is that I need the ability to validate user input variables, either directly on the command line or through tower.

I guess another concern would be around adoption, and whether or not adoption would be improved if this was an extension of assert or a standalone module. I personally prefer them to be separate, but we can have the discussion if there are others with concerns.

bcoca:
Aside from the overlap with assert/fail, this would work better either as an action plugin or even a filter, there is little use to push execution to a host for this kind of thing.

exodusftw:
@bcoca - I actually tend to agree that this would probably be better served as an action plugin - I basically started down this path cause I couldn't find much in the way of documentation regarding input validation. Assert expressions make sense, but it didn't seem to handle some of the things like type enforcement or regex matching that I mentioned above (although I would love to save myself the time and be corrected on that point if i'm mistaken, as i'm admittedly no expert here).

I can work on converting this to an action plugin, but before I set off to work on that, are there serious concerns that this overlaps with `assert` and do any of you feel like there's an actual need for this type of functionality?

bcoca:
assert/fail work with `when:` which can be any jinja2 expression which should be flexible enough to do everything you mention.

exodusftw:
I think it's totally reasonable to utilize assert if it has the proper filters in place to force type casting so there are safeguards in place to force proper user inputs. However, for validation against accept/rejected values lists or regex matching, I would think it's preferable to abstract some of that away, if for no other reason than to not litter the code with jinja2 expressions that I would imagine could get somewhat unwieldy for things like whitelisting/blacklisting and having to use issubset()

jhawkesworth:
I don't mind a bit of jinja2 since its a cornerstone of ansible.  Personally I'd rather see assert extended if there's things it can't do right now (I haven't felt the need to test for issubset on anything yet).  
assert gets used a lot in the integration tests so I suggest you have a look round the integration tests and see what you think about the readability of what is there.  Maybe mocking up some examples either using (extended) assert or proposed validate syntax might help see if its worth digging into.

exodusftw:
I've converted this to an action plugin and will submit to core for further discussion. I'll close this PR and link here to the new PR's for documentation purposes once they're submitted.

exodusftw:
Submitted new PR's to core for a new action plugin:

[Core Module #4820](https://github.com/ansible/ansible-modules-core/pull/4820)
[Core Plugin #17552](https://github.com/ansible/ansible/pull/17552)

PR #2936, opened 2016-09-16
Modification of describe_gateways key so that it is consistent with w…

wimnat:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

ec2_customer_gateway
##### ANSIBLE VERSION

ansible 2.2.0 (detached HEAD d8a3feb976) last updated 2016/09/13 12:03:16 (GMT +000)
  lib/ansible/modules/core: (detached HEAD 7de287237f) last updated 2016/09/13 12:03:45 (GMT +000)
  lib/ansible/modules/extras: (detached HEAD 68ca157f3b) last updated 2016/09/13 12:04:11 (GMT +000)
  config file =
##### SUMMARY

The boto3 describe_gateway api returns a key of CustomerGateways where as the create_gateway api returns a key of CustomerGateway.  This just fixes the key so it is consistent when returned to the user.

Also added AnsibleModule spec to require bgp_ip on state=present as defined in the doc


ryansb:
I'd rather not break the old behavior - can you instead leave the old (weird) pluralization with a copy of the correct singular key? That way folks using the plural key won't have their playbooks broken.

ryansb:
Otherwise LGTM

cloud/amazon/ec2_customer_gateway.py
```diff
@@ -44,6 +44,9 @@
     required: false
     default: present
     choices: [ 'present', 'absent' ]
+notes:
+    - Return values contain customer_gateway and customer_gateways keys which are identical dicts. You should use
+      customer_gateway. See U(https://github.com/ansible/ansible-modules-extras/issues/2773) for details.
```
ryansb, line 49: Thanks for adding the docs here too :+1:

cloud/amazon/ec2_customer_gateway.py
```diff
@@ -208,19 +217,22 @@ def main():
...
+    # key of CustomerGateway. For consistency, change it here
+    existing['CustomerGateway'] = existing['CustomerGateways']
 
     results = dict(changed=False)
     if module.params['state'] == 'present':
-        if existing['CustomerGateways']:
-            results['gateway']=existing
-            if existing['CustomerGateways'][0]['Tags']:
-                tag_array = existing['CustomerGateways'][0]['Tags']
+        if existing['CustomerGateway']:
```
ryansb, line 226: Won't this mean that sometimes `CustomerGateway` will now be an array?

cloud/amazon/ec2_customer_gateway.py
```diff
@@ -208,19 +217,22 @@ def main():
     name = module.params.get('name')
 
     existing = gw_mgr.describe_gateways(module.params['ip_address'])
+    # describe_gateways returns a key of CustomerGateways where as create_gateway returns a
+    # key of CustomerGateway. For consistency, change it here
+    existing['CustomerGateway'] = existing['CustomerGateways']
```
ryansb, line 222: I think this should be `existing['CustomerGateway'] = existing['CustomerGateways'][0]`because Gateways is an array, not an object.

ryansb (changes requested):

gregdek:
Thanks again to @wimnat for this PR, and thanks @michaelbaydoun for reviewing. Marking for inclusion.

ryansb:
bot_broken 

mbaydoun didn't participate at all

cloud/amazon/ec2_customer_gateway.py
wimnat: The way it has been coded further on always points to the first element of the list.  This is OK because boto3s response syntax is a list of dicts (and there will always be only 1 element in the list).

If I made your suggested change then I'd have to remove all the `[0]` references further down so i think it's best to leave as is.

cloud/amazon/ec2_customer_gateway.py
wimnat: It always has been an array (list) with one element.

MichaelBaydoun:
Sorry for the late response.  I agree with leaving the CustomerGateways key for backwards compatibility, we can remove it in 3.0.

resmo:
@ryansb the bot issue has been address to comment without mentioning the maintainer if an ansible member gave the "ship it". I removed your comment so the bot will not label this PR as "bot broken" again. Thanks!

cloud/amazon/ec2_customer_gateway.py
ryansb: Totally fair, thanks!

cloud/amazon/ec2_customer_gateway.py
ryansb: Thank you

ryansb (approved):

PR #2947, opened 2016-09-17
New module = elb_target_group

wimnat:
##### ISSUE TYPE
- New Module Pull Request
##### COMPONENT NAME

elb_target_group
##### ANSIBLE VERSION

2.3
##### SUMMARY

A new module for AWS elastic load balancer target groups.  Goes well with my other PR for ec2_application_lb.

Unfinished but eager to get some feedback from the community.

### DIFF
--- a/cloud/amazon/elb_target_group.py
+++ b/cloud/amazon/elb_target_group.py
@@ -234,7 +234,7 @@ def wait_for_status(connection, module, target_group_arn, targets, status):
             else:
                 time.sleep(polling_increment_secs)
         except botocore.exceptions.ClientError as e:
-            module.fail_json(msg=str(e))
+            module.fail_json(msg=e.message, **camel_dict_to_snake_dict(e.response))
 
     result = response
     return status_achieved, result
--- a/cloud/amazon/elb_target_group.py
+++ b/cloud/amazon/elb_target_group.py
@@ -20,7 +20,7 @@ module: elb_target_group
 short_description: Manage a target group for an Application load balancer
 description:
     - Manage an AWS Application Elastic Load Balancer target group. See U(http://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-target-groups.html) for details.
-version_added: "2.2"
+version_added: "2.3"
 author: "Rob White (@wimnat)"
 options:
   name:
--- a/cloud/amazon/elb_target_group.py
+++ b/cloud/amazon/elb_target_group.py
@@ -117,7 +117,22 @@ EXAMPLES = '''
     name: mytargetgroup
     state: absent
 
-
+# Create a target group with targets
+- elb_target_group:
+  name: mytargetgroup
+  protocol: http
+  port: 81
+  vpc_id: vpc-01234567
+  health_check_path: /
+  successful_response_codes: "200,250-260"
+  targets: 
+   - Id: i-01234567
+     Port: 80
+   - Id: i-98765432
+     Port: 80
+  state: present
+  wait_timeout: 200
+  wait: True
 '''
 
 RETURN = '''
@@ -199,6 +214,31 @@ def get_target_group(connection, module):
         else:
             module.fail_json(msg=e.message, **camel_dict_to_snake_dict(e.response))
 
+def diff_list(a, b):
+    """Find the entries in list a that are not in list b"""
+    b = set(b)
+    return [aa for aa in a if aa not in b]
+
+
+def wait_for_status(connection, module, target_group_arn, targets, status):
+    polling_increment_secs = 5
+    max_retries = (module.params.get('wait_timeout') / polling_increment_secs)
+    status_achieved = False
+
+    for x in range(0, max_retries):
+        try:
+            response = connection.describe_target_health(TargetGroupArn=target_group_arn, Targets=targets)
+            if response['TargetHealthDescriptions'][0]['TargetHealth']['State'] == status:
+                status_achieved = True
+                break
+            else:
+                time.sleep(polling_increment_secs)
+        except botocore.exceptions.ClientError as e:
+            module.fail_json(msg=str(e))
+
+    result = response
+    return status_achieved, result
+
 
 def create_or_update_target_group(connection, module):
 
@@ -208,6 +248,7 @@ def create_or_update_target_group(connection, module):
     params['Protocol'] = module.params.get("protocol").upper()
     params['Port'] = module.params.get("port")
     params['VpcId'] = module.params.get("vpc_id")
+
     # If health check path not None, set health check attributes
     if module.params.get("health_check_path") is not None:
         params['HealthCheckPath'] = module.params.get("health_check_path")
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @wimnat for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

naslanidis:
@wimnat 

Is there anything in particular you would like tested or potentially some time spent on?

I'm keen to help get this module over the line as we're keen to replace all of our ELB's.

naslanidis:
@wimnat 

I wasn't sure of the best way to propose my additions as I'm relatively new to github but I've added some functionality to this PR module file if you're interested: 

https://github.com/naslanidis/ansible-modules-extras/blob/new_AWS_module_elb_target_group/cloud/amazon/elb_target_group.py

I've added the followings support for adding targets for a new target group, as well as adding/removing targets for existing target groups. It could do with some refactoring but it works well. 

Let me know what you think. 

Nick.

noahwilliamsson:
I tested this module and and found the lack of target registration problematic. Fortunately, that's fixed by @naslanidis updated version.  It would be nice if Nick's changes could be merged into this PR.

@naslanidis: Target keys such as `Id` and `Port` are a bit awkward.  To match the convention used in other EC2 modules, would you mind changing these to `instance_id` and `port` instead?
Diffing your version against the one in this PR shows that your version introduced some trailing whitespace in various places.  It would be nice if you could get rid of that in a future version.

One way to propose your changes would be to fork https://github.com/wimnat/ansible-modules-extras and then:

```
git clone <URL to your private fork> ansible-modules-extra
cd ansible-modules-extra
git checkout feature/elb_target_group
git checkout -b target-registration
...add your changes...
git commit
git push
```

Once you've pushed your changes to your private fork you can issue a PR against the  _feature/elb_target_group_ branch of wimnat's _ansible-modules-extra_ repo and hope that the changes are accepted.  Once they're merged they'll automatically appear in the PR.

naslanidis:
Hi @noahwilliamsson @wimnat 

Thanks for testing this module and your suggestions. 

I have updated my own versions of these files in my own repo. When I was looking to fork wimnat's repo I ran into issues because it's already a fork of ansible-extras but I will give it another go.  

On the topic of the case for Id and Port, I don't mind changing it but it raises a bigger issue. All of the Boto3 API responses are CamelCase now so I think a decision will need to be made on the best way to handle this. Previously I've converted between case but when the API responses contain lists and dicts of varying depth it gets trickier and adds quite a bit of code to the modules. The parameters themselves I think should definitely be snake case as they've always been, but the question is whether  dict keys for a dict that is simply passed in as part of a parameter also need to be converted or whether they should match the API. Obviously keeping them the same as the API makes comparisons for idempotence tests easier.  

I've been using the below two modules for a few days now and it's working well for my purposes. 

https://github.com/naslanidis/ansible-modules-extras/blob/new_AWS_module_elb_target_group/cloud/amazon/elb_target_group.py

https://github.com/naslanidis/ansible-modules-extras/blob/new_AWS_module_elb_lb_application/cloud/amazon/elb_lb_application.py

On the ALB module I've added the following: 
- added functions to convert for target group ARNs to target group names and vice verser. Using the ARN made these modules more difficult to use as they change each time. 
- Added support for rules (the main reason we need this module)
- Cleaned up the listener function

These modules are still a bit rough but the functionality is coming along.

shaharmor:
Hey @naslanidis, can you please tell me how I can test your new module? What are the installation instructions?

shaharmor:
@naslanidis 
Some issues I found:

#### general

No docs about `aws_access_key`, `aws_secret_key`, `aws_region`  

#### target group
- health_check_port doesn't work. When setting it to any number I'm getting this error:

```
Parameter validation failed:
Invalid type for parameter HealthCheckPort, value: 3000, type: <type 'int'>, valid types: <type 'basestring'>
```

Solution: wrap `module.params.get("health_check_port")` with `str()`
- health_check_protocol value is required to be either `http` or `https` lowercased, but then I'm getting this error:

```
1 validation error detected: Value 'http' at 'healthCheckProtocol' failed to satisfy constraint: Member must satisfy enum value set: [TCP, HTTP, HTTPS]
```

Solution: add `.upper()` to `module.params.get("health_check_protocol")`

#### application load balancer
- listeners are documented as lowercase, but only works with CamelCase
- Protocol value is supposed to be uppercase, documented as lowercase

shaharmor:
Also, target group is missing attributes configuration

wimnat:
I'll have a look at getting all these in over the weekend

shaharmor:
@wimnat not sure if its related to this PR specifically, but we also need a way to configure the target group on the ASG, which is not supported with ansible's `ec2_asg` module. (Requires boto3)

wimnat:
@shaharmor can you clarify " we also need a way to configure the target group on the ASG"?  What exactly is required?  Do you know which part of the boto3 api is needed?

wimnat:
@shaharmor 

I have added the .upper() for healthcheckprotocol and str() for healthcheckport

Docs for access key, region etc is included by the doc fragment for aws

Fixed up protocol value example case

Let me know of anything else.

shaharmor:
@wimnat this is the boto3 api to attach target group: http://boto3.readthedocs.io/en/latest/reference/services/autoscaling.html#AutoScaling.Client.attach_load_balancer_target_groups

shaharmor:
@wimnat here: https://github.com/ansible/ansible-modules-extras/pull/2947/files#diff-4454f84c0b4518d25ee9652a7d368eceR391 you are removing all currently attached instances from the target group if it already exists... this is very bad as running this module twice will basically terminate all running instances, which should stay

shaharmor:
@wimnat wait_timeout should have default of 300

wimnat:
Hmmm I think asg module might need a boto3 rewrite to include that functionality

jonhadfield:
The asg module ec2_asg.py is written with boto2 which doesn't support target groups. I'm in the process of swapping out boto2 for boto3 code with a plan to fully rewriting it.
@wimnat I'm using both your target group and application elb modules successfully and have some feedback I'll compile when I get a chance. In the meantime, they're working for me! Thanks.

The existing asg module has numerous issues that I'm resisting the urge to correct for now as I simply want it functioning for my own purposes. I've currently got it working in so far as it will create the ASG and register the instances with target groups and I'm now working on the update/delete aspects. Nowhere near ready for PR, but I'll keep my work in progress up-to-date on this gist if you want to play/contribute: https://gist.github.com/jonhadfield/698bbd374643187ee0412a98439e3473

jonhadfield:
There's an issue in the logic around how targets are managed. Here's the scenario...
1) I define a target group with no targets
2) I create an autoscaling group, associating the ASG with the target group so that all instances are registered with that target group as they scale up/down
3) Instances become registered with target group
4) I re-run the playbook and the logic in the module says to remove any that aren't defined, therefore detaching the instances the ASG added.

Suggest a 'purge_undefined' module parameter that defaults to True so it can be idempotent when registered via ASG. 
This works for me: https://gist.github.com/jonhadfield/01f9820e1dfcc59b9a7f48819314261b.
Btw, you're missing 'targets' from the documentation section.

aviboru:
thanks for the modules, they are definitely helpful.
One quick thing to point out which is just a mistypo i believe:
the module being referred to is "module: elb_lb_application" while the module file is "elb_application_lb.py"
please ignore, if this has already been corrected.

When using ALB module i keep getting the below error, am i missing something ??
`"module_name": "elb_application_lb"}, "msg": "An error occurred (InternalFailure) when calling the CreateLoadBalancer operation (reached max retries: 4): Unknown", "response_metadata": {"http_headers": {"connection": "close", "content-length": "241", "content-type": "text/xml", "date": "Fri, 02 Dec 2016 06:44:37 GMT", "x-amzn-requestid": "cb188da0-b85a-11e6-8427-e326c3aae2ba"}, "http_status_code": 500, "max_attempts_reached": true, "request_id": "cb188da0-b85a-11e6-8427-e326c3aae2ba", "retry_attempts": 4}}`

linuxdynasty:
Only suggestion I have, is to use the AWSRetry.backoff() decorator for all of the boto3 connections.

gundalow:
This PR has been moved to https://github.com/ansible/ansible/issues/19492 reviews welcome there

PR #2990, opened 2016-09-20
win_share module now does not attempt to change permissions

adbertram:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

win_share
##### ANSIBLE VERSION

```
ansible 2.1.1.0
  configured module search path = ['/home/vagrant/ansibleinventory/library']

```
##### SUMMARY

Previously, the win_share module would attempt to change various things even if the parameters were not used. This was not only slowing it down but also marking it as changed when nothing had changed.

```

```

…ns defined

The module now only attempts to change various options only if the parameter is defined.


gregdek:

PR #2994, opened 2016-09-21
Create os_keystone_endpoint.py

mlabarre:
##### ISSUE TYPE

- New Module Pull Request
  ##### COMPONENT NAME
  
  os_keystone_endpoint
  ##### ANSIBLE VERSION
  

```
ansible 2.2.0
```
##### SUMMARY

Create / modify and delete endpoints

```

```

### DIFF
--- a/cloud/openstack/os_keystone_endpoint.py
+++ b/cloud/openstack/os_keystone_endpoint.py
@@ -50,7 +50,7 @@ options:
      required: true
    interface:
      description:
-        - Endpoint type : ie 'public', 'internal', 'admin' (v3 format).
+        - Endpoint type (ie 'public', 'internal', 'admin') (v3 format).
      required: false
    url:
      description:
--- a/cloud/openstack/os_keystone_endpoint.py
+++ b/cloud/openstack/os_keystone_endpoint.py
@@ -1,0 +1,334 @@
+#!/usr/bin/python
+
+# Copyright (c) 2016 Helicom
+#
+# This module is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This software is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this software.  If not, see <http://www.gnu.org/licenses/>.
+
+try:
+    import shade
+    HAS_SHADE = True
+except ImportError:
+    HAS_SHADE = False
+
+from distutils.version import StrictVersion
+
+
+DOCUMENTATION = '''
+---
+module: os_api_endpoint
+short_description: Create, update or delete API endpoints.
+extends_documentation_fragment: openstack
+author: "Michel Labarre (@mlabarre)"
+version_added: "2.2"
+description:
+    - Add, update or delete API endpoints.
+    - V2.0 and V3 API version are supported. So arguments as interface,
+      url, public_url, admin_url and internal_url are supported. But
+        - for V2 version, only public_url, internal_url and admin_url
+          can be specified (not interface, neither url).
+        - for V3 version, only interface AND url can be specified
+          (no public_url, internal_url and admin_url).
+options:
+   service_name:
+     description:
+        - Name for the target service (ie 'keystone', etc.).
+     required: true
+   service_type:
+     description:
+        - type for the target service (ie 'identity', etc.).
+     required: true
+   interface:
+     description:
+        - Endpoint type : ie 'public', 'internal', 'admin' (v3 format).
+     required: false
+   url:
+     description:
+        - Endpoint URL (v3 format).
+     required: false
+   public_url:
+     description:
+        - Public endpoint URL (v2 format).
+     required: true
+   internal_url:
+     description:
+        - Internal endpoint URL (v2 format).
+     required: false
+   admin_url:
+     description:
+        - Admin endpoint URL (v2 format).
+     required: false
+   region:
+     description:
+        - Region name or id.
+     required: false
+   enabled:
+     description:
+        - Enable endpoint.
+     required: false
+     default: true
+   state:
+     description:
+       - Should the endpoint be present or absent on the user.
+     choices: [present, absent]
+     default: present
+requirements:
+    - "python >= 2.6"
+    - "shade"
+'''
+
+EXAMPLES = '''
+# Some examples contain a 'cloud' attribute.
+# This attribute is not required if you work always with the same API version.
+
+# Add/Update an endpoint internal url with v3 version
+- os_keystone_endpoint:
+    cloud: cloud_v3_api
+    service_name: keystone
+    service_type: identity
+    interface: internal
+    region: myregion
+    enabled: True
+    url: "http://server:5000/v3"
+    state: present
+
+# Add all endpoints public, admin, internal
+- os_keystone_endpoint:
+    cloud: cloud_v3_api
+    service_name: keystone
+    service_type: identity
+    interface: "{{ item.interface }}"
+    url: "{{ item.interface }}"
+    region: "{{ item.region }}"
+    enabled: "{{ item.enabled }}"
+    state: present
+  with_items:
+    - { interface: 'public', url: 'http://server:5000',
+        region: 'myregion', enabled: true }
+    - { interface: 'internal', url: 'http://server:5000',
+        region: 'myregion', enabled: true }
+    - { interface: 'admin', url: 'http://server:35357',
+        region: 'myregion', enabled: true }
+
+# Add/Update an endpoint with v2.0 version
+- os_keystone_endpoint:
+    cloud: cloud_v2_api
+    service_name: keystone
+    service_type: identity
+    public_url: "http://server:5000/v2.0"
+    internal_url: "http://server:5000/v2.0"
+    admin_url: "http://server:35357/v2.0"
+    region: myregion
+    state: present
+
+# Remove an endpoint with v3 version
+- os_keystone_endpoint:
+    cloud: cloud_v3_api
+    service_name: keystone
+    service_type: identity
+    interface_type: internal
+    url: "http://internal:5000"
+    region: myregion
+    state: absent
+
+# Remove an endpoint with v2 version
+- os_keystone_endpoint:
+    service_name: keystone
+    service_type: identity
+    public_url: "http://server:5000/v2.0"
+    internal_url: "http://server:5000/v2.0"
+    admin_url: "http://server:35357/v2.0"
+    region: myregion
+    state: absent
+
+'''
+
+RETURN = '''
+#
+'''
+
+def get_endpoint(service, interface):
+    """
+    Retrieve existing endpoint(s) depending on service and interface values.
+    :param service: Service id.
+    :param interface: Interface (will be None with v2.0).
+    :return: Endpoint entry.
+    """
+    if interface:
+        endpoints_list = cloud.search_endpoints(
+            filters={'service_id': service, 'interface': interface})
+    else:
+        endpoints_list = cloud.search_endpoints(
+            filters={'service_id': service})
+    if len(endpoints_list) > 1:
+        module.fail_json(msg='Multiple endpoint entries exist service %s.'
+                             % service)
+    return endpoints_list[0] if len(endpoints_list) == 1 else None
+
+
+def main():
+
+    global cloud, module
+
+    argument_spec = openstack_full_argument_spec(
+        service_name=dict(required=True),
+        service_type=dict(required=True),
+        interface=dict(required=False),
+        url=dict(required=False),
+        admin_url=dict(required=False),
+        public_url=dict(required=False),
+        internal_url=dict(required=False),
+        region=dict(required=False),
+        enabled=dict(default=True, required=False),
+        state=dict(default='present', choices=['absent', 'present'])
+    )
+
+    module_kwargs = openstack_module_kwargs(
+        required_one_of=[
+            ['url', 'admin_url', 'internal_url', 'public_url']
+        ],
+        mutually_exclusive=[
+            ['url', 'admin_url'],
+            ['url', 'internal_url'],
+            ['url', 'public_url'],
+        ],
+        required_together=[['interface', 'url']]
+    )
+
+    module = AnsibleModule(argument_spec,
+                           supports_check_mode=True,
+                           **module_kwargs)
+    # role grant/revoke API introduced in 1.11.0
+    if not HAS_SHADE or \
+            (StrictVersion(shade.__version__) < StrictVersion('1.11.0')):
+        module.fail_json(msg='shade 1.11.0 or higher is required for this module')
+
+    try:
+
+        cloud = shade.operator_cloud(**module.params)
+
+        service_name = module.params.get('service_name', None)
+        service_type = module.params.get('service_type', None)
+        interface = module.params.get('interface', None)
+        url = module.params.get('url', None)
+        region = module.params.get('region', None)
+        enabled = "True" == module.params.get('enabled', True)
+        state = module.params.get('state', None)
+        admin_url = module.params.get('admin_url', None)
+        internal_url = module.params.get('internal_url', None)
+        public_url = module.params.get('public_url', None)
+
+        services_list = cloud.search_services(
+            filters={'name': service_name, 'type': service_type})
+        if len(services_list) != 1:
+            module.fail_json(
+               msg='Service %s not found or more than 1 service have this name'
+                    % service_name)
+        # Only service id is capable to identify unique entry.
+        service_id = services_list[0].id
+
+        current_endpoint = get_endpoint(service_id, interface)
+
+        if current_endpoint is None and state == 'absent':
+            module.exit_json(changed=False)
+
+        # Check if changes.
+        changed = False
+        if current_endpoint is None:
+            changed = True
+        else:
+            if url and (enabled is not None
+                        and current_endpoint.enabled != enabled):
+                changed = True
+            if hasattr(current_endpoint, 'region') \
+                    and current_endpoint.region != region or \
+                    not hasattr(current_endpoint, 'region') \
+                    and region is not None:
+                changed = True
+            if url:
+                if current_endpoint.url != url:
+                    changed = True
+            else:
+                if hasattr(current_endpoint, 'internalurl') \
+                        and current_endpoint.internalurl != internal_url \
+                        or not hasattr(current_endpoint, 'internalurl') \
+                        and internal_url is not None:
+                    changed = True
+                if hasattr(current_endpoint, 'adminurl') \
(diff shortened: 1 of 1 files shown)


mlabarre (commented):
Doc correction

gregdek:
Thanks @mlabarre. To the current maintainers, @emonty, @shrews, @juliakreger, @j2sol, @rcarrillocruz please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

cloud/openstack/os_keystone_endpoint.py
```diff
@@ -0,0 +1,334 @@
+#!/usr/bin/python
+
+# Copyright (c) 2016 Helicom
```
omgjlk, line 3: Can we get a contact information on this, in case of future relicensing requests?

cloud/openstack/os_keystone_endpoint.py
```diff
@@ -0,0 +1,334 @@
...
+author: "Michel Labarre (@mlabarre)"
+version_added: "2.2"
+description:
+    - Add, update or delete API endpoints.
+    - V2.0 and V3 API version are supported. So arguments as interface,
+      url, public_url, admin_url and internal_url are supported. But
+        - for V2 version, only public_url, internal_url and admin_url
+          can be specified (not interface, neither url).
+        - for V3 version, only interface AND url can be specified
+          (no public_url, internal_url and admin_url).
```
omgjlk, line 41: Is it not possible to just take in URL and interface as module arguments, and then internally or through shade decide how to represent the data to the API? The input to the module shouldn't change. The admin still has to provide a URL and some indication of what type of URL it is. I don't think the arguments to the ansible module should change depending on API version.

omgjlk (commented):
Not blocking this, as I haven't read through the rest, but I'd rather get questions answered before spending more time reading the code.

mlabarre:
Hello
Yesterday I've post a request for the same module having the following signature
`# Add endpoints for keystone/identity (with V3 API)
- os_api_endpoint:
  cloud: cloud_v3
  auth:
    token: "{{ admin_token }}"
    endpoint: "{{ admin_url_v3 }}"
  service_name: keystone
  service_type: identity
  endpoints:
  - { interface: "public", url: "http://server:5000/v2.0", enabled: true, region: "myregion"}
  - { interface: "internal", url: "http://server:5000/v2.0", enabled: true, region: "myregion" }
  - { interface: "admin", url: "http://server:35357/v2.0", enabled: true, region: "myregion"}
    state: present
    `
    Module refused because I test the API version in code and this kind would be executed by Shade. For admin it was transparent.
    I modified the module to be more agnostic.
    Now, you ask if it's possible to transform publicurl, etc into interface and url. I'm a little lost...
    The module presented yesterday allowed an interface list that allowed in the case of a v2 treat the endpoint as a whole.
    I can redo a third module that accepts only one lot of attributes : interface / url / region [/ enabled] but we must allow me to work with versions in the module to treat all CRUD.
    Thanks

mlabarre:
Format is better below:

Yesterday I've post a request for the same module having the following signature

```
# Add endpoints for keystone/identity (with V3 API)
- os_api_endpoint:
    cloud: cloud_v3
    auth:
      token: "{{ admin_token }}"
      endpoint: "{{ admin_url_v3 }}"
    service_name: keystone
    service_type: identity
    endpoints:
      - { interface: "public", url: "http://server:5000/v2.0", enabled: true, region: "myregion"}
      - { interface: "internal", url: "http://server:5000/v2.0", enabled: true, region: "myregion" }
      - { interface: "admin", url: "http://server:35357/v2.0", enabled: true, region: "myregion"}
    state: present

```

Module refused because I test the API version in code and this kind would be executed by Shade. For admin it was transparent.
I modified the module to be more agnostic.
Now, you ask if it's possible to transform publicurl, etc into interface and url. I'm a little lost...
The module presented yesterday allowed an interface list that allowed in the case of a v2 treat the endpoint as a whole.
I can redo a third module that accepts only one lot of attributes : interface / url / region [/ enabled] but we must allow me to work with versions in the module to treat all CRUD.
Thanks

omgjlk:
I believe @SamYaple objected to the code _inside_ the module that was doing v2/v3 detection and logic. He was not objecting to the module input signature, but I'll let him confirm this.

cloud/openstack/os_keystone_endpoint.py
ghost: Correct j2sol, this should be using ansibles mutually exclusive options. The options can be either public_url, admin_url, internal_url OR interface and url. Both of these types work with either v2.0 or v3. There is no version dependency on making them work.

cloud/openstack/os_keystone_endpoint.py
```diff
@@ -0,0 +1,334 @@
...
+     description:
+        - Name for the target service (ie 'keystone', etc.).
+     required: true
+   service_type:
+     description:
+        - type for the target service (ie 'identity', etc.).
+     required: true
+   interface:
+     description:
+        - Endpoint type (ie 'public', 'internal', 'admin') (v3 format).
```
ghost, line 53: all references to "v3 format" or "v2 format" should be removed

cloud/openstack/os_keystone_endpoint.py
```diff
@@ -0,0 +1,334 @@
...
+    public_url: "http://server:5000/v2.0"
+    internal_url: "http://server:5000/v2.0"
+    admin_url: "http://server:35357/v2.0"
+    region: myregion
+    state: absent
+
+'''
+
+RETURN = '''
+#
```
ghost, line 157: commenting so its not lost, im sure you are just waiting, but this should be filled out before merge

cloud/openstack/os_keystone_endpoint.py
```diff
@@ -0,0 +1,334 @@
...
+            filters={'service_id': service})
+    if len(endpoints_list) > 1:
+        module.fail_json(msg='Multiple endpoint entries exist service %s.'
+                             % service)
+    return endpoints_list[0] if len(endpoints_list) == 1 else None
+
+
+def main():
+
+    global cloud, module
```
ghost, line 181: please do not use globals

cloud/openstack/os_keystone_endpoint.py
```diff
@@ -0,0 +1,334 @@
...
+            ['url', 'internal_url'],
+            ['url', 'public_url'],
+        ],
+        required_together=[['interface', 'url']]
+    )
+
+    module = AnsibleModule(argument_spec,
+                           supports_check_mode=True,
+                           **module_kwargs)
+    # role grant/revoke API introduced in 1.11.0
```
ghost, line 211: this comment is not accurate, specifically it should be that update was added in 1.11.0 and v3 support was fully implemented. But i dont think the comment is needed at all

cloud/openstack/os_keystone_endpoint.py
```diff
@@ -0,0 +1,334 @@
...
+    if not HAS_SHADE or \
```
ghost, line 212: For consistency, please have this section read:

```
if not HAS_SHADE:
    module.fail_json(msg='shade is required for this module')
if StrictVersion(shade.__version__) < StrictVersion('1.11.0'):
    module.fail_json(msg="To utilize this module, the installed version of"
                         "the shade library MUST be >=1.11.0")
```

cloud/openstack/os_keystone_endpoint.py
```diff
@@ -0,0 +1,334 @@
...
+    argument_spec = openstack_full_argument_spec(
+        service_name=dict(required=True),
+        service_type=dict(required=True),
+        interface=dict(required=False),
+        url=dict(required=False),
+        admin_url=dict(required=False),
+        public_url=dict(required=False),
+        internal_url=dict(required=False),
+        region=dict(required=False),
+        enabled=dict(default=True, required=False),
```
ghost, line 192: set type='bool'

cloud/openstack/os_keystone_endpoint.py
```diff
@@ -0,0 +1,334 @@
...
+    try:
+
+        cloud = shade.operator_cloud(**module.params)
+
+        service_name = module.params.get('service_name', None)
+        service_type = module.params.get('service_type', None)
+        interface = module.params.get('interface', None)
+        url = module.params.get('url', None)
+        region = module.params.get('region', None)
+        enabled = "True" == module.params.get('enabled', True)
```
ghost, line 225: the bool type if set correctly will prevent the need to do this

cloud/openstack/os_keystone_endpoint.py
```diff
@@ -0,0 +1,334 @@
...
+            module.fail_json(
+               msg='Service %s not found or more than 1 service have this name'
+                    % service_name)
+        # Only service id is capable to identify unique entry.
+        service_id = services_list[0].id
+
+        current_endpoint = get_endpoint(service_id, interface)
+
+        if current_endpoint is None and state == 'absent':
+            module.exit_json(changed=False)
```
ghost, line 243: See the functions named '_system_state_change' and '_needs_update' in the other ansible shade modules and how those are structured. Please implement that here (rather than in the main() function)

cloud/openstack/os_keystone_endpoint.py
```diff
@@ -0,0 +1,334 @@
...
+                    # So we can only use interface, url and region
+                    if url:
+                        cloud.update_endpoint(current_endpoint.id,
+                                              url=url,
+                                              interface=interface,
+                                              region=region)
+                    else:
+                        # No update possible in V2.0. So we must delete
+                        # and recreate.
+                        cloud.delete_endpoint(current_endpoint.id)
```
ghost, line 311: This prevents the module from being idempotent. We cannot do this. Since this is a module that could only be used by cloud operators it is expected they should know if they are using v2.0 and v3, and the appropriate capabilities. Therefore, if shade returns and error letting them know the cloud doesnt support this feature, it should be passed to the user.

cloud/openstack/os_keystone_endpoint.py
```diff
@@ -0,0 +1,334 @@
...
+
+        args = {}
+        for endpoint in ["public_url", "internal_url", "admin_url"]:
+            args[endpoint] = module.params.get(endpoint)
+
+        # Do changes.
+        if changed:
+            if state == "present":
+                if current_endpoint is None:
+                    cloud.create_endpoint(service_id,
```
ghost, line 292: due to the way shade works, it is safe to pass interface, url, public_url, admin_url, and internal_url for each request, as long as they are set to None. That just means the logic for mutually_exclusive bits must be handled appropriately in the ansible module and then we can safely pass them all through.

cloud/openstack/os_keystone_endpoint.py
```diff
@@ -0,0 +1,334 @@
...
+
+        if current_endpoint is None and state == 'absent':
+            module.exit_json(changed=False)
+
+        # Check if changes.
+        changed = False
+        if current_endpoint is None:
+            changed = True
+        else:
+            if url and (enabled is not None
```
ghost, line 250: this logic is not correct since you can pass interface+url to a v2.0 endpoint through shade (shade handles that on the backed). Therefore it is not safe to assume just because the url is set this will be a v2.0 endpoint like you appear to be

ghost (changes requested):

mlabarre:
ok 
Now a simple question.
Shade accept all arguments for an endpoint creation : interface, url, publicurl and so.
If you supply publicurl, internalurl, adminrul in V2 with state=present, you create it
If you supply publicurl, internalurl, adminrul in V3 with state=present, you create it
If you supply publicurl, internalurl, adminrul in V2 with state=absent, you delete it
If you supply publicurl, internalurl, adminrul in V3 with state=absent, you do nothing because search filters does not returns endpoints.
In this logic/semantics, the module will have to interpret nothing and on the other he has to.
Permit all arguments on creation is very confuse...
Another thing. What that really represents "absent" relating, ie, internalurl ? the presence of internalurl in endpoint or its value ? Great discussion which my customer !

ghost:
Deletes are done with the endpoint id for a reason. You must supply an id to shade to delete an endpoint.

v2 create with the *_url variables returns a list with a single endpoint
v3 create with the *_url variables returns a list with three endpoints

v2 delete is the same as v3, the difference in the ansible module would be it would delete 3 endpoints on v3 and only a single endpoint on v2.

As with all the other modules, if an endpoint search reveals _multiple_ matches, we must abort and tell the user to fix that.

mlabarre:
"v2 delete is the same as v3, the difference in the ansible module would be it would delete 3 endpoints on v3 and only a single endpoint on v2."

Absolutely no. in V3 if when you supply an interface with state = "absent" you would delete only this interface, not the endpoint. And you have not necessary 3 endpoints in V3.
And to do that :
. you indicate interface (search returns empty with V2)
. you indicate ie publicurl (search returns empty with V3)
So in the module, when i have a state "absent" all is relative to endpoint(s) found by the search request which use input arguments. 
So, for me, search must be done on service id only. So we have 0 to 3 entries in list.
1 entry does not say 'V2' since I can have ie simply public url on my V3 endpoint.
So I must analyze the shade return and scan input arguments (interface/url which does not say necessary 'V3' and public/admin/internal url  which does not say necessary 'V2') to try a correspondance.
Ouah !
I use identity v2 and v3 and I want only one version of endpoint module !

ghost:
The return value from shade being normalized for v2.0 and v3 endpoints will be happening, though that doesn't directly influence the flow of this module.

When you specify v2.0 interface+url and set state to absent, then absolutely it will work the same for v2 and v3. With v2.0, a single endpoint has public, admin, and internal_url. But it doesn't have a hard requirement on having all of those set (infact, the only requirement is having public_url set in v2.0). Therefore there will not be a valid match and no change will occur on v2.0 endpoint.

All of this logic is currently baked into shade. It would simply be a matter of properly searching in the ansible module. Remember, the only way to delete is to pass an endpoint id. Searching will be the responsibility of the module.

I have been unfortunately busy in the past few weeks so I have no submitted this PR myself, but if you would like me to submit a PR for this module to show you the logic I can.

mlabarre:
As I said before the creation of the endpoint is not a problem (shade 
determines itself what it needs), the most important action is the 
search for the endpoint on which we must act to modify or delete it .
When the interface / url arguments are provided, we found (or not) the 
endpoint with ServiceId + interface (v3)

Lorsque les arguments publicurl, etc. sont fournis (totalement ou 
partiellement) soit ils concernent un endpoint v2 dans tel cas on ne 
peut le retrouver qu'avec le serviceid (search avec publicurl ne renvoit 
rien). S'ils concernent un v3 (à mon sens c'est une mauvaise 
codification) on peut le retrouver en ajoutant au filtre l'interface 
avec 'public' comme valeur (par exemple) si l'argument est public_url.

When publicurl arguments, etc. are provided (all or part thereof) :
  . if they relate to a v2 endpoint (in this case we can not find them 
with the ServiceId (search publicurl returns with nothing).
  . if they concern an v3 (I think it is a bad coding) can be found by 
adding the filter interface with 'public' as a value (for example) if 
the argument is public_url.

Now suppose a E1 endpoint created in v3:
    -service: S1 (in the test it is the equivalent of the serviceId = 
Service name + type)
    -public = P1
    -internal = I1

and E2 endpoint created in v2:
    -service: S2 (in the test it is the equivalent of the serviceId = 
Service name + type)
    -public = P2
    -internal = I2
1. execute with v2 API:

test 1.1
- os_keystone_api
  Service: S1
  public_url: X
  internal_url: I1
  state: present
  > search with filter = serviceId. It returns one endpoint with 
  publicurl, InternalURL -> X = P1 -> changed -> update: Shade causes an 
  exception

test 1.2
- os_keystone_api
  Service: S1
  interface: admin
  url: A1
  state: present
  > search with filter = serviceId. It returns one endpoint with 
  publicurl, InternalURL -> changed -> update: Shade causes an exception

test 1.3
- os_keystone_api
  Service: S1
  public_url: X
  internal_url: I1
  state: absent
  > search with filter = serviceId. It returns one endpoint with 
  publicurl, InternalURL -> Delete while X = P1?

test 1.4
- os_keystone_api
  Service: S1
  interface: internal
  url: I1
  state: absent
  > search with filter = serviceId. It returns one endpoint with 
  publicurl, InternalURL -> Delete?

test 1.5
- os_keystone_api
  Service: S2
  public_url: X
  internal_url: I2
  state: present
  > search with filter = serviceId. It returns one endpoint with 
  publicurl, InternalURL -> X = P1 -> changed -> update: Shade causes an 
  exception

test 1.6
- os_keystone_api
  Service: S2
  public_url: X
  internal_url: I2
  state: absent
  
  > search with filter = serviceId. It returns only one endpoint with 
  > publicurl, InternalURL but the id of the endpoint is the public_url -> 
  > Delete while X = P2?
  >     If we remove, we remove all interfaces.
1. It runs with API v3:

2.1 test
- os_keystone_api
  Service: S1
  public_url: X
  internal_url: I1
  state: present
  > search with filter = serviceId. It returns 2 endpoints with 
  interface / url -> X = P1 -> changed -> update: public update ok

2.2 test
- os_keystone_api
  Service: S1
  interface: admin
  url: A1
  state: present
  > search with filter = serviceId + interface. It returns 1 endpoints 
  with interface / url -> changed -> update: add admin ok

2.3 test
- os_keystone_api
  Service: S1
  public_url = X
  internal_url = I1
  state: absent
  > search with filter = serviceId. It returns 2 endpoints with 
  interface / url -> Delete public_url while X = P1?  Delete internal_url ok.

test 2.4
- os_keystone_api
  Service: S1
  interface: internal
  url: I1
  state: absent
  > search with filter = serviceId + interface. It returns with one 
  endpoint interface / url -> delete internal ok.

test 2.5
- os_keystone_api
  Service: S2
  public_url = X
  internal_url I2 =
  state: present
  > search with filter = serviceId. It returns 2 endpoints with 
  interface / url -> X = P1 -> changed -> update: public url ok

test 2.6
- os_keystone_api
  Service: S2
  public_url = X
  internal_url I2 =
  state: absent
  > search with filter = serviceId. It returns 2 endpoints with 
  internal / url -> X = P2 -> delete public? delete internal ok.

With the good responses, module would be ok.

ghost:
Keystone v2.0 endpoints cannot be updated. So if a playbook is run that tries to update v2.0 endpoints, it _should_ fail. I don't agree with deleting and recreating the endpoints for v2.0 (nor do we do that in other modules with a similar problem).

In your v2.0 examples, you have several where because it returns a single endpoint, you use it outright, but you should not be. If the endpoint doesn't match what you have, then you should not be using it. In this case, you should be ensuring that the public + admin + internal url all match what the ansible module was given. In the case of passing interface + url to v2.0 api, there is only one time that that will ever match with ansible, and that is when passing the public interface option since the public interface is required for v2.0. It should not be matching internal or admin url individually.

cloud/openstack/os_keystone_endpoint.py
```diff
@@ -0,0 +1,334 @@
...
+    HAS_SHADE = True
+except ImportError:
+    HAS_SHADE = False
+
+from distutils.version import StrictVersion
+
+
+DOCUMENTATION = '''
+---
+module: os_api_endpoint
```
emonty, line 29: os_keystone_endpoint

emonty (changes requested):
I agree with @SamYaple in his reviews.

PR #3010, opened 2016-09-22
Corrects an implied map() usage as list

caphrim007:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

network/f5/bigip_facts.py
##### ANSIBLE VERSION

```
ansible 2.2.0
  config file = 
```
##### SUMMARY

In the six package, the map() function returns an iterator instead
of a list. This code was continuing to use the map() return value
as if it were a list and this broke the address_class facts.

This patch changes the code to use the list() method on the return
value of map().


caphrim007:
@mhite @gundalow more bug files for 2.2, ready_for_review

mhite:
shipit

gundalow:
Merged, thanks for the PR and quick testing.

PR #3038, opened 2016-09-26
zfs: Don't set property on empty values. i.e. when set to None.

techmunk:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

zfs
##### ANSIBLE VERSION

```
ansible 2.1.1.0
```
##### SUMMARY

When setting a ZFS property in ansible to None (i.e. you're using a loop to set options on several ZFS mounts), ZFS will error with `missing value in property=value argument`.
This patch allows setting a property to None to effectively have it ignored, and not error.

### DIFF
--- a/system/zfs.py
+++ b/system/zfs.py
@@ -184,7 +184,7 @@ class Zfs(object):
     def set_properties_if_changed(self):
         current_properties = self.get_current_properties()
         for prop, value in self.properties.iteritems():
-            if current_properties.get(prop, None) != value:
+            if value != '' and current_properties.get(prop, None) != value:
                 self.set_property(prop, value)
 
     def get_current_properties(self):


gregdek:
Thanks @techmunk. To the current maintainers, @johanwiren please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

johanwiren:
I think the error message you get is correct. The zfs module does not do validation of the properties, but leaves that up to the zfs command.

Setting a property to 'none' is supported and used for re-setting quotas etc., however None will be passed as a string and fail as you noticed.

Ansible does have support for omitting parameters to a module (http://docs.ansible.com/ansible/playbooks_filters.html#omitting-parameters), and from what I can tell that would probably suit your use-case. If not, please provide a sample playbook that fails.

techmunk:
Thanks. I didn't know about the omit option. It solves my use case perfectly.

PR #3044, opened 2016-09-27
ovirt_vms: wait for VM to be created for states stopped/suspended

machacekondra:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

ovirt_vms
##### ANSIBLE VERSION

```
2.2
```
##### SUMMARY

This patch add wait for creation of VM in case state is stopped or suspended and VM didn't existed before. The VM can be in image_locked state, so we need to wait for VM to be in DOWN state.

```

```


machacekondra:
@ryansb

PR #3069, opened 2016-09-29
Add Ability to Exlude Paths from Archive

rjsab:
##### ISSUE TYPE

- Feature Pull Request
##### COMPONENT NAME

files/archive.
##### ANSIBLE VERSION

```
ansible 2.1.1.0
  config file = 
```
##### SUMMARY

The archive module allows you to pass a glob to the path argument.
This feature allows end users to continue to pass a glob to the path argument,
but enables a new argument to exclude specific dirs, or a glob of dirs in order to prevent
unwanted paths from being archived. The idea is to enable similar functionality as the
`--exclude=PATTERN` option for the `tar` command.


gregdek:
Thanks @rjsabatini. To the current maintainers, @bendoh please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

files/archive.py
```diff
@@ -158,6 +186,22 @@ def main():
...
+    # Only attempt to expand the exclude paths if it exists
+    if exclude_paths:
+        for i, exclude_path in enumerate(exclude_paths):
+            exclude_path = os.path.expanduser(os.path.expandvars(exclude_path))
+
+            # Expand any glob characters. If found, add the expanded glob to the
+            # list of expanded_paths, which might be empty.
+            if ('*' in exclude_path or '?' in exclude_path):
+                expanded_exclude_paths = expanded_exclude_paths + glob.glob(exclude_path)
+                globby = True
```
bendoh: I don't think we want globby to be set here, since it will force the output as an archive, even if you specified a single path to compress.

I.e., 

`archive: path=/a/single/file.txt exclude_paths=/globby/path/*`

Will cause this to fail as `archive` will get set on L166, but `dest` is not defined so there will be a failure on line 175.

I think just removing this line is fine – `globby` indicates that we _want_ an archived file, but the usage above does not indicate that.

bendoh (changes requested):
Please kill `globby = True` on L198 – Otherwise this looks great!

bendoh:
needs_revision

rjsab:
I've updated to remove `globby = True` from L198. Let me know if you have further suggestions, I'm more than happy to make modifications where needed.

rjsab:
I noticed the test failed, do I need to update L31 `version_added: 2.2` to `version_added: 2.3`?

bendoh:
I think the version_added flag needs to be added to the parameter definition itself, after L54.

The module was added in 2.2 and that value on L31 shouldn't change.

rjsab:
Looks like adding it to that section worked. Thanks!

bendoh:
shipit

haad:
shipit

gregdek:
Thanks again to @rjsabatini for this PR, and thanks @bendoh for reviewing. Marking for inclusion.

PR #3073, opened 2016-10-01
Add module "launchd" to manage Mac OS X services.

bjoernalbers:
##### ISSUE TYPE

New Module Pull Request
##### COMPONENT NAME

osx_service
##### ANSIBLE VERSION

```
ansible 2.1.1.0
  config file = 
```
##### SUMMARY

This module allows to start, stop and restart Mac OS X launchd daemons
via `launchctl`.
It supports system services (stuff that has a plist under
/Library/LaunchDaemons) and must be run with root permissions.

About the implementation...

Service management on Mac OS X is a different horse and deserves its own module.
I build the first draft in Ruby: [osx_service.rb](https://gist.github.com/bjoernalbers/f40b3fc17067959b9e4663be5021c1be).
But with the help of `AnsibleModule` I was able to port it to Python so that others could use this as well.

Here is some example output...

```
TASK [stop crashplan] **********************************************************
changed: [localhost] => {"changed": true, "name": "com.crashplan.engine", "state": "stopped"}
```

Thanks in advance,
Björn

### DIFF
--- a/system/launchd.py
+++ b/system/launchd.py
@@ -30,7 +30,7 @@ author:
 version_added: '2.2'
 short_description: Manage Mac OS X launchd services.
 description:
-    - Manage (start, stop & restart) Mac OS X launchd services via launchctl.
+    - Manage (start, stop and restart) Mac OS X launchd services via launchctl.
 options:
     name:
         required: true
@@ -39,7 +39,7 @@ options:
         aliases: [ 'service', 'label' ]
     state:
         required: true
-        choices: [ 'started', 'loaded', 'stopped', 'unloaded', 'restarted', 'reloaded ]
+        choices: [ 'started', 'loaded', 'stopped', 'unloaded', 'restarted', 'reloaded' ]
         description:
             - C(started), C(loaded), C(stopped) and C(unloaded) are idempotent
               actions that will not run commands unless necessary.
--- a/system/launchd.py
+++ b/system/launchd.py
@@ -1,6 +1,8 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
-
+#
+# (c) 2016, Brian Coca <bcoca@ansible.com>
+# (c) 2106, Theo Crevon (https://github.com/tcr-ableton)
 # (c) 2016, Björn Albers <bjoernalbers@gmail.com>
 #
 # This file is part of Ansible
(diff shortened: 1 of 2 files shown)


bjoernalbers:
CI tells "ERROR: No RETURN documentation provided" but documentation _is_ included?

gregdek:
Thanks @bjoernalbers for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

bjoernalbers:
@gregdek this module is ready_for_review :-)

gregdek:
Thanks @bjoernalbers for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

bcoca:
it seems we have duplicate efforts https://github.com/ansible/ansible-modules-core/pull/3694

bjoernalbers:
@bcoca thanks for your "launchd" module and for pointing this out.

Looks like "launchd" has stronger documentation and "osx_service" a slightly better error handling (`launchctl` might return 0 even on errors).
Should I merge these into one module and mention us three - @bcoca @tcr-ableton and me - as developers?

Here me suggestions about the API / interface.
I'd be glad to hear your ideas and opinions on this.

### Module name: launchd vs. osx_service

**I'd be okay with both and follow your opinion.**
"launchd" better reflects the underlying system while "osx_service" shows the similarty to the "service" module.

### System daemons vs. user daemons

System daemons are controlled when running `launchctl` as root.
When controlling user daemons `launchctl` has to run as that user.

The new implemenation of `launchd` makes this more explicit which is a good thing.
But it's not ready yet and that's why [people](https://github.com/Homebrew/legacy-homebrew/issues/33259) keep using the "legacy subcommands".

**So let us limit the focus on system daemons which is a good default, ok?**
When the new launchd API is stable we can easily expand to controlling user services... and save us from wasting time on workarounds.

### Option "name"

Good idea with the "service" alias!
**Would make an additional "label" alias make sense since this is official launchd vocabulary?**

### Option "enabled"

I guess almost all users would like to enable services they control with ansible.
This key is also stored externally by `launchd` and might differ with the plist:

> ```
> Disabled <boolean>
> This optional key specifies whether the job should be loaded by default. Note that this key may be overridden through the enable subcommand of > launchctl(3).  Previous
> Darwin operating systems would modify the configuration file's value for this key, but now this state is kept externally.
> ```
> 
> -- `man launchd.plist`

**That's why I suggest to leave this option out in order to simplify thinks.**
Then we can just always load/unload with option "-w":

> ```
> -w Overrides the Disabled key and sets it to false or true for the load and
>    unload subcommands respectively. In previous versions, this option would modify
>    the configuration file. Now the state of the Disabled key is stored elsewhere
>    on disk in a location that may not be directly manipulated by any process other
>    than launchd.
> ```
> 
> -- `man launchctl`

Another benefit: we don't have to modify plists which can be managed with the template module anyway.

### Option "state"

**I'm gonna add "reloaded" too, ok?**

Thanks in advance for your feedback!

bcoca:
I wrote launchd w/o having access to an OS X machine, I mostly used the man page. I did not know a bout the errors on `rc == 0` (_sigh_), that code should be easy to incorporate.

### Module name: launchd vs. osx_service

Since other service modules 'systemd, runit, svc' follow the 'management system' I lean towards launchd

### System daemons vs. user daemons

fine with me

### Option "name"

+1 to adding 'label' as an alias

### Option "enabled"

I'm not sure what you want to leave out here, we want enabled/disabled as a general rule to indicate a service will start at boot or not

### Option "enabled"

reloaded is fine

bjoernalbers:
Hello Brian,

sorry for my fuzzy description about the "enabled" problem.
To my understanding Mac OS X / launchd has no reliable way to get the current enable/disable state of a service (yet).
launchd plists do have a disabled key but the actual state is stored somewhere else.
So a safe workaround is to make sure that each requested service is enabled.
This would be a useful default anyway.
Otherwise we'd have to restart the service on each ansible run in order to enforce that its enable (probably unwanted by many users).
If a user wants to disable a service he/she has to simply stop it.

I've merged both and did some integration tests.
Could you please have a look and approve if ok?

Thanks

gregdek:
Thanks again to @bjoernalbers for this PR. This PR was reviewed by an Ansible member. Marking for inclusion.

However, we would still appreciate if the maintainer(s) @ansible/core could look over it.

bjoernalbers:
@gregdek Hello Greg, could you please merge this one? Or is another approval from ansible core staff required? If this PR has still issues I will fix them of course. Thanks

gregdek:
Hi @bjoernalbers -- the issue is that we're in feature freeze for 2.2. This module won't be included until 2.3.

bjoernalbers:
PR has been migrated to https://github.com/ansible/ansible/pull/18888

PR #3080, opened 2016-10-03
Improve verbosity and avoids erroneously succeeding

### LINKED ISSUE #3053: Pear module not verbose enough while failing
##### ISSUE TYPE
- Feature Idea
##### COMPONENT NAME

pecl
##### ANSIBLE VERSION

```
ansible 2.1.1.0
  config file = /home/[...]/ansible/ansible.cfg
```
##### CONFIGURATION

```
$ cat ansible.cfg  | grep -v '^$\|^#'
[defaults]
forks           = 100
vault_password_file = ~/.vault.password
[ssh_connection]
pipelining = False
```
##### OS / ENVIRONMENT

```
$ lsb_release -a
No LSB modules are available.
Distributor ID: Debian
Description:    Debian GNU/Linux 8.6 (jessie)
Release:        8.6
Codename:       jessie
```
##### SUMMARY

I asked for the pecl module to install something but my alternatives pointed /usr/bin/php to php7 which make the module fail without reporting why.
##### STEPS TO REPRODUCE

```
  - name: apt-get install from stretch (testing)
    apt: name={{item}} state=latest default_release=stretch
    with_items:
    - php7.0
    - php-pear

  - name: Use php7 as /usr/bin/php
    alternatives: name=php path=/usr/bin/php7.0

  - name: Install stomp PHP module
    pear: name="pecl/stomp" state=present
```

I'm getting (I tried `-vvv`):

```
failed: [dev] (item=pecl/stomp) => {"failed": true, "invocation": {"module_args": {"name": "pecl/stomp", "state": "present"}, "module_name": "pear"}, "item": "pecl/stomp", "msg": "failed to install pecl/stomp"}
```

Which is not enough to tell why it failed, which looks wrong to hide this.

To diagnose I started with `pecl --help` and boom:

```
$ pecl --help

Parse error: syntax error, unexpected 'new' (T_NEW) in /usr/share/php/PEAR/Frontend.php on line 91
```

This is clearly my issue here, so I updated my alternatives to use php5 and I'm good, that's not really an issue.

BUT I'd liked a lot if ansible told me straight that the PECL module gave a `Parse error: syntax error, unexpected 'new' (T_NEW) in /usr/share/php/PEAR/Frontend.php on line 91`.

jle64:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

pear module
##### ANSIBLE VERSION

```
ansible 2.1.1.0
  config file = /home/jonathan/.ansible.cfg
  configured module search path = ['/home/jonathan/dev/ansible/modules']
```
##### SUMMARY

Fixes #3053

This adds verification that packages are really installed to avoid succeeding when package installation failed (instead of just relying on pear exit value which is often 0 even in case of failure), and this adds pear stdout output to the error message to make the cause more explicit.

### DIFF
--- a/packaging/language/pear.py
+++ b/packaging/language/pear.py
@@ -149,8 +149,11 @@ def install_packages(module, state, packages):
         cmd = "pear %s %s" % (command, package)
         rc, stdout, stderr = module.run_command(cmd, check_rc=False)
 
-        if rc != 0:
-            module.fail_json(msg="failed to install %s" % (package))
+        # verify that the package is actually installed
+        # as pear exit value is unreliable
+        installed, updated = query_package(module, package)
+        if rc != 0 or not installed or (state == 'latest' and not updated):
+            module.fail_json(msg="failed to install %s: %s" % (package, stdout))
 
         install_c += 1
 


packaging/language/pear.py
```diff
@@ -149,8 +149,11 @@ def install_packages(module, state, packages):
         cmd = "pear %s %s" % (command, package)
         rc, stdout, stderr = module.run_command(cmd, check_rc=False)
 
-        if rc != 0:
-            module.fail_json(msg="failed to install %s" % (package))
+        # verify that the package is actually installed
+        # as pear exit value is unreliable
+        installed, updated = query_package(module, package)
+        if rc != 0 or not installed or (state == 'latest' and not updated):
```
bcoca, line 155: isn't it possible for state=='latest' and the package not need updating?

jle64: The "updated" variable is true if the latest version of the package is installed. So this seems OK to me, if the required package state is "latest" and the package is not in the latest version, throw an error.

(But I agree that the "updated" variable is wrongly named. It should probably be named up2date or something.)

PR #2938, opened 2016-09-16
Fixes #2736 - dont overwrite undefined parameters

wimnat:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

s3_bucket
##### ANSIBLE VERSION

ansible 2.2.0
##### SUMMARY

The s3_bucket was applying defaults to a bucket even if they were undefined.  This meant that if it conflicted with what was configured in aws then it would make a change.

With this change, if the parameter is not defined by the user, no change will be made in aws.

PR #3091, opened 2016-10-04
Fix failure to apply bucket policy when creating a bucket from scratch

ryansb:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

s3_bucket
##### ANSIBLE VERSION

[devel]
##### SUMMARY

In [this PR](https://github.com/ansible/ansible-modules-extras/pull/2938/files?w=1) the code for applying a bucket policy was wrapped in a conditional that caused all new buckets to never get their policies applied. This change gets rid of that by defaulting `policy` to `None` and only executing when a policy is specified by the user.


gregdek:
Thanks @ryansb. To the current maintainers, @wimnat please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

amenonsen:
@ryansb it still always reports "changed" for me, but at least it doesn't eat the policy as #2938 did. My invocation looks like this:

```
- s3_bucket:
    name: "x"
    region: "eu-west-1"
    state: present
    tags: "{{ sometags }}"
    policy: >
      {
        "Version": "2012-10-17",
        …
      }
```

ryansb:
The reporting of `changed` is a result of something else. [This check](https://github.com/ansible/ansible-modules-extras/pull/3091/files#diff-d1353778df71e40405b333a88cdb9925R212) is a deep equality test of the policy. Here's an example:

I run a play like this:

```
  - s3_bucket:
      name: "{{ username }}-test-bucket"
      profile: "{{ aws_profile }}"
      region: "{{ region }}"
      policy:
        Id: "ObjectDeliveryPolicy"
        Version: "2012-10-17"
        Statement:
          - Action:
              - "s3:PutObject"
            Effect: Allow
            Resource: "arn:aws:s3:::{{ username }}-test-bucket/*"
            Principal:
                AWS: "arn:aws:iam::[redacted]:root"
```

Note specifically that `Action` accepts a list of actions (s3:PutObject). When AWS saves that policy, it accepts that policy just fine. Then, when we describe the policy later this is what comes back:

```
{
  "Id": "ObjectDeliveryPolicy", 
  "Statement": [
    {
      "Action": "s3:PutObject", 
      "Effect": "Allow", 
      "Principal": {
        "AWS": "arn:aws:iam::[redacted]:root"
      }, 
      "Resource": "arn:aws:s3:::ryansb-test-bucket/*"
    }
  ], 
  "Version": "2012-10-17"
}
```

Note that `Action` got changed to a single item, not a list, which causes the equality check to fail and show "changed" even though the policies are logically equivalent. Unfortunately, the only way to have the task stop reporting changed is to change the playbook to send a policy with `s3:PutObject` as a solo string instead of a list.

ryansb:
So this task will sometimes report `changed` incorrectly because AWS munges the policy when it's stored in IAM.

amenonsen:
Thanks for the explanation, that makes sense now. (Based on the comments
on another open issue, I thought the problem was the order of the list
of principals changing, and was puzzled when my single-principal task
also showed the same behaviour. I didn't notice that the action was also
being converted to a string.)

In light of that explanation, your PR looks good to me. Thanks!

ryansb:
Yes, the order of principles would also cause this to report `changed=True`. Doing a semantic equality on these IAM policies is going to be a bit more complex, since some lists don't have order that matters (like principals) and others are valid as both solo objects and lists. A better deep equality is needed, no doubt. Right now we just use Python's builting `dict`deep equality.

ryansb:
This has also been backported to 2.2 and will be in RC2

PR #3099, opened 2016-10-05
fixes problem where wrong ACL could be selected

privateip:
This fixes a bug where the wrong acl name could be matched in the running
config if the desired acl name is a starting subset.


gregdek:
Thanks @privateip for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

PR #3107, opened 2016-10-06
Stress module

bbyhuy:
##### ISSUE TYPE

- New Module Pull Request
##### COMPONENT NAME

stress
##### ANSIBLE VERSION

```
ansible 2.1.2.0
```
##### SUMMARY

Benchmarking is a field which automation is often sought after. By wrapping these benchmark tools (more to come) as Ansible modules, it enables people to quickly launch benchmarks remotely and at scale. Each benchmark module is formatted with essentially the same boilerplate with specific changes to just the parsing and execution portions. As much of the code is kept the same as possible as to allow adding new benchmarks to be quick and painless.

stress is a deliberately simple workload generator for POSIX systems. It imposes a configurable amount of CPU, memory, I/O, and disk stress on the system.

### DIFF
--- a/benchmarks/stress.py
+++ b/benchmarks/stress.py
@@ -23,7 +23,7 @@ short_description: Execute the Stress benchmark
 description:
  - Use this module to run the Stress benchmark
  - https://people.seas.harvard.edu/~apw/stress/
-version_added: "2.1"
+version_added: "2.3"
 options:
   chdir:
     description:
--- a/benchmarks/stress.py
+++ b/benchmarks/stress.py
@@ -1,0 +1,214 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2016, Hugh Ma <Hugh.Ma@flextronics.com>
+#
+# This module is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This software is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this software.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: stress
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @bbyhuy for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

benchmarks/stress.py
```diff
@@ -0,0 +1,214 @@
...
+module: stress
+short_description: Execute the Stress benchmark
+description:
+ - Use this module to run the Stress benchmark
+ - https://people.seas.harvard.edu/~apw/stress/
+version_added: "2.3"
+options:
+  chdir:
+    description:
+     - change working directory
```
haad, line 30: Generally option entries in doc should be written as sentence.

benchmarks/stress.py
```diff
@@ -0,0 +1,214 @@
...
+     - Total time to run stress for.
+    required: True
+requirements:
+ - stress
+author: "Hugh Ma <Hugh.Ma@flextronics.com>"
+'''
+
+EXAMPLES = '''
+# Run stress for 10 seconds on 1 cpu
+- stress: timeout=10 cpu=1
```
haad, line 60: Can you use ansible native YAML syntax here ? it considered best practice for new modules.

benchmarks/stress.py
```diff
@@ -0,0 +1,214 @@
...
+        missing_params = list()
+        for param in ['timeout', 'cpu']:
+            if not module.params[param]:
+                missing_params.append(param)
+        if len(missing_params) > 0:
+            module.fail_json(msg="missing required arguments: {}".format(missing_params))
+        if module.params['executable']:
+            stress_bin = module.params['executable']
+        else:
+            stress_bin = module.get_bin_path('stress', True, ['/usr/bin/local'])
```
haad, line 203: /usr/bin/local looks like very specific path ? can we use standard system PATH in case if path is not specified or maybe /usr/local/bin ?

haad (changes requested):

benchmarks/stress.py
bbyhuy: I believe that parameter is opt_dirs which just adds '/usr/bin/local' to the list of pre-defined paths to search. So I don't think there should be too much of a problem here.

[https://github.com/ansible/ansible/blob/fda933723c8cb60248d41b5691edd86fe0a40807/lib/ansible/module_utils/basic.py#L1723-L1728]

PR #3113, opened 2016-10-07
kubernetes_templating

astropuffin:
##### ISSUE TYPE
- New Module Pull Request
##### COMPONENT NAME

kubernetes
##### ANSIBLE VERSION

Tested a slightly tweaked implementation using the below version, but updated to be compatible with devel.

```
ansible 2.0.0.0
  config file = 
```
##### SUMMARY

Features:
- template files (with variables)
- relative paths (so it can reference ansible directories rather than abs paths)
- token authentication
- removes code that explicitly doesn't work (code clean up)

Almost all of the templating code was copied from the template action plugin. However, that plugin has substantially more code that handles file writing to a remote host, which this plugin doesn't need at all. As a result, I've probably imported more things than I need to. In the future a more generic "grab from file template" action plugin should be written, as this module/plugin doesn't need any special logic beyond what is already offered.

running the examples:

```
> ansible-p  ansible-playbook kubernetes.yml -vvv
No config file found; using defaults
 [WARNING]: provided hosts list is empty, only localhost is available

1 plays in kubernetes.yml

PLAY [Kube ns test] ************************************************************

TASK [testing : Create a kubernetes namespace from file] ***********************
task path: /Users/j2/work/src/github.com/dronedeploy/kube_deploy/roles/testing/tasks/main.yml:3
ok: [localhost] => {"changed": false, "invocation": {"module_args": {"api_endpoint": "redacted", "state": "present", "template": "ns.yml", "token": "redacted"}, "module_name": "kubernetes"}}

TASK [testing : Create a kubernetes namespace from template] *******************
task path: /Users/j2/work/src/github.com/dronedeploy/kube_deploy/roles/testing/tasks/main.yml:10
ok: [localhost] => (item=ansible-test) => {"changed": false, "invocation": {"module_args": {"api_endpoint": "kube-master.ddops.cool", "state": "present", "template": "ns.yml.j2", "token": "redacted"}, "module_name": "kubernetes"}, "item": "ansible-test"}

TASK [testing : Create a kubernetes namespace from inline] *********************
task path: /Users/j2/work/src/github.com/dronedeploy/kube_deploy/roles/testing/tasks/main.yml:19
ok: [localhost] => {"changed": false, "invocation": {"module_args": {"api_endpoint": "redacted", "inline": {"apiVersion": "v1", "kind": "Namespace", "metadata": {"annotations": {"a1": "value1", "a2": "value2"}, "labels": {"label_env": "test", "label_ver": "latest"}, "name": "ansible-test"}}, "state": "present", "token": "redacted"}, "module_name": "kubernetes"}}

TASK [testing : Create a kubernetes namespace from inline with variable] *******
task path: /Users/j2/work/src/github.com/dronedeploy/kube_deploy/roles/testing/tasks/main.yml:36
ok: [localhost] => (item=ansible-test) => {"changed": false, "invocation": {"module_args": {"api_endpoint": "redacted", "inline": {"apiVersion": "v1", "kind": "Namespace", "metadata": {"annotations": {"a1": "value1", "a2": "value2"}, "labels": {"label_env": "test", "label_ver": "latest"}, "name": "ansible-test"}}, "state": "present", "token": "redacted"}, "module_name": "kubernetes"}, "item": "ansible-test"}

localhost                  : ok=4    changed=0    unreachable=0    failed=0   

```

Note: I couldn't figure out where to put the action plugin code. I don't think this is the right spot for it. Please let me know where it goes and I'll update the PR.

### DIFF
--- a/clustering/action_plugins/kubernetes.py
+++ b/clustering/action_plugins/kubernetes.py
@@ -193,6 +193,7 @@ class ActionModule(ActionBase):
     def get_template_source(self, filepath):
         try:
             source = self._find_needle('templates', source)
+            b_source = to_bytes(source)
             with open(b_source, 'r') as f:
                 template_data = to_text(f.read())
 
--- a/clustering/action_plugins/kubernetes.py
+++ b/clustering/action_plugins/kubernetes.py
@@ -1,0 +1,340 @@
+# (c) 2015, Michael DeHaan <michael.dehaan@gmail.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+import base64
+import datetime
+import os
+import pwd
+import time
+import json
+import requests
+import yaml
+
+from ansible import constants as C
+from ansible.errors import AnsibleError
+from ansible.module_utils._text import to_bytes, to_native, to_text
+from ansible.plugins.action import ActionBase
+from ansible.utils.hashing import checksum_s
+from ansible.utils.boolean import boolean
+from ansible.module_utils.urls import fetch_url
+
+############################################################################
+############################################################################
+# For API coverage, this Anislbe module provides capability to operate on
+# all Kubernetes objects that support a "create" call (except for 'Events').
+# In order to obtain a valid list of Kubernetes objects, the v1 spec file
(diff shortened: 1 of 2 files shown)


astropuffin:
no idea why the compile is failing. I'm using the same syntax as: https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/action/template.py#L91

It compiles fine locally:

```
> python -m py_compile action_plugins/kubernetes.py
> echo $?
0
```

astropuffin:
ok, so reason it's not compiling is because tests are running python2.4...

astropuffin:
ok so it's still having issues with python 2.4, currently `from __future__ import (absolute_import, division, print_function)`. It's an action plugin, so it should only have to comply with pythn 2.6+
https://docs.ansible.com/ansible/developing_modules_python3.html

can someone direct me as to how I can designate that properly?

gregdek:
Thanks @astropuffin for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

astropuffin:
ready_for_review I believe this will work once the action plugin piece is moved to the correct location. Currently failing on python2.4 syntax checking, but the part that is failing was copied from an already passing action plugin. https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/action/template.py#L17

astropuffin:
Passing 2.4 compile check. Now failing the "in the wrong spot" checks.

PR #3124, opened 2016-10-07
Fix `archive` truncating archived file names based on prefix length

ryansb:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

archive
##### ANSIBLE VERSION

2.2-rc1
##### SUMMARY

When archiving multiple files, the full length of the calculated
`arcroot` would be removed from the beginning of all file names. If
there was no arcroot, the first two characters of all files would be
removed, so `file.txt` would become `le.txt`.

This patch uses the regular expressions substitution anchored to the
start of the string, so the arcroot is only removed if it is actually
present.


gregdek:
Thanks @ryansb. To the current maintainers, @bendoh please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

Issue #3128, opened 2016-10-10
virtual server rule ... already exists in partition


perbly:
##### ISSUE TYPE
- Bug Report
##### COMPONENT NAME

network/f5/bigip_virtual_server.py
##### ANSIBLE VERSION

```
2.2.0
```
##### OS / ENVIRONMENT

N/A
##### SUMMARY

When a virtual server exists and nothing should be done the module is trying to update the virtual server and gives error message "virtual server rule (rulename) already exists in partition"
##### STEPS TO REPRODUCE

Create a virtual server using module. Run the playbook/task again without changing any vars.
##### EXPECTED RESULTS

Nothing
##### ACTUAL RESULTS

Module is trying to modify rules

```
01020066:3: The requested virtual server rule (/parititon/vsname
 /partition/irule_name) already exists in partition partition.'"  
```

The problem is in this line:

`current_rules = map(lambda x: (x['priority'], x['rule_name']), get_rules(api, name))`

`current_rules = list(map(lambda x: (x['priority'], x['rule_name']), get_rules(api, name)))`

This is added in pull request: https://github.com/ansible/ansible-modules-extras/pull/3129

Regards
Per Bly

ansibot:
@Etienne-Carriere ping, this issue is waiting for your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

caphrim007:
under review

ansibot:
@Etienne-Carriere, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@Etienne-Carriere, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@Etienne-Carriere, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
This repository has been locked. All new issues and pull requests should be filed in https://github.com/ansible/ansible

Please read through the <a href="http://docs.ansible.com/ansible/dev_guide/repomerge.html#move-issues-and-prs-to-new-repo">repomerge</a> page in the dev guide. The guide contains links to tools which automatically move your issue or pull request to the ansible/ansible repo.

[...]

PR #3129, opened 2016-10-10
Functionality to have default persistence profile none if needed.

perbly:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

network/f5/bigip_virtual_server.py
##### ANSIBLE VERSION

```
2.2.0
```
##### SUMMARY

Changed the behaviour of default persistence profile.
If a user is using the same task on multiple virtual servers and would like the default persistence profile to be none on some of them this works now.

Fixed issue (https://github.com/ansible/ansible-modules-extras/issues/3128) 
`current_rules = 'list(map(lambda x: (x['priority'], x['rule_name']), get_rules(api, name)))`


gregdek:
Thanks @perbly. To the current maintainers, @mhite, @caphrim007, @Etienne-Carriere please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

caphrim007:
under review. please squash those commits in the meantime

caphrim007:
this change in behavior is counter intuitive to the existing module behavior as well as the behavior of other modules.

looking at the code here, it seems that what we would probably want to do instead is to have another parameter that is mutually exclusive with `default_persistence_profile` that acts as a boolean to allow the user to say "scorch the earth".

Since that's what the iControl SOAP call is doing anyway, it seems more intuitive to me to instead do the following
- add a `remove_all_persistence_profiles`, type=BOOLEANS, default=False
- make it mutually exclusive with `default_persistence_profile` so that a user can enter one or the other, but not both. Entering both wouldn't make sense anyway because you're already deleting the existing ones when you set a new one.

@mhite what do you think?

The map -> list bugfix is a valid fix though if you want to send that in a separate PR.

network/f5/bigip_virtual_server.py
```diff
@@ -508,6 +509,15 @@ def set_default_persistence_profiles(api, name, persistence_profile):
                 profiles=[[{'profile_name': persistence_profile, 'default_profile': True}]]
             )
             updated = True
+        # If a virtual server has a persistence profile and would like this to be None this section enabled this
+        # When persistence_profile is not defined in task or set to Null in vars the persistence profile will be
+        # set to None.
+        if persistence_profile is None or False:
```
mhite, line 515: Does this really work as expected? Keep in mind that "is" has a higher operator precedence than "or". 

```
>>> persistence_profile = None
>>> persistence_profile is None or False
True
>>> persistence_profile = False
>>> persistence_profile is None or False
False
>>> persistence_profile = 'myprofile'
>>> persistence_profile is None or False
False
```

I'm also slightly confused at why you are checking if "persistence_profile is None" here when we will have already returned to the calling function in lines 491-492.

network/f5/bigip_virtual_server.py
```diff
@@ -236,7 +237,7 @@ def set_rules(api, name, rules_list):
         return False
     rules_list = list(enumerate(rules_list))
     try:
-        current_rules = map(lambda x: (x['priority'], x['rule_name']), get_rules(api, name))
+        current_rules = list(map(lambda x: (x['priority'], x['rule_name']), get_rules(api, name)))
```
mhite, line 240: As @caphrim007 mentioned, this should go in as a separate PR as it does fix a bug introduced recently in Ansible that introduces new default map() behavior.

network/f5/bigip_virtual_server.py
caphrim007: @mhite what are your thoughts on a separate param to zero out those persistence profiles in a similar fashion to the `user` module's use of a binary parameter for generating ssh keys

https://github.com/ansible/ansible-modules-core/blob/devel/system/user.py#L143

mhite:
See my inline comments regarding some potential logic issues in the code.

I would like to see a better summary to start with so I can better understand the intended behavior.

```
Changed the behaviour of default persistence profile.
If a user is using the same task on multiple virtual servers and would like the default persistence profile to be none on some of them this works now.
```

It looks like the change removes the persistence profile if a play is run without default_persistence_profile set (because the implicit default value of params['default_persistence_profile'] is None)? If this were the case, then this parameter should be required instead of optional else we are going to have some pretty terrible unintended side effects.

That being said, I'm not a fan of this particular approach. I'd rather the parameter reflect the explciit intended state while not doing anything with the parameter if it is not supplied by the user. So if you want to remove the profile, have the user pass '' as the default_persistent_profile. 

Or am I totally off the mark here? 

Thanks.

network/f5/bigip_virtual_server.py
mhite: Maybe I'm misunderstanding the intent of this PR, but why not just let someone pass `''` as the `default_persistence_profile` in order to remove it?

caphrim007: I could go for that too. It falls back on the module author to interpret emptiness, and not None-ness, as meaning "remove all the profiles"

mhite: I've always subscribed to the notion of the playbook containing the state you want to implement, rather than using verb-type attributes to trigger things like 'go delete this thing'. Just explicitly set that thing _to_ empty. If you don't want to inspect or implement state on an attribute, don't include it in the playbook.

caphrim007: I'm ok with that.

PR #3179, opened 2016-10-15
Fix gce module to compile on python 3

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

gce_tag gce_img


abadger:
These modules require python-2.6 so we can use the "as" syntax but that's not a big deal.  We'll be finding everything that imports pycompat24 and removing the cruft for Ansible-2.4 anyway.

abadger:
Merge to devel and cherry-picked to stable-2.2

PR #3182, opened 2016-10-15
Port ovirt to python 3

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

ovirt
##### SUMMARY

Since ovirt sdk is not running on python 2.4, we can use python 2.6
syntax directly for exceptions.


abadger:

PR #3184, opened 2016-10-15
Make xenserver_facts compile on python 3

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

xenserver_facts
##### SUMMARY

Since the xenapi is not needed on python 2.4, we can use the
regular exception handling code


abadger:

PR #3190, opened 2016-10-15
Fix proxmox for python 3

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

proxmox
##### ANSIBLE VERSION

```
devel and 2.2
```
##### SUMMARY

This is mscherer's update to proxmox to work on python3 rebased so that it doesn't conflict on the update to sanity tests.

Since it doesn't work on python 2.4, we can use the native
exception handling way for python 3


abadger:

PR #3195, opened 2016-10-16
Added powershell script for win_reg_stat module and documentation for…

jborean93:
##### ISSUE TYPE
- New Module Pull Request
##### COMPONENT NAME

win_reg_stat
##### ANSIBLE VERSION

```
ansible 2.1.2.0
  config file =
```
##### SUMMARY

When dealing with installing programs with Windows there are numerous cases when it is best to check for the existence of a registry key or property. Currently the only way of doing this is to use the raw command with a powershell script to get this value. This module is designed to make it easier to both check if a registry key exists but also get the subkeys/property information such as the type and value.


gregdek:
Thanks @jborean93 for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

jhawkesworth:
HI,

Thanks for this, I like the idea of this module.  You could probably make use of setup module's current ability to run arbitrary powershell, but this is cleaner and I think you can guess its purpose from the name.

I've got a couple of thoughts / questions though.

You set error action preference to stop.  I'm not sure if it is going to actually make a difference, but I think I'm right in saying that modules are expected to either Exit-Json or Fail-Json and handle error failures that would otherwise stop them from returning valid json to STDOUT, so that ansible always has some kind of result back from the module (even if its a failure).  So not sure if setting error action preference to stop is the right thing to do in a module.

Only other thought I've had is that other modules combine the hive and the rest of the path and I wonder if it would be worth fitting in with the existing convention.  Also I'm curious what difference it makes using the full name instead of the powershell drive-type names.

Jon

jborean93:
Hey

The only reason I set the error action preference to stop is because I saw it in a newer windows module https://github.com/ansible/ansible-modules-core/blob/devel/windows/win_command.ps1 and thought it was the way forward, I can remove it if that isn't the case. 

In all honesty I separated the hive and path as separate parameters as it made it easier for me to check that the hive matches the know list of hives and can easily convert that known list to the .net names in select cases. There is nothing that stops me from combining it together and just splitting the path by `:` and getting the first entry, let me know your thoughts.

As to why it currently only works with the short name there were 2 reasons for this;
- both win_regmerge and win_regedit seems to use this style awnd I wanted to keep it consistent
- it makes it easier to check that the hive supplied it valid (this isn't hard to change)
- I initially had issues with using the short name and the values being returned were different, this turned out to be the way I was getting the data and it just stuck in the end

Happy to go either way and can change it however is needed. One question I have is how do I add tests for this module? I've created a playbook that goes through the various key types and asserts that the responses are correct but not sure how to add it to Ansible for future changes?

jhawkesworth:
Interesting.  I need to read up more on error action preference but I think stop is benign.

I think it would be nice to make it so that you can specify the hive + path the same way as the other modules if you don't mind changing it.

Adding tests is something you have to do in the main ansible git project at the moment.  They go in test/integration/target.  Its actually just changed recently but I think there is still a make file in there which will run all of the windows integration tests if you run 'make test_winrm'.  You can set TEST_FLAGS envrionment variable to specific specific tags (which will then just run the tests).  Well worth a look through the existing registry tests (and improving them, by all means) to get a feel for how the tests operate.

Hope this helps,

Jon

jborean93:
I will have a look at combining the hive and path together and split it with `:\` like a normal windows path. Once this is in I will also have a look at adding in a test to the main Ansible repo. After looking at win_regedit I have also been thinking of 
- changing the raw_value return value for a REG_BINARY and REG_NONE string to a list of hex strings
- change path to key to align with win_regedit

jhawkesworth:
Thanks for this.  Its looking great now.  I'm looking forward to trying it out... although I don't know when right now.  Would be great to have this for stuff like checking that all the server hardening is still in place.

jborean93:
Thanks @jhawkesworth I use it a lot where I work as part of our role development testing and to check whether I need to run a particular step in the role itself.

jborean93:
Created separate pull request on the ansible repo to test this new module https://github.com/ansible/ansible/pull/18329

jborean93:
@jhawkesworth is this waiting on anything from me to continue forward, I'm not sure what the process is with new modules?

jhawkesworth:
Hi, code-wise there's nothing else needed from you.  What's needed now is to get others to test it out.
The only way I know to do this is to test out other (windows) PRs and bug fixes, and leave feedback on them, and hope that other module contributors will then do the same for you.

I had a quick look through your tests for this and I particularly like your 'get-content file | set-content file' trick for enforcing windows file endings (I'm slightly annoyed I didn't think of that myself).

I want to test this (and your other PRs out) but I have a lot on at the moment so can't make any promised about when I might get to it.

crazyguyonabike:
Tested and works well. Its confusing to have win_regedit only support PSDrive short names and this module does both. Please strive towards consistency and I guess for the ideal of less code to maintain and debug, I'd lean towards win_regedit's functionality

https://github.com/ansible/ansible-modules-extras/blob/devel/windows/win_regedit.ps1#L45

jborean93:
@crazyguyonabike I originally only supported the PSDrive short names in the module to keep in line with the win_regedit's functionality but did add support for the long names just in case people preferred that. I am happy to go either way as I understand the maintainability concerns but I think the main driver would be consistency as functionally the code converts the long to short in the beginning.

jborean93:
Changes moved to https://github.com/ansible/ansible/pull/19145 after repo merge

PR #3235, opened 2016-10-19
Do not import splitter, since we do not use it

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

blockinfile


gregdek:

abadger:
Merged to devel and stale-2.2 branches.  This will be in either 2.2.0 or 2.2.1 depending on whether we need to make another 2.2.0rc.

PR #3283, opened 2016-10-24
Added new module gitlab_tag

dj-wasabi:
##### ISSUE TYPE
- New Module Pull Request
##### COMPONENT NAME

gitlab_tag
##### ANSIBLE VERSION

```
ansible 2.1.2.0
  config file = 
```
##### SUMMARY

This module will create a git tag in Gitlab on a project. The group and project should exists, otherwise this module will complain that either one of these does not exists. This module can only create or delete a tag, not update a tag.


gregdek:
Thanks again to @dj-wasabi. This module is going into community review.

We notice that you are one of the maintainers (@dj-wasabi) of this module, so when you think it's ready, please comment "shipit" and we will consider it for merging.

mscherer:
Since I suspect that gitlab module is not python 2.4 compatible, I think you can drop the py2.4 syntax.

source_control/gitlab_tag.py
```diff
@@ -0,0 +1,268 @@
...
+    validate_certs=false
+    login_token="WnUzDsxjy8230-Dy_k"
+    group: my_first_group
+    project: my_first_project
+    name: v1.0
+    ref: master
+    description: My First Release
+    state: present
+
+- name: "Delete tag"
```
mscherer, line 107: I think the example would be clearer if shorter (ie, description do not make much sense if it will be removed, so maybe it can be removed, maybe the same for ref ?)

source_control/gitlab_tag.py
```diff
@@ -0,0 +1,268 @@
...
+    if project_tag:
+        if state == "present":
+            module.exit_json(changed=False, result="Tag is already created.")
+        else:
+            if module.check_mode:
+                module.exit_json(changed=True, result="Tag should have been deleted.")
+            if tag.deleteTag():
+                module.exit_json(changed=True, result="Tag is deleted.")
+            else:
+                module.exit_json(changed=False, result="Something went wrong deleting the tag.")
```
mscherer: Is there a way to get a error message or something, as a user would be puzzled by this error message (maybe we can't, I didn't looked at the code)

source_control/gitlab_tag.py
dj-wasabi: Yes, you are right. Thanks!

PR #3311, opened 2016-10-27
maven_artifact: Operation fails because status != 200 when using a maven repo with a file URL

chui101:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

maven_artifact
##### ANSIBLE VERSION

```
ansible 2.1.2.0
```
##### SUMMARY

When using a file: url with maven_artifact plugin **info['status']** is **None** causing plugin execution to fail even if file was copied successfully.

e.g. 

```
- maven_artifact: group_id=com.example artifact_id=foo version=1.2.3.4 dest=foo/pom.xml extension=pom repository_url=file:/bar/baz
```

yields output 

```
Failed to download maven-metadata.xml because of OK (1234 bytes)for URL file:/bar/baz/com/example/foo/maven-metadata.xml
```

While **info['status']** is **None** in this case, **info['msg']** contains **OK (### bytes)** when a file transfer is successful with this method, so if **info['msg'][:2] == 'OK'** then this indicates a successful operation.

_Edited for formatting/grammar_

### DIFF
--- a/packaging/language/maven_artifact.py
+++ b/packaging/language/maven_artifact.py
@@ -257,7 +257,7 @@ class MavenDownloader:
         self.module.params['http_agent'] = self.module.params.get('user_agent', None)
 
         response, info = fetch_url(self.module, url_to_use, timeout=req_timeout)
-        if info['status'] != 200:
+        if info['status'] != 200 or info['msg'][:2] != 'OK':
             raise ValueError(failmsg + " because of " + info['msg'] + "for URL " + url_to_use)
         else:
             return f(response)


gregdek:
Thanks @chui101. To the current maintainers, @chrisisbeef please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

packaging/language/maven_artifact.py
```diff
@@ -257,7 +257,7 @@ def _request(self, url, failmsg, f):
         self.module.params['http_agent'] = self.module.params.get('user_agent', None)
 
         response, info = fetch_url(self.module, url_to_use, timeout=req_timeout)
-        if info['status'] != 200:
+        if info['status'] != 200 or info['msg'][:2] != 'OK':
```
bdellegrazie, line 260: Shouldn't this be:
if info['status'] != 200 and info['msg'][:2] != 'OK':
?

chui101: It might seem awkward, but these are actually the two cases where we want it to fail. It works as intended here... 

If the info['status'] != 200, it shouldn't automatically short circuit the logic and fail - it should perform a fallback check because maybe info['msg'] says OK.

bdellegrazie: I understand now, its because `None != 200` evaluates to false and you're relying on _or_ short circuiting. Thank you for pointing out my error.

As a pure question only (not suggesting you change anything) would the intent be clearer and more obvious by reversing the result as in:

``` python
if info['status'] == 200 or info['msg'][:2] == 'OK':
    return f(response)
else:
    raise ValueError(failmsg + " because of " + info['msg'] + "for URL " + url_to_use)
```

Thanks again.

PR #3319, opened 2016-10-29
Minor fixes for openwrt_init for busybox ps and worthless exit codes

agaffney:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

`openwrt_init` module
##### ANSIBLE VERSION

```
ansible 2.2.0.0
```
##### SUMMARY

This commit addresses a few minor issues with the `openwrt_init` module.

The original `ps` command did not work, because of busybox `ps`'s limited options. The module would always show the service as not running when using the `pattern` parameter.

The other issue is that the `enable` command on an init script can return a non-zero exit code when the init script does not set a `STOP` value. Instead of checking the exit code, the module now re-runs the `enabled` command on the init script and compares to the desired state.


agaffney:
@bcoca

PR #3325, opened 2016-10-31
dnf: fix compatibility with DNF 2.0

### LINKED ISSUE #3310: dnf module doesn't work with dnf-2.0 API
##### ISSUE TYPE

- Bug Report
##### COMPONENT NAME

packaging/os/dnf.py
##### ANSIBLE VERSION

```
devel, 2.2, 2.1

Current RPM versions:
* ansible-2.1.2.0-1.fc24.noarch
* python2-dnf-2.0.0-0.rc1.4.fc26.noarch (on the managed host)
```
##### OS / ENVIRONMENT

Fedora Rawhide
##### SUMMARY

Originally reported here: https://bugzilla.redhat.com/show_bug.cgi?id=1387621

When trying to use the 'dnf' module in a task, we're getting tracebacks instead of successful install.  Diagnosed in the bugzilla report as being usage of functions in DNF-1.x that have not been kept in the DNF-2.0 API.  Since the module will run against both dnf-1.x and dnf-2.x hosts, the fix for this will need to be able to do the right thing for both.
##### STEPS TO REPRODUCE

On a fedora rawhide host:

```
ansible localhost -m dnf -a 'state=present name=python3-q ' 
```

##### EXPECTED RESULTS

Successful install of the package
##### ACTUAL RESULTS

```
fatal: [IP_ADDRESS_REMOVED]: FAILED! => {"changed": false, "failed": true, "module_stderr": "", "module_stdout": "No handlers could be found for logger \"dnf\"\r\nTraceback (most recent call last):
  File \"/tmp/ansible_LDnNIF/ansible_module_dnf.py\", line 355, in <module>
    main()
  File \"/tmp/ansible_LDnNIF/ansible_module_dnf.py\", line 349, in main
    ensure(module, base, params['state'], params['name'])
  File \"/tmp/ansible_LDnNIF/ansible_module_dnf.py\", line 240, in ensure
    pkg_specs, group_specs, filenames = cli.commands.parse_spec_group_file(
AttributeError: 'module' object has no attribute 'parse_spec_group_file'\r\n", "msg": "MODULE FAILURE"}

```

igor-raits:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
dnf

##### ANSIBLE VERSION
```
ansible 2.1.2.0
```

##### SUMMARY
Adopt to changes in DNF 2.0.

* Reimplement (copy) old dnf.cli.commands.parse_spec_group_file(),
  upstream uses argparse since 2.0.
* add_remote_rpm() has been changed to the add_remote_rpms()

Fixes: #3310


packaging/os/dnf.py
```diff
@@ -258,13 +258,35 @@ def _mark_package_install(module, base, pkg_spec):
         module.fail_json(msg="No package {} available.".format(pkg_spec))
 
 
+def _parse_spec_group_file(names):
+    pkg_specs, grp_specs, filenames = []
```
abadger: One change needed here:

```
    pkg_specs = []
    grp_specs = [] 
    filenames = []
```

Python raises a ValueError without this change because it's expecting that it's supposed to unpack the right hand side into three variables on the left hand side.

Doing the following would fix the ValueError but it would have a logic error instead:

```
pkg_specs = grp_specs = filenames = []
```

That's because the above is the same as

```
filenames = []
grp_specs = filenames
pkg_specs = grp_specs
```

which sets all three variables to the same list.  Since we want three separate lists, we need to define them all separately.

abadger:
Merged to devel and cherry-picked to stable-2.2  Will be in the 2.2.1 release.

PR #3349, opened 2016-11-03
Fix python3 syntax for octal numbers

abadger:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME
crypto/openssl_privatekey.py

##### ANSIBLE VERSION
0600 is no longer a valid octal literal.  Use the syntax that's correct for both python2.4 and python3.


abadger:
Merged to devel.

PR #3382, opened 2016-11-07
Remove need for translate/maketrans due to py3 differences

abadger:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME
vmware_guest.py

##### ANSIBLE VERSION

```
devel 2.2
```

##### SUMMARY

* translate() has a different api for text vs byte strings
* maketrans must be imported from a different location on py2 vs py3

Since this is such a small string outside of a loop we don't have to
worry too much about speed so it's better to have a single piece of code
that works on both py2 and py3

### DIFF
--- a/cloud/vmware/vmware_guest.py
+++ b/cloud/vmware/vmware_guest.py
@@ -161,7 +161,6 @@ except ImportError:
     pass
 
 import os
-import string
 import time
 
 from ansible.module_utils.urls import fetch_url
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @abadger. To the current maintainers, @jctanner, @jcpowermac, @mtnbikenc, @chrrrles please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

PR #3383, opened 2016-11-07
Refactor firewalld

maxamillion:
##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
system/firewalld.py

##### ANSIBLE VERSION

```
ansible 2.3.0 (devel e90fe3bd74) last updated 2016/11/07 12:31:16 (GMT -500)
```

##### SUMMARY

Fixes source handling (previously was reporting a change was successfully happening when it wasn't in certain scenarios). Code refactor to make addition of new functionality much easier and general code flow more simple.

```
No change to command or output.
```

### DIFF
--- a/system/firewalld.py
+++ b/system/firewalld.py
@@ -108,6 +108,7 @@ from ansible.module_utils.basic import AnsibleModule
 
 import sys
 
+
 #####################
 # Globals
 #
--- a/system/firewalld.py
+++ b/system/firewalld.py
@@ -108,17 +108,199 @@ from ansible.module_utils.basic import AnsibleModule
 
 import sys
 
-
 #####################
 # Globals
 #
 fw = None
+msgs = None
 module = None
-fw_offline = False
+changed = False
+immediate = None
+permanent = None
 Rich_Rule = None
+fw_offline = False
+desired_state = None
 FirewallClientZoneSettings = None
 
 module = None
+
+
+#####################
+# action logic
+#
+def action_runner(
+    action,
+    get_enabled_immediate, get_enabled_immediate_args,
+    get_enabled_permanent, get_enabled_permanent_args,
+    set_enabled_immediate, set_enabled_immediate_args,
+    set_enabled_permanent, set_enabled_permanent_args,
+    set_disabled_immediate, set_disabled_immediate_args,
+    set_disabled_permanent, set_disabled_permanent_args
+    ):
+    """
+    action_runner
+
+    This fuction contains the "transaction logic" where as all operations
+    follow a similar pattern in order to perform their action but simply
+    call different functions to carry that action out.
+
+    :param action: str, string of the type of firewalld module action this is (example: service, port, etc)
+    :param get_enabled_immediate: func, getter function for immediate actions
+    :param get_enabled_immediate_args: tuple, args for get_enabled_immediate function
+    :param get_enabled_permanent: func, getter function for permanent actions
+    :param get_enabled_permanent_args: tuple, args for get_enabled_permanent function
+    :param set_enabled_immediate: func, setter function for immediate actions
+    :param set_enabled_immediate_args: tuple, args for set_enabled_immediate function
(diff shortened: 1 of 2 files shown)


test/utils/shippable/sanity-skip-python24.txt
```diff
@@ -10,6 +10,7 @@
 /packaging/os/dnf.py
 /packaging/os/layman.py
 /remote_management/ipmi/
+/system/firewalld.py
```
abadger, line 13: Is python-2.6 now a requirement because python2-firewalld needs python2.6+?  If so this is fine but you need to add python-2.6 to the requirements section in the module documentation

system/firewalld.py
```diff
@@ -112,13 +112,197 @@
...
+msgs = None
 module = None
-fw_offline = False
+changed = False
+immediate = None
+permanent = None
 Rich_Rule = None
+fw_offline = False
+desired_state = None
 FirewallClientZoneSettings = None
```
abadger, line 123: I think this goes a bit overboard with use of globals.  I do think trying a global for modules is a good experiment because we use parts of it (the fail_json() method) globally (in lieu of an exception handler hook really).  Many of these others, though, can be easily passed into functions or be return values from functions as they are only used in main and one other function.

In general, globals make it harder to unittest a piece of code and can obscure the inputs and side effects of the code for other people reading it later.

system/firewalld.py
```diff
@@ -112,13 +112,197 @@
...
+# action logic
+#
+def action_runner(
+    action,
+    get_enabled_immediate, get_enabled_immediate_args,
+    get_enabled_permanent, get_enabled_permanent_args,
+    set_enabled_immediate, set_enabled_immediate_args,
+    set_enabled_permanent, set_enabled_permanent_args,
+    set_disabled_immediate, set_disabled_immediate_args,
+    set_disabled_permanent, set_disabled_permanent_args
```
abadger, line 138: I'd be tempted to use functools.partial here if you already require python-2.6+. https://docs.python.org/2/library/functools.html#functools.partial  functools was added in python2.5 so you can't use it if you're stuck on python-2.4.

I do see that you use set_enabled_permanent_args and set_enabled_immediate_args to construct the enabled_msg and disabled_msg... I guess if you use partial the msg strings will have to be constructed outside and then passed into the function.

system/firewalld.py
```diff
@@ -112,13 +112,197 @@
...
+    global immediate
+    global permanent
+    global changed
+    global desired_state
+    global permanent
+    global immediate
+
+    ## These are actions that modify msgs during the transaction for more
+    ## verbose output to the user.
+    custom_msg_actions = [
```
abadger, line 172: You can make this a frozenset().  Performance isn't that important here but sets are the best way to deal with lists that are used just for containment tests.

system/firewalld.py
abadger: Or just pass in msgs_tuple as one extra parameter to this function... Have to look at the callers to see which would work better.

abadger:
Taking a step back -- I wonder if this code is actually an object-oriented pattern.

Base class with the action_handler() method.  Takes attributes for permanent, immediate and other inputs that you're trying to make global variables.  Subclasses for Zone, Masquerade, etc implement enabled, disabled, and get using the functions that you're currently passing into action_runner() along with self.permanent and self.immediate.

test/utils/shippable/sanity-skip-python24.txt
maxamillion: Yes, I'll be sure to add that.

maxamillion:
@abadger agreed, I like that pattern better and it would remove a reasonable amount of the hacks I'm doing here. I'll try to find some cycles to pivot to that and re-up the PR. Thanks for the review!

gregdek:
Thanks again to @maxamillion. This module is going into community review.

We notice that you are one of the maintainers (@maxamillion) of this module, so when you think it's ready, please comment "shipit" and we will consider it for merging.

PR #3400, opened 2016-11-09
Make the 'make' module run on python 3

mscherer:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
make
##### SUMMARY
Traceback:

      File \"/tmp/ansible_d28_6uwl/ansible_module_make.py\", line 153, in <module>
        main()
      File \"/tmp/ansible_d28_6uwl/ansible_module_make.py\", line 119, in main
        rc, out, err = run_command(base_command + ['--question'], module, check_rc=False)
      File \"/tmp/ansible_d28_6uwl/ansible_module_make.py\", line 79, in run_command
        return rc, sanitize_output(out), sanitize_output(err)
      File \"/tmp/ansible_d28_6uwl/ansible_module_make.py\", line 95, in sanitize_output
        return output.rstrip(b(\"\\r\\n\"))
    TypeError: rstrip arg must be None or str

There is also a six.iteritems issue, fixed using six.

### DIFF
--- a/system/make.py
+++ b/system/make.py
@@ -65,6 +65,9 @@ EXAMPLES = '''
 # fix this
 RETURN = '''# '''
 
+from ansible.module_utils.six import iteritems
+from ansible.module_utils.basic import AnsibleModule
+
 
 def run_command(command, module, check_rc=True):
     """
@@ -90,9 +93,9 @@ def sanitize_output(output):
     :return: sanitized output
     """
     if output is None:
-        return b('')
+        return ''
     else:
-        return output.rstrip(b("\r\n"))
+        return output.rstrip("\r\n")
 
 
 def main():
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @mscherer. To the current maintainers, @LinusU please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

system/make.py
```diff
@@ -90,9 +93,9 @@ def sanitize_output(output):
     :return: sanitized output
     """
     if output is None:
-        return b('')
+        return ''
```
LinusU, line 96: Are you sure that this still works on python2 though?

mscherer: I wasn't sure, but I did a test and it did seems fine. But that's also why we have a review.

abadger: Yep, now that module.run_command() returns native strings this looks like the correct behaviour.

abadger:
Merged to devel.  Not needed in 2.2 because we don't return stdout and stderr there.

mscherer:
So do we want to make a different fix on 2.2, or we consider not important to have this module running on python 3 there ?

PR #3432, opened 2016-11-11
Fix issue in adding RAW disk in block storage domain

nijinashok:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
Ovirt

##### ANSIBLE VERSION

```
ansible 2.3.0 (devel)
```

##### SUMMARY
Ovirt won't allow to add RAW sparse disk in block storage domain and is only available in file based storage domain. By default, sparse option is true in ovirt. So currently the raw disk creation using this module in a block storage domain will fail with error "Disk configuration (RAW Sparse) is incompatible with the storage domain type". 

The commit adds sparse option where user can pass it as "True/False". So while creating RAW disk in block storage domain, we can pass "sparse = False" to make it work.

### DIFF
--- a/cloud/ovirt/ovirt_disks.py
+++ b/cloud/ovirt/ovirt_disks.py
@@ -65,7 +65,10 @@ options:
         default: 'virtio'
     format:
         description:
-            - "Format of the disk. Either copy-on-write or raw."
+            - Specify format of the disk.
+            - If (cow) format is used, disk will by created as sparse, so space will be allocated for the volume as needed, also known as I(thin provision).
+            - If (raw) format is used, disk storage will be allocated right away, also known as I(preallocated).
+            - Note that this option isn't idempotent as it's not currently possible to change format of the disk via API.
         choices: ['raw', 'cow']
     storage_domain:
         description:
@@ -168,6 +171,7 @@ class DisksModule(BaseModule):
             format=otypes.DiskFormat(
                 self._module.params.get('format')
             ) if self._module.params.get('format') else None,
+            sparse=False if self._module.params.get('format') == 'raw' else True,
             provisioned_size=convert_to_bytes(
                 self._module.params.get('size')
             ),
@@ -198,7 +202,6 @@ class DisksModule(BaseModule):
         return (
             equal(self._module.params.get('description'), entity.description) and
             equal(convert_to_bytes(self._module.params.get('size')), entity.provisioned_size) and
-            equal(self._module.params.get('format'), str(entity.format)) and
             equal(self._module.params.get('shareable'), entity.shareable)
         )
 
@@ -234,7 +237,6 @@ def main():
         vm_id=dict(default=None),
         size=dict(default=None),
         interface=dict(default=None,),
-        allocation_policy=dict(default=None),
         storage_domain=dict(default=None),
         profile=dict(default=None),
         format=dict(default=None, choices=['raw', 'cow']),


gregdek:
Thanks @nijinashok. To the current maintainers, @machacekondra please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

cloud/ovirt/ovirt_disks.py
```diff
@@ -199,6 +204,7 @@ def update_check(self, entity):
             equal(self._module.params.get('description'), entity.description) and
             equal(convert_to_bytes(self._module.params.get('size')), entity.provisioned_size) and
             equal(self._module.params.get('format'), str(entity.format)) and
+            equal(self._module.params.get('sparse'), entity.sparse) and
```
machacekondra: Currently it's not possible to change allocation policy of the disk in API. so this shouldn't be here, as we can't update it.

cloud/ovirt/ovirt_disks.py
```diff
@@ -238,6 +244,7 @@ def main():
         storage_domain=dict(default=None),
         profile=dict(default=None),
         format=dict(default=None, choices=['raw', 'cow']),
+        sparse=dict(default=None, type='bool'),
```
machacekondra: On line 244 there is `allocation_policy` parameter, which I forgot to implement, which should handle this action. Can you please remove it?

cloud/ovirt/ovirt_disks.py
```diff
@@ -67,6 +67,10 @@
         description:
             - "Format of the disk. Either copy-on-write or raw."
         choices: ['raw', 'cow']
+    sparse:
+        description:
+            - "I(True) create disk with sparse allocation. By default ovirt-engine create sparse allocated disk"
```
machacekondra: Can you please change to following description:
- If (True) disk will by createad as sparse, so space will be allocated for the volume as needed, also known as I(thin provision).
- If (False) storage will be allocated right away, also known as I(preallocated).
- Note that this option isn't idempotent as it's not currently possible to change allocation policy of disk via API.

machacekondra (changes requested):
Thank you very much for the fix. Can you please check out my comments?

machacekondra:
Taking back my comments. There shouldn't be any `sparse` parameter, but we should send `sparse = True` or `sparse = False` depend on the `format` parameter. So if user will send the `raw` we will send `sparse=False` and if user specify `cow` we will send `sparse=True`.

Also we should change documentation of `format` parameter accordingly:

```
- Specify format of the disk.
- If (cow) format is used, disk will by created as sparse, so space will be allocated for the volume as needed, also known as I(thin provision).
- If (raw) format is used, disk storage will be allocated right away, also known as I(preallocated).
- Note that this option isn't idempotent as it's not currently possible to change format of the disk via API.
```

machacekondra:
Also if you can please remove checking of `format` parameter in `update_check` method. As it don't make any sense. And also as previously noted please remove `allocation_policy` parameter. Do you have time to do it? If not I will do it.

nijinashok:
Thanks a lot for reviewing the request. I have made changes as per your suggestion.

machacekondra:
Thank you, looks good to me. Maybe I would replace `False if self._module.params.get('format') == 'raw' else True` with `self._module.params.get('format').lower() != 'raw'`, but not critical I guess.

@mperina Can you please review? As oVirt support following matrix:

file based storage:
- RAW preallocated
- RAW sparse
- qcow2 sparse

block storage:
- RAW preallocated
- qcow2 sparse

So in my opinion we should create always this options:
- RAW preallocated
- qcow2 sparse

unless oVirt will support more.

machacekondra:
@mwperina

machacekondra:
shipit

gregdek:
Thanks again to @nijinashok for this PR, and thanks @machacekondra for reviewing. Marking for inclusion.

mwperina (approved):
:+1:

PR #3522, opened 2016-11-22
Add check mode support to ec2_elb_facts

tomislacker:
##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
ec2_elb_facts

##### ANSIBLE VERSION

```
ansible 2.2.0.0
  config file = 
  configured module search path = Default w/o overrides
```

##### SUMMARY
Enables `ec2_elb_facts` to run in check mode. This then allows the ability to use the module to collect instance information, dynamically add those instances to the inventory, and do real checks with other tasks against the ELB instances.

Fixes #3521

PR #3523, opened 2016-11-22
Add check mode support to ec2_remote_facts

tomislacker:
##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
ec2_remote_facts

##### ANSIBLE VERSION

```
ansible 2.2.0.0
  config file = 
```

##### SUMMARY
Enables `ec2_remote_facts` to run in check mode. This then allows the ability to use the module to collect instance information, dynamically add those instances to the inventory, and do real checks with other tasks against the instances.


gregdek:
Thanks @tomislacker. To the current maintainers, @michaeljs1990, @wimnat please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

wimnat:
@tomislacker can you please help clarify the use case?

Your PR doesn't do anything other than add check_mode to the argument spec.  You would also need to add dry_run in to the boto get_only_instance call to make use of this.

However, is this necessary?  I don't quite understand the summary you included.

needs_revision

gregdek:
Thanks @tomislacker for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

tomislacker:
@wimnat the thought being that a playbook like this would be possible:

```yaml
---
- hosts: ec2-elb-services
  connection: local
  gather_facts: no

  vars:
    ansible_python_interpreter: "/usr/bin/env python2.7"

  tasks:
    - name: Collect Varnish ELB facts
      ec2_elb_facts:
        names: "{{ env }}-{{ stack }}-elb"
        region: "{{ aws_region }}"
      register: elbfacts
      tags: always

    - name: Get ELB instances facts
      ec2_remote_facts:
        filters:
          instance-id: "{{ elbfacts.elbs[0].instances }}"
        region: "{{ aws_region }}"
      register: instancefacts
      tags: always

    - name: Set EC2 Address
      add_host:
        name: "{{ item.private_ip_address }}"
        groups: "ec2-elb-service-instances"
      with_items: "{{ instancefacts.instances }}"
      tags: always

- hosts: ec2-elb-service-instances
  become: yes

  roles:
    - role: myservice-config
```

This would then allow you to:

1. Enumerate instances in an ELB
1. Get the IPs of the instances & add them to dynamic inventory
1. Assess the pending changes to the actual instances.

Given that `ec2_remote_facts` is strictly a read-only operation, it seems to me that it should be run in full during check mode as it's analogous to `gather_facts`, so we wouldn't want to pass `dry_run` to `get_only_instances`.

There's also similar precedent for this proposed behavior such as in [ec2_group:275](https://github.com/ansible/ansible-modules-core/blob/stable-2.2/cloud/amazon/ec2_group.py#L275) where before check mode is ever evaluated, the module is conducting reconnaissance and making AWS API calls.

EDIT: I have a related PR open at #3522 to make an identical change to the `ec2_elb_facts` module.

wimnat:
Sorry if i'm being stupid here but i still don't get it.  Check mode is to run the module and see if a change would be made without actually making the change.  But as you correctly stated, ec2_remote_facts is strictly read-only anyway so check mode is really not going to have an effect on what the module does.

In the example you gave, i don't see why it wouldn't work as the module stands, without your PR

tomislacker:
@wimnat 

> In the example you gave, i don't see why it wouldn't work as the module stands, without your PR

Because if the module doesn't indicate to Ansible that check mode is supported in *any* capacity, Ansible won't conduct any execution for the module. Meaning that my above noted playbook will not function as desired.

tomislacker:
@wimnat did you have a chance to try out an example like I provided to see what I'm attempting to facilitate?

ryansb:
👍 LGTM. Since it's read-only, you don't need extra code to support check mode. Just the flag to tell Ansible it's supported.

wimnat:
Yep I get what's going on now :D

shipit

tomislacker:
@wimnat @ryansb thank you much gentlemen. If this continues the precedent, I believe there's some other modules that would also be in-scope for modifications as well though I haven't had an explicit need to modify them myself yet.

PR #3552, opened 2016-11-28
add win_shortcut module

Rajeshkumar90:
ISSUE TYPE

New Module Pull Request
COMPONENT NAME

Adding shortcut to windows host

ANSIBLE VERSION

ansible 2.1.2.0

SUMMARY

Adding windows shortcut to the windows host

Example :
win_shortcut:
src: 'C:\Program Files\Notepad++\notepad++.exe'
dest: 'C:\Users\588378\Desktop\notepad++.lnk'


windows/win_shortcut.ps1
```diff
@@ -0,0 +1,59 @@
...
+{
+ Fail-Json (New-Object psobject) "missing required argument: Either src or  dest File path contains Null Value"
+}
+
+if((Test-Path  $TargetFile ) -and (Test-Path $ShortcutFile))
+{
+  $result.changed = $FALSE
+}
+else
+{
```
jhawkesworth, line 49: I think it would be better if this was a try {  ...  } catch {   .. }} block so any exceptions can be caught and returned using Fail-Json.

windows/win_shortcut.py
```diff
@@ -0,0 +1,55 @@
...
+# this is a windows documentation stub.  actual code lives in the .ps1
+# file of the same name
+
+DOCUMENTATION = '''
+---
+module: win_shortcut
+version_added: "2.3"
+short_description: Adds Windows shortcut on windows hosts.
+description:
+    - Uses .net Environment to set path environment variables and can set at User, Machine and Process level.
```
jhawkesworth: Looks like description needs updating.

jhawkesworth:
Almost there, just a couple of little things to address (see inline comments).
Many thanks,
Jon

windows/win_shortcut.py
Rajeshkumar90: Updated with new file

windows/win_shortcut.ps1
Rajeshkumar90: added try and catch
tested in env

Rajeshkumar90 (commented):
Added try and catch for ps file

gregdek:
Thanks @Rajeshkumar90 for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

windows/win_shortcut.py
```diff
@@ -0,0 +1,55 @@
...
+author: "SyedShabir, RajeshKumar"
+notes:
+   - This module is helpful in deploying application which requires windows shortcut.
+'''
+
+EXAMPLES = '''
+  # Creates shortcut for given exe
+  win_shortcut:
+    src: 'C:\\Program Files\\Mozilla Firefox\\Firefox.exe'
+    dest: 'C:\\users\\Test\\Desktop\\Firefox.exe'
```
nitzmahone: We're trying to standardize on unquoted, unescaped paths for Windows module docs going forward, so these should be like `src: C:\Program Files\Mozilla Firefox\Firefox.exe`

windows/win_shortcut.ps1
```diff
@@ -0,0 +1,65 @@
...
+
+$params = Parse-Args $args
+
+$result = New-Object psobject @{
+    changed = $FALSE
+}
+
+$TargetFile = Get-Attr $params "src" $FALSE
+
+If ($TargetFile -eq $FALSE)
```
nitzmahone: Use -failifempty arg to Get-Attr and you can get rid of this

windows/win_shortcut.ps1
```diff
@@ -0,0 +1,65 @@
...
+{
+   Fail-Json (New-Object psobject) "missing required argument: src File Path"
+
+}
+$ShortcutFile = Get-Attr $params "dest" $FALSE
+
+If ($ShortcutFile -eq $FALSE)
```
nitzmahone: Same- use -failifempty.

windows/win_shortcut.ps1
```diff
@@ -0,0 +1,65 @@
...
+{
+   Fail-Json (New-Object psobject) "missing required argument: dest File Path"
+}
+
+If(($TargetFile -or $ShortcutFile) -eq $null)
```
nitzmahone: This will be unnecessary with -failifempty usage above

windows/win_shortcut.ps1
```diff
@@ -0,0 +1,65 @@
...
+{
+ Fail-Json (New-Object psobject) "missing required argument: Either src or  dest File path contains Null Value"
+}
+try
+{
+ if((Test-Path  $TargetFile ) -and (Test-Path $ShortcutFile))
```
nitzmahone: This isn't fully testing the change conditions. What if the target path is different than the existing shortcut? The module should be able to handle this case easily (luckily, CreateShortcut also allows you to inspect/edit an existing shortcut, so should be very easy to add this behavior).

nitzmahone (changes requested):
Biggest thing is ensuring it properly updates an existing shortcut. Couple other little things that will make the code a lot smaller.

nitzmahone:
Only other comment: given how small the actual code is, adding this to win_file as `state: shortcut` with a `src` arg (ala *nix `file` module) might make more sense than a standalone module... Thoughts, @jhawkesworth @dagwieers @trondhindenes ?

dagwieers:
Oh wait, I have my own implementation of this that I hadn't shared yet.

https://github.com/crombeen/ansible/blob/master/library/win_shortcut.ps1

You can set a shortcut, change the (visible) name of the link and add an icon. I don't think we want this to be part of the win_file module as well. A shortcut is certainly not the same as a symlink on Linux, you can have shortcuts going to a URL.

nitzmahone:
@dagwieers Good point- if you take it to the logical conclusion by adding all the other shortcut properties, we definitely don't want all that in win_file. OK, nevermind on that part. :)

dagwieers:
Shall I add my win_shortcut implementation as an alternative in a PR ?

I use it to manage desktop icons at the elementary school so preschoolers can easily go to a selection of educational websites at that start of computer class.

nitzmahone:
@dagwieers Yours doesn't look like it fully handles change detection on the various properties- that'd be a merge blocker for me personally. I figure it's trivial to add the other properties to this one later once the target change detection is in place.

dagwieers:
True, that's the reason why I hadn't added it :-)

Rajeshkumar90:
'ready_for_review'

jhawkesworth:
closing alternative win_shortcut module has now been merged.

PR #3561, opened 2016-11-29
Add update parameter to os_stack

TristanCacqueray:
##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
cloud/openstack/os_stack

##### ANSIBLE VERSION
```
ansible 2.2.0.0
```

##### SUMMARY
Add update parameter to os_stack.
```
Before:
TASK [Deploy stack] ************************************************************
changed: [localhost] => (item=upstream)

After:
TASK [Deploy stack] ************************************************************
ok: [localhost] => (item=upstream)
```

This change enables using the os_stack module to get stack information
without doing an update.

### DIFF
--- a/cloud/openstack/os_stack.py
+++ b/cloud/openstack/os_stack.py
@@ -41,6 +41,13 @@ options:
       choices: ['present', 'absent']
       required: false
       default: present
+    update:
+      description:
+        - Indicate whenever the stack is updated
+      choices: ['yes', 'no']
+      required: false
+      default: yes
+      version_added: "2.3"
     name:
       description:
         - Name of the stack that should be created, name could be char and digit, no space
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @TristanCacqueray for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

ryansb:
This seems like it would let users skip updates to a stack. Instead of this way (which adds a parameter), I think users could use a `when` statement on the task to skip the update under certain conditions.

TristanCacqueray:
The goal is to register the stack object without doing an update. Would a new "info" state be more appropriate ?

PR #3562, opened 2016-11-29
Remove iteritems from portage module for python3

abadger:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME
portage.py

##### ANSIBLE VERSION

```
devel, 2.2
```

##### SUMMARY
python3 dicts do not have an iteritems() method.  Since the dicts are small, just use the items() method.


abadger:
Merged to devel and stable-2.2.  Will be in either 2.2.1 or 2.2.2 depending on whether we have a 2.2.1rc2.

PR #3571, opened 2016-11-30
Adding new module to create/remove users with optional roles in ms sq…

jpeck-resilient:
##### ISSUE TYPE
 - New Module Pull Request

##### COMPONENT NAME
mssql_user

##### ANSIBLE VERSION
```
ansible 2.2.0.0

```

##### SUMMARY

New module to work with ms sql databases to create or remove users, optionally with a role.

This reuses connection methodology from mssql_db.  I have other modules around creating availability groups and extensions to the mssql_db that i would like to add as well if this looks suitable.  It would be good if the code to create a connection could be moved into a helper class (like has been done with mysql too, if there is appetite i can do this).

Thanks for looking!

…l server

### DIFF
--- a/database/mssql/mssql_user.py
+++ b/database/mssql/mssql_user.py
@@ -1,0 +1,238 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Ansible module to manage mssql databases
+# (c) 2016, Resilient PLC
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @jpeck-resilient for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

PR #3588, opened 2016-11-30
Use native YAML - Packaging

Fale:
##### ISSUE TYPE
 - Docs Pull Request

##### COMPONENT NAME
packaging/*

##### ANSIBLE VERSION
```
2.2
```

##### SUMMARY
Make YAML code use the native style

@gundalow﻿⁠⁠⁠⁠

### DIFF
--- a/packaging/language/bower.py
+++ b/packaging/language/bower.py
@@ -74,7 +74,7 @@ EXAMPLES = '''
 - name: Install "bootstrap" bower package on version 3.1.1.
  bower:
     name: bootstrap
-    version: 3.1.1
+    version: '3.1.1'
 
 - name: Remove the "bootstrap" bower package.
   bower:
--- a/packaging/language/bower.py
+++ b/packaging/language/bower.py
@@ -67,24 +67,37 @@ options:
 '''
 
 EXAMPLES = '''
-description: Install "bootstrap" bower package.
-- bower: name=bootstrap
+- name: Install "bootstrap" bower package.
+  bower:
+    name: bootstrap
 
-description: Install "bootstrap" bower package on version 3.1.1.
-- bower: name=bootstrap version=3.1.1
+- name: Install "bootstrap" bower package on version 3.1.1.
+ bower:
+    name: bootstrap
+    version: 3.1.1
 
-description: Remove the "bootstrap" bower package.
-- bower: name=bootstrap state=absent
+- name: Remove the "bootstrap" bower package.
+  bower:
+    name: bootstrap
+    state: absent
 
-description: Install packages based on bower.json.
-- bower: path=/app/location
+- name: Install packages based on bower.json.
+  bower:
+    path: /app/location
 
-description: Update packages based on bower.json to their latest version.
-- bower: path=/app/location state=latest
+- name: Update packages based on bower.json to their latest version.
+  bower:
+    path: /app/location
+    state: latest
 
-description: install bower locally and run from there
-- npm: path=/app/location name=bower global=no
-- bower: path=/app/location relative_execpath=node_modules/.bin
+# install bower locally and run from there
+- npm:
+    path: /app/location
+    name: bower
+    global: no
+- bower:
+    path: /app/location
+    relative_execpath: node_modules/.bin
 '''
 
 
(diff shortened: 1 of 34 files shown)


gregdek:
Thanks @Fale for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

packaging/os/zypper_repository.py
```diff
@@ -114,22 +114,36 @@
...
 # Add a repo and add it's gpg key
-- zypper_repository: repo=http://download.opensuse.org/repositories/systemsmanagement/openSUSE_Leap_42.1/ auto_import_keys=yes
+- zypper_repository:
+    repo: 'http://download.opensuse.org/repositories/systemsmanagement/openSUSE_Leap_42.1/'
+    auto_import_keys: yes
  
 # Force refresh of a repository
-- zypper_repository: repo=http://my_internal_ci_repo/repo name=my_ci_repo state=present runrefresh=yes
+- zypper_repository:
+    repo: 'http://my_internal_ci_repo/repo name=my_ci_repo'
```
abadger: Is this one right?

packaging/language/cpanm.py
```diff
@@ -91,29 +91,43 @@
...
+- cpanm:
+    name: Dancer
+    system_lib: yes
 
 # install Dancer if it's not already installed
 # OR the installed version is older than version 1.0
-- cpanm: name=Dancer version=1.0
+- cpanm:
+    name: Dancer
+    version: 1.0
```
abadger: In general I'd leave more quotes in the yaml (because I can never remember quoting rule ;-)

In this case we actually should add quotes.  That's because the key=value parser always ends up with strings.  But the yaml parser will end up with other types.  So in the key=value example version is a string but in the yaml example version is a string.  Using quotes corrects that.

packaging/os/zypper_repository.py
Fale: Nope. Thanks for noticing it

packaging/language/bower.py
```diff
@@ -67,24 +67,37 @@
...
+- name: Update packages based on bower.json to their latest version.
+  bower:
+    path: /app/location
+    state: latest
+
+# install bower locally and run from there
+- npm:
+    path: /app/location
+    name: bower
+    global: no
```
gundalow, line 97: `global:` isn't defined in `DOCUMENTATION:` no in `arg_spec` so it needs to be removed from the examples

packaging/language/cpanm.py
```diff
@@ -91,29 +91,43 @@
...
+- cpanm:
+    name: Dancer
+    system_lib: yes
 
 # install Dancer if it's not already installed
 # OR the installed version is older than version 1.0
-- cpanm: name=Dancer version=1.0
+- cpanm:
+    name: Dancer
+    version: '1.0'
```
gundalow, line 130: Why is `'1.0'` quotes here, but not in other PRs that you've done?

gundalow (changes requested):
needs_revision

packaging/language/cpanm.py
gundalow: We don't appear to be consistent in the recent PRs on this.

packaging/language/cpanm.py
Fale: It should have been, because as @abadger pointed out that could become a float. I think I was careful on those 'risky' values, but some could have gone through. Do you remember specific cases?

packaging/language/bower.py
Fale: Actually it is: http://docs.ansible.com/ansible/npm_module.html

packaging/language/cpanm.py
Fale: I can create a regex to find all cases with and without quotes and then standardize them all

packaging/language/bower.py
gundalow: Oh, my mistake. I was looking at `bower`'s docs, rather than `npm`'s.

packaging/language/cpanm.py
Fale: In #3609 should be fixed

gundalow (approved):

PR #3639, opened 2016-12-02
Katello module support for repository_sets without a releasever

cliles:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
katello

##### ANSIBLE VERSION
```
ansible 2.3.0
```

##### SUMMARY
The existing logic will fail to add a katello/satellite repository set if that repository set does not have a release version. Some repository sets, like the satellite tools set, do not have a release version. By adding some conditionals we can also support these types of repository sets.

```
"WARNING:nailgun.client:Received HTTP 422 response: {\"displayMessage\":\"[\\\"releasever\\\"] cannot be specified for Red Hat Satellite Tools 6.2 (for RHEL 6 Server) (RPMs) as that information is not substituable
```

- Allow support of repository_sets that do not have a releasever, like the
Red Hat Satellite Tools 6.2 for RHEL 7 Server RPMs x86_64 set.

### DIFF
--- a/infrastructure/foreman/katello.py
+++ b/infrastructure/foreman/katello.py
@@ -241,7 +241,7 @@ class NailGun(object):
             e = get_exception()
             
             if "Import is the same as existing data" in e.message:
-                return True
+                return False
             else:
                 self._module.fail_json(msg="Manifest import failed with %s" % e)
 
@@ -304,7 +304,7 @@ class NailGun(object):
             formatted_name = [params['name'].replace('(', '').replace(')', '')]
             formatted_name.append(params['basearch'])
 
-            if params['releasever']:
+            if 'releasever' in params:
                 formatted_name.append(params['releasever'])
 
             formatted_name = ' '.join(formatted_name)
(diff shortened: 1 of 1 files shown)


cliles:
@ehelms, please review and let me know if I'm doing it wrong.

gregdek:
Thanks @cliles. To the current maintainers, @ehelms please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

infrastructure/foreman/katello.py
```diff
@@ -241,7 +241,7 @@ def manifest(self, params):
             e = get_exception()
             
             if "Import is the same as existing data" in e.message:
-                return True
+                return False
```
ehelms, line 244: Why do you want to return False here?

cliles: This was actually another change that has nothing to do with repository_sets, my bad :)

In the current behavior, even though the manifest does not change, ansible reports in the play recap that task did change . By returning False when the response indicates that there was no change in the manifest data we get that the task did not change.

```
PLAY RECAP ***************
x.x.x.x             : ok=1   changed=0    unreachable=0    failed=0
```

Instead of

```
PLAY RECAP ***************
x.x.x.x             : ok=0   changed=1    unreachable=0    failed=0
```

I had originally intended to modify the module to search and compare most entities prior to attempting to update/create them, but this case was handled with this simpler change.

ehelms: Oh right! Thanks for explaining.

ehelms (approved):

ehelms:
Thanks @cliles ! This is ready for merge.

cliles:
@ehelms - seems we didn't ship it in time, and now the repo is locked, I will be moving it over to the main ansible repo :/

PR #3679, opened 2016-12-06
Fix the usage of map since converting to using python six package

caphrim007:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

network/f5/bigip_virtual_server.py

##### ANSIBLE VERSION

```
ansible 1.9.2
```

##### SUMMARY
six changed the map function to return an iterable. So we need to
send that through the list method to get the value we need


caphrim007:
@gundalow what's wrong with shippable?

gundalow:
Repos have been merged, so you need to raise this against ansible/ansible

